PR #58, opened 2024-07-01
[torchcodec] Add support for Nvidia GPU Decoding

ahmadsharif1:
Summary:
1. Add CUDA support to VideoDecoder.cpp. This is done by checking what device is passed into the options and using CUDA if the device type is cuda.
2. Add -DENABLE_CUDA flag in cmake.
3. Check ENABLE_CUDA environment variable in setup.py and pass it down to cmake if it is present.
4. Add a unit test to demonstrate that CUDA decoding does work. This uses a different tensor than the one from CPU decoding because hardware decoding is intrinsically a bit inaccurate. I generated the reference tensor by dumping the tensor from the GPU on my devVM. It is possible different Nvidia hardware show different outputs. How to test this in a more robust way is TBD.
5. Added a new parameter for cuda device index for `add_video_stream`. If this is present, we will use it to do hardware decoding on a CUDA device.

Differential Revision: D59121006


Ilyabasharov:
When it will be available for using ?

ahmadsharif1:
It's available if you build torchcodec from source by installing pre-reqs and running:

```
CMAKE_BUILD_PARALLEL_LEVEL=8 CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 pip install -e . --no-build-isolation -vv
```

We plan to release it to pip later this year.

Ilyabasharov:
@ahmadsharif1 Thanks for your answer! Do you plan to support RTSP gpu decoding ?

PR #177, opened 2024-08-12
Update build command

NicolasHug:
https://github.com/pytorch/test-infra/pull/5538 was merged in `test-infra`, we can now explicitly define the command we want to run to build the wheels, and avoid resorting to the decentralized and less-obvious `env-var-script`.


ahmadsharif1:
Why are tests failing?

ahmadsharif1 (approved):

PR #180, opened 2024-08-14
Use our own index to seek more accurately when it is available

ahmadsharif1:
Before this change when the user asks us to seek to some timestamp T, we pass that timestamp down to FFMPEG.
However, we have noticed that sometimes FFMEPG seeks past that timestamp for certain files:
https://github.com/pytorch/torchcodec/issues/179
https://trac.ffmpeg.org/ticket/11137
After this change if we have our own index, we use that to seek to the last keyframe before the target frame. This avoids the buggy behavior and seeks are accurate even for the H265 encoded videos.

Benchmarked this change and didn't find changes above the noise level.

<details>

```py
# Before
python ~/personal/torchcodec/benchmarks/decoders/benchmark_decoders2.py 
med = 48.34ms +- 7.39

# After
python ~/personal/torchcodec/benchmarks/decoders/benchmark_decoders2.py 
med = 44.17ms +- 6.23

# Benchmark code
$ cat ~/personal/torchcodec/benchmarks/decoders/benchmark_decoders2.py
from torchcodec.decoders import SimpleVideoDecoder
import torch
import os


def bench(f, *args, num_exp=100, warmup=10, **kwargs):
    video_path = os.path.join(os.path.dirname(__file__), "../../test/resources/nasa_13013.mp4")
    decoder = SimpleVideoDecoder(video_path)
    pts_list = [x * 0.5 for x in range(20)]
    for _ in range(warmup):
        frames = [decoder.get_frame_displayed_at(pts) for pts in pts_list]

    times = []
    for _ in range(num_exp):
        start = perf_counter_ns()
        frames = [decoder.get_frame_displayed_at(pts) for pts in pts_list]
        end = perf_counter_ns()
        times.append(end - start)

    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    return med

if __name__ == "__main__":
    times = bench(lambda x: x, num_exp=1000, warmup=10)
    report_stats(times, unit="ms")
```

</details>


test/decoders/test_simple_video_decoder.py
```diff
@@ -322,11 +322,7 @@ def test_get_frame_displayed_at(self):
 
     def test_get_frame_displayed_at_h265(self):
         decoder = SimpleVideoDecoder(H265_VIDEO.path)
-        # Note that for H265, FFMPEG's seeking is not precise. Even though we ask to
-        # seek with a max_ts=0.5, FFMPEG will seek beyond that point.
-        # TODO: Revert use frame5 in the test below once it's fixed upstream:
-        # https://trac.ffmpeg.org/ticket/11137
-        ref_frame6 = H265_VIDEO.get_frame_by_name("frame000006")
+        ref_frame6 = H265_VIDEO.get_frame_by_name("frame000005")
```
NicolasHug, line 325: Now that the comment was removed, it's not immediately obvious why this test exists. It might be worth clarifying its purpose by adding e.g.

```
# Non-regression test for https://github.com/pytorch/torchcodec/issues/179
```

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -732,6 +732,18 @@ void VideoDecoder::maybeSeekToBeforeDesiredPts() {
   int firstActiveStreamIndex = *activeStreamIndices_.begin();
   const auto& firstStreamInfo = streams_[firstActiveStreamIndex];
   int64_t desiredPts = *maybeDesiredPts_ * firstStreamInfo.timeBase.den;
+
+  // For some encodings like H265, FFMPEG sometimes seeks past the point we
+  // set as the max_ts. So we use our own index to give it the exact pts we
+  // want.
```
NicolasHug, line 738: Correct me if I'm wrong - after the newly added `if` block, `desiredPts` isn't "the exact pts we want", right?

NicolasHug (approved):
Thanks a ton for the fix and for the benchmark @ahmadsharif1.

Unrelated to this PR, but I'm surprised the `stddev` is so large even over 1000 runs!

PR #183, opened 2024-08-15
Added colorspace and color_range to torchcodec

ahmadsharif1:
Without this change, with ffmpeg7 we get warning messages like the following:

Changing video frame properties on the fly is not supported by all filters

For more info see this:

https://trac.ffmpeg.org/ticket/10823


scotts:
I also experimented with a similar change, but I believe it's not supported by all versions of FFmpeg. If we do something like this, we'll need to guard it appropriately.

PR #187, opened 2024-08-15
Add CI for linux cuda

ahmadsharif1:
This diff adds GPU CI to TorchCodec.

It installs the NVFFMPEG headers, builds FFMPEG from source, builds TorchCodec and then runs test on the GPU tests.

This will be useful to catch GPU recording regressions.

I also added an explicit dependency on NPP image processing library because it was failing to build inside the container environment.

### DIFF
--- a/.github/workflows/test_linux_cuda.yaml
+++ b/.github/workflows/test_linux_cuda.yaml
@@ -32,10 +32,10 @@ jobs:
         pwd
         conda create --yes --name test
         conda activate test
-        conda install --yes pip cmake pkg-config nasm
+        conda install --quiet --yes pip cmake pkg-config nasm
 
-        pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu124
-        conda install --yes nvidia::libnpp
+        pip install --quiet --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu124
+        conda install --quiet --yes nvidia::libnpp
 
         # Build and install FFMPEG from source with CUDA enabled.
         # The one on conda doesn't have CUDA enabled.
@@ -54,7 +54,7 @@ jobs:
         which pkg-config
         pkg-config --list-all
         ./configure --prefix=$CONDA_PREFIX --enable-nonfree --enable-cuda-nvcc --disable-static --enable-shared --optflags=-fno-omit-frame-pointer --disable-stripping --enable-cuvid
-        make -j install
+        make --quiet -j install
         popd
 
         CMAKE_BUILD_PARALLEL_LEVEL=8 CXXFLAGS="" LDFLAGS="-Wl,--allow-shlib-undefined -Wl,-rpath,$CONDA_PREFIX/lib -Wl,-rpath-link,$CONDA_PREFIX/lib -L$CONDA_PREFIX/lib" CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 ENABLE_NVTX=1 pip install -e . --no-build-isolation -vv --debug
--- a/.github/workflows/test_linux_cuda.yaml
+++ b/.github/workflows/test_linux_cuda.yaml
@@ -57,8 +57,7 @@ jobs:
         make --quiet -j install
         popd
 
-        CMAKE_BUILD_PARALLEL_LEVEL=8 CXXFLAGS="" LDFLAGS="-Wl,--allow-shlib-undefined -Wl,-rpath,$CONDA_PREFIX/lib -Wl,-rpath-link,$CONDA_PREFIX/lib -L$CONDA_PREFIX/lib" CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 ENABLE_NVTX=1 pip install -e . --no-build-isolation -vv --debug
-        pip install pytest
+        CMAKE_BUILD_PARALLEL_LEVEL=8 CXXFLAGS="" LDFLAGS="-Wl,--allow-shlib-undefined -Wl,-rpath,$CONDA_PREFIX/lib -Wl,-rpath-link,$CONDA_PREFIX/lib -L$CONDA_PREFIX/lib" CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 ENABLE_NVTX=1 pip install -e ".[dev]"" --no-build-isolation -vv --debug
 
         # We skip certain tests because they are not relevant to GPU decoding and they always fail with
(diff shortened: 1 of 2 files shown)


.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - nightly
+      - main
+      - release/*
+  workflow_dispatch:
+
+jobs:
+  tests:
+    strategy:
+      matrix:
+        python_version: ["3.8", "3.9", "3.10"]
```
NicolasHug, line 17: We shouldn't test on 3.8 as we don't support it.
For now, a single version should be enough on this job, so let's drop 3.10 too.

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+        cuda_arch_version: ["12.4"]
```
NicolasHug, line 18: Eventually we'll also want 12.0 and 11.X but that's OK for now.

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+      gpu-arch-version: ${{ matrix.cuda_arch_version }}
+      timeout: 120
+
+      script: |
+        nvidia-smi
+        ls -l
+        pwd
+        conda create --yes --name test
+        conda activate test
+        conda install --yes pip cmake pkg-config nasm
```
NicolasHug, line 34: Also use `--quiet` where possible in every conda call, otherwise the logs get really long

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+        # This fails because conda reactivate fails inside this env
+        # conda install --yes conda-forge::compilers
```
NicolasHug, line 37: Remove

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+        pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cu124
```
NicolasHug, line 39: We shouldn't need to install torchaudio, and we shouldn't need to install torchvision at this point as well

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+        pushd ffmpeg
+        git checkout origin/release/6.1
+        which pkg-config
+        pkg-config --list-all
+        ./configure --prefix=$CONDA_PREFIX --enable-nonfree --enable-cuda-nvcc --disable-static --enable-shared --optflags=-fno-omit-frame-pointer --disable-stripping --enable-cuvid
+        make -j install
+        popd
+
+        CMAKE_BUILD_PARALLEL_LEVEL=8 CXXFLAGS="" LDFLAGS="-Wl,--allow-shlib-undefined -Wl,-rpath,$CONDA_PREFIX/lib -Wl,-rpath-link,$CONDA_PREFIX/lib -L$CONDA_PREFIX/lib" CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 ENABLE_NVTX=1 pip install -e . --no-build-isolation -vv --debug
+        pip install pytest
```
NicolasHug, line 63: You should be using `pip install -e ".[dev]"` which will install pytest (and remove this line)

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+        # We skip certain tests because they are not relevant to GPU decoding and they always fail with
+        # a custom FFMPEG build.
```
NicolasHug, line 66: We probably want to address those failures though

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,68 @@
...
+        pytest -k "not (test_get_metadata or get_ffmpeg_version)"
```
NicolasHug, line 67: So, fun fact: it's currently hard to tell whether the GPU tests are being ran at all, because they're protected within an "if" block. I'll share pointers on how to address this offline

.github/workflows/test_linux_cuda.yaml
ahmadsharif1: Yeah, these are the version string issues. I will fix them later as they are not GPU related.

.github/workflows/test_linux_cuda.yaml
ahmadsharif1: This is accurate and a real problem. Should I address this in a different diff?

.github/workflows/test_linux_cuda.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+      script: |
+        nvidia-smi
+        ls -l
+        pwd
+        conda create --yes --name test
+        conda activate test
+        conda install --quiet --yes pip cmake pkg-config nasm
+
+        pip install --quiet --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu124
```
ahmadsharif1, line 37: Torchvision is required for the gpu benchmark which uses functional transforms from vision. See gpu_benchmark.py for details.

https://github.com/pytorch/torchcodec/blob/f4065f1b477148cfb0ef94167fb0bf3a63803e55/benchmarks/decoders/gpu_benchmark.py#L8

NicolasHug: but we don't need it to build, right? So we might prefer to install it later. This is OK though, since this job is just a temporary setup I guess.

.github/workflows/test_linux_cuda.yaml
ahmadsharif1: Torchvision is needed by gpu_benchmark.py

NicolasHug (approved):
Thanks @ahmadsharif1 . Let's follow-up about https://github.com/pytorch/torchcodec/pull/187/files#r1723560554 in a separate PR.

Eventually we'll want to make sure wheels can be built (and tested) and integrate that with our existing `wheel.yml` workflows, but this is a great first step to test GPU-related changes.

.github/workflows/test_linux_cuda.yaml
ahmadsharif1: That's a good point. Should I move it below after the build step so it doesn't accidentally leak as a build-time dep? WDYT?

PR #194, opened 2024-08-27
Add flags to set benchmark video file paths

ahmadsharif1:
We don't actually have a large video file in the repo by design (to keep the repo lightweight).

So we add a flag to specify it.

```
python benchmarks/decoders/benchmark_decoders.py --bm_large_video_path=/home/ahmads/jupyter/853.mp4
[------------------------------- decode latency for function call pattern for 700KB video ------------------------------]
                                             |  10 seek()+next()  |  1 next()  |  10 next()  |  100 next()  |  200 next()
1 threads: --------------------------------------------------------------------------------------------------------------
      DecordNonBatchDecoderAccurateSeek      |        71.2        |    22.5    |     29.8    |     94.7     |    135.5   
      TorchCodecDecoderNonCompiled           |       111.9        |    36.0    |     41.0    |     85.5     |    130.9   
      TCNonCompiled:ffmpeg_thread_count=1    |       173.5        |    20.1    |     26.1    |     90.3     |    169.7   
      TorchCodecDecoderCompiled              |       101.3        |    33.3    |     40.3    |     85.1     |    132.0   
      TVNewAPIDecoderWithBackendVideoReader  |       437.5        |    15.2    |     21.2    |     64.3     |    121.2   
      TorchAudioDecoder                      |       395.9        |     8.0    |     16.9    |     84.2     |    172.6   

[--------------- decode latency for function call pattern for 50MB video ----------------]
                                             |  5 seek()+next()  |  1 next()  |  10 next()
1 threads: -------------------------------------------------------------------------------
      DecordNonBatchDecoderAccurateSeek      |       1475.0      |   331.7    |     635.3 
      TorchCodecDecoderNonCompiled           |       1360.8      |   296.1    |     466.1 
      TCNonCompiled:ffmpeg_thread_count=1    |       7412.2      |   224.7    |     769.1 
      TorchCodecDecoderCompiled              |       1253.5      |   270.3    |     425.4 
      TVNewAPIDecoderWithBackendVideoReader  |      14022.6      |   298.3    |    1149.8 
      TorchAudioDecoder                      |      13059.1      |   315.1    |    1047.5 

[ creation latency for function call pattern for 50MB video ]
                         |  creation+next time
1 threads: -----------------------------------
      TorchCodecDecoder  |        293.6       

```

### DIFF
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -27,6 +27,10 @@ torch._dynamo.config.cache_size_limit = 100
 torch._dynamo.config.capture_dynamic_output_shape_ops = True
 
 
+def in_fbcode() -> bool:
+    return "FB_PAR_RUNTIME_FILES" in os.environ
+
+
 class AbstractDecoder:
     def __init__(self):
         pass
@@ -198,14 +202,13 @@ class TorchAudioDecoder(AbstractDecoder):
 
 
 def get_test_resource_path(filename: str) -> str:
-    if not __package__:
-        return os.path.join(
-            os.path.dirname(__file__), "..", "..", "test", "resources", filename
-        )
-
-    resource = importlib.resources.files(__package__).joinpath(filename)
-    with importlib.resources.as_file(resource) as path:
-        return os.fspath(path)
+    if in_fbcode():
+        resource = importlib.resources.files(__package__).joinpath(filename)
+        with importlib.resources.as_file(resource) as path:
+            return os.fspath(path)
+    return os.path.join(
+        os.path.dirname(__file__), "..", "..", "test", "resources", filename
+    )
 
 
 def create_torchcodec_decoder_from_file(video_file):
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -197,8 +197,12 @@ class TorchAudioDecoder(AbstractDecoder):
         return frames
 
 
-def get_video_path(filename: str) -> str:
-    return os.path.join("/home/ahmads/personal/torchcodec/test/resources", filename)
+def get_test_resource_path(filename: str) -> str:
+    if not __package__:
+        return os.path.join(
+            os.path.dirname(__file__), "..", "..", "test", "resources", filename
+        )
+
     resource = importlib.resources.files(__package__).joinpath(filename)
     with importlib.resources.as_file(resource) as path:
         return os.fspath(path)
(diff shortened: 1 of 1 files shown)


benchmarks/decoders/benchmark_decoders.py
```diff
@@ -198,6 +198,7 @@ def get_consecutive_frames_from_video(self, video_file, numFramesToDecode):
 
 
 def get_video_path(filename: str) -> str:
+    return os.path.join("/home/ahmads/personal/torchcodec/test/resources", filename)
```
scotts, line 201: Debugging remnant. :)

ahmadsharif1: Good catch.  Actually I meant to fix this because this file is not part of the package -- __package__ is actually None since we run this from the source dir.

scotts (approved):

benchmarks/decoders/benchmark_decoders.py
```diff
@@ -197,7 +197,12 @@ def get_consecutive_frames_from_video(self, video_file, numFramesToDecode):
         return frames
 
 
-def get_video_path(filename: str) -> str:
+def get_test_resource_path(filename: str) -> str:
+    if not __package__:
```
NicolasHug, line 201: I don't understand why `__package__` is present here and not in fbcode. I also don't understand why test resources are available internally as `resource = importlib.resources.files(__package__).joinpath(filename)`, but not on OSS.
Before merging this PR, I think we should either: 

- understand and document what each block is doing and why it works
- replace the `if not __package__` check with a simpler `if IN_FBCODE:` block (or something similar) which would be more explicit.

scotts: Agreed that `if IN_FBCODE` is a better check. That is what we do in a similar chunk of code for testing: [utils.py#L34](https://github.com/pytorch/torchcodec/blob/e6a4161edeed22d1157817c42bc8ceaa4f47a787/test/utils.py#L34).

My understanding of why we have different behavior depending on fbcode or not:

* In fbcode, we use `buck run //path/to/target:` to execute the benchmark. Buck creates a temporary directory with all relevant files for executing the benchmark, including all resources. Where these files actually end up is owned by buck. Hence, we have to ask `importlib` to tell use where the resources are. The code we're using as an internal best-practice.
* Externally, we just run benchmark on the command line: `python3 ./benchmarks/decoders/benchmark_decoders.py`. In that case, the file system is as it appears to be, and we can just access the files as one would expect.

ahmadsharif1: I added a function that checks for the presence of an environment variable that seems to be present running inside fbcode.

scotts (approved):

PR #201, opened 2024-09-05
Add Mac to FFmpeg build CI

scotts:
Adds a Mac build of the FFmpeg non-GPL libraries. The code is lifted almost verbatim from [TorchAudio's setup](https://github.com/pytorch/audio/blob/97ed7b36b7a741253d4e41e4da3c901d83294503/.github/workflows/ffmpeg.yml#L60), which reuses [pytorch/test-infra](https://github.com/pytorch/test-infra). Note there are subtle differences between our current Linux job spec and this new one for Mac - those differences are necessary to work with pytorch/test-infra's setup.

Since this action will only execute once a week on Sunday, I tested it by temporarily hacking it to run on a pull request. See the result of one of those runs to see successful execution: https://github.com/pytorch/torchcodec/actions/runs/10730119629

### DIFF
--- a/.github/workflows/build_ffmpeg.yaml
+++ b/.github/workflows/build_ffmpeg.yaml
@@ -6,7 +6,7 @@
 #
 # This job does not include the uploading part.
 # Upload needs to be done manually, and it should be done only once
-# par new major release of FFmepg.
+# per new major release of FFmepg.
 name: Build non-GPL FFmpeg from source
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/build_ffmpeg.yaml
```diff
@@ -43,3 +43,26 @@ jobs:
         with:
           name: "ffmpeg_${{ matrix.ffmpeg-version }}_linux_x86_64"
           path: artifact_dir
+  LGPL-macOS:
+    strategy:
+      fail-fast: false
+      matrix:
+        ffmpeg-version: ["4.4.4", "5.1.4", "6.1.1", "7.0.1"]
+        runner: ["macos-m1-stable", "macos-12"]
```
NicolasHug, line 51: Pytorch doesn't support x86 for mac anymore

```suggestion
        runner: ["macos-m1-stable"]
```

NicolasHug (approved):

.github/workflows/build_ffmpeg.yaml
scotts: Oh, interesting. The curious bit here is that my system is an x86 Mac, so if I'm going to do any local testing of Mac wheels, then I'll need to keep this in. I can just use CI for everything instead. Let me know which you think is more important: aligning with general PyTorch, or enabling local testing.

scotts:
As I look forward to how to how we'll want to pull these artifacts to use in builds, I think it will actually be easier if we build Linux and Mac ffmpeg non-GPL binaries using the same infra, and upload them together in the same tar file. I'm going to change this PR to reflect that. I'll request a re-review.

@NicolasHug, please let me know if you think we should prioritize alignment with PyTorch's support, or keep the repo in a state we can all locally test.

.github/workflows/build_ffmpeg.yaml
NicolasHug: You shouldn't need to keep this in for local testing. This just instructs the CI to spin x86 instances. This is never executed on your machine. The actual build script should execute just fine for you locally.

NicolasHug:
> how we'll want to pull these artifacts to use in builds

We won't be pulling those artifacts into other builds. We'll manually download those artifacts to S3, and other builds will pull from S3.

> it will actually be easier if we build Linux and Mac ffmpeg non-GPL binaries using the same infra

I was going to suggest that we add a "TODO: we should migrate the linux job to pytorch infra as well". This doesn't need to happen now, but if you think it makes more sense, that's perfectly fine too!

.github/workflows/build_ffmpeg.yaml
scotts: Yes, agreed, this never executes on my machine. But I will want to locally define `BUILD_AGAINST_ALL_FFMPEG_FROM_S3` and build on my machine to test that flow locally. And if we don't build non-GPL FFmpeg x86 binaries for Mac in this step in CI, then they won't be available to upload to S3, and I won't be able to test that flow locally.

We can remove it after I've gotten everything setup, but I think at least while I'm working on this, it's better that these binaries are available.

scotts:
I moved Linux over to PyTorch infra as well. All binaries now get uploaded into a single tar file. Example: https://github.com/pytorch/torchcodec/actions/runs/10744678413

scotts:
The failing test is unrelated to these changes, see Issue #203. Merging to unblock; we can remove macos-12 after the entire Mac wheel flow works.

PR #205, opened 2024-09-09
Allow option to use the swscale library for color conversion instead of filtergraph

ahmadsharif1:
This PR does the following:

1. Refactor the VideoDecoder.cpp code so we can decouple frame allocation from decoding and color-conversion.
2. Allow the option of using sws_scale for color conversion instead of filtergraph. sws_scale is faster than filtergraph at color converting smaller videos. Moreover the setup and destruction time overhead of filtergraph is much larger than sws_scale's setup and destruction time.
3. benchmark_decoders.py: Added some code to the benchmark to print more detailed stats if a flag is turned on.
4. benchmark_decoders.py: Added code to pass in options to torchcodec decoders.
5. Use bilinear interpolation whenever we are scaling. I will test this against torchvision's bilinear scaling in a subsequent diff.

Benchmark results show performance improvement:

```
python benchmarks/decoders/benchmark_decoders.py --bm_video_paths=/home/ahmads/personal/lerobot/benchmarks/video/outputs/video_benchmark/videos/libx264_yuv420p_None_None/lerobot_pusht_image.mp4,/home/ahmads/personal/lerobot/benchmarks/video/outputs/video_benchmark/videos/libx265_yuv420p_None_None/lerobot_pusht_image.mp4,/home/ahmads/jupyter/carmel1.mp4 --decoders=torchvision,tcoptions,tcoptions:color_conversion_library=swsscale+num_threads=1,tcoptions:color_conversion_library=filtergraph+num_threads=1,tcbatchoptions,tcbatchoptions:num_
threads=1+color_conversion_library=swsscale,tcbatchoptions:num_threads=1+color_conversion_library=filtergraph --bm_video_speed_min_run_seconds=10
[ video=/home/ahmads/personal/lerobot/benchmarks/video/outputs/video_benchmark/videos/libx264_yuv420p_None_None/lerobot_pusht_image.mp4 h264 96x96, 16.1s 10.0fps ]
                                                                                     |  10 seek()+next()  |  1 next()  |  10 next()
1 threads: ------------------------------------------------------------------------------------------------------------------------
      TorchcodecNonCompiled:color_conversion_library=filtergraph+num_threads=1       |        12.5        |    4.5     |     5.2   
      TVNewAPIDecoderWithBackendVideoReader                                          |        66.8        |    4.4     |     5.8   
      TorchcodecNonCompiled:color_conversion_library=swsscale+num_threads=1          |         8.9        |    1.1     |     1.8   
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=filtergraph  |        12.6        |    4.9     |     5.8   
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=swsscale     |         8.7        |    1.5     |     2.1   

[ video=/home/ahmads/personal/lerobot/benchmarks/video/outputs/video_benchmark/videos/libx265_yuv420p_None_None/lerobot_pusht_image.mp4 hevc 96x96, 16.1s 10.0fps ]
                                                                                     |  10 seek()+next()  |  1 next()  |  10 next()
1 threads: ------------------------------------------------------------------------------------------------------------------------
      TorchcodecNonCompiled:color_conversion_library=filtergraph+num_threads=1       |        13.5        |    4.2     |     5.5   
      TVNewAPIDecoderWithBackendVideoReader                                          |        68.7        |    3.8     |     5.2   
      TorchcodecNonCompiled:color_conversion_library=swsscale+num_threads=1          |         9.7        |    1.0     |     1.8   
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=filtergraph  |        13.9        |    4.6     |     5.6   
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=swsscale     |         9.2        |    1.4     |     2.0   

[------------------------------- video=/home/ahmads/jupyter/carmel1.mp4 h264 640x360, 1.3s 30.0fps -------------------------------]
                                                                                     |  10 seek()+next()  |  1 next()  |  10 next()
1 threads: ------------------------------------------------------------------------------------------------------------------------
      TorchcodecNonCompiled:color_conversion_library=filtergraph+num_threads=1       |        44.6        |    9.2     |     27.5  
      TVNewAPIDecoderWithBackendVideoReader                                          |       255.2        |    9.7     |     28.3  
      TorchcodecNonCompiled:color_conversion_library=swsscale+num_threads=1          |        40.8        |    4.5     |     14.0  
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=filtergraph  |        37.4        |    8.7     |     17.9  
      TorchCodecNonCompiledBatch:num_threads=1+color_conversion_library=swsscale     |        31.7        |    4.0     |     12.5  

[ video=/home/ahmads/personal/lerobot/benchmarks/video/outputs/video_benchmark/videos/libx264_yuv420p_None_None/lerobot_pusht_image.mp4 h264 640x360, 1.3s 30.0fps ]
                             |  create()+next()
1 threads: ------------------------------------
      TorchcodecNonCompiled  |        11.6     

```


test/decoders/test_video_decoder_ops.py
NicolasHug, line 1: There doesn't seem to be any new test covering the new color conversion option?

NicolasHug (commented):
Thanks for the PR, I made a quick review. I have further questions / discussion points that I'll take offline for now

NicolasHug (commented):
Thanks for the PR, I made a quick review. I have further questions / discussion points that I'll take offline for now

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -139,6 +146,7 @@ class VideoDecoder {
     // is the same as the original video.
     std::optional<int> width;
     std::optional<int> height;
+    std::optional<ColorConversionLibrary> colorConversionLibrary = AUTO;
```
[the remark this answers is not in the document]
ahmadsharif1: no

test/decoders/test_video_decoder_ops.py
ahmadsharif1: added tests now

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -123,7 +123,14 @@ class VideoDecoder {
...
   // ADDING STREAMS API
   // --------------------------------------------------------------------------
-
+  enum ColorConversionLibrary {
+    // Let the decoder decide the color conversion library to use.
+    AUTO,
+    // Use the libswscale library for color conversion.
+    FILTERGRAPH,
+    // Use the libswscale library for color conversion.
+    SWSSCALE
```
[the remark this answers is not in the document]
ahmadsharif1: Good point. I made the option sw_scale and also changed the title of the PR

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -782,8 +825,43 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
...
+      rawOutput.data = tensor.data_ptr<uint8_t>();
+      convertFrameToBufferUsingSwsScale(rawOutput);
+      auto convertDone = std::chrono::high_resolution_clock::now();
+      auto startToAllocateDone =
+          std::chrono::duration_cast<std::chrono::microseconds>(
+              allocateDone - start);
+      auto allocateToConvertDone =
+          std::chrono::duration_cast<std::chrono::microseconds>(
+              convertDone - allocateDone);
+      auto total = std::chrono::duration_cast<std::chrono::microseconds>(
```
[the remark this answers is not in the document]
ahmadsharif1: I will refactor this in a different diff. It's probably not that bad. A self-profiling decoder is great imo

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -401,7 +418,18 @@ void VideoDecoder::addVideoStreamDecoder(
   activeStreamIndices_.insert(streamNumber);
   updateMetadataWithCodecContext(streamInfo.streamIndex, codecContext);
   streamInfo.options = options;
-  initializeFilterGraphForStream(streamNumber, options);
+  int width = options.width.value_or(codecContext->width);
+  // There is a bug in sws_scale where it doesn't handle non-multiple of 16
+  // widths.
+  // https://stackoverflow.com/questions/74351955/turn-off-sw-scale-conversion-to-planar-yuv-32-byte-alignment-requirements
+  // In that case we are forced to use a filtergraph to do the color conversion.
```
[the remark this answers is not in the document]
ahmadsharif1: At the moment we wont warn anyone since there are no users. Idealy we don't warn and just do the right thing under the AUTO enum

benchmarks/decoders/benchmark_decoders.py
```diff
@@ -132,30 +166,54 @@ def get_frames_from_video(self, video_file, pts_list):
...
         return frames
 
 
-class TorchCodecDecoderNonCompiledBatch(AbstractDecoder):
-    def __init__(self, num_threads=None):
+class TorchCodecNonCompiledBatch(AbstractDecoder):
+    def __init__(self, num_threads=None, color_conversion_library=None):
         self._print_each_iteration_time = False
-        self._num_threads = num_threads
+        self._num_threads = int(num_threads) if num_threads else None
```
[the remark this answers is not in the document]
ahmadsharif1: Because now we read in options as a "+"-separarted string from the cli and pass them as **kwargs. See the changes in main() below

benchmarks/decoders/benchmark_decoders.py
```diff
@@ -86,38 +86,72 @@ def get_consecutive_frames_from_video(self, video_file, numFramesToDecode):
...
+        self._color_conversion_library = color_conversion_library
 
     def get_frames_from_video(self, video_file, pts_list):
         decoder = create_from_file(video_file)
-        add_video_stream(decoder, num_threads=self._num_threads)
+        add_video_stream(
```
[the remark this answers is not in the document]
ahmadsharif1: Good point. Done

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -123,7 +123,14 @@ class VideoDecoder {
   // --------------------------------------------------------------------------
   // ADDING STREAMS API
   // --------------------------------------------------------------------------
-
+  enum ColorConversionLibrary {
+    // Let the decoder decide the color conversion library to use.
```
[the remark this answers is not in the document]
ahmadsharif1: Decoder selection is an implementation detail that we can change at any time.

By decoder I meant the C++ `VideoDecoder`

I changed the wording

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -887,8 +967,22 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndexes(
...
           "Invalid frame index=" + std::to_string(frameIndex));
     }
-    torch::Tensor frame = getFrameAtIndex(streamIndex, frameIndex).frame;
-    output.frames[i++] = frame;
+    int64_t pts = stream.allFrames[frameIndex].pts;
+    setCursorPtsInSeconds(ptsToSeconds(pts, stream.timeBase));
+    auto rawSingleOutput = getNextRawDecodedOutputNoDemux();
+    if (stream.filterState.sourceContext == nullptr) {
+      // We are using sws_scale to conver the frame to tensor. So we can
+      // convert directly to the output tensor.
```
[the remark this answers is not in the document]
ahmadsharif1: Updated the comment

src/torchcodec/decoders/_core/VideoDecoder.cpp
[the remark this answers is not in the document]
NicolasHug: Things tend to look more obvious in the eyes of the code author :) 

In this block there are 12 statements, and 5 of those are dedicated to the core logic. The 7 remaining ones are about benchmarking.

Do we expect such benchmark info to be forever useful, or is this more of a one-time thing?

Before leaving this out for a follow-up diff, I would suggest to protect the benchmarking logic within `#ifdef` blocks, so that they can easily be collapsed on the IDE to ease code review / reading.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -123,7 +123,14 @@ class VideoDecoder {
   // --------------------------------------------------------------------------
   // ADDING STREAMS API
   // --------------------------------------------------------------------------
-
+  enum ColorConversionLibrary {
+    // Let the implementation decide the color conversion library to use.
+    AUTO,
```
NicolasHug, line 128: Since AUTO is currently just FILTERGRAPH, I'd suggest to remove this and only add it back when we intend to change the default policy. It was a source of confusion to me (as in previous comments).

test/decoders/test_video_decoder_ops.py
```diff
@@ -322,6 +323,46 @@ def test_frame_pts_equality(self):
...
+        assert_tensor_equal(frame0, reference_frame0)
+        reference_frame1 = NASA_VIDEO.get_frame_data_by_index(1)
+        frame1, _, _ = get_next_frame(decoder)
+        assert_tensor_equal(frame1, reference_frame1)
+        seek_to_pts(decoder, 6.0)
+        frame_time6, _, _ = get_next_frame(decoder)
+        reference_frame_time6 = NASA_VIDEO.get_frame_by_name("time6.000000")
+        assert_tensor_equal(frame_time6, reference_frame_time6)
+
+    def test_color_conversion_library_with_scaling(self):
```
NicolasHug, line 341: Just a note: it's worth testing both upscaling and downscaling. Some scaling options like antialising are only relevant when down-scaling.

Also, I guess this isn't the point of this test at this time, but before we enable the `AUTO` strategy to automatically switch between swscale and filtergraph, we should get more coverage of the different resizing options. We need to avoid discrepancies between the resizing strategies of these 2 libraries. I think it's worth mentioning that as a comment in this test.

ahmadsharif1: Good callout. I parameterized this test to test upscaling, downscaling and stretching in one dimension while compressing in another.

src/torchcodec/decoders/_core/VideoDecoder.h
ahmadsharif1: Done.

src/torchcodec/decoders/_core/VideoDecoder.cpp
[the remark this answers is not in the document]
ahmadsharif1: I think benchmark would be useful forever.

I think #ifdefs may make it even harder to read.

Later on I plan to add a timing class that will have its own code #ifdefs commented out for non-developer builds.

Should I really add #ifdefs in this diff? Leaving this unresolved.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -325,7 +339,7 @@ void VideoDecoder::initializeFilterGraphForStream(
     width = *options.width;
     height = *options.height;
   }
-  std::snprintf(description, sizeof(description), "scale=%d:%d", width, height);
+  std::snprintf(description, sizeof(description), "scale=%d:%d:sws_flags=bilinear", width, height);
```
scotts, line 342: Not a priority now, but we can do this in a safer way with `std::stringstream`; it won't require allocating a static sized buffer.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -401,7 +415,18 @@ void VideoDecoder::addVideoStreamDecoder(
...
-  initializeFilterGraphForStream(streamNumber, options);
+  int width = options.width.value_or(codecContext->width);
+  // There is a bug in sws_scale where it doesn't handle non-multiple of 32
+  // widths.
+  // https://stackoverflow.com/questions/74351955/turn-off-sw-scale-conversion-to-planar-yuv-32-byte-alignment-requirements
+  // In that case we are forced to use a filtergraph to do the color conversion.
+  auto colorConversionLibrary = options.colorConversionLibrary.value_or(
+      ColorConversionLibrary::FILTERGRAPH);
+  bool useFilterGraph =
+      (colorConversionLibrary != ColorConversionLibrary::SWSCALE);
```
scotts, line 426: Based on the comment above, I would expect to see something like:
```
if (width % 32 != 0) {
  useFilterGraph = true;
}
```
I think I understand why: for benchmarking, we want absolute to use sws_scale even if the results are wrong. That is, we want the benchmark to control which color conversion library we use, not the C++ library.

I think that's okay, so long as it only happens through `_add_video_stream()`. I would prefer that we defensively add the above check in `add_video_stream()` now, rather than waiting until we implement `ColorConversionLibrary::AUTO`. This sort of bug is the kind that is easy to miss, and I'd like us to be extra defensive. The comment above, with the reference to StackOverflow, would then make the most sense there.

src/torchcodec/decoders/_core/VideoDecoder.cpp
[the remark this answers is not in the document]
scotts: I actually agree with @NicolasHug that this code is now hard to read. My reasons, which some are the same as what Nicolas said:

1. The timing code takes up more lines than the actual logic.
2. Because we're using `auto` as the type, I have to actually read each timing line to realize it's a timing line. I can't just skim over them and focus on the actual logic. (Unfortunately, [explicitly stating the return type](https://fburl.com/r4137aac) would be even noisier.)
3. We're not using any new lines to group code together. "Paragraphs" of code would help me figure out what I can ignore as timing, as I expect a paragraph of code to have a start and stop timing call.

This timing code is very granular. I understand that in this PR, the amount of time just a few statements took was key to getting a big perf win, but I don't think that necessarily means we should keep it all the time. I consider it similar to putting in print statements when debugging. Some print statements graduate to actual logging, but most don't, and we delete most debugging print statements at commit time.

I would prefer either:

1. We remove the timing code in this PR, and wait until we have a readable abstraction to commit it. _OR_:
2. We limit the timing code to function boundaries. Yes, that will not be enough to fully diagnose perf problems, but it's enough to know where to start digging.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -887,8 +964,24 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndexes(
       throw std::runtime_error(
           "Invalid frame index=" + std::to_string(frameIndex));
     }
-    torch::Tensor frame = getFrameAtIndex(streamIndex, frameIndex).frame;
-    output.frames[i++] = frame;
+    int64_t pts = stream.allFrames[frameIndex].pts;
+    setCursorPtsInSeconds(ptsToSeconds(pts, stream.timeBase));
+    auto rawSingleOutput = getNextRawDecodedOutputNoDemux();
+    if (stream.filterState.sourceContext == nullptr) {
```
scotts, line 970: It is not obvious to me that `stream.filterState.sourceContext` being null means that we should use sws_scale. It would be more obvious if the else branch explicitly used the value in this function, but the use is buried in the call to `convertFrameToTensorUsingFilterGraph()`.

I think we should store a `ColorConversionLibrary` value in the decoder's state, and use that value for control flow. We can always do not-null assertions as needed upon actual use.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1053,6 +1152,67 @@ double VideoDecoder::getPtsSecondsForFrame(
...
+        frame->height,
+        frameFormat,
+        outputWidth,
+        outputHeight,
+        AV_PIX_FMT_RGB24,
+        SWS_BILINEAR,
+        nullptr,
+        nullptr,
+        nullptr);
+    const int *invTable, *table;
```
scotts, line 1176: As a rule in C++, we should declare one variable per line, and it should be pronounced `const int* invTable`. I'd also ideally like to see us explicitly initialize everything, even if it's going to be set in a function we're caling.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1053,6 +1152,67 @@ double VideoDecoder::getPtsSecondsForFrame(
...
+    int srcRange, dstRange, brightness, contrast, saturation;
+    sws_getColorspaceDetails(
+        swsContext,
+        (int**)&invTable,
```
scotts, line 1180: This cast is only necessary because we declared `invTable` as a const pointer, but I don't see any value in doing so. We should declare it and `table` as non-const pointers, eliminating the need for the casts.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -153,6 +160,34 @@ class VideoDecoder {
...
+  struct RawDecodedOutput {
+    // The actual decoded output as a unique pointer to an AVFrame.
+    UniqueAVFrame frame;
+    // The stream index of the decoded frame.
+    int streamIndex;
+    // This is an unowned pointer that we copy the frame data to after color
+    // conversion.
+    // For a single tensor this points to the start of data_ptr. For a batch
+    // tensor it may point to the middle of the allocated batch tensor.
+    void *data = nullptr;
```
scotts, line 186: Let's declare it as `void* data`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: I removed the comment as it was just a remnant from before.

I hardcoded "filtergraph" in the public API.

src/torchcodec/decoders/_core/VideoDecoder.cpp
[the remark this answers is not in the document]
ahmadsharif1: Disagree and commit. Removed those lines

benchmarks/decoders/benchmark_decoders.py
```diff
@@ -86,38 +86,72 @@ def get_consecutive_frames_from_video(self, video_file, numFramesToDecode):
...
         reader = self.torchvision.io.VideoReader(video_file, "video")
+        create_done = timeit.default_timer()
         frames = []
         for pts in pts_list:
             reader.seek(pts)
             frame = next(reader)
             frames.append(frame["data"].permute(1, 2, 0))
+        frames_done = timeit.default_timer()
+        if self._print_each_iteration_time:
+            del reader
```
scotts, line 106: Are you intending to time how long it takes to do the deallocations of memory on the C++ side? Note that `del reader` on the Python side will only decrement a reference counter. You can make an explicit call to the garbage collector (see https://docs.python.org/3/library/gc.html#gc.collect), but that's going to do a full collection of all garbage.

ahmadsharif1: This code was for my own debugging/edification and is turned off by default. A private variable controls it.

I can delete it if you want. It's developer-only code -- not user-facing.

Let me know.

I timed the code btw -- when using pytorch to time it, the aggregate profile doesn't change when I call `del decoder` on it. It's just that if you print the iters, you can see a timing difference once in every few iters.

scotts: I think the benchmark is easier to understand without it, honestly. If we keep it, then we need a comment explaining that we know this would happen anyway when the function exits, and this does not actually cause the GC to happen, but doing this here before the function exists allows us to sometimes see the cost of GC.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: @ahmadsharif1, where is it hardcoded? I was expecting to find it in `core.add_video_stream()`, but I didn't see it there.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -785,8 +824,22 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
   output.durationSeconds = ptsToSeconds(
       getDuration(frame), formatContext_->streams[streamIndex]->time_base);
   if (output.streamType == AVMEDIA_TYPE_VIDEO) {
-    output.frame =
-        convertFrameToTensorUsingFilterGraph(streamIndex, frame.get());
+    if (streamInfo.colorConversionLibrary == ColorConversionLibrary::SWSCALE) {
```
scotts, line 827: Nit: when checking enums, I prefer explicit value checks on all branches:
```
if (val == Kind::A) {
  handle_a(val);
} else if (val == Kind::B) {
  handle_b(val);
} else {
  throw exception("Unhandled Kind: " + to_string(val));
}
```
The reason is that if we ever change the enum values, then we'll fail loudly if there is any place we did not handle it. If we handle a case by default in a bare-else branch, we can easily miss this.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -123,7 +123,13 @@ class VideoDecoder {
   // --------------------------------------------------------------------------
   // ADDING STREAMS API
   // --------------------------------------------------------------------------
-
+  enum ColorConversionLibrary {
+    // TODO: Add an AUTO option later.
+    // Use the libswscale library for color conversion.
```
scotts, line 128: Comment looks wrong; should say something about filtergraph.

test/decoders/test_video_decoder_ops.py
```diff
@@ -322,6 +323,54 @@ def test_frame_pts_equality(self):
             )
             assert pts_is_equal
 
+    @pytest.mark.parametrize("color_conversion_library", ("filtergraph", "swscale"))
+    def test_color_conversion_library(self, color_conversion_library):
+        decoder = create_from_file(str(NASA_VIDEO.path))
+        _add_video_stream(decoder, color_conversion_library=color_conversion_library)
+        frame0, _, _ = get_next_frame(decoder)
```
scotts, line 330: `frame0, *_ = get_next_frame(decoder)` will also work, and is more future-proof.

benchmarks/decoders/benchmark_decoders.py
ahmadsharif1: I removed the del line now

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Done

scotts (approved):

test/convert_image_to_tensor.py
```diff
@@ -16,9 +16,29 @@
...
     # Get base filename without extension
     base_filename = os.path.splitext(img_file)[0]
-    pil_image = Image.open(img_file)
-    img_tensor = torch.from_numpy(np.asarray(pil_image))
-    print(img_tensor.shape)
-    print(img_tensor.dtype)
-    # Save tensor to disk
-    torch.save(img_tensor, base_filename + ".pt", _use_new_zipfile_serialization=True)
+    extension = os.path.splitext(img_file)[1]
+    if extension == ".pt":
```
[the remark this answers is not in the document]
ahmadsharif1: I can revert this but I want this script to convert from image to tensor and from tensor to image. As coded it does the action based on the extension automatically. It's not easy to get wrong -- I find manual options are sometimes easier to get wrong. WDTY?

test/decoders/test_video_decoder_ops.py
ahmadsharif1: Great idea. Done.

src/torchcodec/decoders/_core/VideoDecoder.h
ahmadsharif1: Good catch. Done.

PR #209, opened 2024-09-10
Pass thread count to ffmpeg's filtergraph

romanmeta:
Summary: This ensures that ffmpeg does not create it's own thread pool when asked not to. Otherwise we pay the cost of passing work around between these 2 thread pools.

Differential Revision: D62141413


ahmadsharif1 (approved):
Do not submit the internal diff. I can merge this here once tests pass and it should be imported internally using difftrain.

PR #210, opened 2024-09-11
Adding Mac to S3 non-GPL FFmpeg compile

scotts:
Adds CI workflows so that we build from S3 on Mac arm64 (the architecture for M1 processors). Enabling Mac arm64 builds required changes in several places:

1. Modified `fetch_and_expose_non_gpl_ffmpeg_libs.cmake` to handle Mac OS on arm64. I had to generalize the logic so it can handle both Mac and Linux. This includes pulling the right tar files from S3 and specifying the right library names. (Linux uses .so; Mac uses .dylib.)
2. Added `.github/workflows/macos_wheel.yaml`. The goal for this PR is to get it to the state where it builds from the S3 binaries and runs the Python tests. Subsequent PRs will turn this into the actual wheel building and testing.
3. Renamed `.github/workflows/wheel.yaml` to `linux_wheel.yaml`.
4. Modified `test/utils.py` so that when testing on Macs, we allow an absolute tolerance of 3 when comparing tensors. *This means that we are okay with not having bit-for-bit-equality between Mac and Linux.* Based on what I can find, I think trying to maintain bit-for-bit equality between systems and architectures is unrealistic. We will maintain bit-for-bit equality on Linux x86.
5. Modified `VideoDecoderOps.cpp` to throw a pybind11 exception into one specific to PyTorch. The pybind11 exception was supposed to become a `StopIteration` exception in Python. That worked on Linux, but it segfaulted on Mac. PyTorch does not have an equivalent of `StopIteration`, but it does have one that becomes `IndexError`. I think that's close enough. From what I can tell, PyTorch does not always play nice with pybind11 exceptions; see https://github.com/pytorch/pytorch/issues/34172. I opened Issue #212 to track using PyTorch errors everywhere in the C++ code.


src/torchcodec/decoders/_core/fetch_and_expose_non_gpl_ffmpeg_libs.cmake
```diff
@@ -8,33 +8,150 @@ if (CMAKE_VERSION VERSION_GREATER_EQUAL "3.24.0")
...
 )
+
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    set(
+        platform_url
+        ${base_url}/linux_x86_64
+    )
+
+    set(
+        f4_hash
```
ahmadsharif1, line 24: nit: maybe use a more descriptive name here?

test/utils.py
```diff
@@ -10,12 +11,14 @@
...
-# frames from media because we expect our decoding to exactly match what a user can
-# do on the command line with ffmpeg.
+# For use with decoded data frames. On Linux, we expect exact, bit-for-bit equality. On
+# all other platforms, we allow a small tolerance.
 def assert_tensor_equal(*args, **kwargs):
-    torch.testing.assert_close(*args, **kwargs, atol=0, rtol=0)
+    if sys.platform == "linux":
+        absolute_tolerance = 0
+    else:
+        absolute_tolerance = 3
```
ahmadsharif1, line 20: Is there an issue here? Or are we just accepting that this will always remain here?

Maybe you can add the result of your investigation here so future code readers will know why we did this.

ahmadsharif1 (approved):

test/utils.py
scotts: The issue, I believe, is that FFmpeg does not guarantee bit-for-bit equality across architectures. I believe some version of this will always remain, unless we decide to partition our test tensors by system and architecture.

src/torchcodec/decoders/_core/fetch_and_expose_non_gpl_ffmpeg_libs.cmake
scotts: Sure, let's call it fn_sha256.

test/utils.py
ahmadsharif1: I meant if we have a github issue here for reference. You should put this comment in the code as a comment because more people look at the code than the PR thread. That's what I meant -- add a comment here for future code readers (in the code itself, not the PR comment thread).

test/utils.py
```diff
@@ -10,12 +11,14 @@
 import torch
 
 
-# For use with decoded data frames, or in other instances were we are confident that
-# reference and test tensors should be exactly equal. This is true for decoded data
-# frames from media because we expect our decoding to exactly match what a user can
-# do on the command line with ffmpeg.
+# For use with decoded data frames. On Linux, we expect exact, bit-for-bit equality. On
+# all other platforms, we allow a small tolerance.
```
ahmadsharif1, line 15: Specifically this comment doesn't explain the "why" -- only the "what".

"What" comments aren't as useful for future readers because the "what" can be redundant with the code itself.

test/utils.py
scotts: Yup, I expanded on the comment that was already there.

PR #213, opened 2024-09-18
Build Mac wheels in CI

scotts:
WIP. Current status:

1. Building the wheel from the non-GPL FFmpeg S3 binaries seemingly succeeds. The wheel is successfully uploaded as a part of its job.
2. Downloading the wheel in a subsequent job succeeds.
3. All installations succeed, including the wheel and all runtime dependencies.
4. Smoke test fails. See [a failing job](https://github.com/pytorch/torchcodec/actions/runs/10940949170/job/30374618751?pr=213) for logs.
    * We get a segfault at runtime. I've hacked the C++ code to determine where, exactly, the segfault happens, and it's [a line of code](https://github.com/pytorch/torchcodec/blob/main/src/torchcodec/decoders/_core/VideoDecoder.cpp#L660) that does not seem to be a logic error of any kind.
    * In support of the notion that it's not a logic error, when everything is built and tested in the same environment using the S3 binaries, all tests pass.
    * Memory corruption of some kind seems more likely, and that would indicate there is something wrong with how the wheel is constructed. That is, if it was the binaries alone, I think that would show itself in what we have currently committed, which builds from S3 binaries and tests in the same job.
    * I cannot find anything obvious when I dig into the libraries in the wheel. All of the FFmpeg libraries are loaded dynamically, and it's very strange that we fail at this point in the code, as opposed to when we try to call an FFmpeg function.


scotts:
Closing in favor of PR #230. I accidentally did this work directly on main; #230 moves the changes to a different branch.

Issue #203, opened 2024-09-06
Docs CI is broken


scotts:
### 🐛 Describe the bug

Our CI for docs is broken. See example: https://github.com/pytorch/torchcodec/actions/runs/10744233389/job/29800698755?pr=202. The failure is:
```
generating gallery for generated_examples... [100%] basic_example.py
Warning, treated as error:

../../examples/basic_example.py unexpectedly failed to execute correctly:

    Traceback (most recent call last):
      File "/home/runner/work/torchcodec/torchcodec/examples/basic_example.py", line 66, in <module>
        decoder = SimpleVideoDecoder(raw_video_bytes)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
      File "/home/runner/work/torchcodec/torchcodec/src/torchcodec/decoders/_simple_video_decoder.py", line 1[31](https://github.com/pytorch/torchcodec/actions/runs/10743607733/job/29798699894?pr=201#step:8:32), in __init__
        core.scan_all_streams_to_update_metadata(self._decoder)
      File "/usr/share/miniconda3/envs/test/lib/python3.12/site-packages/torch/_ops.py", line 667, in __call__
        return self_._op(*args, **kwargs)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^
    NotImplementedError: Could not run 'torchcodec_ns::scan_all_streams_to_update_metadata' with arguments from the 'CPU' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build). If you are a Facebook employee using PyTorch on mobile, please visit https://fburl.com/ptmfixes for possible resolutions. 'torchcodec_ns::scan_all_streams_to_update_metadata' is only available for these backends: [CUDA, Meta, BackendSelect, Python, FuncTorchDynamicLayerBackMode, Functionalize, Named, Conjugate, Negative, ZeroTensor, ADInplaceOrView, AutogradOther, AutogradCPU, AutogradCUDA, AutogradXLA, AutogradMPS, AutogradXPU, AutogradHPU, AutogradLazy, AutogradMeta, Tracer, AutocastCPU, AutocastXPU, AutocastCUDA, FuncTorchBatched, BatchedNestedTensor, FuncTorchVmapMode, Batched, VmapMode, FuncTorchGradWrapper, PythonTLSSnapshot, FuncTorchDynamicLayerFrontMode, PreDispatch, PythonDispatcher].
```
[...]

PR #216, opened 2024-09-20
Fix doc build by using local video

scotts:
We've been hitting a 403 in our docs build. That is not initially why we opened Issue #203, but it does seem to be the current problem. Building the docs in CI and locally hits a 403 on the video. Rather than trying to debug that, I think it's easier and more robust to always use a local version of the video.


NicolasHug (commented):
I think we need to be mindful of the repo's size. We're already storing a bunch of videos for testing and adding a new one just for docs may not be necessary.

Over time and with more history, repos can become extremely large (x GBs), and that can create friction. Should we just try to find another video?

scotts:
@NicolasHug, I understand your concern. In this case, I think it's worth it.

The problem I encountered was that the video _is_ available. I can download it locally on my Mac and my devserver through `wget`. It was just failing (getting a 403) during the doc build, both in CI and on my devserver. There's probably something about the environment of the sphinx build that, somehow, is not okay with doing a download. So I don't think we can quickly fix this problem by using a different video - I believe the root problem is that we're trying to grab a video from the internet. I'm afraid that figuring out why we can't download something during our sphinx build will require a deep investigation.

I also think it's not great that our docs build is non-determistic. I think it leads to these kinds of problems, and I'm worried we'll constantly play whack-a-mole trying to keep them healthy.

NicolasHug:
I'll give this a look next week, but I don't think we should vendor the video in the repo. This forces users to manually download the video locally to run the example, and prevents them from running this example on colab. That's not a pleasant or convenient first experience with torchcodec, so I think it's important to address with care.

scotts:
@NicolasHug, if you think you can keep the docs build robust while still downloading a video, that's great! I'm going to continue using this draft PR to see if I can fix the original reason we opened #203. Now that we're no longer getting a 403, I can see that the CI is failing with the same problem as before.

scotts:
@NicolasHug, note that I changed the PyTorch install to also include `torchvision` and `torchaudio`. That is, I copied it exactly from https://pytorch.org/get-started/locally/. That seemed to make the problem go away. If you can pair that change with your future fix, we should be able to get the docs building in CI again.

scotts:
Abandoning in favor of @NicolasHug's fix in #217.

PR #217, opened 2024-09-23
Fix video download in basic example

NicolasHug:
Closes https://github.com/pytorch/torchcodec/pull/216
Partially addresses https://github.com/pytorch/torchcodec/issues/203

Looks like pexels started requiring the GET request to have a "User-Agent" header field. The solution was inspired from https://stackoverflow.com/questions/38489386/how-to-fix-403-forbidden-errors-when-calling-apis-using-python-requests.

I prefer going for this solution rather than https://github.com/pytorch/torchcodec/pull/216, but happy to reconsider if downloading the video keeps bringing flakyness to our CI.

### DIFF
--- a/examples/basic_example.py
+++ b/examples/basic_example.py
@@ -26,7 +26,7 @@ import requests
 # Video source: https://www.pexels.com/video/dog-eating-854132/
 # License: CC0. Author: Coverr.
 url = "https://videos.pexels.com/video-files/854132/854132-sd_640_360_25fps.mp4"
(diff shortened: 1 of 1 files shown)


NicolasHug:
Doc build job is still failing due to pre-existing issue mentioned in https://github.com/pytorch/torchcodec/issues/203, but that's to be expected.

I can confirm that building the docs locally fails on `main` but succeeds with this PR.

scotts (approved):

PR #218, opened 2024-09-23
Docs CI build installs PyTorch using official command

scotts:
The Docs build was installing PyTorch but not TorchVision and TorchAudio. The official instructions say to install both of them, see: https://pytorch.org/get-started/locally/. This seems to fix our docs build. This PR, with #217, should resolve Issue #203.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -28,7 +28,9 @@ jobs:
         run: python -m pip install --upgrade pip
       - name: Install dependencies and FFmpeg
         run: |
-          python -m pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cpu
+          # TODO: torchvision and torchaudio shouldn't be needed. They were only added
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
It's weird that this is fixing anything. The official installation instructions also specify `torchvision` and `torchaudio`, but that's not because they're pre-requisite for pytorch to work. `pip install torch` on its own is expected to work fine. They're only part of the official install instructions because the assumption is that users will need those anyway, and they need to be installed at the same time of torch for them to run.

There is likely an underlying problem upstream with pytorch. That being said, I don't think we have enough resources to figure out a proper solution :( so I'll approve. I can confirm too that the torch version is exactly the same with this PR as in `main`, which is currently failing.

In order to avoid having bloated steps that we don't understand, please just add a TODO comment e.g.

```
# TODO: torchvision and torchaudio shouldn't be needed. They were only added
# to silence an error as seen in https://github.com/pytorch/torchcodec/issues/203
```

PR #221, opened 2024-09-24
Add Random index-based clip sampler - part 1

NicolasHug:
This PR adds a random index-based clip sampler.

Leaving the following for other PRs:

- Edge case policy support, i.e. what to do when there aren't enough frames to sample? Right now, we raise a loud error.
- Docs


src/torchcodec/samplers/__init__.py
```diff
@@ -0,0 +1 @@
+from ._implem import clips_at_random_indices
```
NicolasHug, line 1: I'm creating a new `torchcodec.samplers` namespace, keeping the existing `torchcodec._samplers` namespace untouched to avoid conflicts on internal codebase.

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,62 @@
+from typing import List
+
+import torch
+
+from torchcodec.decoders import (  # TODO: move FrameBatch to torchcodec.FrameBatch?
```
NicolasHug, line 5: @ahmadsharif1 @scotts  Thoughts?

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,62 @@
...
+    # num_indices_between_frames = 1, clip = ffff      , span = 4
+    # num_indices_between_frames = 2, clip = fxfxfxf   , span = 7
+    # num_indices_between_frames = 3, clip = fxxfxxfxxf, span = 10
+    clip_span = num_indices_between_frames * (num_frames_per_clip - 1) + 1
+
+    # TODO: We should probably not error.
+    if clip_span > len(decoder):
+        raise ValueError(
+            f"Clip span ({clip_span}) is larger than the number of frames ({len(decoder)})"
+        )
```
NicolasHug, line 43: @ahmadsharif1 @scotts 

I am errorring here out of convenience, but I'm not sure this leads to the best UX. I think we should try as hard as we can to return something, even if it may not respect `num_indices_between_frames` or `num_frames_per_clip`.

Typically, users will apply consistent values for `num_indices_between_frames` and `num_frames_per_clip` across all videos. However, given the significant variability in video lengths, it's probable that any reasonable configuration for these parameters might not work for at least one video in a large dataset. And that would lead to a loud error mid-training, which can be problematic and frustrating.

In general what we want to avoid are silently-wrong result. I don't think there is a risk of silently wrong result here. E.g. if the output shape is *really* wrong, then the model will fail loudly later in the pipeline.

If `clip_span > len(decoder)`,  the strategy I would suggest is to:
- decrease `num_indices_between_frames` and try to return something that respects `num_frames_per_clip`
- if not, `reduce_num_frames_per_clip`

Of course, this should be properly documented.

Looking for your thoughts and input before I start implementing such a thing :)

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,62 @@
...
+
+    last_clip_start_index = len(decoder) - clip_span
+    clip_start_indices = torch.randint(
+        low=0, high=last_clip_start_index + 1, size=(num_clips,)
+    )
+
+    # TODO: This is inefficient as we are potentially seeking backwards.
+    # We should sort by clip start before querying, and re-shuffle.
+    # Note: we may still have to seek backwards if we have overlapping clips.
```
NicolasHug, line 52: @ahmadsharif1 @scotts 

This is the simplest implementation that works, but of course it's not efficient. I'm seeing 2 potential ways of improving irt:

1. What's described in the comment above i.e. sort the clip start, call `get_frames_at`, and re-shuffle the clips by clip start. This makes sure clips are decoded in order, but we may still have to seek backwards if a clip ends *after* the next one starts. I think this can be implemented fairly easily with the current APIs that we have.

2. flatten all frame indices across all clips, sort them, decode all the frames in sorted order, re-order the frames to their original shuffled order, and then re-assemble the frames into clips (this is easy since we know all clips are of the same size). I suspect this is the fastest and most efficient. I think this can be done with our current APIs, but it may benefit from a new `decode_frames_at_those_indices(indices)` API.

LMK you thoughts on the proposed strategies, and if you have any alternative suggestion.

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,62 @@
...
+    if num_clips <= 0:
+        raise ValueError(f"num_clips ({num_clips}) must be strictly positive")
+    if num_frames_per_clip <= 0:
+        raise ValueError(
+            f"num_frames_per_clip ({num_frames_per_clip}) must be strictly positive"
+        )
+    if num_indices_between_frames <= 0:
+        raise ValueError(
+            f"num_indices_between_frames ({num_indices_between_frames}) must be strictly positive"
+        )
```
scotts, line 27: Is it possible for our decoder to have length 0? That is, we have a video file with metadata and no frames? If yes, we should probably check that here. That is the one condition where it will not be possible for us to respect `num_frames_per_clip`, because there are no frames at all.

src/torchcodec/samplers/_implem.py
scotts: Agreed that this should not be an error condition for the reasons you describe.

The behavior that I think users will want might be annoying to implement. I think we have to respect `num_frames_per_clip`, even if it means we produce duplicate frames. Returning back clips that always respect `num_frames_per_clip` is, I think, one of the main values of the sampler. It prevents users from having to implement this annoying logic to make sure they have the right data shape for their model inputs. What I think that we should return:

* As many frames as possible that obey the users constraints. Let's call this number of frames `possible`.
* The remaining frames, `num_frames_per_clip - possible`, will be duplicates from within that clip. Users should be able to control this behavior with a policy. I can think of two reasonable policies to start with: `REPEAT_LAST_FRAME` and `REPEAT_WHOLE_CLIP`.

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,62 @@
...
+    if clip_span > len(decoder):
+        raise ValueError(
+            f"Clip span ({clip_span}) is larger than the number of frames ({len(decoder)})"
+        )
+
+    last_clip_start_index = len(decoder) - clip_span
+    clip_start_indices = torch.randint(
+        low=0, high=last_clip_start_index + 1, size=(num_clips,)
+    )
+
```
scotts, line 49: (I'm putting the comment here so that we can still read the logic without a giant textbox in the middle of it.)

I'm not sure if we want to compute and then use `last_clip_start_index`. From a probability perspective, I think we're actually making it much less likely that the frames in the range `[last_clip_start_index, len(decoder))` are ever returned to the user. For all other frames, consider that they could be the start of a clip, or included in a clip. For last set of frames, they're only included if we happen to select `last_clip_start_index` as a clip start.

An alternative approach is that we don't do this, and just set `low=0, high=len(decoder)`. Then we let our clip-too-small policy from above handle the case when the clip start is is in the range `len(decoder) - clip_span`.

Both options result in some quirky behavior. What is currently implemented makes it less likely (I think) that we'll ever return frames in the range `[last_clip_start_index, len(decoder))`. My suggestion above means that selecting those frames is just as likely as the frames in the rest of the video, but we'll always have some duplicates - and maybe biasing the earlier part of the video is better than having duplicates. I'm not sure, sampling is actually a hard problem. `¯\_(ツ)_/¯`

src/torchcodec/samplers/_implem.py
scotts: I think Option 1 is a reasonable optimization that is quick and intuitive enough to do for a first implementation.

Option 2 is something we can do if we discover Option 1 is too slow. The C++ API will probably be generally useful outside of our sampler - that is, for users who have their own bespoke sampler. (I have seen one instance of this in the wild.) I believe we would require the decoder to be scanned for this behavior, of course deferring to @ahmadsharif1's understanding.

Note that `SimpleVideoDecoder` always scans the file - although I think our future decoders may not always scan the file, so we should still consider this case. (Getting a proposal up for `SingleStreamVideoDecoder` is one of my TODOs for this week.)

src/torchcodec/samplers/_implem.py
ahmadsharif1: The default should probably be to not error, but at the user's request (maybe as a parameter), there should be a strict mode where it errors out if there aren't enough frames to satisfy the requirements of the user.

src/torchcodec/samplers/_implem.py
ahmadsharif1: I think uniformly sampling the clip_starts in the range where they are valid matches my own intuition.

Duplicate frames may not be good for training.

If @scotts has a strong opinion, maybe we can expose an option but the current behavior seems good to me as the default.

src/torchcodec/samplers/_implem.py
ahmadsharif1: For the fastest implementation, you would need to flatten, sort and dedupe frames, then ask the decoder to decode in that order, and then fill in the FrameBatch.

I think torch.rand can return duplicate values and even without duplicate frame starts, based on dilation you can get duplicate frames. You want to decode them just once, for the fastest performance (since we don't cache them in the C++ code).

Seeking backwards is slow.

I think even getting frame count reliably is hard without scanning the file. At least I don't know of a reliable way that works 100% of the time without scanning.

However if the frame count is in the header and you want to trust it for a fast sampler you could do that from the core API.

src/torchcodec/samplers/_implem.py
NicolasHug: I'll do it. We'll have to move the `FrameBatch` implementation within `torchcodec._something`, otherwise this is creating cyclic imports. I'll create a PR to tackle that immediately after this one is merge. I prefer not doing it here to avoid complicating the PR and the reviews.

src/torchcodec/samplers/_implem.py
NicolasHug: After chatting offline we concluded the best way to handle this was to let the user choose the sampling range, similarly to the "equaly_spaced_in_time" sampler.

src/torchcodec/samplers/_implem.py
NicolasHug: We decided to go with option 1 for now as it's simpler to implement, and immediately follow-up with option 2.

src/torchcodec/samplers/_implem.py
NicolasHug: We chatted offline, we'll introduce a policy parameter and potentially another boolean flag, this is still TBD. Let me resolve this thread to clear up the long list of open comments.

test/conftest.py
```diff
@@ -0,0 +1,22 @@
+import random
+
+import pytest
+import torch
+
+
+@pytest.fixture(autouse=True)
+def prevent_leaking_rng():
+    # Prevent each test from leaking the rng to all other test when they call
+    # torch.manual_seed() or random.seed().
```
NicolasHug, line 10: This is taken from torchvision

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
+import random
+from typing import List, Optional
+
+import torch
+
+from torchcodec.decoders import FrameBatch, SimpleVideoDecoder
+
+
```
NicolasHug, line 8: Below I factored-out some validation logic into functions. This isn't strictly necessary for this PR, but we'll need the exact same logic for the `clips_at_regular_indices()` sampler. So I prefer extracting the logic now, so that we can minimize the diff changes of the `clips_at_regular_indices()` PR.

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
...
+def _validate_sampling_range(
+    *, sampling_range_start, sampling_range_end, num_frames, clip_span
+):
+    if sampling_range_start < 0:
+        sampling_range_start = num_frames + sampling_range_start
+
+    # TODO: or max(sampling_range_start, num_frames - 1)?
+    sampling_range_start = sampling_range_start % num_frames
```
NicolasHug, line 36: Looking for input on this. I'm in a "let's try to avoid mid-train crashes as much as possible" mindset.

test/conftest.py
scotts: My understanding of what this is doing, and why it works:

1. Because we have `autouse=True`, this fixture is used for _all_ tests, including our decoder tests.
2. As this is autoused, we execute lines 8--15 before all tests. That effectively saves the current RNG state for torch, random and torch.cuda.
3. We yield on line 17, which lets the actual test execute. That test may itself change the state of the above.
4. After the test runs, we return back on line 19, and restore the state we captured above.

Is my understanding correct?

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
+import random
```
ahmadsharif1, line 1: random nit: why not call it implementation.py? It's not user-facing, so we can have long descriptive names here

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
...
+        raise ValueError(
+            f"num_indices_between_frames ({num_indices_between_frames}) must be strictly positive"
+        )
+
+
+def _validate_sampling_range(
+    *, sampling_range_start, sampling_range_end, num_frames, clip_span
+):
+    if sampling_range_start < 0:
+        sampling_range_start = num_frames + sampling_range_start
```
ahmadsharif1, line 33: what if it's still negative after this?

src/torchcodec/samplers/_implem.py
ahmadsharif1: why aren't we doing the same for sampling_range_end if it's not None?

Maybe you can extract out a function like _coerce_index() or something and apply it for both start and end?

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
...
+        )
+        for clip_start_index in clip_start_indices
+    ]
+
+    # This an ugly way to shuffle the clips using pytorch RNG *without*
+    # affecting the python builtin RNG.
+    builtin_random_state = random.getstate()
+    random.seed(torch.randint(0, 2**32, (1,)).item())
+    random.shuffle(clips)
+    random.setstate(builtin_random_state)
```
ahmadsharif1, line 136: I wonder if there is a contextmanager for this? I couldn't find it myself but maybe there is a way using the `with` statement here that auto saves/restores?

test/samplers/test_samplers.py
```diff
@@ -0,0 +1,202 @@
...
+    # Call with a different seed, expect different results
+    torch.manual_seed(1)
+    clips_3 = clips_at_random_indices(decoder, num_clips=num_clips)
+    with pytest.raises(AssertionError, match="Tensor-likes are not"):
+        assert_tensor_equal(clips_1[0].data, clips_3[0].data)
+
+    # Make sure we didn't alter the builtin Python RNG
+    builtin_random_state_end = random.getstate()
+    assert builtin_random_state_start == builtin_random_state_end
+
```
ahmadsharif1, line 148: I wonder if we can fuzz test this to make sure we don't return errors mid-training.

Can pytest help with passing in random values to the parameters (with certain constraints like positive-only values, etc. for things like frames_per_clip, etc.) to make sure we are robust to errors?

test/samplers/test_samplers.py
```diff
@@ -0,0 +1,202 @@
...
+        sampling_range_start=sampling_range_start,
+        sampling_range_end=sampling_range_end,
+    )
+
+    cm = (
+        contextlib.nullcontext()
+        if assert_all_equal
+        else pytest.raises(AssertionError, match="Tensor-likes are not")
+    )
+    with cm:
```
ahmadsharif1, line 81: A comment would be useful here. I don't understand the purpose or even function of what this is doing (as a non-python user).

test/samplers/test_samplers.py
```diff
@@ -0,0 +1,202 @@
...
+    ),
+)
+def test_random_sampler_range(
+    sampling_range_start, sampling_range_end, assert_all_equal
+):
+    # Test the sampling_range_start and sampling_range_end parameters by
+    # asserting that all clips are equal if the sampling range is of size 1,
+    # and that they are not all equal if the sampling range is of size 2.
+
+    # Since this has a low but non-zero probability of failing, we hard-code a
```
ahmadsharif1, line 62: Can you document under what scenario we fail here?

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
...
+
+def clips_at_random_indices(
+    decoder: SimpleVideoDecoder,
+    *,
+    num_clips: int = 1,
+    num_frames_per_clip: int = 1,
+    num_indices_between_frames: int = 1,
+    sampling_range_start: int = 0,
+    sampling_range_end: Optional[int] = None,  # interval is [start, end).
+) -> List[FrameBatch]:
```
ahmadsharif1, line 81: Did we agree to return a List here? I thought some users had a need to get a single tensor back? Would those users just stack it manually? Note that stacking could be an expensive operation.

src/torchcodec/samplers/_implem.py
scotts: The question as I understand it is: what should the behavior be when the sampling start is greater than the total number of frames? I'm inclined to make this an error for now, for the following reasons:

1. None of the non-failure options (mod it or max it with number of frames) seem intuitive to me.
2. We already have a resilient way to do the right thing: _don't_ provide a start and end.
3. It's easy to relax what was once an error condition, but it's hard to change the behavior of non-error conditions.
4. This feels tied to the problem of what to do when there are not enough frames to meet the required frames per clip. We know we're going to do something there, but we're currently making it an error. When we have a mechanism to solve that problem, we could use it here, too.

src/torchcodec/samplers/_implem.py
```diff
@@ -0,0 +1,138 @@
...
+        for clip_start_index in clip_start_indices
+    ]
+
+    # This an ugly way to shuffle the clips using pytorch RNG *without*
+    # affecting the python builtin RNG.
+    builtin_random_state = random.getstate()
+    random.seed(torch.randint(0, 2**32, (1,)).item())
+    random.shuffle(clips)
+    random.setstate(builtin_random_state)
+
```
scotts, line 137: Why is it necessary that we don't affect the random module's state? I assume we're trying not to affect the random values seen by models during actual training, but why do we want to do that?

src/torchcodec/samplers/_implem.py
NicolasHug: When we discussed this, we decided to "let the implementation decide".
I do agree that stacked tensors make more sense in general, but we also don't want to stack on behalf of the user as it may introduce an unnecessary copy.
The current implementation leads to a list, but it's possible by implementing the "better frame shuffling" strategy (left as a TODO at the bottom of this function), we'll end up with a stack. This is still TBD.

src/torchcodec/samplers/_implem.py
NicolasHug: That's typically the kind of scenario where a CM can be useful indeed. I thought about writing one, but decided against it considering this logic is just 2 lines of codes, and we'll likely remove it when we address the TODO.

test/conftest.py
NicolasHug: Yup, that's it. This is a use of pytest' fixtures as a scoped setup/teardown mechanism.

test/samplers/test_samplers.py
NicolasHug: There is the `hypothesis` package that I think does what you're suggesting, but I'm not sure how useful it would be in this specific instance? What do you have in mind with 

> make sure we don't return errors mid-training.

?

src/torchcodec/samplers/_implem.py
NicolasHug: OK, let's just error for now and stay open for an automated behavior upon user request.

NicolasHug: > why aren't we doing the same for sampling_range_end if it's not None?

Good catch, thanks. I think we should cap it to `num_frames`, as this is a fairly common strategy for the upper bound (as in slice indexing).

src/torchcodec/samplers/_implem.py
NicolasHug: It would result in an error because `sampling_range_end <= sampling_range_start` .

If we don't error, we can either:

- decide that it's equivalent to 1 (so that it can work when sampling_range_start is 0)
- keep wrapping-around (modulo).

Neither sounds like a great intuitive option, so I it might be best to error, similar to what we decided in https://github.com/pytorch/torchcodec/pull/221/files#r1781347650

src/torchcodec/samplers/_implem.py
NicolasHug: I personally don't find `implementation` substantially more descriptive than `implem`.
I'm OK to address, but only at the very end before merging, because the renaming would cause GitHub to mark ongoing comments as "outdated".

src/torchcodec/samplers/_implem.py
scotts: Right, I'm on board with not hardcoding a seed. I guess what I'm confused about is why we can't just call `random.shuffle(clips)` on its own. Why do we need to set a seed before shuffling? (Without that need, I assume it's okay to _consume_ values in the RNG stream.)

NicolasHug: We're a pytorch library, so in general we want our RNG stream to come from pytorch, not from other RNGs.
In other words we want `torch.manual_seed(123)` to affect the RNG of the sampler.

scotts (approved):

src/torchcodec/samplers/_implem.py
scotts: Ohhhh, got it, so we're porting over torch's RNG state to random. Since your last response was the aha moment for me, let's say something to that effect in a comment. :)

test/samplers/test_samplers.py
ahmadsharif1: I was requesting fuzz testing so we can be robust to errors so we don't return errors in the middle of training (which is what your original philosophy was).

image

Now if we are switching to errors mindset, you do seem to have tests for exact errors, so you can resolve this comment.

src/torchcodec/samplers/_implem.py
ahmadsharif1: Decoding frames to a single tensor doesn't involves a copy -- and that can be done for max speed if we want.

Stacking them after the fact is inefficient.

PR #223, opened 2024-09-24
Update install instructions to explicitly mention PyTorch 2.4

scotts:
We did not explicitly mention that we need PyTorch 2.4. It was only implied by telling folks to get the latest stable version of PyTorch.

I'm also doing some minor editing on the text to list requirements first, then ending with how to install TorchCodec.

### DIFF
--- a/docs/source/install_instructions.rst
+++ b/docs/source/install_instructions.rst
@@ -5,27 +5,27 @@ Installation Instructions
     TorchCodec is only available on Linux for now. We plan to support other
     platforms in the future.
 
-Installing torchcodec should be as simple as:
+There are three steps to installing TorchCodec:
 
-.. code:: bash
+1. Install the latest stable version of PyTorch following the
+   `official instructions <https://pytorch.org/get-started/locally/>`_. TorchCodec
+   requires `PyTorch 2.4 <https://pytorch.org/docs/2.4/>`_.
 
-    pip install torchcodec
+2. Install FFmpeg, if it's not already installed. Your Linux distribution probably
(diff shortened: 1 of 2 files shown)


docs/source/install_instructions.rst
```diff
@@ -5,18 +5,13 @@ Installation Instructions
...
-    pip install torchcodec
-
-You will need a working PyTorch installation, which you can install following
-the `official instructions <https://pytorch.org/get-started/locally/>`_.
-
-You will also need FFmpeg installed on your system, and TorchCodec decoding
+You will also need FFmpeg installed on your system. TorchCodec's decoding
 capabilities are determined by your underlying FFmpeg installation. There are
-different options to install FFmpeg e.g.:
+different options to install FFmpeg, e.g.:
```
ahmadsharif1, line 14: Instead of converting a list to English, it would be more compact and easy to read something like this:

TorchCodec installation steps:

1. Install pytorch using ...
2. Install ffmpeg using ...
3. Install torchcodec using `pip install torchcodec`

Seems more compact and readable.

ahmadsharif1 (approved):
What you wrote is okay with me as well, but as a person who likes to skim, I always appreciate terse instructions that get to the point quickly without a lot of words. Instead of "first" "then", etc. I prefer numbered lists.

docs/source/install_instructions.rst
scotts: I did that initially, then reverted it because it felt like too much ceremony. I had tried sub-subsections. I'm not sure how well it will work having code blocks in numbered lists. Let me try and see how it renders.

Issue #215, opened 2024-09-19
SimpleVideoDecoder should support numpy integers


scotts:
### 🚀 The feature

Currently, `SimpleVideoDeocder` has two ways of getting frames by an index: `__getitem__()` (array access) and `get_frame_at()`. Both currently accept an `int`. We should extend to to numpys' integers: https://numpy.org/doc/stable/user/basics.types.html#relationship-between-numpy-data-types-and-c-data-types. What I think we should probably support:

* [numpy.int32](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int32)
* [numpy.uint32](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uint32)
* [numpy.int64](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.int64)
* [numpy.uint64](https://numpy.org/doc/stable/reference/arrays.scalars.html#numpy.uint64)

The smaller integer sizes are probably (?) not appropriate for indexing into a video, as there will be many videos with more than 2^16 frames.

### Motivation, pitch

Users are likely to use values directly from numpy to index into videos.

PR #228, opened 2024-09-25
Allow numpy integers to be used as indices for SimpleVideoDecoder

scotts:
Addresses Issue #215. The method `SimpleVideoDecoder.__getitem__(self, key)` used to do an explicit check:
```
isinstance(key, int)
```
That would fail when users provided an int-like type that was not actually an int, such as the numpy family of integers: https://numpy.org/doc/stable/reference/arrays.scalars.html#sized-aliases. We can easily handle those cases by instead doing a check for:
```
isinstance(key, numbers.Integral)
```
The key characteristic of [numbers.Integral](https://docs.python.org/3/library/numbers.html#numbers.Integral) is that it can always be turned into an `int` through the `int()` constructor.

Note that we're also adding tests for `SimpleVideoDecoder.get_frame_at(self, index)` and `SimpleVideoDecoder.get_frames_at(self, start, stop, step)` to make sure they can handle numpy integers as well. However, we don't need to change any actual logic, as they just work through duck typing. Because `__getitem__()` needs to dispatch on the key type, we need to do some type inspection. This PR makes that type inspection much more relaxed.

### DIFF
--- a/src/torchcodec/decoders/_simple_video_decoder.py
+++ b/src/torchcodec/decoders/_simple_video_decoder.py
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import dataclasses
+import numbers
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Iterable, Iterator, Literal, Tuple, Union
@@ -186,17 +187,17 @@ class SimpleVideoDecoder:
         )
         return frame_data
 
-    def __getitem__(self, key: Union[int, slice]) -> Tensor:
+    def __getitem__(self, key: Union[numbers.Integral, slice]) -> Tensor:
         """Return frame or frames as tensors, at the given index or range.
 
         Args:
-            key(int or slice): The index or range of frame(s) to retrieve.
+            key(numbers.Integral or slice): The index or range of frame(s) to retrieve.
 
         Returns:
             torch.Tensor: The frame or frames at the given index or range.
         """
-        if isinstance(key, int):
-            return self._getitem_int(key)
+        if isinstance(key, numbers.Integral):
+            return self._getitem_int(int(key))
         elif isinstance(key, slice):
             return self._getitem_slice(key)
 
--- a/test/decoders/test_simple_video_decoder.py
+++ b/test/decoders/test_simple_video_decoder.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import numpy
 import pytest
 import torch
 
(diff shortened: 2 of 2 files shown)


src/torchcodec/decoders/_simple_video_decoder.py
```diff
@@ -186,17 +187,17 @@ def _getitem_slice(self, key: slice) -> Tensor:
...
         Args:
-            key(int or slice): The index or range of frame(s) to retrieve.
+            key(numbers.Integral or slice): The index or range of frame(s) to retrieve.
 
         Returns:
             torch.Tensor: The frame or frames at the given index or range.
         """
-        if isinstance(key, int):
-            return self._getitem_int(key)
+        if isinstance(key, numbers.Integral):
```
ahmadsharif1, line 199: One question: would it be more general to try to cast to an int and seeing if we get an exception instead of explicitly checking numbers.Integral here?

My concern is that the user may pass in a type that's castable to an int but not an instance of numbers.Integral (is that even possible?). Do you know if that is the case?

test/decoders/test_simple_video_decoder.py
```diff
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import numpy
```
ahmadsharif1, line 7: numpy is now a test-time dependency. Do we need to update any docs?

src/torchcodec/decoders/_simple_video_decoder.py
scotts: Casting to an int would _succeed_ if the user provided a float or a string (for example, `decoder[2.3]` or `decoder["3"]`), but we want to fail in those cases. We don't want to check if an object can be castable to an int, because `int(2.3)` and `int("3")` will both succeed. We want to have the same behavior as Python lists, and both `lst[2.3]` and `lst["3"]` result in a TypeError. Rather, we want to check if the object has a conversion to int, which is what numbers.Integral does.

The other option is `if int(key) == key`, but I think that's a little too subtle. I'd rather do this, which I think is the blessed, modern way of checking for exactly what we want.

To answer your question: yes, I believe that is possible. But I think this handles the majority of cases we want to handle.

test/decoders/test_simple_video_decoder.py
scotts: It's already a test time dependency. :) See https://github.com/pytorch/torchcodec/blob/1fffd024cbf12c3e6116f1555c676ff7adc86d40/test/decoders/test_video_decoder_ops.py#L13 and https://github.com/pytorch/torchcodec/blob/1fffd024cbf12c3e6116f1555c676ff7adc86d40/test/utils.py#L9

ahmadsharif1 (approved):

ahmadsharif1 (approved):

PR #229, opened 2024-09-26
Use swscale for color conversion unless we are forced to do otherwise

ahmadsharif1:
Before this diff, we would use filtergraph for color conversion.

After this diff, we prefer to use swscale for color conversion because it is faster.

We fall back to filtergraph if the width is not 32-byte aligned.

If the user forces us to use a specific color-conversion library we use that (i.e. for benchmarks).

Benchmark results:

```
python benchmarks/decoders/benchmark_decoders.py --decoders=tcoptions:color_conversion_library=swscale,tcoptions:color_conversion_library=filtergraph --bm_video_paths=/home/ahmads/jupyter/carmel1.mp4,/home/
ahmads/jupyter/853.mp4
[------------------------------- video=/home/ahmads/jupyter/carmel1.mp4 h264 640x360, 1.3s 30.0fps --------------------------------]
                                                                  |  10 seek()+next()  |  1 next()  |  10 next()  |  create()+next()
1 threads: -------------------------------------------------------------------------------------------------------------------------
      TorchcodecNonCompiled:color_conversion_library=filtergraph  |        28.0        |    13.8    |     24.2    |                 
      TorchcodecNonCompiled:color_conversion_library=swscale      |        21.6        |    11.8    |     18.7    |                 
      TorchcodecNonCompiled                                       |                    |            |             |        11.9     

[------------- video=/home/ahmads/jupyter/853.mp4 h264 3840x2160, 4.988317s 59.94005593469701fps --------------]
                                                                  |  10 seek()+next()  |  1 next()  |  10 next()
1 threads: -----------------------------------------------------------------------------------------------------
      TorchcodecNonCompiled:color_conversion_library=filtergraph  |       1958.3       |   295.5    |    577.0  
      TorchcodecNonCompiled:color_conversion_library=swscale      |       2232.9       |   301.3    |    621.7  

```

### DIFF
--- a/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
+++ b/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
@@ -7,7 +7,6 @@
 #include "src/torchcodec/decoders/_core/FFMPEGCommon.h"
 
 #include <c10/util/Exception.h>
-#include <iostream>
 
 namespace facebook::torchcodec {
 
--- a/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
+++ b/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
@@ -28,14 +28,6 @@ int64_t getDuration(const AVFrame* frame) {
 #endif
 }
 
-bool canSwsScaleHandleUnalignedData() {
-#if LIBSWSCALE_VERSION_MAJOR < 6
-  return false;
-#else
-  return true;
-#endif
-}
-
 AVIOBytesContext::AVIOBytesContext(
     const void* data,
     size_t data_size,
--- a/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
+++ b/src/torchcodec/decoders/_core/FFMPEGCommon.cpp
@@ -7,6 +7,7 @@
 #include "src/torchcodec/decoders/_core/FFMPEGCommon.h"
 
 #include <c10/util/Exception.h>
+#include <iostream>
 
 namespace facebook::torchcodec {
 
@@ -25,6 +26,14 @@ int64_t getDuration(const AVFrame* frame) {
   return frame->pkt_duration;
 #else
   return frame->duration;
+#endif
+}
+
+bool canSwsScaleHandleUnalignedData() {
+#if LIBSWSCALE_VERSION_MAJOR < 6
+  return false;
+#else
+  return true;
 #endif
 }
 
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -424,15 +424,32 @@ void VideoDecoder::addVideoStreamDecoder(
   updateMetadataWithCodecContext(streamInfo.streamIndex, codecContext);
   streamInfo.options = options;
   int width = options.width.value_or(codecContext->width);
-  auto colorConversionLibrary = options.colorConversionLibrary.value_or(
-      ColorConversionLibrary::FILTERGRAPH);
-  bool useFilterGraph =
-      (colorConversionLibrary != ColorConversionLibrary::SWSCALE);
-  if (useFilterGraph) {
+
+  // Use swscale for color conversion by default because it is faster.
+  VideoDecoder::ColorConversionLibrary autoColorConversionLibrary =
+      ColorConversionLibrary::SWSCALE;
+  // However, swscale requires widths to be multiples of 32:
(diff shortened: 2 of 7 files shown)


src/torchcodec/decoders/_core/FFMPEGCommon.cpp
```diff
@@ -7,6 +7,7 @@
 #include "src/torchcodec/decoders/_core/FFMPEGCommon.h"
 
 #include <c10/util/Exception.h>
+#include <iostream>
```
scotts, line 10: Probably a remnant from debugging, although also not horrible to keep in.

ahmadsharif1: Good catch. Removed now.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -424,15 +424,32 @@ void VideoDecoder::addVideoStreamDecoder(
...
   streamInfo.options = options;
   int width = options.width.value_or(codecContext->width);
-  auto colorConversionLibrary = options.colorConversionLibrary.value_or(
-      ColorConversionLibrary::FILTERGRAPH);
-  bool useFilterGraph =
-      (colorConversionLibrary != ColorConversionLibrary::SWSCALE);
-  if (useFilterGraph) {
+
+  // Use swscale for color conversion by default because it is faster.
+  VideoDecoder::ColorConversionLibrary autoColorConversionLibrary =
```
scotts, line 429: The actual logic of what we're doing here is pretty straight-forward, but I had to read it very carefully because of the length of the variable names their similarity to other locals. I think readability would be improved by pulling this into a function:
```
VideoDecoder::ColorConversionLibrary detectColorConversionLibrary(int width) {
  if (width % 32 != 0 && !canSwsScaleHandleUnalignedData()) {
    return ColorConversionLibrary::FILTERGRAPH;
  }
  return ColorConversionLibrary::SWSCALE;
}
```
And then locally:
```
auto colorConversionLibrary =
      options.colorConversionLibrary.value_or(detectColorConversionLibrary());
```

test/decoders/test_video_decoder_ops.py
```diff
@@ -371,6 +383,68 @@ def test_color_conversion_library_with_down_scaling(
...
         assert_tensor_equal(filtergraph_frame0, swscale_frame0)
 
+    @pytest.mark.parametrize(
+        "width_scaling_factor,height_scaling_factor",
+        ((1.31, 1.5), (0.71, 0.5), (1.31, 0.7), (0.71, 1.5), (1.0, 1.0)),
+    )
+    @pytest.mark.parametrize("width", [30, 32, 300])
+    @pytest.mark.parametrize("height", [128])
+    def test_color_conversion_library_with_generated_videos(
+        self, tmp_path, width, height, width_scaling_factor, height_scaling_factor
```
scotts, line 393: TIL that pytest's tmp_path is basically magic. :) https://docs.pytest.org/en/stable/how-to/tmp_path.html

test/decoders/test_video_decoder_ops.py
```diff
@@ -371,6 +383,68 @@ def test_color_conversion_library_with_down_scaling(
...
+        filtergraph_decoder = create_from_file(str(video_path))
+        _add_video_stream(
+            filtergraph_decoder,
+            width=target_width,
+            height=target_height,
+            color_conversion_library="filtergraph",
+        )
+        filtergraph_frame0, _, _ = get_next_frame(filtergraph_decoder)
+
+        auto_decoder = create_from_file(str(video_path))
```
scotts, line 439: Let's add a comment explaining what we're doing here, because it's kinda subtle.

* We consider filtergraph to the reference color conversion library. Whatever results it produces, we trust.
* We want to use swscale as much as possible, because it's faster. However, it does not work with widths are not a multiple of 32.
* We have logic in the core library to detect when we can safely use swscale based on video dimension.
* In this test, we compare filtergraph to whatever the result of our auto detection is. That means we are _intentionally_ comparing filtergraph to itself sometimes.

Note that I _almost_ made a comment that I was surprised we weren't comparing filtergraph to swscale, and then I realized the above.

We may want to pull this comment all the up to the top of the test function header, since it's really the core thing we're testing.

scotts (approved):

test/decoders/test_video_decoder_ops.py
ahmadsharif1: Done.

PR #230, opened 2024-09-27
Build Mac wheels in CI

scotts:
Note that this summary was written a month after opening the PR. As evidenced by the commit history below, getting this in shape has been a journey. A summary of the actual changes we'll commit:

1. A separate workflow for compiling and testing Mac wheels. While it has a lot of overlap with the workflow for Linux wheels, there is still some specialization needed when building the wheel. In particular, we are actually still using the [test-infra@macbuildwheel](https://github.com/pytorch/test-infra/tree/macbuildwheel) branch. Eventually that branch will get merged, and we can use main, just like with Linux. At that point, it will be worth seeing if we can consolidate the workflows and make Mac vs Linux a matrix option. (And maybe CUDA as well, see #272.)
2. Minor changes to the Linux wheel workflow to keep it aligned with Mac wheels.
3. Generalization of some of our packaging scripts to be aware of "Linux" versus "Darwin."
4. The removal of logging calls (`VLOG` and `LOG`) from our C++ code. For reasons that we have not been able to fully explain, we get segfaults in those logging calls only when we ran from a Mac wheel. We do not segfault there when we build and test Mac from source, or in any situation with Linux. That might indicate a problem in PyTorch core. If we want logging back, we can decide how best to it then.


.github/workflows/macos_wheel.yaml
```diff
@@ -24,6 +24,36 @@ defaults:
     shell: bash -l -eo pipefail {0}
 
 jobs:
+  generate-matrix:
+    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@main
```
atalman, line 28:
```suggestion
    uses: pytorch/test-infra/.github/workflows/generate_binary_build_matrix.yml@macbuildwheel
```

.github/workflows/macos_wheel.yaml
```diff
@@ -24,6 +24,36 @@ defaults:
...
+      test-infra-ref: main
+      with-xpu: disable
+      with-rocm: disable
+      with-cuda: disable
+  build:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+    name: Build and Upload Mac wheel
+    uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@main
```
atalman, line 42:
```suggestion
    uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@macbuildwheel
```

.github/workflows/macos_wheel.yaml
```diff
@@ -32,7 +62,12 @@ jobs:
         python-version: ['3.9']
         ffmpeg-version-for-tests: ['4.4.2', '5.1.2', '6.1.1', '7.0.1']
     if: ${{ always() }}
+    needs: build
     steps:
+      - uses: actions/download-artifact@v3
+        with:
+          name: pytorch_torchcodec__${{ matrix.python-version }}_cpu_
+          path: pytorch/torchcodec/dist/
       - name: Setup conda env
```
atalman, line 71: Looks like this step is not working as intended. I get something like this:
```
....
  # To activate this environment, use
  #
  #     $ conda activate test
  #
  # To deactivate an active environment, use
  #
  #     $ conda deactivate
  

setup-miniconda ran successfully
```

However next step:
```
Run python3 -m pip install --upgrade pip
  python3 -m pip install --upgrade pip
  shell: /bin/bash -l -eo pipefail {0}
  env:
    INPUT_RUN_POST: true
    CONDA: /Users/ec[2](https://github.com/pytorch/torchcodec/actions/runs/11260601615/job/31315172014?pr=230#step:4:2)-user/miniconda3
    CONDA_PKGS_DIR: /Users/ec2-user/conda_pkgs_dir
Requirement already satisfied: pip in /Library/Frameworks/Python.framework/Versions/[3](https://github.com/pytorch/torchcodec/actions/runs/11260601615/job/31315172014?pr=230#step:4:3).10/lib/python3.10/site-packages (24.2)
```

It picks python 3.10

.github/workflows/macos_wheel.yaml
```diff
@@ -69,9 +99,30 @@ jobs:
 
           conda install "ffmpeg=${{ matrix.ffmpeg-version-for-tests }}" -c conda-forge
           ffmpeg -version
+
+      - name: Install test dependencies
```
atalman, line 103: For this step I suggest we onboard to validation framework:
https://github.com/pytorch/test-infra/blob/main/.github/workflows/validate-domain-library.yml

This is exactly the use case we wrote validation framework for:
https://github.com/pytorch/test-infra/wiki/Validation-Framework

.github/workflows/macos_wheel.yaml
```diff
@@ -32,46 +62,118 @@ jobs:
...
       - name: Install ffmpeg, post build
         run: |
-          # Ideally we would have checked for that before installing the wheel,
-          # but we need to checkout the repo to access this file, and we don't
-          # want to checkout the repo before installing the wheel to avoid any
-          # side-effect. It's OK.
-          source packaging/helpers.sh
-          assert_ffmpeg_not_installed
-
           conda install "ffmpeg=${{ matrix.ffmpeg-version-for-tests }}" -c conda-forge
```
atalman, line 164: Here all conda install need to be done via ${CONDA_RUN}

atalman:
Hi @scotts 
Looks like problem with the build 
```
pip install TorchCodec-0.0.4.dev0-cp39-cp39-macosx_11_0_arm64.whl
pip install --pre torchvision --index-url https://download.pytorch.org/whl/nightly/cpu
conda install "ffmpeg=4.4.2" -c conda-forge
```
and then executing smoke tests:
```
python test/decoders/manual_smoke_test.py  
zsh: segmentation fault  python test/decoders/manual_smoke_test.py
```

.github/workflows/linux_wheel.yaml
```diff
@@ -34,11 +35,13 @@ jobs:
       with-xpu: disable
       with-rocm: disable
       with-cuda: disable
+      build-python-only: "disable"
```
scotts, line 38: I learned about this option from @ahmadsharif1. I turned it on for both Linux and Mac, as I think it fits our use-case where we're also building C++ libraries.

.github/workflows/linux_wheel.yaml
```diff
@@ -121,7 +123,7 @@ jobs:
           ls
       - name: Smoke test
         run: |
-          python test/decoders/manual_smoke_test.py
+          python -X faulthandler test/decoders/manual_smoke_test.py
```
scotts, line 126: In case we run into any problems again on Linux or Mac, I think running the smoke test with the faulthandler on will be helpful.

.github/workflows/macos_wheel.yaml
```diff
@@ -24,15 +24,53 @@ defaults:
...
+
+  build:
+    needs: generate-matrix
+    strategy:
+      fail-fast: false
+    name: Build and Upload Mac wheel
+    uses: pytorch/test-infra/.github/workflows/build_wheels_macos.yml@macbuildwheel
+    with:
+      repository: pytorch/torchcodec
+      ref: ""
```
ahmadsharif1, line 48: Does this build the wheel against nightly?

I noticed we are testing against nightly so probably have to build against nightly too?

.github/workflows/macos_wheel.yaml
```diff
@@ -45,33 +83,28 @@ jobs:
...
-          # Ideally we would have checked for that before installing the wheel,
-          # but we need to checkout the repo to access this file, and we don't
-          # want to checkout the repo before installing the wheel to avoid any
-          # side-effect. It's OK.
-          source packaging/helpers.sh
-          assert_ffmpeg_not_installed
-
-          conda install "ffmpeg=${{ matrix.ffmpeg-version-for-tests }}" -c conda-forge
-          ffmpeg -version
+          rm -r src/
```
NicolasHug, line 103: Nit: either copy the comment made in the linux file https://github.com/NicolasHug/torchcodec/blob/d43dd9120d227ad922079830c05b3c8fa3662de4/.github/workflows/linux_wheel.yaml#L110-L119 or just point to it, to explain why we do this?

Once we merge this PR and when we have the CUDA wheel builds, it'll be time to refactor the 3 separate scripts into something more centralized

.github/workflows/macos_wheel.yaml
NicolasHug: Yes, I think the `build_wheels_macos` job of test-infra is setup to build against nightly by default, and we can see from the logs that nightly is pickedup

NicolasHug (approved):
What a ride. Thank you so much @scotts for pushing through!
Nothing blocking from me, just nits and one Q

.github/workflows/macos_wheel.yaml
```diff
@@ -24,15 +24,53 @@ defaults:
...
+      test-infra-repository: pytorch/test-infra
+      test-infra-ref: macbuildwheel
+      build-matrix: ${{ needs.generate-matrix.outputs.matrix }}
+      post-script: packaging/post_build_script.sh
+      smoke-test-script: packaging/fake_smoke_test.py
+      runner-type: macos-m1-stable
+      package-name: torchcodec
+      trigger-event: ${{ github.event_name }}
+      build-platform: "python-build-package"
+      build-command: "BUILD_AGAINST_ALL_FFMPEG_FROM_S3=1 ${CONDA_RUN} python3 -m build --wheel -vvv --no-isolation"
```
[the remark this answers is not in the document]
scotts: No, it shouldn't still be needed. Let me remove and check. If it is still needed, I'll make a comment.

ahmadsharif1 (approved):

.github/workflows/macos_wheel.yaml
scotts: I don't know, but this code is common between Mac and Linux, and was inherited from Linux. Let's investigate that as a follow-up.

PR #233, opened 2024-10-01
Rename SimpleVideoDecoder to VideoDecoder

scotts:
Renaming `SimpleVideoDecoder` to `VideoDecoder`. The rationale of "simple" was that we wanted an obvious entry point for users who were unfamiliar with video encoding structure. There will be many options on a full video decoder that are going to introduce concepts such users will not be familiar with. We now think we can provide reasonable defaults for all of those options, and those users can just as easily use the same class as power users.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -18,10 +18,10 @@ body:
       # All necessary imports at the beginning
       import torch
       import torchcodec
-      from torchcodec.decoders import SimpleVideoDecoder
+      from torchcodec.decoders import VideoDecoder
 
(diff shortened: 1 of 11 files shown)


NicolasHug (approved):
FYI there are 2 instances left:

```
~/dev/torchcodec (rename_simple) » git grep -ni SimpleVideoDecoder               nicolashug@nicolashug-fedora-PF2MM9CP
.github/ISSUE_TEMPLATE/bug-report.yml:21:      from torchcodec.decoders import SimpleVideoDecoder
.github/ISSUE_TEMPLATE/bug-report.yml:24:      decoder = SimpleVideoDecoder("path/to/video.mp4")  # Help! This fails!
```

Approving now to unblock as @ahmadsharif1 also shared his agreement

PR #234, opened 2024-10-02
Added skeleton code for cuda builds

ahmadsharif1:
This diff does not implement CUDA decoding at all. It only provides the skeleton for CUDA decoding.

Specifically:

1. It adds device as a parameter to the private _core API.
2. In the C++ code: It adds the concept of a device API (DeviceInterface.h) that CUDA or other devices can implement. It also adds a CPUOnlyDevice.cpp file for the case where no device is selected.
3. setup.py and hence pip now accepts an env var called ENABLED_CUDA. Set it to 1 to enable CUDA.
4. ENABLE_CUDA is passed down to CMakeLists.txt where it can include or exclude files from the compilation process.
5. Add a decorator called @needs_gpu that, if set, will skip tests if CUDA is not enabled at runtime.
6. it adds a basic CI for GPU. Right now it builds torchcodec with ENABLE_CUDA=1 and runs pytests.

Note that the GPU CI is the only place where @needs_cuda tests are run. They are skipped outside of that.

This PR by itself is large enough that I have excluded the actual implementation from it. The next diff will actually implement GPU decoding.


test/decoders/test_video_decoder_ops.py
```diff
@@ -39,6 +39,16 @@
 torch._dynamo.config.capture_dynamic_output_shape_ops = True
 
 
+def needs_cuda(test_item):
```
NicolasHug, line 42: I suspect we'll need it in other places, so move it to https://github.com/NicolasHug/torchcodec/blob/main/test/utils.py ?

Also please add small description comment e.g.

```suggestion
def needs_cuda(test_item):
    # Decorator for skipping CUDA tests when CUDA isn't available
```

src/torchcodec/decoders/_core/video_decoder_ops.py
```diff
@@ -135,6 +136,7 @@ def add_video_stream_abstract(
     num_threads: Optional[int] = None,
     dimension_order: Optional[str] = None,
     stream_index: Optional[int] = None,
+    device: Optional[str] = None,
```
NicolasHug, line 139: Eventually we'll want to accept a `torch.device` object as well as a str, but I assume that's for later?

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -0,0 +1,17 @@
+#include <torch/types.h>
+
+namespace facebook::torchcodec {
+
+void maybeInitializeDeviceContext(const torch::Device& device) {
+  if (device.type() == torch::kCPU) {
+    return;
+  } else if (device.type() == torch::kCUDA) {
+    // TODO: https://github.com/pytorch/torchcodec/issues/238: Implement CUDA
+    // device.
```
NicolasHug, line 10: Q - just for my understanding, is this where you'll implement the context caching? Or is it about something else?

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -16,7 +16,13 @@ function(make_torchcodec_library library_name ffmpeg_target)
         VideoDecoder.cpp
         VideoDecoderOps.h
         VideoDecoderOps.cpp
+        DeviceInterface.h
     )
+    if(ENABLE_CUDA)
+        list(APPEND sources CudaDevice.cpp)
+    else()
+        list(APPEND sources CPUOnlyDevice.cpp)
```
NicolasHug, line 24: Super Nit: just curious why one is called "CudaDevice" and the other is "CPUOnlyDevice"? That seems to suggest that "CudaDevice" isn't "Cuda Only" itself?

.github/workflows/cuda_tests.yaml
```diff
@@ -0,0 +1,66 @@
...
+        ./configure --prefix=$CONDA_PREFIX --enable-nonfree --enable-cuda-nvcc --disable-static --enable-shared --optflags=-fno-omit-frame-pointer --disable-stripping --enable-cuvid --enable-rpath
+        make --silent -j install
+        popd
+
+        CMAKE_BUILD_PARALLEL_LEVEL=8 CXXFLAGS="" LDFLAGS="-Wl,--allow-shlib-undefined -Wl,-rpath,$CONDA_PREFIX/lib -Wl,-rpath-link,$CONDA_PREFIX/lib -L$CONDA_PREFIX/lib" CMAKE_BUILD_TYPE=Release ENABLE_CUDA=1 ENABLE_NVTX=1 pip install -e ".[dev]" --no-build-isolation -vv --debug
+        # Ensure our compiled ffmpeg binary is in the path.
+        which ffmpeg
+        # We skip certain tests because they are not relevant to GPU decoding and they always fail with
+        # a custom FFMPEG build.
+        pytest -k "not (test_get_metadata or get_ffmpeg_version)" -vvv
```
NicolasHug, line 63: Nit: separate the script into substeps (like the other workflows) to ease debugging / viewing on the UX. Right now, viewing the test results requires a ton of scrolling to go past the compiling/build logs.

.github/workflows/cuda_tests.yaml
NicolasHug, line 1: Super Nit: the other files have a `yml` extension, not `yaml`.

NicolasHug (approved):
Thank you @ahmadsharif1 , only nits and questions from me. Nothing blocking!

src/torchcodec/decoders/_core/CudaDevice.cpp
ahmadsharif1: Yes, eventually this is where that device init and caching will go.

src/torchcodec/decoders/_core/CMakeLists.txt
ahmadsharif1: Yes CudaDevice will support both CPU and CUDA. For CPU it will just do what it normally does. For CUDA it will do something special.

CPUOnly will error if it sees any other device.

.github/workflows/cuda_tests.yaml
ahmadsharif1: This is not true, right:

image

Or are you comparing it against a different branch?

src/torchcodec/decoders/_core/video_decoder_ops.py
ahmadsharif1: I think for now we can only accept strings/tensors/ints, not a torch.Device. That's part of the limitation of C++ custom ops AFAIU.

.github/workflows/cuda_tests.yaml
NicolasHug: Sorry I was comparing with torchvision!

PR #237, opened 2024-10-03
Use `int` instead of `numbers.integral` in docstring

NicolasHug:
In general we want to avoid having type-annotation-like types within the docstring. Type annotations are meant to be read by a machine, not by humans, and over time they will clutter the docs.

In this instance, `int` is a more common and familiar term than `numbers.Integral`, and will help keep the doc more user-friendly. I don't believe there is any loss of accuracy keeping `int` in the docstring.

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -191,7 +191,7 @@ class VideoDecoder:
         """Return frame or frames as tensors, at the given index or range.
 
         Args:
-            key(numbers.Integral or slice): The index or range of frame(s) to retrieve.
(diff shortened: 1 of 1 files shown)


ahmadsharif1:
Does `int` convey the message to the user that numpy integers are okay to pass? I believe that was @scotts intent here.

NicolasHug:
I would argue that:

- We don't need to convey that point
- `numbers.Integral` doesn't convey that point either.

I really think numpy integers are an edge case that isn't in the mind of users. I've been using numpy for 10 years and I still don't know what these numpy types are. When `int` is the documented type, users expect duck typing to prevail, and expect `int`-like objects to work.

scotts (approved):

PR #240, opened 2024-10-04
Add regular index-based sampler - part 1

NicolasHug:
This PR adds an index-based sampler that returns equally-spaced clips.
It is a partial implementation, just like the random sampler in https://github.com/pytorch/torchcodec/pull/221. In particular it is missing 2 key parts:

- The implementation of a policy to deal with edge cases when the number of available frames is smaller than what is requested by the sampler
- fast and efficient "sort and decode indices" strategy

Both these parts are left as TODO and are tracked in https://github.com/pytorch/torchcodec/issues/239. I am hoping to merge this PR *before* we tackle those, because I want to make sure the solution we come up with will satisfy both samplers (and not just the existing random sampler).


src/torchcodec/samplers/_implem.py
```diff
@@ -140,3 +140,61 @@ def clips_at_random_indices(
     random.setstate(builtin_random_state)
 
     return clips
+
+
+def clips_at_regular_indices(
+    decoder: VideoDecoder,
+    *,
+    num_clips: int = 1,
```
NicolasHug, line 148: I feel like `clips_at_regular_indices(num_clips=1)` is an anti pattern. Same with 2. Maybe we should make 3 the default, and enforce `num_clips>=3`?

I don't feel too strongly about it either way - just being pedantic.

ahmadsharif1: If this is what other libraries do we can do the same here.

I was thinking of sampling the middle of the video if num_clips=1 but that would make it slower than existing libraries

test/samplers/test_samplers.py
```diff
@@ -150,59 +165,87 @@ def test_random_sampler_randomness():
     assert builtin_random_state_start == builtin_random_state_end
 
 
-def test_random_sampler_errors():
+@pytest.mark.parametrize(
+    "num_clips, sampling_range_size",
+    (
+        # Ask for 50 clips while the sampling range is 10 frames wide
```
ahmadsharif1, line 171: Is this 50 or 10?

test/samplers/test_samplers.py
```diff
@@ -150,59 +165,87 @@ def test_random_sampler_randomness():
...
+        sampling_range_end=sampling_range_size,  # because sampling_range_start=0
+    )
+
+    assert len(clips) == num_clips
+
+    clip_starts_seconds = torch.tensor([clip.pts_seconds[0] for clip in clips])
+    assert len(torch.unique(clip_starts_seconds)) == sampling_range_size
+
+    # Assert clips starts are ordered, i.e. the start indices don't just "wrap
+    # around". They're duplicated *and* ordered.
```
ahmadsharif1, line 195: Mention reason for not wrapping around?

ahmadsharif1 (approved):

src/torchcodec/samplers/_implem.py
NicolasHug: OK. Other libs allow `num_clips=1`. Let's revisit if needed

test/samplers/test_samplers.py
NicolasHug: There's no particular reason honestly, it's just an assertion of the behavior of `torch.linspace`. I'll add a comment near the Note to clarify that this is somewhat arbitrary.

PR #241, opened 2024-10-04
Samplers: add support for edge-case policies

NicolasHug:
This PR addresses the first big item in https://github.com/pytorch/torchcodec/issues/239, namely: support a "policy" to apply when the sampler tries to sample a frame beyond the end of the video.  We don't automatically error anymore, which leads to more flexibility. As discussed offline, the available policies are "repeat_last", "wrap", and "error".

Important points:

- I factorized the samplers into this `_abtract_sampler()` helper function. The only difference between the 2 samplers is the call to `torch.randint` vs `torch.linspace`, and I wanted to make that clear.
- The different policies are implemented as well-defined Python function. This leaves the door open for us to let the user define their own policies by providing a function, should we want to in the future.
- The crux of this PR is the `_build_all_clips_indices()` function. It's simple, but it does a lot of things. Look at it carefully.
- The other important part is the `_decode_all_clips_indices()` function, which we plan to re-write in a much more efficient way, with a C++ core.

Note: This PR is already complex enough, so I have *not* implemented the  "try to reduce `num_indices_between_frames`" strategy that I originally suggested https://github.com/pytorch/torchcodec/pull/221#discussion_r1773127663. This is orthogonal to the policies implemented here, so can revisit later if we want it.


src/torchcodec/samplers/_implem.py
```diff
@@ -74,130 +78,215 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
...
+    kind: Literal["random", "regular"],
     decoder: VideoDecoder,
     *,
     num_clips: int = 1,
     num_frames_per_clip: int = 1,
     num_indices_between_frames: int = 1,
     sampling_range_start: int = 0,
     sampling_range_end: Optional[int] = None,  # interval is [start, end).
+    # Important note: sampling_range_end defines the upper bound of where a clip
+    # can *start*, not where a clip can end.
```
NicolasHug, line 192: Why?
Because if `sampling_range_end` define the upper bound of where a clip can *end*, then the last few frames in that sampling range will *always* have a lower probability of being sampled. I.e. we would always be in this scenario that this comment pointed out https://github.com/pytorch/torchcodec/pull/221#discussion_r1773371164.

If sampling_range_end` define the upper bound of where a clip can *end*, then technically, there really there is no need for a `policy`. Hope that makes sense.

scotts: Almost entirely agreed, but the wrinkle is that even if we defined `sampling_range_end` to be the end of a clip, we'd still need a policy unless we also defined `len(decoder)` as the default last possible frame for the end of a clip. If we did that for both, we'd also ensure only whole clips.

Agreed that this is the right approach. We'll need to document carefully. :) I also think there's no way around needing to document carefully; whatever behavior we choose will end up with subtle behavior.

src/torchcodec/samplers/_implem.py
```diff
@@ -74,130 +88,211 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
...
+    all_decoded_frames: list[Frame] = [
+        decoder.get_frame_at(index) for index in all_clips_indices
+    ]
+    all_clips: list[list[Frame]] = chunk_list(
+        all_decoded_frames, chunk_size=num_frames_per_clip
+    )
+
+    return [to_framebatch(clip) for clip in all_clips]
+
+
```
scotts, line 191: Nit: I think a better name is `_generic_sampler()`. I tend to think of things called "abstract" as defining desired behavior without an actual implementation. I think of things that are called "generic" as having an implementation is _generic_ to one or several parameters.

src/torchcodec/samplers/_implem.py
```diff
@@ -25,21 +24,36 @@ def _validate_params(
...
         )
 
+    clip_span = _get_clip_span(
+        num_indices_between_frames=num_indices_between_frames,
+        num_frames_per_clip=num_frames_per_clip,
+    )
+
     if sampling_range_end is None:
-        sampling_range_end = num_frames - clip_span + 1
+        sampling_range_end = max(num_frames_in_video - clip_span + 1, 1)
```
scotts, line 56: Based on the discussion below, shouldn't this just be `num_frames_in_video`? That is, the default last index at which a clip can start is the final frame in the video. And if we're worried that `num_frames_in_video` is not at least 1, maybe we should explicitly check for (and error?) on that?

Even if we decide line 56 is correct, I can't reconcile it and 67. There, it looks like we're making sure that `sampling_range_end` is not larger than the video itself - which makes sense to me. But there, we don't subtract off the size of a clip, but here at line 56, we do.

src/torchcodec/samplers/_implem.py
```diff
@@ -74,130 +88,211 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
...
+
+def _decode_all_clips_indices(
+    decoder: VideoDecoder, all_clips_indices: list[int], num_frames_per_clip: int
+) -> list[FrameBatch]:
+    # This takes the list of all the frames to decode, decode all the frames,
+    # and then packs them into clips of length num_frames_per_clip.
+    # This is slow, unoptimized, and u.g.l.y. It is not meant to stay.
+    # TODO:
+    # - sort the frames to avoid backward seeks, dedup, decode, and re-organize frames.
+    # - write most of this in C++
```
NicolasHug, line 168: If you want a peek of what this will look like: https://github.com/pytorch/torchcodec/pull/245/commits/3dcbe1e4b52980fadbd8fa4a2922fecd8304c9c9

I'm leaving it for a follow-up PR to ease reviewing. Basic benchmark shows 5X speedups.

src/torchcodec/samplers/_implem.py
```diff
@@ -74,130 +88,211 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
...
+    # frame_indices = [1, 2, 3], num_frames_per_clip = 5
+    # output = [1, 2, 3, 1, 2]
+    return (frame_indices * (num_frames_per_clip // len(frame_indices) + 1))[
+        :num_frames_per_clip
+    ]
+
+
+def _error_policy(frames_indices: list[int], num_frames_per_clip: int) -> list[int]:
+    raise ValueError(
+        "You set the 'error' policy, and the sampler tried the decode a frame "
```
scotts, line 110: `"tried the decode"` -> `"tried to decode"`

src/torchcodec/samplers/_implem.py
NicolasHug: I think the lack of documentation is making this significantly harder to review, sorry about that.

So, my reasoning is the following:

- what happens at line 56 is the default behavior. In line 67, this isn't the default behavior anymore,  `sampling_range_end` is passed by the user
- the default behavior (56) is: "automatically set the upper bound of where a clip can start to a value such that we never try to sample beyond the last frame."
- the reason this is the default behavior is because there is *no* way for users to explicitly set such value, unless they do the match of computing the `clip_span` (and we don't want them to do that).
- Now, at line 67 when the user explicitly passes `sampling_range_end`, we just make sure it doesn't go above `num_frames_in_video` to avoid errors on short videos. If we were to substract `clip_span` there, it would change the semantic of `sampling_range_end` from "upper bound of the start" to "upper bound of the end"

src/torchcodec/samplers/_implem.py
```diff
@@ -74,130 +88,211 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
...
+    # above). When the frame indices go beyond num_frames_in_video, we force the
+    # frame indices back to valid values by applying the user's policy (wrap,
+    # repeat, etc.).
+    all_clips_indices: list[int] = []
+
+    clip_span = _get_clip_span(
+        num_indices_between_frames=num_indices_between_frames,
+        num_frames_per_clip=num_frames_per_clip,
+    )
+
```
scotts, line 148: The fact that this loop is so easy to implement - we can just call Python's `range()` directly, and how we're using our policy is obvious and still generic - makes me more confident we're doing the right thing.

src/torchcodec/samplers/_implem.py
ahmadsharif1: I was trying to prove to myself that that comment about probabilities made sense so I wrote a simple simulator:

image

It seems like the distribution is symmetric in the sense that even the start few frames have smaller chance of being selected.

I think that is because when you pick frame X you pick frame X + 1 as well, but you never pick frame -1 for example. So frames at the start also have lower chance. Note that this with with 0 dilation so that may change up things too.

I don't know if it breaks your assumptions but you can think about this.

ahmadsharif1: It may make sense to have a "slow" test with a simulator to see if the fair policy is within some threshold of an "even" distribution.

This comment isn't blocking or anything. It was mostly for my own curiosity since we don't have a closed-form mathematical expression for the probability of being selected.

NicolasHug: Ah, it's true that the first frames are also less likely to be sampled. Oh boy. I don't think we can or should do anything about that side though?

ahmadsharif1: It's kind of like how Conv1D adds padding to both sides:

https://pytorch.org/docs/stable/generated/torch.nn.Conv1d.html

NicolasHug: I did write a test to illustrate the difference of behavior depending on the value of `sampling_range_end`.

Writing an actual statistical test is challenging because in fact when `sampling_range_end=len(decoder)`, the last frames tend to be over-represented. I.e. the frames still don't follow a uniform probability, there's no distribution to check against.

test/samplers/test_samplers.py
```diff
@@ -131,6 +134,80 @@ def test_sampling_range_negative(sampler):
         assert_tensor_equal(clip.data, clips_1[0].data)
 
 
+@pytest.mark.parametrize("sampler", (clips_at_random_indices, clips_at_regular_indices))
+def test_sampling_range_default_behavior(sampler):
```
scotts, line 138: Okay, based on this comment, I think that explains the behavior I was confused about in [my comment above](https://github.com/pytorch/torchcodec/pull/241/files#r1790330006).

This means that effectively, the default values for the sampler are:
```
  sampling_range_start = 0,
  sampling_range_end = max(len(decoder) - clip_span + 1, 1),
```
I do think it may be simpler to just make them:
```
  sampling_range_start = 0,
  sampling_range_end = len(decoder),
```
And then let the policy deal with the case that we start a clip in the range `[len(encoder) - clip_size, len(encoder))`. (I may be off by 1 somewhere.) I don't have a strong preference for this, but I think it will make our code more straight-forward, and I think it's easier to explain the behavior. I'm unsure if it's actually a _better_ result for the user, though.

test/samplers/test_samplers.py
```diff
@@ -131,6 +134,80 @@ def test_sampling_range_negative(sampler):
...
+    # the number of frames in the video. That means that the last few frames of
+    # the video are less likely to be part of a clip.
+    # When sampling_range_end is set manually to e.g. len(decoder), the last
+    # frames are way more likely to be part of a clip, since there is no
+    # restriction on the sampling range (and the user-defined policy comes into
+    # action, potentially repeating that last frame).
+    #
+    # In this test we assert that the last sampled frame occurs significantly
+    # more often when sampling_range_end=len(decoder) than when it's None.
+    # This is only a proxy, for lack of better testing oppportunities.
```
NicolasHug, line 151: Looking into this more I think I went overboard here. It would be far simpler to just check the *beginning* of the last clip, instead of the last frame across all clips. That's a more direct check of the expected behavior. I'll re-write this test but please don't let that block the review :)

src/torchcodec/samplers/_implem.py
ahmadsharif1: Thinking about it a bit more, if you want frames to all be totally equal in terms of output probability, you will have to pick first first and last few frames with higher probability. It's not as simple as repeating the last frame, etc. because the second last frame also needs to be weighted higher, etc.

And that's why your test doesn't follow a uniform distribution.

For now this implementation is easy and simple. Later on if there is a need, we can implement something complex and "fair".

test/samplers/test_samplers.py
```diff
@@ -131,6 +134,80 @@ def test_sampling_range_negative(sampler):
...
+    all_pts_manual = torch.concat([clip.pts_seconds for clip in clips_manual]).tolist()
+    largest_pts_manual = max(all_pts_manual)
+    largest_pts_counts_manual = Counter(all_pts_manual)[largest_pts_manual]
+
+    # Assert that the probability of occurence of the "last" sampled frame is
+    # way higher when setting sampling_range_end=len(decoder) than when relying
+    # on the default behavior.
+    # Note: ideally we would directly assert the number of occurrences of the
+    # last frame based on its index, but our APIs don't return indices, only pts
+    # (yet?)
```
scotts, line 193: Yes, we should do that. Created #246.

src/torchcodec/samplers/_implem.py
scotts: Yes, that all tracks. I _think_ that is consistent with what I surmised below, when I read the test.

src/torchcodec/samplers/_implem.py
NicolasHug: Yeah, enforcing a uniform distribution across all frames is pretty hard.
That being said, we can already enforce a uniform distribution for **clip starts**: all the user has to do is to set `sampling_range_end=len(decoder)`. With that, all the frames have an equal probability of being picked to start a clip.

scotts (approved):

src/torchcodec/samplers/_implem.py
NicolasHug: SG, thanks. I don't have a super strong preference either. Let me merge now to unblock, but happy to revisit.

PR #242, opened 2024-10-04
Add CUDA decoding support 

ahmadsharif1:
Actually implement cuda decoding in C++:

1. Initialize a cuda device if requested. We create a small tensor on the device to initialize the context.
2. Use the cuda device to decode the video to NV12 format.
3. Use libNPP to convert from NV12 to RGB. We make sure to wait on this cuda event so there are no race conditions in accessing this tensor from the downstream consumer (that is on a different stream than libNPP).

Note that the GPU decodes frames that are not bit-accurate. This is by design and we ensure tensors are approximately equal rather than fully accurate. The actual tensor values depends on the GPU architecture because GPU math is not precise.

Also added a gpu_benchmark with the following results:

```
python benchmarks/decoders/gpu_benchmark.py --video /tmp/frame_numbers_1920x1080_100.mp4
[--------------------- Decode+Resize Time --------------------]
                       |  video=frame_numbers_1920x1080_100.mp4
1 threads: ----------------------------------------------------
      D=cuda R=cuda    |                   12.1                
      D=cuda R=cpu     |                  148.0                
      D=cuda R=native  |                   11.5                
      D=cuda R=none    |                   11.5                
      D=cpu R=cuda     |                   16.9                
      D=cpu R=cpu      |                  134.2                
      D=cpu R=native   |                   23.2                
      D=cpu R=none     |                    9.4                

Times are in seconds (s).

Key: D=Decode, R=Resize
Native resize is done as part of the decode step
none resize means there is no resize step -- native or otherwise

```

Results show that a single NVDec is slower than 22 core CPU without resizing, but faster with resizing.

I also added a "throughput mode" for the benchmark that decodes W videos in parallel using T threads. Results of this "throughput mode" shows that A100 has higher decode throughput than my 22-core CPU:

```
python benchmarks/decoders/gpu_benchmark.py --video /tmp/frame_numbers_1920x1080_100.mp4 --devices=cuda:0,cpu --resize_
devices=none --num_threads 10 --num_videos 10

[---------------------------------- Decode+Resize Time ----------------------------------]
                               |  threads=10 work=10 video=frame_numbers_1920x1080_100.mp4
1 threads: -------------------------------------------------------------------------------
      D=cuda R=none T=10 W=10  |                            29.0                          
      D=cpu R=none T=10 W=10   |                            38.8                          

Times are in seconds (s).

Key: D=Decode, R=Resize T=threads W=work (number of videos to decode)
Native resize is done as part of the decode step
none resize means there is no resize step -- native or otherwise

```

nvidia-smi shows 99% NVDEC utilization :)

```
# gpu         pid   type     sm    mem    enc    dec    jpg    ofa    command 
# Idx           #    C/G      %      %      %      %      %      %    name 
    0    2180816     C     70      5      -     99      -      -    python      
```

This throughput mode is representative of video decoding using the dataloader with multiple threads.

### DIFF
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import pathlib
 import time
 from concurrent.futures import ThreadPoolExecutor
 
@@ -101,7 +102,7 @@ def main():
     parser.add_argument(
         "--video",
         type=str,
-        default=os.path.dirname(__file__) + "/../../test/resources/nasa_13013.mp4",
+        default=pathlib.Path(__file__).parent / "../../test/resources/nasa_13013.mp4",
     )
     parser.add_argument(
         "--use_torch_benchmark",
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -1,6 +1,7 @@
 import argparse
 import os
 import time
+from concurrent.futures import ThreadPoolExecutor
 
 import torch.utils.benchmark as benchmark
 
@@ -62,6 +63,17 @@ def decode_full_video(video_path, decode_device_string, resize_device_string):
     return frame_count, end_time - start_time
 
 
+def decode_videos_using_threads(
+    video_path, decode_device_string, resize_device_string, num_videos, num_threads
+):
+    executor = ThreadPoolExecutor(max_workers=num_threads)
+    for i in range(num_videos):
+        executor.submit(
+            decode_full_video, video_path, decode_device_string, resize_device_string
+        )
+    executor.shutdown(wait=True)
+
+
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument(
@@ -90,6 +102,18 @@ def main():
             "autorange. Without this we just run one iteration without warmup "
             "to measure the cold start time."
         ),
+    )
+    parser.add_argument(
+        "--num_threads",
+        type=int,
+        default=1,
+        help="Number of threads to use for decoding. Only used when --use_torch_benchmark is set.",
+    )
+    parser.add_argument(
+        "--num_videos",
+        type=int,
+        default=50,
+        help="Number of videos to decode in parallel. Only used when --num_threads is set.",
     )
     args = parser.parse_args()
     video_path = args.video
@@ -120,22 +144,40 @@ def main():
                 resize_label = "cuda"
             print("decode_device", decode_device_string)
             print("resize_device", resize_device_string)
-            t = benchmark.Timer(
-                stmt="decode_full_video(video_path, decode_device_string, resize_device_string)",
-                globals={
-                    "decode_device_string": decode_device_string,
-                    "video_path": video_path,
-                    "decode_full_video": decode_full_video,
-                    "resize_device_string": resize_device_string,
-                },
-                label=label,
-                description=f"video={os.path.basename(video_path)}",
-                sub_label=f"D={decode_label} R={resize_label}",
-            ).blocked_autorange()
-            results.append(t)
+            if args.num_threads > 1:
+                t = benchmark.Timer(
+                    stmt="decode_videos_using_threads(video_path, decode_device_string, resize_device_string, num_videos, num_threads)",
+                    globals={
+                        "decode_device_string": decode_device_string,
+                        "video_path": video_path,
+                        "decode_full_video": decode_full_video,
+                        "decode_videos_using_threads": decode_videos_using_threads,
+                        "resize_device_string": resize_device_string,
+                        "num_videos": args.num_videos,
+                        "num_threads": args.num_threads,
+                    },
+                    label=label,
+                    description=f"threads={args.num_threads} work={args.num_videos} video={os.path.basename(video_path)}",
+                    sub_label=f"D={decode_label} R={resize_label} T={args.num_threads} W={args.num_videos}",
+                ).blocked_autorange()
+                results.append(t)
+            else:
+                t = benchmark.Timer(
+                    stmt="decode_full_video(video_path, decode_device_string, resize_device_string)",
+                    globals={
+                        "decode_device_string": decode_device_string,
+                        "video_path": video_path,
+                        "decode_full_video": decode_full_video,
+                        "resize_device_string": resize_device_string,
+                    },
+                    label=label,
+                    description=f"video={os.path.basename(video_path)}",
+                    sub_label=f"D={decode_label} R={resize_label}",
+                ).blocked_autorange()
+                results.append(t)
     compare = benchmark.Compare(results)
     compare.print()
-    print("Key: D=Decode, R=Resize")
+    print("Key: D=Decode, R=Resize T=threads W=work (number of videos to decode)")
     print("Native resize is done as part of the decode step")
     print("none resize means there is no resize step -- native or otherwise")
 
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -102,7 +102,9 @@ def main():
     parser.add_argument(
         "--video",
         type=str,
-        default=pathlib.Path(__file__).parent / "../../test/resources/nasa_13013.mp4",
+        default=str(
+            pathlib.Path(__file__).parent / "../../test/resources/nasa_13013.mp4"
+        ),
     )
     parser.add_argument(
         "--use_torch_benchmark",
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -3,6 +3,8 @@ import os
 import time
 from concurrent.futures import ThreadPoolExecutor
 
+import torch
+
 import torch.utils.benchmark as benchmark
 
 import torchcodec
@@ -64,12 +66,20 @@ def decode_full_video(video_path, decode_device_string, resize_device_string):
 
 
 def decode_videos_using_threads(
-    video_path, decode_device_string, resize_device_string, num_videos, num_threads
+    video_path,
+    decode_device_string,
+    resize_device_string,
+    num_videos,
+    num_threads,
+    use_multiple_gpus,
 ):
     executor = ThreadPoolExecutor(max_workers=num_threads)
     for i in range(num_videos):
+        actual_decode_device = decode_device_string
+        if "cuda" in decode_device_string:
+            actual_decode_device = f"cuda:{i % torch.cuda.device_count()}"
         executor.submit(
-            decode_full_video, video_path, decode_device_string, resize_device_string
+            decode_full_video, video_path, actual_decode_device, resize_device_string
         )
     executor.shutdown(wait=True)
 
@@ -115,6 +125,12 @@ def main():
         default=50,
         help="Number of videos to decode in parallel. Only used when --num_threads is set.",
     )
+    parser.add_argument(
+        "--use_multiple_gpus",
+        action=argparse.BooleanOptionalAction,
+        default=True,
+        help=("Use multiple GPUs to decode multiple videos in multi-threaded mode."),
+    )
     args = parser.parse_args()
     video_path = args.video
 
@@ -146,7 +162,7 @@ def main():
             print("resize_device", resize_device_string)
             if args.num_threads > 1:
                 t = benchmark.Timer(
-                    stmt="decode_videos_using_threads(video_path, decode_device_string, resize_device_string, num_videos, num_threads)",
+                    stmt="decode_videos_using_threads(video_path, decode_device_string, resize_device_string, num_videos, num_threads, use_multiple_gpus)",
                     globals={
                         "decode_device_string": decode_device_string,
                         "video_path": video_path,
@@ -155,6 +171,7 @@ def main():
                         "resize_device_string": resize_device_string,
                         "num_videos": args.num_videos,
                         "num_threads": args.num_threads,
+                        "use_multiple_gpus": args.use_multiple_gpus,
                     },
                     label=label,
                     description=f"threads={args.num_threads} work={args.num_videos} video={os.path.basename(video_path)}",
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -77,7 +77,7 @@ def decode_videos_using_threads(
     executor = ThreadPoolExecutor(max_workers=num_threads)
     for i in range(num_videos):
         actual_decode_device = decode_device_string
-        if "cuda" in decode_device_string:
+        if "cuda" in decode_device_string and use_multiple_gpus:
             actual_decode_device = f"cuda:{i % torch.cuda.device_count()}"
         executor.submit(
             decode_full_video, video_path, actual_decode_device, resize_device_string
--- a/benchmarks/decoders/gpu_benchmark.py
+++ b/benchmarks/decoders/gpu_benchmark.py
@@ -1,0 +1,144 @@
+import argparse
+import os
+import time
+
+import torch.utils.benchmark as benchmark
+
+import torchcodec
+import torchvision.transforms.v2.functional as F
+
+RESIZED_WIDTH = 256
+RESIZED_HEIGHT = 256
+
+
+def transfer_and_resize_frame(frame, resize_device_string):
+    # This should be a no-op if the frame is already on the target device.
+    frame = frame.to(resize_device_string)
+    frame = F.resize(frame, (RESIZED_HEIGHT, RESIZED_WIDTH))
+    return frame
+
+
+def decode_full_video(video_path, decode_device_string, resize_device_string):
+    # We use the core API instead of SimpleVideoDecoder because the core API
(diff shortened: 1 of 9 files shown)


ronghanghu:
Looking forward to this!

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +52,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  cudaStream_t nppStream = nppGetStream();
+  cudaStream_t torchStream = at::cuda::getCurrentCUDAStream().stream();
+  status = nppiNV12ToRGB_8u_P2C3R(
+      input,
+      src->linesize[0],
+      static_cast<Npp8u*>(dst.data_ptr()),
+      dst.stride(0),
+      oSizeROI);
+  // Make the pytorch stream wait for the npp kernel to finish before using the
+  // output.
```
NicolasHug, line 106: I don't have a strong opinion on this, but maybe using the higher-level pytorch APIs might simplify this code a bit?

https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CUDAEvent.h

example usage: 

https://github.com/pytorch/vision/blob/80d85179ed378a37d25223b6da4f9ce95ce5384f/torchvision/csrc/io/image/cuda/encode_jpegs_cuda.cpp#L97-L98

benchmarks/decoders/gpu_benchmark.py
```diff
@@ -0,0 +1,203 @@
...
+    parser.add_argument(
+        "--resize_devices",
+        default="cuda:0,cpu,native,none",
+        type=str,
+        help="Comma-separated devices to test preroc (resize) on. Use 'none' to specify no resize.",
+    )
+    parser.add_argument(
+        "--video",
+        type=str,
+        default=os.path.dirname(__file__) + "/../../test/resources/nasa_13013.mp4",
```
NicolasHug, line 104: Nit: use the more powerful and modern `pathlib` instead of `os.path` for handling path. E.g. this would be

```py
pathlib.Path(__file__).parent / "../../test/resources/nasa_13013.mp4"
```

(and as a bonus the `..` is cross-platform... but we don't care that much since we don't target Windows)

NicolasHug: I have opened https://github.com/pytorch/torchcodec/pull/249 to migrate existing uses of `os.path`

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -34,12 +34,17 @@ function(make_torchcodec_library library_name ffmpeg_target)
         ${Python3_INCLUDE_DIRS}
     )
 
+    set(NEEDED_LIBRARIES ${ffmpeg_target} ${TORCH_LIBRARIES} ${Python3_LIBRARIES})
+    if(ENABLE_CUDA)
+        list(APPEND NEEDED_LIBRARIES ${CUDA_CUDA_LIBRARY} ${CUDA_nppi_LIBRARY} ${CUDA_nppicc_LIBRARY} )
+    endif()
+    if(ENABLE_NVTX)
+        list(APPEND NEEDED_LIBRARIES nvtx3-cpp)
+    endif()
```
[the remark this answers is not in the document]
ahmadsharif1: Good catch. That was a remnant from the previous iteration. We aren't using NVTX anymore

test/decoders/test_video_decoder_ops.py
```diff
@@ -464,8 +464,14 @@ def test_color_conversion_library_with_generated_videos(
...
-        with pytest.raises(RuntimeError, match="CUDA device is unimplemented"):
-            add_video_stream(decoder, device="cuda")
+        add_video_stream(decoder, device="cuda")
+        frame0, *_ = get_next_frame(decoder)
+        assert frame0.device.type == "cuda"
+        frame0_cpu = frame0.to("cpu")
+        reference_frame0 = NASA_VIDEO.get_frame_data_by_index(0)
+        # We pass in atol of 60 because the CUDA decoder is not bit-accurate
+        # compared to the CPU decoder.
+        torch.testing.assert_close(frame0_cpu, reference_frame0, atol=60, rtol=0)
```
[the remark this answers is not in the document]
ahmadsharif1: Good idea. Done

src/torchcodec/decoders/_core/CPUOnlyDevice.cpp
```diff
@@ -13,7 +14,19 @@ void throwUnsupportedDeviceError(const torch::Device& device) {
   throw std::runtime_error("Unsupported device: " + device.str());
 }
 
-void initializeDeviceContext(const torch::Device& device) {
+VideoDecoder::DecodedOutput convertAVFrameToDecodedOutputOnDevice(
+    const torch::Device& device,
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    AVCodecContext* codecContext,
+    VideoDecoder::RawDecodedOutput& rawOutput) {
+  throwUnsupportedDeviceError(device);
```
scotts, line 22: Should this function and the function below always throw? If yes, then we should just do something like `TORCH_CHECK(false, "Unsupported device.");`. In order avoid the need for a return value, mark the function as `[[noreturn]]`: https://en.cppreference.com/w/cpp/language/attributes/noreturn. We should rely on a TORCH macro to do the throwing for us rather than doing the throw ourselves, and we should make it obviously one that will always fail its check.

src/torchcodec/decoders/_core/DeviceInterface.h
```diff
@@ -23,6 +28,14 @@ namespace facebook::torchcodec {
 
 // Initialize the hardware device that is specified in `device`. Some builds
 // support CUDA and others only support CPU.
-void initializeDeviceContext(const torch::Device& device);
+void initializeDeviceContext(
+    const torch::Device& device,
+    AVCodecContext* codecContext);
+
+VideoDecoder::DecodedOutput convertAVFrameToDecodedOutputOnDevice(
```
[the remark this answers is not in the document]
ahmadsharif1: Since this is an interface, I want it to be generic so we can support AMD, etc. in the future. That's why I call it "device"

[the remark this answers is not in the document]
scotts: Agreed that we should say "OnCUDA" if the implementation only supports CUDA. We may support other kinds of devices in the future.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -1,6 +1,48 @@
...
+namespace {
+AVBufferRef* getCudaContext(const torch::Device& device) {
+  enum AVHWDeviceType type = av_hwdevice_find_type_by_name("cuda");
+  TORCH_CHECK(type != AV_HWDEVICE_TYPE_NONE, "Failed to find cuda device");
+  int err = 0;
+  AVBufferRef* hw_device_ctx;
+  torch::DeviceIndex deviceIndex = device.index();
+  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
+  std::string deviceOrdinal = std::to_string(deviceIndex);
+  err = av_hwdevice_ctx_create(
```
scotts, line 24: Nit: define and declare `err` here where it's defined.

src/torchcodec/decoders/_core/CPUOnlyDevice.cpp
ahmadsharif1: Good suggestion. Done

scotts: I think we maybe should also annotate `convertAVFrameToDecodedOutputOnDevice()` and `initializeDeviceContext()` with `[[noreturn]]`. Let's also avoid two `TORCH_CHECK` calls. Whatever message we want to put on stderr, we can do it in one check.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +51,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  at::DeviceIndex deviceIndex = device.index();
+  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
+  at::DeviceIndex originalDeviceIndex = at::cuda::current_device();
+  cudaSetDevice(deviceIndex);
+
+  auto start = std::chrono::high_resolution_clock::now();
+
+  cudaStream_t nppStream = nppGetStream();
+  cudaStream_t torchStream = at::cuda::getCurrentCUDAStream().stream();
+  status = nppiNV12ToRGB_8u_P2C3R(
```
scotts, line 99: Nit: define and declare `status` here.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +51,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  TORCH_CHECK(
+      src->format == AV_PIX_FMT_CUDA,
+      "Expected format to be AV_PIX_FMT_CUDA, got " +
+          std::string(av_get_pix_fmt_name((AVPixelFormat)src->format)));
+  int width = options.width.value_or(codecContext->width);
+  int height = options.height.value_or(codecContext->height);
+  NppStatus status;
+  NppiSize oSizeROI;
+  oSizeROI.width = width;
+  oSizeROI.height = height;
```
scotts, line 84: Nit: we can do this cleaner with:
```suggestion
  NppiSize oSizeROI = {width, height];
```
A theme to my nits is that if possible, I prefer to initialize and declare values at the same time. :)

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +51,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  Npp8u* input[2];
+  input[0] = (Npp8u*)src->data[0];
+  input[1] = (Npp8u*)src->data[1];
```
scotts, line 87: Same here:
```suggestion
  Npp8u* input[2] = {
    static_cast<Npp8u*>(src->data[0]),
    static_cast<Npp8u*>(src->data[1])
  };
```
The reason I prefer this style is that it's a signal to the reader that we are introducing `input`, and that from now on we will (likely) only ever read from it. For me, it helps to reduce the context I need to carry across statements.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +51,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  torch::Tensor& dst = output.frame;
+  dst = allocateDeviceTensor({height, width, 3}, options.device);
+  at::DeviceIndex deviceIndex = device.index();
+  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
+  at::DeviceIndex originalDeviceIndex = at::cuda::current_device();
+  cudaSetDevice(deviceIndex);
```
scotts, line 93: I can see that later on line 121 we restore the device index. Why? Can we explain why we need to set and then restore?

ahmadsharif1: I ended up using CUDADeviceGuard. Callers are assuming we are not interfering with the cuda device in this function

src/torchcodec/decoders/_core/CudaDevice.cpp
ahmadsharif1: Great suggestion. This simplifies the code a lot. Done now

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -856,6 +856,25 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
   output.duration = getDuration(frame);
   output.durationSeconds = ptsToSeconds(
       getDuration(frame), formatContext_->streams[streamIndex]->time_base);
+  if (streamInfo.options.device.type() != torch::kCPU) {
```
scotts, line 859: Is this logic truly general to all non-CPU devices? If no, then here and elsewhere, we should do something closer to:
```
if (streamInfo.options.device.type() == torch::kCUDA) {
  logicSpecificToCUDA();
}
else if (streamInfo.options.device.type() == torch::kCPU) {
  logicSpecificToCPU();
}
else {
  TORCH_CHECK(false, "Unsupported device");
}
```

ahmadsharif1: Right now we only support cuda.

I am assuming if we support AMD we will use the same interface, and use cmake or #ifdefs to link in the correct device code.

So VideoDecoder.cpp just assumes cmake or the linker will do the right thing and just calls the device code for any type of device.

At the moment cuda device is linked by cmake for cuda builds. How that will be done for AMD is TBD.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -10,13 +51,83 @@ void throwErrorIfNonCudaDevice(const torch::Device& device) {
...
+  NppiSize oSizeROI;
+  oSizeROI.width = width;
+  oSizeROI.height = height;
+  Npp8u* input[2];
+  input[0] = (Npp8u*)src->data[0];
+  input[1] = (Npp8u*)src->data[1];
+  torch::Tensor& dst = output.frame;
+  dst = allocateDeviceTensor({height, width, 3}, options.device);
+  at::DeviceIndex deviceIndex = device.index();
+  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
```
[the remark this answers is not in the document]
ahmadsharif1: For a single GPU libtorch returns -1 while ffmpeg assumes it will be 0. So we have to bridge that gap.

For multi-GPU setup, I haven't seen -1 being returned by torch -- so there we wont have to do a max.

The -1 seems to be a libtorch specific thing.

src/torchcodec/decoders/_core/CPUOnlyDevice.cpp
ahmadsharif1: The two checks are there because one is a programming/logic error on our part -- we should never pass in a CPU device for device functions.

The other is the check for passing in a non-compiled device.

benchmarks/decoders/gpu_benchmark.py
ahmadsharif1: I updated my code to use pathlib as it's new code

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: I see, that makes sense. That also means that we don't need to have CPU versions tof functions hat throw for all N devices we support.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -1,6 +1,49 @@
...
+#include <libavutil/pixdesc.h>
+}
 
 namespace facebook::torchcodec {
+namespace {
+AVBufferRef* getCudaContext(const torch::Device& device) {
+  enum AVHWDeviceType type = av_hwdevice_find_type_by_name("cuda");
+  TORCH_CHECK(type != AV_HWDEVICE_TYPE_NONE, "Failed to find cuda device");
+  torch::DeviceIndex deviceIndex = device.index();
+  // FFMPEG cannot handle negative device indices.
```
scotts, line 21: I understand now what instigated this code, but I still can't evaluate if it's correct. Looking at the docs, a negative value indicates the "current device": https://pytorch.org/cppdocs/api/structc10_1_1_device.html#_CPPv4N3c106DeviceE

Is it safe to map all values of "current device" to 0? Is this a mapping we need to track? What happens when we are on a system with multiple GPUs? I'm assuming we don't fully understand the answers to these questions, and I don't want to block progress. So I think we should have a meatier comment both explaining what we do know, and indicating this may be a problem in the future.

scotts (approved):

src/torchcodec/decoders/_core/CudaDevice.cpp
ahmadsharif1: I have added a longer comment with a TODO to investigate that it works properly with multi-GPU setup. I am sure once users start using it, we will hit more edge cases.

ronghanghu:
Hi @ahmadsharif1, thanks for the great PR to add back GPU support! Wondering if it's possible to also add back the `device` parameter into `SimpleVideoDecoder`, which was previously removed in https://github.com/pytorch/torchcodec/pull/196/files#diff-5ff4f051479ffd5d021001e2a101973746feda3a3f579bf2d072629329c421dc?

ahmadsharif1:
Hi @ronghanghu 

We will be adding GPU support to SimpleVideoDecoder before releasing CUDA-enabled binaries, so stay tuned. However it will be done in a different PR.

Out of curiosity, can you share how you get TorchCodec? Is it through pip install torchcodec or do you build it from source (or do build it from internal fbcode)?

We want to add support for GPU decoding in SimpleVideoDecoder before releasing it, but the release may take a while. If you build from source you can get it earlier. So let us know how you get TorchCodec so we can help you get it asap.

ronghanghu:
Thanks @ahmadsharif1, we are building TorchCodec internally

PR #245, opened 2024-10-07
Speed-up samplers by avoiding backwards seeks

NicolasHug:
We now first sort and dedup the frame indices to be decoded within the sampler (see code for details). We can still implement this in C++ to optimize this further but this already leads to strong speed-ups.

Benchmark results - TL;DR: 5X faster when num_clips is large.

```
Using the following values:
sampler = clips_at_random_indices
num_frames_per_clip = 10
num_indices_between_frames = 2


num_clips = 1
when num_clips=1 there should be no speed-up.
We just need to make sure the new logic didn't add any overhead
main: med = 92.16ms +- 9.36
PR:   med = 89.27ms +- 7.90


num_clips = 50
With num_clips = 50 there are potentially a lot of overlap and backwards-seeks.
We expect significant speed-ups.
main: med = 1527.26ms +- 839.48
PR:   med = 331.37ms +- 170.58
```

Benchmark code:

<details>

```py
from torchcodec.samplers import clips_at_random_indices
import torch

    for _ in range(warmup):
        f(*args, **kwargs)

    times = []
    for _ in range(num_exp):
        start = perf_counter_ns()
        f(*args, **kwargs)
        end = perf_counter_ns()
        times.append(end - start)

    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    return med


def sample():
    decoder = VideoDecoder("test/resources/nasa_13013.mp4")
    clips_at_random_indices(
        decoder,
        num_clips=1,
        num_frames_per_clip=10,
        num_indices_between_frames=2,
    )

times = bench(sample, num_exp=30, warmup=2)
report_stats(times, unit="ms")
```

</details>


src/torchcodec/samplers/_implem.py
```diff
@@ -160,12 +160,17 @@ def _build_all_clips_indices(
...
+    # decode all the frames, and then packs them into clips of length
+    # num_frames_per_clip.
+    #
+    # To avoid backwards seeks (which are slow), we:
+    # - sort all the frame indices to be decoded
+    # - dedup them
+    # - decode all unique frames in sorted order
+    # - re-assemble the decoded frames back to their original order
+    #
+    # TODO: Write this in C++ so we can avoid the copies that happen in `to_framebatch`
```
NicolasHug, line 173: Apologies for the use of  `i` and `j` indices. Kinda hard to find good names for "indices within an array of indices"

src/torchcodec/samplers/_implem.py
```diff
@@ -179,9 +184,24 @@ def to_framebatch(frames: list[Frame]) -> FrameBatch:
...
+            # Avoid decoding the same frame twice.
+            decoded_frame = previous_decoded_frame
+        else:
+            decoded_frame = decoder.get_frame_at(index=frame_index)
+        previous_decoded_frame = decoded_frame
+        all_decoded_frames[j] = decoded_frame
+
     all_clips: list[list[Frame]] = chunk_list(
         all_decoded_frames, chunk_size=num_frames_per_clip
     )
```
NicolasHug, line 207: Note that we don't *have to* chunk the clips. The implementation already allows us to return a single 5D `FrameBatch` instead of a `list[4D FrameBatch]`.  I'll just leave this for another PR so we can discuss.

src/torchcodec/samplers/_implem.py
```diff
@@ -179,9 +184,24 @@ def to_framebatch(frames: list[Frame]) -> FrameBatch:
...
+    previous_decoded_frame = None
+    all_decoded_frames = [None] * len(all_clips_indices)
+    for i, j in enumerate(argsort):
+        frame_index = all_clips_indices_sorted[i]
+        if (
+            previous_decoded_frame is not None  # then we know i > 0
+            and frame_index == all_clips_indices_sorted[i - 1]
+        ):
+            # Avoid decoding the same frame twice.
+            decoded_frame = previous_decoded_frame
```
ahmadsharif1, line 199: This is setting it to the same python object, right?

Will there be any issues with that? Example, if the user modifies that tensor or something else in FrameBatch -- they will modify both entries in the list, right?

src/torchcodec/samplers/_implem.py
```diff
@@ -179,9 +184,24 @@ def to_framebatch(frames: list[Frame]) -> FrameBatch:
             data=data, pts_seconds=pts_seconds, duration_seconds=duration_seconds
         )
 
-    all_decoded_frames: list[Frame] = [
-        decoder.get_frame_at(index) for index in all_clips_indices
-    ]
+    all_clips_indices_sorted, argsort = zip(
+        *sorted((j, i) for (i, j) in enumerate(all_clips_indices))
```
ahmadsharif1, line 188: Nit: i, j makes both look like indexes in the same range. Maybe call them batch_index, frame_index?

ahmadsharif1 (approved):

src/torchcodec/samplers/_implem.py
NicolasHug, line 199: To be slightly safer w.r.t. future changes this should be

```py
decoded_frame = copy(previous_decoded_frame)
```

but we don't implement `__copy__` on Frame.

Note that a copy *still* happens within `to_framebatch`, so this is currently safe, but admittedly subject to an implementation detail that will change.

We can either:

- be OK with this since we'll re-implement it in C++ anyway
- implement `__copy__`.

LMK.

ahmadsharif1:
optionally you could also check in the benchmark code because we probably want to track the performance of this and make sure it doesn't regress.

src/torchcodec/samplers/_implem.py
ahmadsharif1: I am OK with this as-is.

NicolasHug: OK, I'll add a comment in `to_framebatch()` so we don't accidentally mess it up.

PR #247, opened 2024-10-07
Refactor Frame and FrameBatch into their own source file

scotts:
This will simplify my future PRs where we start implementing `SingleStreamDecoder`.

### DIFF
--- a/docs/source/api_ref_decoders.rst
+++ b/docs/source/api_ref_decoders.rst
@@ -20,6 +20,4 @@ torchcodec.decoders
     :nosignatures:
     :template: dataclass.rst
 
-    Frame
-    FrameBatch
     VideoStreamMetadata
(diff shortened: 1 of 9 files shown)


src/torchcodec/decoders/_frame.py
NicolasHug, line 1: Can you actually move this file one level up into `src/torchcodec/`? We'll need that to expose them in `torchcodec` as discussed in https://github.com/pytorch/torchcodec/pull/221#discussion_r1773652722

NicolasHug (approved):
LGTM with one small request to move it one level up

src/torchcodec/decoders/_frame.py
scotts: I remembered that discussion, but got the conclusion wrong. :)

ahmadsharif1 (approved):

PR #250, opened 2024-10-08
Add time-based regular sampler

NicolasHug:
This PR adds a time-based regular sampler

CI failures are spurious and unrelated, caused by infra issues.

### DIFF
--- a/src/torchcodec/samplers/_implem.py
+++ b/src/torchcodec/samplers/_implem.py
@@ -443,23 +443,14 @@ def _build_all_clips_timestamps(
 
     all_clips_timestamps: list[float] = []
     for start_seconds in clip_start_seconds:
-        # clip_timestamps = [
-        #   start_seconds + 0 * seconds_between_frames,
-        #   start_seconds + 1 * seconds_between_frames,
-        #   start_seconds + 2 * seconds_between_frames,
-        #   ...
-        # ]
-        clip_timestamps = torch.full(
-            size=(num_frames_per_clip,), fill_value=start_seconds, dtype=torch.float
-        )
-        clip_timestamps += (
-            torch.arange(num_frames_per_clip, dtype=torch.float)
-            * seconds_between_frames
-        )
+        clip_timestamps = [
+            start_seconds + i * seconds_between_frames
+            for i in range(num_frames_per_clip)
+        ]
+        clip_timestamps = [
+            timestamp for timestamp in clip_timestamps if timestamp < end_stream_seconds
+        ]
 
-        # We clip to valid values, so that we can call the same policies as
-        # for index-based samplers.
-        clip_timestamps = clip_timestamps[clip_timestamps < end_stream_seconds].tolist()
         if len(clip_timestamps) < num_frames_per_clip:
             clip_timestamps = policy_fun(clip_timestamps, num_frames_per_clip)
         all_clips_timestamps += clip_timestamps
--- a/src/torchcodec/samplers/_implem.py
+++ b/src/torchcodec/samplers/_implem.py
@@ -9,32 +9,33 @@ from torchcodec.decoders import VideoDecoder
 _EPS = 1e-4
 
 
-def _validate_params(
-    *, decoder, num_clips, num_frames_per_clip, num_indices_between_frames, policy
-):
+def _validate_params(*, decoder, num_frames_per_clip, policy):
     if len(decoder) < 1:
         raise ValueError(
             f"Decoder must have at least one frame, found {len(decoder)} frames."
         )
 
-    if num_clips <= 0:
-        raise ValueError(f"num_clips ({num_clips}) must be strictly positive")
     if num_frames_per_clip <= 0:
         raise ValueError(
             f"num_frames_per_clip ({num_frames_per_clip}) must be strictly positive"
         )
-    if num_indices_between_frames <= 0:
-        raise ValueError(
-            f"num_indices_between_frames ({num_indices_between_frames}) must be strictly positive"
-        )
-
     if policy not in _POLICY_FUNCTIONS.keys():
         raise ValueError(
             f"Invalid policy ({policy}). Supported values are {_POLICY_FUNCTIONS.keys()}."
         )
 
 
-def _validate_sampling_range(
+def _validate_params_index_based(*, num_clips, num_indices_between_frames):
+    if num_clips <= 0:
+        raise ValueError(f"num_clips ({num_clips}) must be strictly positive")
+
+    if num_indices_between_frames <= 0:
+        raise ValueError(
+            f"num_indices_between_frames ({num_indices_between_frames}) must be strictly positive"
+        )
+
+
+def _validate_sampling_range_index_based(
     *,
     num_indices_between_frames,
     num_frames_per_clip,
@@ -235,13 +236,15 @@ def _generic_index_based_sampler(
 
     _validate_params(
         decoder=decoder,
-        num_clips=num_clips,
         num_frames_per_clip=num_frames_per_clip,
-        num_indices_between_frames=num_indices_between_frames,
         policy=policy,
     )
+    _validate_params_index_based(
+        num_clips=num_clips,
+        num_indices_between_frames=num_indices_between_frames,
+    )
 
-    sampling_range_start, sampling_range_end = _validate_sampling_range(
+    sampling_range_start, sampling_range_end = _validate_sampling_range_index_based(
         num_frames_per_clip=num_frames_per_clip,
         num_indices_between_frames=num_indices_between_frames,
         sampling_range_start=sampling_range_start,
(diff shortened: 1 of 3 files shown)


src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+
+    return seconds_between_frames
+
+
+def _validate_sampling_range_time_based(
+    *,
+    num_frames_per_clip,
+    seconds_between_frames,
+    sampling_range_start,
+    sampling_range_end,
```
NicolasHug, line 377: Note: we don't support negative value for `sampling_range_*` right now. Leaving it out as future improvement.

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+        sampling_range_end=sampling_range_end,
+        begin_stream_seconds=decoder.metadata.begin_stream_seconds,
+        end_stream_seconds=decoder.metadata.end_stream_seconds,
+    )
+
+    clip_start_seconds = torch.arange(
+        sampling_range_start,
+        sampling_range_end,  # excluded
+        seconds_between_clip_starts,
+    )
```
NicolasHug, line 552: I'm fairly confident that implementing a random time-based sampler should be trivial, i.e. just changing this to

```py
clip_start_seconds = torch.rand(num_clips) * (sampling_range_end - sampling_range_start) + sampling_range_start
```

LMK if we want this and I'll follow-up with a PR.

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+        all_decoded_frames, chunk_size=num_frames_per_clip
+    )
+
+    return [to_framebatch(clip) for clip in all_clips]
+
+
+def clips_at_regular_timestamps(
+    decoder,
+    *,
+    seconds_between_clip_starts: float,
```
ahmadsharif1, line 515: The user will have to do some math to arrive at a number here that ensures the full video is used and we don't apply a policy function, right?

Thinking about this a bit more, I think this may be easy to get wrong for users and it is not consistent with `clilps_at_regular_indices` which takes in a num_clips.

For 2 clips, a naive user may just use seconds_between_clip_starts = video_duration (or video_duration / 2)

For 3 clips a naive user may use seconds_between_clip_starts = video_duration / 2 (or video_duration/3).

It seems easy to get wrong (off by one), etc. And doesn't offer an easy way to transition between index-based and time-based (params have to be changed).

I think the params here should be:

```
    num_clips: int = 1,
    num_frames_per_clip: int = 1,
    seconds_between_clip_frames: float = 1,  # This is the equivalent of indices_between_clip_frames
```

With this construction we probably don't even need a policy function if we exclude the upper bound properly like torch.arange, etc. (We may need a policy function if we want to sample all frames with equal probability).

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+        # clip_timestamps = [
+        #  clip_start + 0 * seconds_between_frames,
+        #  clip_start + 1 * seconds_between_frames,
+        #  clip_start + 2 * seconds_between_frames,
+        #  ...
+        #  clip_start + (num_frames_per_clip - 1) * seconds_between_frames,
+        # ]
+        # To guarantee that any such value is < end_stream_seconds, we only need
+        # to guarantee that
+        # clip_start < end_stream_seconds - (num_frames_per_clip - 1) * seconds_between_frames
```
ahmadsharif1, line 414: Can you mention that we use torch.arange() here so it excludes the end?

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+
+    return [to_framebatch(clip) for clip in all_clips]
+
+
+def clips_at_regular_timestamps(
+    decoder,
+    *,
+    seconds_between_clip_starts: float,
+    num_frames_per_clip: int = 1,
+    seconds_between_frames: Optional[float] = None,
```
ahmadsharif1, line 517: This is a bit misleading because the seconds between frames within a clip are different from across clips.

Maybe rename this to seconds_between_frames_within_clip

test/samplers/test_samplers.py
```diff
@@ -61,20 +94,80 @@ def test_sampler(sampler, num_indices_between_frames):
     )
 
 
-@pytest.mark.parametrize("sampler", (clips_at_random_indices, clips_at_regular_indices))
+@pytest.mark.parametrize("seconds_between_frames", [None, 3])
```
ahmadsharif1, line 97: It would be nice to write a test that derives seconds_between_clip_starts from the video duration like a real user would. In that test we should also make sure that there are no repeated frames in the sampled frames.

ahmadsharif1 (commented):
I have one large comment for the interface. Feel free to ignore it, but I do feel that we should keep the interfaces similar for users who want to transition from index-based to time-based when they realize fps is variable.

I'll leave it to @scotts for the approval

src/torchcodec/samplers/_implem.py
NicolasHug: I don't find `seconds_between_clips` particularly intuitive either.

I'm happy to introduce a `num_clips` parameter that would  be mutually exclusive with `seconds_between_clip_starts`.
 However, as discussed during the design phase, we'll have to keep `seconds_between_clip_starts` for feature-parity with existing samples.

ahmadsharif1: I think some guidance on how to derive this seconds_between_clip_starts should be useful as a comment or function since I do think it may be easy to get wrong. Add it to the test as well, so users can see how it's used.

ahmadsharif1: I don't know enough about python to see how mutually exclusive args would be constructed, but num_clips seems better for users who want to transition from fixed FPS to variable FPS videos.

In C++ one solution may be to do function overloading where the same name either accepts num_clips or seconds_between_clip_starts as two different types. I don't know enough python to know what's pythonic acceptable in python.

src/torchcodec/samplers/_implem.py
NicolasHug: The same applies to the existing `num_indices_between_frames` parameter that exists for index-based samplers. I don't think it's worth extending this already-long parameter name, it should be obvious from the docs what it means.

test/samplers/test_samplers.py
NicolasHug: I don't think users will derive `seconds_between_clip_starts` from the video duration.

src/torchcodec/samplers/_implem.py
NicolasHug: I don't think this is particularly relevant here.
I did add a comment on `torch.arange` and insisted in a few places that `sampling_range_end` should be an open bound. But here in this function, whether `torch.arange` is used should not be relevant.

src/torchcodec/samplers/_implem.py
```diff
@@ -89,32 +90,37 @@ def _get_clip_span(*, num_indices_between_frames, num_frames_per_clip):
     return num_indices_between_frames * (num_frames_per_clip - 1) + 1
 
 
+_LIST_OF_INT_OR_FLOAT = Union[list[int], list[float]]
```
scotts, line 93: Random aside: since we bumped our minimum Python version up to 3.9, I think we can use the `|` operator to indicate unions now. This would apply to a lot of other places as well. There's also some other type hint nicetities I think we can take advatnage of, like (I think) not needing `typing.List` and `typing.Dict` anymore. Not worth dealing with in this PR, I'm just making the comment to raise awareness for some code cleanup efforts in the future.

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+
+    all_clips_timestamps: list[float] = []
+    for start_seconds in clip_start_seconds:
+        clip_timestamps = [
+            start_seconds + i * seconds_between_frames
+            for i in range(num_frames_per_clip)
+        ]
+        clip_timestamps = [
+            timestamp for timestamp in clip_timestamps if timestamp < end_stream_seconds
+        ]
```
scotts, line 452: I'm unsure if this alternative is better:
```suggestion
        clip_timestamps = [
            timestamp
            for i in range(num_frames_per_clip)
            if (timestamp := start_seconds + i * seconds_between_frames) < end_stream_seconds
        ]
```
It'll avoid a list creation, but it may be too confusing.

src/torchcodec/samplers/_implem.py
NicolasHug: Ah, I tried but sadly the use of `|` is 3.10 https://peps.python.org/pep-0604/

Seems like we can still use the modern `|`  if we do `from __future__ import annotations` https://stackoverflow.com/questions/76579085/python-3-9-use-or-operator-for-union-types

src/torchcodec/samplers/_implem.py
```diff
@@ -323,3 +331,236 @@ def clips_at_regular_indices(
...
+        all_clips_timestamps += clip_timestamps
+
+    return all_clips_timestamps
+
+
+def _decode_all_clips_timestamps(
+    decoder: VideoDecoder, all_clips_timestamps: list[float], num_frames_per_clip: int
+) -> list[FrameBatch]:
+    # This is 99% the same as _decode_all_clips_indices. The only change is the
+    # call to .get_frame_displayed_at(pts) instead of .get_frame_at(idx)
```
scotts, line 465: Could we at least extract out `chunk_list()` and `to_framebatch()` as reusable components?

scotts: We could also pass in the method to call for the actual decoding. But we can work on that later.

test/samplers/test_samplers.py
```diff
@@ -1,21 +1,66 @@
...
+
+    if expected_seconds_between_clip_starts is not None:
+        # This can only be asserted with the time-based sampler, where
+        # seconds_between_clip_starts is known since it's passed by the user.
+        torch.testing.assert_close(
+            diff, expected_seconds_between_clip_starts, atol=0.01, rtol=0
+        )
+
+    assert (diff > 0).all()  # Also assert clips are sorted by start time
+
```
scotts, line 59: Is line 51 supposed to be indented one level? As written, I think it will only apply to the last value `diff` takes on in the loop above. (I didn't even realized iteration variables remained in scope after their loop in Python until I tested it just now.)

I'm also confused by line 58 - like line 51 above, I would expect it to be indented one level and just be `assert diff > 0`. As written, though, I'm very confused how it passes, as I thought `(diff > 0)` would become a boolean, and booleans don't have an `all()` method.

Is there something obvious I'm missing?

scotts (approved):

test/samplers/test_samplers.py
NicolasHug: Good catch, thank you!

I botched it while I did the refactoring *and* the original code wasn't even correct either. 

> As written, though, I'm very confused how it passes, as I thought (diff > 0) would become a boolean, and booleans don't have an all() method

`diff` was a tensor so the code didn't fail, but my assertion was still asserting the wrong thing.

src/torchcodec/samplers/_implem.py
NicolasHug: Done.

> We could also pass in the method to call for the actual decoding. But we can work on that later.

I thought about doing this. We can factorize this code to death, even the input validation and the `_build_all_clips_indices` / `_build_all_clips_timestamps` could be made generic.

For `_decode_all_clips_timestamps` / `_decode_all_clips_indices`, I feel like it's not worth it because we'll be re-writing them in C++ soon anyway. For the rest, let's leave it out for later. There will be a trade-off between compactness and clarity of the code.

src/torchcodec/samplers/_implem.py
NicolasHug: Oh nice, I didn't realize we could use `:=` in list comprehension like this!

PR #254, opened 2024-10-09
Add stream_index as an option to VideoDecoder

scotts:
Adds stream_index as an option to VideoDecoder. Example usage:
```
decoder = VideoDecoder("myfile.mp4", stream_index=2)
```
While this is a modest change in the interface, it required:

1. Adding the concept of streams to our test file generation.
2. Adding new test files for the new stream.
3. Adding the concept of streams to our testing utils.
4. Enhancing some existing tests to parameterize on the stream index.

I suggest reviewing the changes to VideoDecoder first. Everything else is a consequence of that.


src/torchcodec/decoders/_video_decoder.py
```diff
@@ -273,14 +282,23 @@ def get_frames_displayed_at(
...
     video_metadata = core.get_video_metadata(decoder)
 
     best_stream_index = video_metadata.best_video_stream_index
-    if best_stream_index is None:
+    if best_stream_index is None and stream_index is None:
         raise ValueError(
-            "The best video stream is unknown. " + _ERROR_REPORTING_INSTRUCTIONS
+            "The best video stream is unknown and there is no specified stream. "
+            + _ERROR_REPORTING_INSTRUCTIONS
         )
```
NicolasHug, line 294: Nit
We have 2 instances of `if stream_index` is None, and we're retrieving `best_stream_index` unconditionally while it should only be needed when `streaming_index` is None. I'd suggest to move this block within a single `if stream_index is None` below, starting at line 296

NicolasHug (approved):
Thanks @scotts , LGTM.

As an aside, I find that our testing framework is becoming more and more... frameworky. I fear that writing even simple test is going to get increasingly difficult, as one has to familiarize themselves with the testing framework, which itself grows in complexity over time.

scotts:
@NicolasHug, point taken about the testing framework being frameworky. My goal here is to make something that models the parts of video formats that we care about so that we can declaratively state such things when defining a new kind of test video. Basically, our testing logic is going to be a very simple model of the underlying FFmpeg file formats.

In general, adding a new fundamental capability in our decoder _will probably_ require changing some testing logic. For example, I anticipate having to make some more changes when we add audio decoding. This tendency will level-out as our public capabilities match the underlying capabilities in the C++ layer. This current change is the most disruptive change I can think of, as having to support multiple streams is far-reaching. Most future changes should be less involved.

In general, when adding a new media file for testing, _we will probably not_ require changing testing logic. We'll probably be able to just add a new declaration. I'm anticipating much more of adding new test files in the future, and much less of new fundamental capabilities. That is, I want it to be easy for us to add a new test file and state the metadata we need for tests in a declarative manner.

PR #259, opened 2024-10-11
Split samplers implementation across different files

NicolasHug:
This PR splits the existing `_implem.py` file into:

- `_time_based.py`
- `_index_based.py`
- `_common.py`

The policy-related tests are also extracted into a separate `test_policy` file.

Zero logic was changed. Only copy/pasting.


ahmadsharif1 (approved):
assuming 0 actual changes

PR #264, opened 2024-10-15
Unbreak CI by using C++17 code

ahmadsharif1:
std::string's starts_with method is c++20

So we use a different incantation that's C++17 compliant

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoderOps.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoderOps.cpp
@@ -168,7 +168,7 @@ void _add_video_stream(
   if (device.has_value()) {
     if (device.value() == "cpu") {
       options.device = torch::Device(torch::kCPU);
-    } else if (device.value().starts_with("cuda")) {
+    } else if (device.value().rfind("cuda", 0) == 0) { // starts with "cuda"
       std::string deviceStr(device.value());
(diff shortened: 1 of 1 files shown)


scotts (approved):

src/torchcodec/decoders/_core/VideoDecoderOps.cpp
```diff
@@ -168,7 +168,7 @@ void _add_video_stream(
   if (device.has_value()) {
     if (device.value() == "cpu") {
       options.device = torch::Device(torch::kCPU);
-    } else if (device.value().starts_with("cuda")) {
+    } else if (device.value().rfind("cuda", 0) == 0) { // starts with "cuda"
```
scotts, line 171: Since `rfind()` starts from the end ("reverse find"), this is technically "ends with 'cuda'". See https://en.cppreference.com/w/cpp/string/basic_string/rfind.

scotts:
Approving and merging to unbreak CI. We can figure out later if it should be `rfind()` or `find()`.

src/torchcodec/decoders/_core/VideoDecoderOps.cpp
ahmadsharif1: rfind(*, 0) looks for a prefix

PR #266, opened 2024-10-16
Pre-allocate tensors when possible to avoid copies

NicolasHug:


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -945,7 +956,8 @@ VideoDecoder::DecodedOutput VideoDecoder::getFrameDisplayedAtTimestampNoDemux(
         return seconds >= frameStartTime && seconds < frameEndTime;
       });
   // Convert the frame to tensor.
-  return convertAVFrameToDecodedOutput(rawOutput);
+  auto preAllocatedOutputTensor = torch::empty({0});
```
NicolasHug, line 959: We are now creating "fake pre-allocated tensors" in a bunch of places. Do I extract that out in a `Decoder::MakeDummyPreAllocatedTensor()` method to clarify the meaning behind all this?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -846,7 +846,8 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
 }
 
 VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
-    VideoDecoder::RawDecodedOutput& rawOutput) {
+    VideoDecoder::RawDecodedOutput& rawOutput,
+    torch::Tensor& preAllocatedOutputTensor) {
```
NicolasHug, line 850: I dislike that we always have to pass a tensor. The current logic is that if an empty tensor was passed, then it's up to `convertAVFrameToDecodedOutputOnCPU` to do the allocation.

I tried `std::optional<torch::Tensor>& tensor = std::nullopt` but unfortunately, this requires the tensor to be `const` :(

I tried overloading, but that would lead to overloading a tons of different functions (the calling stack is long) and make things way more complex.

I tried passing a tensor pointer, but that make lifetime management impossible.

Do we have a better option than what is currently implemented?

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -215,14 +215,19 @@ class VideoDecoder {
   // Decodes the frame where the current cursor position is. It also advances
   // the cursor to the next frame.
   DecodedOutput getNextDecodedOutputNoDemux();
+  DecodedOutput getNextDecodedOutputNoDemux(
+      torch::Tensor& preAllocatedOutputTensor);
```
NicolasHug, line 219: Note: I had to overload this one (only), because it's called in a ton of places in the C++ tests. Forcing to pass an empty tensor at all call-sites would be quite noisy

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Why do we need to create these tensors? I thought having an optional would allow us to only pass this in when we actually need it.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: I think we do.

rawOutput itself has a pointer in it that callers populate. You can set the value of that instead of passing in a pre-allocated tensor as an extra parameter.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: See https://github.com/pytorch/torchcodec/pull/266/files#r1803042354 for why I wasn't able to use optional here

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: The problem is that rawOutput isn't yet places like in `getFramesInRange`, where the output tensor is being pre-allocated. Can you confirm?

scotts: A potential solution to this and the question of creating a bunch of empty tensors is to create a single empty tensor shared by the class:
```
Tensor EMPTY_TENSOR = torch::empty({0});
```
And then we could change this function definition to be:
```
VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
    VideoDecoder::RawDecodedOutput& rawOutput,
    torch::Tensor& preAllocatedOutputTensor = &EMPTY_TENSOR) {
```
And then we do an identity-check on `EMPTY_TENSOR` in other places. We, of course, have to be very careful that we never modify `EMPTY_TENSOR`.

The other possibility is that we could just pass a pointer. That does not make lifetime management impossible, it just means we have to be very careful. In Rust terms, this would be a "borrow." The caller is responsible for the lifetime management of the object, but we are letting `convertAVFrameToDecodedOutput` _borrow_ the object to manipulate as it wants to. Once `convertAVFrameToDecodedOutput` returns back to us, we own the object again. In this world, the definition would be:
```
VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
    VideoDecoder::RawDecodedOutput& rawOutput,
    torch::Tensor* preAllocatedOutputTensor = nullptr) {
```
And then we do a `nullptr` check to know if we have a preallocated tensor. I prefer the pointer approach, as the reference approach kinda nullifies (ha!) the point (ha!) of references, which is that they always refer to a real object.

If we were using [`shared_ptr`](https://en.cppreference.com/w/cpp/memory/shared_ptr) elsewhere, we would use that, but we're not. We really only make use of [`unique_ptr`](https://en.cppreference.com/w/cpp/memory/unique_ptr), which I think is fine.

ahmadsharif1: As discussed IMHO we should use std::optional<torch::Tensor> t = std::nullopt (note that there is no reference here).

It really is an input-only parameter, not an inout parameter so there is no need for a tensor.

The caller wont really look at this this parameter again -- it should only look at the return value.

ahmadsharif1:
I will review the code in detail in a bit, but since this is a rather large C++ change can you run the perf benchmarks to make sure we don't regress?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -878,16 +881,24 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
...
   auto& streamInfo = streams_[streamIndex];
   if (output.streamType == AVMEDIA_TYPE_VIDEO) {
     if (streamInfo.colorConversionLibrary == ColorConversionLibrary::SWSCALE) {
-      int width = streamInfo.options.width.value_or(frame->width);
-      int height = streamInfo.options.height.value_or(frame->height);
-      torch::Tensor tensor = torch::empty(
-          {height, width, 3}, torch::TensorOptions().dtype({torch::kUInt8}));
+      torch::Tensor tensor;
+      if (preAllocatedOutputTensor.has_value()) {
+        // TODO: check shape of preAllocatedOutputTensor?
```
ahmadsharif1, line 893: I think we should TORCH_CHECK for height, width, shape, etc. here.

ahmadsharif1 (commented):
This change looks a lot better now. Thanks @NicolasHug 

I think we should fail loudly if the shapes don't match up

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1061,8 +1073,9 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesInRange(
   BatchDecodedOutput output(numOutputFrames, options, streamMetadata);
 
   for (int64_t i = start, f = 0; i < stop; i += step, ++f) {
-    DecodedOutput singleOut = getFrameAtIndex(streamIndex, i);
-    output.frames[f] = singleOut.frame;
+    auto preAllocatedOutputTensor = output.frames[f];
+    DecodedOutput singleOut =
+        getFrameAtIndex(streamIndex, i, preAllocatedOutputTensor);
```
scotts, line 1078: Since we're not using pointers, we can make this even simpler than before:
```
DecodedOutput singleOut =
        getFrameAtIndex(streamIndex, i, output.frames[f]);
```
That is, everything should work fine and be clear without a local alias.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: I have this a try thinking it would be a simple assert like

```py
assert `shape[-3] == H, W, 3`
```
 
But it turns out it's not as simple. Some tensors come as HWC while some other come pas HWC. This is because the pre-allocated batched tensors are allocated as such:

https://github.com/pytorch/torchcodec/blob/c6a0a5a079da408df56fd7c06e3b801cbada4db1/src/torchcodec/decoders/_core/VideoDecoder.cpp#L171-L186

It then me realize that everything works, but it's pretty magical. We end up doing the `.pemute()` calls in different places, but I think it would be a lot cleaner if we allocated batched output only as `NHWC`, and then permute this entire `NHWC` tensor in one go. What we do right now is that we permute all the N `HWC` tensors, and that's probably not as efficient (or as clean).

I want to fix this as an immediate follow-up if that's OK. I gave it a try here, but it's not trivial and it might be preferable not to overcomplexify this PR.

scotts: I'm in favor of @NicolasHug suggestion. The logic he points out is legacy from way back when, and it wasn't necessarily throught through in terms of long term maintenance and code health. Always doing it one way, and then permuting as needed on the way out, sounds easier and cleaner.

ahmadsharif1: Sounds good to me

ahmadsharif1 (approved):

ahmadsharif1:
Don't forget to run the benchmark. Sometimes small changes can cause regressions

test/decoders/test_video_decoder_ops.py
```diff
@@ -383,6 +384,48 @@ def test_color_conversion_library_with_scaling(
...
+        decoder = create_from_file(str(NASA_VIDEO.path))
+        _add_video_stream(
+            decoder,
+            color_conversion_library=color_conversion_library,
+            dimension_order=dimension_order,
+        )
+        scan_all_streams_to_update_metadata(decoder)
+
+        frame0_ref = NASA_VIDEO.get_frame_data_by_index(0)
+        C, H, W = frame0_ref.shape
```
NicolasHug, line 401:
```suggestion
```

NicolasHug:
Benchmark results look... good, but also fishy. The variance on the `main` branch is crazy high. I can reproduce the same results consistently. If anything, this PR reduces the variance *drastically* (and the decoding time too). But I would take those results with a pinch of salt. On my laptop, `main` and this PR give the same timings, within stddev.

```
PR devvm
get_frames_in_range, swscale
med = 67.56ms +- 4.25
get_frames_by_pts_in_range, swscale
med = 78.34ms +- 4.12
get_frames_in_range, filtergraph
med = 161.04ms +- 97.86
get_frames_by_pts_in_range, filtergraph
med = 199.17ms +- 114.61

main devvm

get_frames_in_range, swscale
med = 537.86ms +- 133.63
get_frames_by_pts_in_range, swscale
med = 594.93ms +- 149.99
get_frames_in_range, filtergraph
med = 410.08ms +- 99.76
get_frames_by_pts_in_range, filtergraph
med = 308.44ms +- 160.11
```

Benchmark code:
<details>

```py
import torch

    _add_video_stream,
    create_from_file,
    get_frames_by_pts_in_range,
    get_frames_in_range,
)

import torch


def bench(f, color_conversion_library, num_exp=100):

    times = []
    for _ in range(num_exp):
        VIDEO_PATH = "./test/resources/nasa_13013.mp4"
        decoder = create_from_file(VIDEO_PATH)
        _add_video_stream(
            decoder,
            color_conversion_library=color_conversion_library,
        )

        start = perf_counter_ns()
        f(decoder)
        end = perf_counter_ns()
        times.append(end - start)

    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    return med



NUM_EXP = 100
stream_index=3

def _get_frames_in_range(decoder):
    get_frames_in_range(decoder=decoder, stream_index=stream_index, start=0, stop=100)

def _get_frames_by_pts_in_range(decoder):
    get_frames_by_pts_in_range(decoder=decoder, stream_index=stream_index, start_seconds=0, stop_seconds=4)

for color_conversion_library in ("swscale", "filtergraph"):
    print(f"get_frames_in_range, {color_conversion_library}")
    times = bench(_get_frames_in_range, color_conversion_library=color_conversion_library, num_exp=NUM_EXP)
    report_stats(times)
    print(f"get_frames_by_pts_in_range, {color_conversion_library}")
    times = bench(_get_frames_by_pts_in_range, color_conversion_library=color_conversion_library, num_exp=NUM_EXP)
    report_stats(times)
```

</details>

PR #270, opened 2024-10-16
Add Nvidia licenses link to README.md

ahmadsharif1:


README.md
```diff
@@ -139,3 +139,7 @@ guide](CONTRIBUTING.md) for more details.
 ## License
 
 TorchCodec is released under the [BSD 3 license](./LICENSE).
+
+If you build TorchCodec with ENABLE_CUDA=1 or use the CUDA-enabled release of
+torchcodec please review
```
NicolasHug, line 144: I think we should wait until we have cuda wheels before we mention the "CUDA-enabled release", otherwise this might be confusing for users?

```suggestion
If you build TorchCodec with ENABLE_CUDA=1, please review
```

ahmadsharif1: I want to get the legal approval for it and not change it last minute. Hence I added a blanket clause. I can add a (when available) to it if you want.

NicolasHug: Can we get legal approval for the full sentence now, land with the suggestion, and re-add the "CUDA-enable release" at the time of release? It doesn't seem much more work than adding "when available" and then having to remove it?

ahmadsharif1: I am inclined not to change legally approved sentences.

We can delay merging this PR though

README.md
```diff
@@ -139,3 +139,8 @@ guide](CONTRIBUTING.md) for more details.
 ## License
 
 TorchCodec is released under the [BSD 3 license](./LICENSE).
+
+However, TorchCodec may be used with code not written by Meta which may be
+distributed under different licenses.
+
+For example, if you build TorchCodec with ENABLE_CUDA=1 or use the CUDA-enabled release of torchcodec, please review CUDA's license here: [Nvidia licenses](https://docs.nvidia.com/cuda/eula/index.html).
```
NicolasHug, line 146: Consider keeping the line length similar to the rest

NicolasHug (approved):

PR #273, opened 2024-10-17
Fix cuda builds for lower than 6.1 ffmpeg versions

ahmadsharif1:
Previously we were always compiling code that contained a macro introduced in a new ffmpeg version. That would fail because that macro was not defined.

This change guards `getFFMPEGContextFromExistingCudaContext` for older libavutil versions because it contains a macro that's introduced in FFMPEG 6 onwards.

Also add more builders for different ffmpeg versions


scotts:
Approving to unblock because we have builds breaking, but we want to refactor this logic so it's more obviously the correct thing. That is, we would ideally do this kind of macro check once, not in several places.

scotts (approved):

PR #277, opened 2024-10-18
Rework HWC / CHW dimension order conversions

NicolasHug:
This PR simplifies the HWC -> CHW dimension conversion, and reduces the expected input/output dimension order of some functions.

- Batched output tensors are now always created as NHWC. They are converted to NCHW in one single step, instead of converting HWC sub-tensors N times.
- As a consequence, the expected shape of input/allocated tensors within `convertAVFrameToDecodedOutputOnCPU` is now always `HWC`. And we can now enforce it (we couldn't before).
- `convertFrameToTensorUsingFilterGraph` now always returns a HWC tensor. Similarly, `convertFrameToBufferUsingSwsScale` now always expects (pointer over) a HWC tensor.
- The [N]HWC -> [N]CHW permutation is now centralized in a new function helper.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -878,7 +878,9 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
 
 void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
     VideoDecoder::RawDecodedOutput& rawOutput,
-    DecodedOutput& output) {
+    DecodedOutput& output,
+    torch::Tensor* tensor
+    ) {
   int streamIndex = rawOutput.streamIndex;
   AVFrame* frame = rawOutput.frame.get();
   auto& streamInfo = streams_[streamIndex];
@@ -886,15 +888,17 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
     if (streamInfo.colorConversionLibrary == ColorConversionLibrary::SWSCALE) {
       int width = streamInfo.options.width.value_or(frame->width);
       int height = streamInfo.options.height.value_or(frame->height);
-      torch::Tensor tensor = torch::empty(
+        auto tmp = torch::empty(
           {height, width, 3}, torch::TensorOptions().dtype({torch::kUInt8}));
-      rawOutput.data = tensor.data_ptr<uint8_t>();
+        tensor = &(tmp);
+      rawOutput.data = tensor->data_ptr<uint8_t>();
       convertFrameToBufferUsingSwsScale(rawOutput);
 
       if (streamInfo.options.dimensionOrder == "NCHW") {
-        tensor = tensor.permute({2, 0, 1});
+        auto tmp = tensor->permute({2, 0, 1});
+        tensor = &(tmp);
       }
-      output.frame = tensor;
+      output.frame = *tensor;
     } else if (
         streamInfo.colorConversionLibrary ==
         ColorConversionLibrary::FILTERGRAPH) {
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1184,7 +1184,7 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getNextRawDecodedOutputNoDemux() {
       getDecodedOutputWithFilter([this](int frameStreamIndex, AVFrame* frame) {
         StreamInfo& activeStream = streams_[frameStreamIndex];
         return frame->pts >=
-            activeStream.discardFramesBeforePts.value_or(INT64_MIN);
+            activeStream.discardFramesBeforePts;
       });
   return rawOutput;
 }
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -34,6 +34,31 @@ double ptsToSeconds(int64_t pts, const AVRational& timeBase) {
   return ptsToSeconds(pts, timeBase.den);
 }
 
+// Returns a [N]CHW *view* of a [N]HWC input tensor, if the options require so.
+// The [N] leading batch-dimension is optional i.e. the input tensor can be 3D
+// or 4D.
+// Calling permute() is guaranteed to return a view as per the docs:
+// https://pytorch.org/docs/stable/generated/torch.permute.html
+torch::Tensor MaybeHWC2CHW(
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    torch::Tensor& hwcTensor) {
+  if (options.dimensionOrder == "NHWC") {
+    return hwcTensor;
+  }
+  auto numDimensions = hwcTensor.dim();
+  auto shape = hwcTensor.sizes();
+  if (numDimensions == 3) {
+    TORCH_CHECK(shape[2] == 3, "Not a HWC tensor: ", shape);
+    return hwcTensor.permute({2, 0, 1});
+  } else if (numDimensions == 4) {
+    TORCH_CHECK(shape[3] == 3, "Not a HWC tensor: ", shape);
+    return hwcTensor.permute({0, 3, 1, 2});
+  } else {
+    TORCH_CHECK(
+        false, "Expected tensor with 3 or 4 dimensions, got ", numDimensions);
+  }
+}
+
 struct AVInput {
   UniqueAVFormatContext formatContext;
   std::unique_ptr<AVIOBytesContext> ioBytesContext;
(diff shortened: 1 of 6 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -916,6 +931,14 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
           std::to_string(static_cast<int>(streamInfo.colorConversionLibrary)));
     }
+    if (!preAllocatedOutputTensor.has_value()) {
+      // We only convert to CHW if a pre-allocated tensor wasn't passed. When a
+      // pre-allocated tensor is passed, it's up to the caller (typically a
+      // batch API) to do the conversion. This is more efficient as it allows
+      // batch NHWC tensors to be permuted only once, instead of permuting HWC
+      // tensors N times.
+      output.frame = MaybeHWC2CHW(streamInfo.options, output.frame);
+    }
```
NicolasHug, line 941: I still think there's some smell to this. Whether the tensor was pre-allocated and whether it should be permuted should be orthogonal concepts.
I think it would make sense for all the low-level decoding capabilities (including this function `convertAVFrameToDecodedOutputOnCPU` ) to only ever accept and return HWC tensors.

And it should be up to the higher-level decoding entry-points (basically the moral equivalent of the public methods) to do the conversion. It's not trivial because `getFrameAtIndex` is *both* an entry point *and* a sub-function of other entry-points. Maybe that also means we should already let *all* entry-points do their own allocation and always pass pre-allocated tensors?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -890,22 +900,27 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
+        auto shape = tensor.sizes();
+        TORCH_CHECK(
+            (shape.size() == 3) && (shape[0] == height) &&
+                (shape[1] == width) && (shape[2] == 3),
+            "Expected tensor of shape ",
+            height,
+            "x",
+            width,
+            "x3, got ",
+            shape);
```
NicolasHug, line 916: Any idea how to make this single call shorter :thinking: ?

scotts: I think you can do:
```
TORCH_CHECK(
            (shape.size() == 3) && (shape.equals({height, width, 3}),
            "Expected tensor of shape ",
            height,
            "x",
            width,
            "x3, got ",
            shape);
```
But I'm not sure. The main thing I'm not sure about is if an array literal will auto-convert into a the corresponding `ArrayRef`: https://pytorch.org/cppdocs/api/classc10_1_1_array_ref.html#_CPPv4NK3c108ArrayRef6equalsE8ArrayRef

NicolasHug: Ah, I was mainly hoping to avoid the 

```
            height,
            "x",
            width,
            "x3, got ",
            shape);
```

stack :p

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Agreed on your reasoning and the principles.

One way to square the circle is to split the public facing part of `getFrameAtIndex` from the actual work being done. We would have a public member ('getFrameAtIndex'). function and a private member function ('getFrameAtIndexInternal`, or something like that).
  * `getFrameAtInedex` would call `getFrameAtIndexInternal` for the actual work, and after, it would do the conversion check and the actual conversion if needed.
  * `getFrameAtIndexInternal` would just assume HWC tensors and do the real work.

Then all of the internal calls to `getFrameAtIndex` would become calls to `getFrameAtIndexInternal`. I've implementing variants this pattern several times before. We would repeat this for any public entry points that also need to do work for other public entry points, as needed.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -34,6 +34,31 @@ double ptsToSeconds(int64_t pts, const AVRational& timeBase) {
   return ptsToSeconds(pts, timeBase.den);
 }
 
+// Returns a [N]CHW *view* of a [N]HWC input tensor, if the options require so.
+// The [N] leading batch-dimension is optional i.e. the input tensor can be 3D
+// or 4D.
+// Calling permute() is guaranteed to return a view as per the docs:
+// https://pytorch.org/docs/stable/generated/torch.permute.html
+torch::Tensor MaybeHWC2CHW(
```
ahmadsharif1, line 42: Function names should have verbs in the name associated with the action they are doing.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: I am fine with low-level functions only dealing with HWC. AFAICT, most (all?) low-level code deals with HWC because it has better performance.

NicolasHug: Yeah, that sounds good. I'll try to implement that in a follow-up PR

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -167,28 +192,13 @@ VideoDecoder::BatchDecodedOutput::BatchDecodedOutput(
...
-         3},
-        {torch::kUInt8});
-  } else if (options.dimensionOrder == "NCHW") {
-    frames = torch::empty(
-        {numFrames,
-         3,
-         options.height.value_or(*metadata.height),
-         options.width.value_or(*metadata.width)},
-        torch::TensorOptions()
-            .memory_format(torch::MemoryFormat::ChannelsLast)
```
ahmadsharif1, line 185: I am wondering if using this is identical to permuting a NHWC tensor at the end. I am not 100% sure. Do you?

ahmadsharif1: I meant in terms of performance, is it identical? Specifically this is a bit concerning:




We really should have a batch benchmark because the whole point of batch is to do things in a performant way for many frames.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -34,6 +34,31 @@ double ptsToSeconds(int64_t pts, const AVRational& timeBase) {
...
+torch::Tensor MaybeHWC2CHW(
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    torch::Tensor& hwcTensor) {
+  if (options.dimensionOrder == "NHWC") {
+    return hwcTensor;
+  }
+  auto numDimensions = hwcTensor.dim();
+  auto shape = hwcTensor.sizes();
+  if (numDimensions == 3) {
+    TORCH_CHECK(shape[2] == 3, "Not a HWC tensor: ", shape);
```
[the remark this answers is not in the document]
NicolasHug: It will have false positive, but this is the very very best we can do at this stage. The alternative is to *not* check anything.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: Sorry I don't understand what you mean. What image did you mean to link to?

Note that this PR should e strictly more efficient:

> Batched output tensors are now always created as NHWC. They are converted to NCHW in one single step, instead of converting HWC sub-tensors N times.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: > FAICT, most (all?) low-level code deals with HWC

That's not quite the case. In `main` , `convertAVFrameToDecodedOutputOnCPU` accepts both HWC and CHW - this is actually what this PR is fixing.

It may still return both HWC and CHW (instead of just HWC), and this is what I want to fix as a follow-up

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Sorry the image wasn't uploaded properly:

image

I am not sure about the performance implications of doing a permute instead of `.to(channels_last)`

NicolasHug: Sorry I still don't really understand where you're coming from. Can you please share a link? Is this relevant for this PR? 

Again, the change involved in this PR is:

> Batched output tensors are now always created as NHWC. They are converted to NCHW in one single step, instead of converting HWC sub-tensors N times

NicolasHug: Thanks for the link.

We're not concerned about memory format (contiguous vs channels-last) in this PR. This is a related but distinct concern to the dimension order.

ahmadsharif1: Link in the edited comment above. It's from the channels-last page.

I am not 100% sure if creating a NHWC and permuting it is the same as creating a NCHW with channels-last and working with that. The code that you deleted was doing the latter.

A benchmark may show a difference -- or not. Do you know?

PR #280, opened 2024-10-22
Add sort and dedup logic in C++ to `getFramesAtIndices`

NicolasHug:
Benchmark shows no regression against main, and up to 10X speedup when indices are shuffled.

```
On main:
ordered indices: True
med = 22.55ms +- 1.75 

ordered indices: False
med = 309.52ms +- 11.83 

This PR:
ordered indices: True
med = 22.35ms +- 1.59 

ordered indices: False
med = 22.64ms +- 1.88 
```

<details>

```py

import random
import torch

    _add_video_stream,
    create_from_file,
    get_frames_at_indices,
)

import torch


def bench(f, num_exp=100, **kwargs):

    times = []
    for _ in range(num_exp):
        VIDEO_PATH = "./test/resources/nasa_13013.mp4"
        decoder = create_from_file(VIDEO_PATH)
        _add_video_stream(decoder)

        start = perf_counter_ns()
        f(decoder, **kwargs)
        end = perf_counter_ns()
        times.append(end - start)

def report_stats(times, unit="ms", suff=""):
    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    print(f"{med = :.2f}{unit} +- {std:.2f} {suff}")
    return med



NUM_EXP = 50

def _get_frames_at_indices(decoder, **kwargs):
    get_frames_at_indices(decoder=decoder, stream_index=3, **kwargs)

for ordered in (True, False):
    print(f"ordered indices: {ordered}")
    frame_indices = list(range(100))
    if not ordered:
        random.shuffle(frame_indices)
    # For main:
    # times = bench(_get_frames_at_indices, frame_indices=frame_indices,  num_exp=NUM_EXP)
    # report_stats(times)
    # For PR:
    for sort_indices in (True, False):
        times = bench(_get_frames_at_indices, frame_indices=frame_indices, sort_indices=sort_indices, num_exp=NUM_EXP)
        report_stats(times, suff=f"{sort_indices = }")
    print()
```

</details>

Still TODO for other PRs:

- add similar functionality pts
- expose publicly as method in `VideoDecoder`
- let the samplers rely on those C++ APIs.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1050,8 +1050,8 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
     if (options.colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH) {
       output.frames[f] = singleOut.frame;
     }
-    // Note that for now we ignore the pts and duration parts of the output,
-    // because they're never used in any caller.
+    output.ptsSeconds[f] = singleOut.ptsSeconds;
+    output.durationSeconds[f] = singleOut.durationSeconds;
   }
   output.frames = MaybePermuteHWC2CHW(options, output.frames);
   return output;
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1030,8 +1030,7 @@ VideoDecoder::DecodedOutput VideoDecoder::getFrameAtIndex(
 
 VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
     int streamIndex,
-    const std::vector<int64_t>& frameIndices,
-    const bool sortIndices) {
+    const std::vector<int64_t>& frameIndices) {
   validateUserProvidedStreamIndex(streamIndex);
   validateScannedAllStreams("getFramesAtIndices");
 
@@ -1040,12 +1039,15 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   const auto& options = stream.options;
   BatchDecodedOutput output(frameIndices.size(), options, streamMetadata);
 
-  // if frameIndices is [13, 10, 12, 11]
-  // when sorted, it's  [10, 11, 12, 13] <-- this is the sorted order we want
-  //                                         to use to decode the frames
-  // and argsort is     [ 1,  3,  2,  0]
+  auto indicesAreSorted =
+      std::is_sorted(frameIndices.begin(), frameIndices.end());
+
   std::vector<size_t> argsort;
-  if (sortIndices) {
+  if (!indicesAreSorted) {
+    // if frameIndices is [13, 10, 12, 11]
+    // when sorted, it's  [10, 11, 12, 13] <-- this is the sorted order we want
+    //                                         to use to decode the frames
+    // and argsort is     [ 1,  3,  2,  0]
     argsort.resize(frameIndices.size());
     for (size_t i = 0; i < argsort.size(); ++i) {
       argsort[i] = i;
@@ -1058,7 +1060,7 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
 
   auto previousIndexInVideo = -1;
   for (auto f = 0; f < frameIndices.size(); ++f) {
-    auto indexInOutput = sortIndices ? argsort[f] : f;
+    auto indexInOutput = indicesAreSorted ? f : argsort[f];
     auto indexInVideo = frameIndices[indexInOutput];
     if (indexInVideo < 0 || indexInVideo >= stream.allFrames.size()) {
       throw std::runtime_error(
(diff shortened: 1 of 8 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1030,7 +1030,8 @@ VideoDecoder::DecodedOutput VideoDecoder::getFrameAtIndex(
 
 VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
     int streamIndex,
-    const std::vector<int64_t>& frameIndices) {
+    const std::vector<int64_t>& frameIndices,
+    const bool sortIndices) {
```
NicolasHug, line 1034: The default of `sortIndices` is False right now. But I think it should be True. The benchmark shows that sorting an already-sorted frameIndices array is basically zero cost (within noise):

```
This PR:
ordered indices: True
med = 21.82ms +- 1.50 sort_indices = True
med = 21.71ms +- 1.38 sort_indices = False
```

So there seem to be no downside to always sorting by default.

OTOH, if we keep the default to False, some users might be missing out on a 10X speed-up, and that sounds overall worse.

scotts: Agreed on decision and reasoning.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1039,19 +1040,49 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
...
+          output.ptsSeconds[previousIndexInOutput];
+      output.durationSeconds[indexInOutput] =
+          output.durationSeconds[previousIndexInOutput];
+    } else {
+      DecodedOutput singleOut = getFrameAtIndex(
+          streamIndex, indexInVideo, output.frames[indexInOutput]);
+      if (options.colorConversionLibrary ==
+          ColorConversionLibrary::FILTERGRAPH) {
+        output.frames[indexInOutput] = singleOut.frame;
+      }
```
ahmadsharif1, line 1081: Why is this needed when we are passing output.frames[indexInOutput] in `getFrameAtIndex`?

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: IMHO we should just call std::is_sorted before sorting them (if they are found to be not sorted). I cannot imagine this parameter being useful to callers.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: It's because the pre-allocated buffer isn't used with filtergraph, only with swscale.

(Just a note that this is not something that was introduced in this PR, you'll see the same pattern in other callers)

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: I guess it's fine not to expose the `sortIndices` parameter at all.
But then, do we need to check for `is_sorted()` before we sort? Should we just always sort, and assume that the underlying implement is such that sorting an already-sorted array is O(n) anyway?

NicolasHug: We discussed offline, let's just do what Ahmad suggested.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1039,19 +1039,52 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   const auto& options = stream.options;
   BatchDecodedOutput output(frameIndices.size(), options, streamMetadata);
 
+  auto indicesAreSorted =
+      std::is_sorted(frameIndices.begin(), frameIndices.end());
+
+  std::vector<size_t> argsort;
+  if (!indicesAreSorted) {
```
scotts, line 1046: Digging a bit, I think we're probably better off not checking to see if the sequence is already sorted, and just always sorting. Modern implementations of `std::sort` seem to be [Introsort](https://en.wikipedia.org/wiki/Introsort), which was designed to be nearly linear with an already sorted sequence. I'm also fine if we commit this as is. We can always investigate more later if it becomes important.

scotts: Nit: I like to keep object definitions as close as possible to their initial use. So I'd prefer to see lines 1037-1040 appear after the sorting, right before we use `output` in the loop. (I just ran into this trying to find the definition of what `output` is when reading the loop.)

scotts (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: This seems like the wrong behavior. When a user passes in a pre-allocated tensor it should work with either color conversion library.

I guess this behavior is introduced by the PR to add pre-allocated tensor. That PR should have done it for either color conversion library. Maybe make that change first before merging in this PR? That would be my vote because otherwise the caller has to think about what color conversion library was used.

NicolasHug: I agree with you that the current behavior is potentially surprising.

I am open to re-working this eventually, but I want us to acknowledge that https://github.com/pytorch/torchcodec/pull/266 (and https://github.com/pytorch/torchcodec/pull/277) are *significantly* improving the existing code-base. They're not perfect, but they're clear improvements.

PR #282, opened 2024-10-22
Add sort and dedup logic in C++ with new `getFramesDisplayedByTimestamps` method / core API

NicolasHug:
Same as https://github.com/pytorch/torchcodec/pull/280 , but takes a list of pts as input instead of indices. This is needed for the samplers: https://github.com/pytorch/torchcodec/issues/256

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1090,11 +1090,11 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   return output;
 }
 
-VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtPtss(
+VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesDisplayedByTimestamps(
     int streamIndex,
-    const std::vector<double>& framePtss){
+    const std::vector<double>& framePtss) {
   validateUserProvidedStreamIndex(streamIndex);
-  validateScannedAllStreams("getFramesAtPtss");
+  validateScannedAllStreams("getFramesDisplayedByTimestamps");
 
   // The frame displayed at timestamp t and the one displayed at timestamp `t +
   // eps` are probably the same frame, with the same index. The easiest way to
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1050,8 +1050,8 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
     if (options.colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH) {
       output.frames[f] = singleOut.frame;
     }
-    // Note that for now we ignore the pts and duration parts of the output,
-    // because they're never used in any caller.
+    output.ptsSeconds[f] = singleOut.ptsSeconds;
+    output.durationSeconds[f] = singleOut.durationSeconds;
   }
   output.frames = MaybePermuteHWC2CHW(options, output.frames);
   return output;
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1100,6 +1100,8 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesDisplayedByTimestamps(
   // eps` are probably the same frame, with the same index. The easiest way to
   // avoid decoding that unique frame twice is to convert the input timestamps
   // to indices, and leverage the de-duplication logic of getFramesAtIndices.
+  // This means this function requires a scan.
+  // TODO: longer term, we should implement this without requiring a scan
 
   const auto& streamMetadata = containerMetadata_.streams[streamIndex];
   const auto& stream = streams_[streamIndex];
@@ -1119,8 +1121,8 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesDisplayedByTimestamps(
         stream.allFrames.begin(),
         stream.allFrames.end() - 1,
         framePts,
-        [&stream](const FrameInfo& info, double start) {
-          return ptsToSeconds(info.nextPts, stream.timeBase) <= start;
+        [&stream](const FrameInfo& info, double framePts) {
+          return ptsToSeconds(info.nextPts, stream.timeBase) <= framePts;
         });
     int64_t frameIndex = it - stream.allFrames.begin();
     frameIndices[i] = frameIndex;
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1040,11 +1040,13 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   const auto& options = stream.options;
   BatchDecodedOutput output(frameIndices.size(), options, streamMetadata);
 
-  std::vector<size_t> argsort(frameIndices.size());
-  for (size_t i = 0; i < argsort.size(); ++i) {
-    argsort[i] = i;
-  }
+  std::vector<size_t> argsort;
+
   if (sortIndices) {
+    argsort.resize(frameIndices.size());
+    for (size_t i = 0; i < argsort.size(); ++i) {
+      argsort[i] = i;
+    }
     std::sort(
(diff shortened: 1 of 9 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   return output;
 }
 
+VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesDisplayedByTimestamps(
```
NicolasHug, line 1093: For the **function** names, I based it on the existing `getFramesDisplayedByTimestampsInRange()`. I'm happy to bikeshed, but not on this PR please.
Note that our naming is becoming a bit inconsistent, e.g. the python names, ops names and C++ names are not always aligned. We probably want to clean that up, but that's for later.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
...
+    int streamIndex,
+    const std::vector<double>& framePtss) {
```
NicolasHug, line 1095: I'm not super happy about this parameter name `framePtss` in C++, `frame_ptss` in Python. It's the mirror of `frame_indices`. Happy to consider alternatives. The "single frame" equivalent in `get_frame_displayed_at` is just `seconds`.

ahmadsharif1: Maybe just `timestamps` since it's in the function name too?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
...
+  validateUserProvidedStreamIndex(streamIndex);
+  validateScannedAllStreams("getFramesDisplayedByTimestamps");
+
+  // The frame displayed at timestamp t and the one displayed at timestamp `t +
+  // eps` are probably the same frame, with the same index. The easiest way to
+  // avoid decoding that unique frame twice is to convert the input timestamps
+  // to indices, and leverage the de-duplication logic of getFramesAtIndices.
```
ahmadsharif1, line 1102: The long term thing to do here is to not require a scan for this function.

Scan should only be required for index-based functions where indexes need to be exact.

The index-based function can then call this function if the de-duping is done here and this doesn't need a scan.

That said you can merge this as-is and do the long-term thing later

NicolasHug: @ahmadsharif1 do you think it will be possible to correctly de-dup pts-based frame queries without going through indices?

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: It should probably be `framePtses` and `frame_ptses`, following the English rule of adding "es" to make something that ends in "s" plural. That's still not great, but the current version definitely looks like a typo.

scotts: I'm happy with `timestamps` - there was a race condition in commenting. :)

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: This is what I mean by having explicit seeking modes (approximate versus exact) is a big conceptual change. :) It might only requires changing a few dozen lines of code, but it will require changes in many places, and we'll always have to be aware of what mode is active.

At the moment, we're implicitly exact when we need to do things with indices. I think we should just keep being that way as needed, and when we implement the different modes, we can reason about what changes to make wholistically.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: I think `timestamps` is a bit more intuitive to me than `frame_timestamps` because these are random time points given by the user -- not necessarily corresponding to `frame.pts` or `frame.pts + frame.duration` (they could be somewhere in between).

Heck they could even be outside the min and max pts.

But whatever you choose is okay

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
   return output;
 }
 
+VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesDisplayedByTimestamps(
+    int streamIndex,
+    const std::vector<double>& framePtss) {
+  validateUserProvidedStreamIndex(streamIndex);
```
ahmadsharif1, line 1096: Maybe add a TODO saying long term we should not require scanning the file for time-based frame extraction

ahmadsharif1 (approved):

ahmadsharif1 (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
...
+    auto framePts = framePtss[i];
+    TORCH_CHECK(
+        framePts >= minSeconds && framePts < maxSeconds,
+        "frame pts is " + std::to_string(framePts) + "; must be in range [" +
+            std::to_string(minSeconds) + ", " + std::to_string(maxSeconds) +
+            ").");
+
+    auto it = std::lower_bound(
+        stream.allFrames.begin(),
+        stream.allFrames.end() - 1,
```
scotts, line 1120: Are you intentionally not including the last frame? Note that by convention, the `.end()` method on C++ containers is actually an iterator _past_ the end: https://en.cppreference.com/w/cpp/container/vector/end. So this code is actually eliminating the final element from the range.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: SG, I'll go with `timestamps` then

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1090,6 +1090,45 @@ VideoDecoder::BatchDecodedOutput VideoDecoder::getFramesAtIndices(
...
+        framePts,
+        [&stream](const FrameInfo& info, double start) {
+          return ptsToSeconds(info.nextPts, stream.timeBase) <= start;
```
scotts, line 1123: Nit: let's rename `start` to something more indicative of what we're looking for. Using `framePts` is fine, as it will reinforce that it will take on the `framePts` value passed into `lower_bound`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Hmmm. I think you're right. We're eliminating the final element, but if we should have gotten the final element, it will become that anyway. This works for, I think, two resons:

1. Because we've already eliminated the possibility of a pts value outside of the valid range because of the `TORCH_CHECK` above.
2. Because we're converting iterators to indices, and `.end()` becomes equivalent to `.size()` with the subtraction method we're using.

But, I think I actually prefer using `min` to fix things up after. That this works is so subtle, and doing `.end() - 1` in an iterator range is so unusual, I'd rather have the `min` approach with a comment explaining why we need it. I think it makes the code more understandable. The comment would be something like:
```
// If the frame index is larger than the size of all frames, that means we 
// couldn't match the pts value to the pts value of a NEXT FRAME. And 
// that means that this timestamp falls during the time between when the 
// last frame is displayed, and the video ends. Hence, it should map to the
// index of the last frame.

File: test/decoders/test_video_decoder_ops.py [hunk 7]
```diff
@@ -155,6 +156,36 @@ def test_get_frames_at_indices_unsorted_indices(self):
...
+            stream_index=stream_index,
+            timestamps=timestamps,
+        )
+        for frame, expected_frame in zip(frames, expected_frames):
+            assert_tensor_equal(frame, expected_frame)
+
+        # # first and last frame should be equal, at pts=2 [+ eps]. We then
+        # modify the # first frame and assert that it's now different from the
+        # last frame.  # This ensures a copy was properly made during the
+        # de-duplication logic.
```
scotts, line 183: Nit: looks like auto-formatting had some problems when merging line comments.

scotts (approved):

PR #283, opened 2024-10-23
Frame and FrameBatch improvements

NicolasHug:
This PR:

- Adds a bunch of input checks when creating a `Frame` or a `FrameBatch`. We enforce `Frame` data to be 3D, and `FrameBatch` data to be >= 4D. We also ensure consistency of leading dimensions between data, pts_seconds and duration_seconds
- Adds indexing support for `FrameBatch`. This supports pytorch fancy indexing naturally and intuitively. Note that indexing a 4D `FrameBatch` returns a `Frame`.
- Adds iteration support for `FrameBatch`. This removes the "tuple unpacking" behavior that we had for `FrameBatch`, but luckily this is not something we have been using at all. The unpacking behavior of `Frame` is preserved.

These changes are mostly necessary in order for us to change the output of the samples from `List[FrameBatch(4D)]` to `FrameBatch (5D)`, as done in https://github.com/pytorch/torchcodec/pull/284

### DIFF
--- a/src/torchcodec/_frame.py
+++ b/src/torchcodec/_frame.py
@@ -39,6 +39,8 @@ class Frame(Iterable):
     """The duration of the frame, in seconds (float)."""
 
     def __post_init__(self):
+        # This is called after __init__() when a Frame is created. We can run
+        # input validation checks here.
         if not self.data.ndim == 3:
             raise ValueError(f"data must be 3-dimensional, got {self.data.shape = }")
         self.pts_seconds = float(self.pts_seconds)
@@ -64,6 +66,8 @@ class FrameBatch(Iterable):
     """The duration of the frame, in seconds (1-D ``torch.Tensor`` of floats)."""
 
     def __post_init__(self):
+        # This is called after __init__() when a FrameBatch is created. We can
+        # run input validation checks here.
         if self.data.ndim < 4:
             raise ValueError(
                 f"data must be at least 4-dimensional. Got {self.data.shape = } "
--- a/src/torchcodec/_frame.py
+++ b/src/torchcodec/_frame.py
@@ -38,6 +38,13 @@ class Frame(Iterable):
     duration_seconds: float
     """The duration of the frame, in seconds (float)."""
 
+    def __post_init__(self):
+        if not self.data.ndim == 3:
+            raise ValueError(f"data must be 3-dimensional, got {self.data.shape = }")
+
+        self.pts_seconds = float(self.pts_seconds)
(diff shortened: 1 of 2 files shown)


src/torchcodec/_frame.py
```diff
@@ -57,9 +63,52 @@ class FrameBatch(Iterable):
...
+                data=data,
+                pts_seconds=float(pts_seconds.item()),
+                duration_seconds=float(duration_seconds.item()),
+            )
+        else:
+            return FrameBatch(
+                data=data,
+                pts_seconds=pts_seconds,
+                duration_seconds=duration_seconds,
+            )
```
ahmadsharif1, line 108: Tensor has a .item() method for returning the underlying dtype.

Should we have something like that here? i.e. always return a FrameBatch but return a Frame if .item() is called?

src/torchcodec/_frame.py
```diff
@@ -38,6 +38,12 @@ class Frame(Iterable):
     duration_seconds: float
     """The duration of the frame, in seconds (float)."""
 
+    def __post_init__(self):
```
ahmadsharif1, line 41: Can you describe in a comment why this dunder method was added? (for python newbies)

src/torchcodec/_frame.py
NicolasHug: > always return a FrameBatch but return a Frame if .item() is called?

I feel like this is what we're already doing, but perhaps I'm misunderstanding?

BTW, this quirk is only needed for mypy (sigh). Originally the code was just:

```py
        cls = Frame if self.data.ndim == 4 else FrameBatch
        return cls(
            self.data[key],
            self.pts_seconds[key],
            self.duration_seconds[key],
        )
```

and everything was fine, and the Frame would get proper float value because of what we do in its `post_init`. But mypy was complaining so I had to go for this in [`4661237` (#283)](https://github.com/pytorch/torchcodec/pull/283/commits/46612374677ff4dbbea5fdf2b175bc1090c4d1fa)

scotts (approved):

src/torchcodec/_frame.py
ahmadsharif1: > I feel like this is what we're already doing, but perhaps I'm misunderstanding?

Don't we return a Frame for the special case of dimensions=4?

What I am saying is we should return a FrameBatch even in that case (of size 1). So we are consistent with Tensor

I'll leave it to you

NicolasHug: I don't have a super strong preference on this - let me open an issue so we can discuss during one of the meetings

PR #287, opened 2024-10-24
Fix pts -> index conversion in `get_frame_at_timestamps[_in_range]`

NicolasHug:
This is a fix for the bug described in https://github.com/pytorch/torchcodec/pull/286. We use a sightly different solution as suggested in https://github.com/pytorch/torchcodec/pull/286#issuecomment-2435362727.

This PR fixes the behavior of both `get_frame_at_timestamps()` and `get_frames_at_timestamps_in_range()`.

### DIFF
--- a/test/decoders/test_video_decoder_ops.py
+++ b/test/decoders/test_video_decoder_ops.py
@@ -186,6 +186,66 @@ class TestOps:
         with pytest.raises(AssertionError):
             assert_tensor_equal(frames[0], frames[-1])
 
+    def test_pts_apis_against_index_ref(self):
+        # Non-regression test for https://github.com/pytorch/torchcodec/pull/286
+        # Get all frames in the video, then query all frames with all time-based
+        # APIs exactly where those frames are supposed to start. We assert that
+        # we get the expected frame.
+        decoder = create_from_file(str(NASA_VIDEO.path))
+        scan_all_streams_to_update_metadata(decoder)
+        add_video_stream(decoder)
+
+        metadata = get_json_metadata(decoder)
+        metadata_dict = json.loads(metadata)
+        num_frames = metadata_dict["numFrames"]
+        assert num_frames == 390
+
+        stream_index = 3
+        _, all_pts_seconds_ref, _ = zip(
+            *[
+                get_frame_at_index(
+                    decoder, stream_index=stream_index, frame_index=frame_index
+                )
+                for frame_index in range(num_frames)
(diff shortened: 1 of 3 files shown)


test/decoders/test_video_decoder_ops.py
```diff
@@ -186,6 +186,66 @@ def test_get_frames_by_pts(self):
...
+        )
+        pts_seconds = torch.tensor(pts_seconds)
+        assert_tensor_equal(pts_seconds, all_pts_seconds_ref)
+
+        _, pts_seconds, _ = get_frames_by_pts_in_range(
+            decoder,
+            stream_index=stream_index,
+            start_seconds=0,
+            stop_seconds=all_pts_seconds_ref[-1] + 1e-4,
+        )
```
NicolasHug, line 227: Note: all test-cases above were passing on main.

The ones *below* this line were failing. We never caught the one on `get_frames_by_pts_in_range` because it was only triggered when `start_seconds` was later than the second-to-last frame. And the error message is quite.... interesting!

```py
    def __call__(self, /, *args, **kwargs):
>       return self._op(*args, **kwargs)
E       RuntimeError: Trying to create tensor with negative dimension -1: [-1, 270, 480, 3]

```

test/decoders/test_video_decoder_ops.py
```diff
@@ -186,6 +186,66 @@ def test_get_frames_by_pts(self):
...
+        assert_tensor_equal(pts_seconds, all_pts_seconds_ref)
+
```
NicolasHug, line 229: Note: all test-cases above were passing on main.

The ones *below* this line were failing. We never caught the one on `get_frames_by_pts_in_range` because it was only triggered when `start_seconds` was later than the second-to-last frame. And the error message is quite.... interesting!

```py
    def __call__(self, /, *args, **kwargs):
>       return self._op(*args, **kwargs)
E       RuntimeError: Trying to create tensor with negative dimension -1: [-1, 270, 480, 3]

```

ahmadsharif1 (approved):
Great catch!

NicolasHug:
Thanks for the review Ahmad. @scotts , I'll merge this now so I can move https://github.com/pytorch/torchcodec/pull/284 forward, but I'll happily address any further comment you may have

scotts:
Just commenting to chime in on the actual merged PR that is is a great catch. Thanks for fixing my bug. :)

PR #291, opened 2024-10-24
Add option for the user to pass in ffmpeg thread count

ahmadsharif1:
This can control the number of FFMPEG threads and can improve throughput if `num_threads=1` is used (which is the default now).

For a comprehensive benchmark analysis look at this:

https://github.com/ahmadsharif1/torchcodec/blob/bench1/README.md#benchmark-results

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -36,6 +36,11 @@ class VideoDecoder:
             This can be either "NCHW" (default) or "NHWC", where N is the batch
             size, C is the number of channels, H is the height, and W is the
             width of the frames.
+        num_ffmpeg_threads (int, optional): The number of threads to use for decoding.
+            Use 1 for single-threaded decoding which is best if you are running multiple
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_video_decoder.py
```diff
@@ -36,6 +36,11 @@ class VideoDecoder:
             This can be either "NCHW" (default) or "NHWC", where N is the batch
             size, C is the number of channels, H is the height, and W is the
             width of the frames.
+        num_ffmpeg_threads (int, optional): The number of threads to use for decoding.
+            Use 1 for single-threaded decoding which is best if you are running multiple
```
NicolasHug, line 40: Nit, as we don't want to  be overly prescriptive. E.g. using `num_ffmpeg_threads=2` with num_data_loader_workers=2 should be completely fine if the user has more than 4 cores.

```suggestion
            Use 1 for single-threaded decoding which may be best if you are running multiple
```

NicolasHug (approved):
Thank you @ahmadsharif1 !

PR #292, opened 2024-10-24
Use test-infra's main branch now that https://github.com/pytorch/test-infra/pull/5789 is merged

ahmadsharif1:
We were only using my fork because https://github.com/pytorch/test-infra/pull/5789 was not merged


NicolasHug (approved):
Sweet, thank you

PR #294, opened 2024-10-25
Add fps info to samplers benchmark

NicolasHug:
Output now looks like

```
----------
num_clips = 1
clips_at_random_indices     med = 12.57ms +- 3.51  med fps = 795.5
clips_at_regular_indices    med = 8.60ms +- 0.55  med fps = 1163.3
clips_at_random_timestamps  med = 16.37ms +- 3.86  med fps = 610.9
clips_at_regular_timestamps med = 9.77ms +- 0.97  med fps = 1023.7
----------
num_clips = 50
clips_at_random_indices     med = 95.78ms +- 3.56  med fps = 5220.2
clips_at_regular_indices    med = 101.75ms +- 2.90  med fps = 4914.1
clips_at_random_timestamps  med = 100.78ms +- 2.34  med fps = 4961.4
clips_at_regular_timestamps med = 106.48ms +- 3.71  med fps = 4601.9
```

### DIFF
--- a/benchmarks/samplers/benchmark_samplers.py
+++ b/benchmarks/samplers/benchmark_samplers.py
@@ -16,16 +16,22 @@ def bench(f, *args, num_exp=100, warmup=0, **kwargs):
     for _ in range(warmup):
         f(*args, **kwargs)
 
+    num_frames = None
     times = []
     for _ in range(num_exp):
         start = perf_counter_ns()
-        f(*args, **kwargs)
+        clips = f(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


ahmadsharif1:
Can you output the video as well?

Long term you can output the codec and resolution as well. We do the same in benchmark_decoders.py and should probably make it into a common function later.

NicolasHug:
If that's OK I'll wait until the benchmark supports arbitrary videos. It's definitely in scope for the future

PR #295, opened 2024-10-25
Honor `preAllocatedOutputTensor` with filtergraph

NicolasHug:
Up until now, the `preAllocatedOutputTensor` parameter of our low-level decoding functions was only considered for `swscale`. It is now honored in all cases, i.e. when `filtergraph` is used.


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -890,6 +890,14 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
   return output;
 }
 
+// Note [preAllocatedOutputTensor with swscale and filtergraph]:
+// Callers may pass a pre-allocated tensor, where the output frame tensor will
+// be stored. This parameter is honored in any case, but it only leads to a
+// speed-up when swscale is used. With swscale, we can tell ffmpeg to place the
+// decoded frame directly into `preAllocatedtensor.data_ptr()`. That's not
+// possible with filtegraph, leading to an extra copy.
```
ahmadsharif1, line 898: I don't think it's impossible with filtergraph -- we just don't know and I haven't investigated this further.

If ffmpeg allows us to pass in pre-allocated frames to filtergraph we can do this more efficiently for filtergraph too. I would add a TODO to investigate and fix this for filtergraph later.

ahmadsharif1:
Can you add a TODO to write a test that confirms the data_ptr in the pre-allocated tensor is the same as the return value in both color-conversion library cases? Otherwise this can regress and we would not know any better.

Doesn't have to be this diff. You can do it later.

ahmadsharif1 (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -926,6 +934,9 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
         streamInfo.colorConversionLibrary ==
         ColorConversionLibrary::FILTERGRAPH) {
       output.frame = convertFrameToTensorUsingFilterGraph(streamIndex, frame);
+      if (preAllocatedOutputTensor.has_value()) {
+        preAllocatedOutputTensor.value().copy_(output.frame);
```
NicolasHug, line 938: @ahmadsharif1 you suggested

> Can you add a TODO to write a test that confirms the data_ptr in the pre-allocated tensor is the same as the return value in both color-conversion library cases? Otherwise this can regress and we would not know any better.


Just to clarify, you're suggesting to assert that `output.data_ptr() == preAllocatedOutputTensor.data_ptr()` ?

I think this will hold for swscale, but not for filtegraph. In the `copy_` above my understanding is that we copy the *data* at `output.data_ptr() ` into `preAllocatedOutputTensor.data_ptr()`, but the pointers themselves may live at different locations.

I guess that means that the caller should *not* rely on the `output.frame` if a pre-allocated tensor was passed. Are we OK with that?

test/decoders/VideoDecoderTest.cpp
```diff
@@ -223,6 +223,28 @@ TEST_P(VideoDecoderTest, DecodesFramesInABatchInNCHW) {
   EXPECT_TRUE(torch::equal(tensor[1], tensorTime6FromFFMPEG));
 }
 
+void testDecoderWithColorConversion(const std::string& colorConversionLibrary) {
+  std::string path = getResourcePath("nasa_13013.mp4");
+  auto preAllocatedOutputTensor = torch::empty({270, 480, 3}, {torch::kUInt8});
+
+  std::unique_ptr<VideoDecoder> ourDecoder =
+      createDecoderFromPath(path, GetParam());
```
NicolasHug, line 231: So compilation fails on this line because I assume `GetParam` is only valid within a `TEST_P` function.

@ahmadsharif1 @scotts any tip on how to write this test over both `swscale` and `filtergraph` without just duplicating the code?

ahmadsharif1: gtest has parameterized tests:

https://google.github.io/googletest/reference/testing.html#TEST_P

ahmadsharif1: You could call getparam from the parent and pass it down here?

The other way of doing this is parameterized classes but that seems overkill to me.

NicolasHug: Thanks for your feedback. I tried a few things, but couldn't get it to work (new kind of error each time). I just went for duplicating the whole thing, I hope that's OK. We already have some duplicated tests in this class.

PR #296, opened 2024-10-25
Indexing 4D FrameBatch now returns FrameBatch

NicolasHug:
Previously it returned a `Frame`.

Closes https://github.com/pytorch/torchcodec/issues/288, follow-up from https://github.com/pytorch/torchcodec/pull/283#discussion_r1815047315

I don't have a strong preference on this, I see value with both options. I'm opening this PR so we can decide.

The good news is that this change does not require any change to the samplers test, which means the UX should overall be largely unchanged.


ahmadsharif1 (commented):
I like this change because it makes FrameBatch consistent with Tensor itself.

Should we have an `.item()` method similar to Tensor to get the underlying Frame for a single-element FrameBatch?

scotts:
I have a slight preference for this change over the existing code, mainly because this new code is conceptually simpler: iterating over an N-dimensional `FrameBatch` always yields an N-1-dimensional `FrameBatch`. I feel that's easier to reason about and document.

If users end-up writing their own code to "if N == 3, turn into `Frame`" then we can revisit. On `.item()`, that's the kind of I'd want to see more use of before adding. In general, I don't have a strong preference here because I'm just not sure how it will be used. Absent not knowing how it will be used, I fall back on what seems simpler.

scotts (approved):

PR #299, opened 2024-10-25
Delete cuda_tests.yaml -- we have cuda tests in linux_cuda_wheel.yaml

ahmadsharif1:
Deleting this file to reduce maintenance burden.

Initially I thought of keeping this one for C++ tests, etc. but we don't really have cuda C++ tests anymore


NicolasHug (approved):
LGTM but do we actually still need this job now that we have the wheels?

ahmadsharif1:
I don't think we need this.

We had some C++ CUDA tests in my original August implementation that weren't tested by any other yml file.

I actually got rid of those and now we only have pytests.

I'll delete it to reduce the maintenance burden

NicolasHug (approved):

NicolasHug:
Just rename this PR title before merging :) !

PR #303, opened 2024-10-28
Expose new `get_frames_at` and `get_frames_displayed_at` methods

NicolasHug:
As in title. Needs https://github.com/pytorch/torchcodec/pull/302 first.

### DIFF
--- a/test/decoders/test_video_decoder.py
+++ b/test/decoders/test_video_decoder.py
@@ -459,7 +459,7 @@ class TestVideoDecoder:
             lambda decoder: decoder.get_frames_in_range(0, 4).data,
             lambda decoder: decoder.get_frame_displayed_at(0).data,
             # TODO: uncomment once D60001893 lands
-            # lambda decoder: decoder.get_frames_displayed_at(0, 1).data,
+            # lambda decoder: decoder.get_frames_displayed_in_range(0, 1).data,
         ),
     )
     def test_dimension_order(self, dimension_order, frame_getter):
@@ -487,7 +487,7 @@ class TestVideoDecoder:
         decoder = VideoDecoder(NASA_VIDEO.path, stream_index=stream_index)
 
         # Note that we are comparing the results of VideoDecoder's method:
-        #   get_frames_displayed_at()
+        #   get_frames_displayed_in_range()
         # With the testing framework's method:
         #   get_frame_data_by_range()
         # That is, we are testing the correctness of a pts-based range against an index-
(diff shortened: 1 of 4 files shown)


test/decoders/test_video_decoder.py
```diff
@@ -365,6 +401,47 @@ def test_get_frame_displayed_at_fails(self):
         with pytest.raises(IndexError, match="Invalid pts in seconds"):
             frame = decoder.get_frame_displayed_at(100.0)  # noqa
 
+    def test_get_frames_displayed_at(self):
+
+        decoder = VideoDecoder(NASA_VIDEO.path)
+        ref_frame6 = NASA_VIDEO.get_frame_by_name("time6.000000")
```
scotts, line 407: I prefer that we use the testing utilities rather than hard-coding these values. What I did when testing pts-based APIs is that I still compared them to frames I got by index in the testing framework - we know what frame it's going to be, and that alleviates the testing framework from also supporting pts-based access.

It's difficult to verify hard-coded values, and it's particularly difficult to know if the same hard-coded values are supposed to represent the same actual thing. The framework has these values in one place, and also explains how we got the values, and potentially how we would update them if we ever need to.

test/decoders/test_video_decoder.py
```diff
@@ -340,6 +341,41 @@ def test_get_frame_at_fails(self):
...
+        frames = decoder.get_frames_at(indices)
+
+        assert isinstance(frames, FrameBatch)
+
+        for i in range(len(indices)):
+            assert_tensor_equal(
+                frames[i].data, NASA_VIDEO.get_frame_data_by_index(indices[i])
+            )
+
+        expected_pts_seconds = torch.tensor([1.1678, 0.8342], dtype=torch.float64)
```
scotts, line 357: Same comment below about hard-coded values.

test/decoders/test_video_decoder.py
```diff
@@ -301,9 +302,12 @@ def test_get_frame_at(self):
...
         assert_tensor_equal(ref_frame9, frame9.data)
         assert isinstance(frame9.pts_seconds, float)
-        assert frame9.pts_seconds == pytest.approx(0.3003)
+        expected_frame_info = NASA_VIDEO.get_frame_info(9)
+        assert frame9.pts_seconds == pytest.approx(expected_frame_info.pts_seconds)
         assert isinstance(frame9.duration_seconds, float)
-        assert frame9.duration_seconds == pytest.approx(0.03337, rel=1e-3)
+        assert frame9.duration_seconds == pytest.approx(
+            expected_frame_info.duration_seconds, rel=1e-3
+        )
```
NicolasHug, line 310: This is just a drive-by clean-up to remove hard-coded values.

scotts (approved):

PR #308, opened 2024-10-28
Relax test constraint for ffmpeg versions as they can be arbitrary

ahmadsharif1:
ffmpeg could be built from source in which case the version may not be a release version and the version could just be a short hash.

Moreover some ffmepg versions on conda-forge don't have a minor or micro version.


scotts (approved):

NicolasHug (approved):
LGTM.

I know we discussed this in the past, but I still think it's be a better UX if we returned a 3-tuple **when we can** (and then a str otherwise).

PR #310, opened 2024-10-28
Remove `get_frame_by_name` test util

NicolasHug:
Closes https://github.com/pytorch/torchcodec/issues/304


test/resources/nasa_13013.mp4.stream3.frame000180.pt
NicolasHug, line 1: These are symlinks, so they're super cheap. I'll figure out in another PR whether we can just plainly rename/remove the existing "timexxx" frames.

scotts (approved):

PR #313, opened 2024-10-28
Make ffmpeg build workflow execute when script changes

scotts:
Before this, we were hacking the workflow definition to execute on a PR to test it. We don't want this running on _every_ PR, because that would be wasteful and noisy. But we can easily make it run when the FFMpeg building script has changed, which is what we wanted before.

Previous versions of this PR had a trivial change to trigger CI. One of those runs: https://github.com/pytorch/torchcodec/actions/runs/11560100235?pr=313. The current one does not have that trivial change to the script, and the test does not run.

### DIFF
--- a/.github/workflows/build_ffmpeg.yaml
+++ b/.github/workflows/build_ffmpeg.yaml
@@ -11,6 +11,9 @@ name: Build non-GPL FFmpeg from source
 
 on:
   workflow_dispatch:
+  pull_request:
+    paths:
+      - packaging/build_ffmpeg.sh
   schedule:
     - cron: '0 0 * * 0'  # on sunday
 
@@ -46,13 +49,12 @@ jobs:
       fail-fast: false
       matrix:
         ffmpeg-version: ["4.4.4", "5.1.4", "6.1.1", "7.0.1"]
-        runner: ["macos-m1-stable"]
     uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
     with:
       job-name: Build
       upload-artifact: ffmpeg-lgpl
       repository: pytorch/torchcodec
-      runner: "${{ matrix.runner }}"
+      runner: macos-14-xlarge
       script: |
         export FFMPEG_VERSION="${{ matrix.ffmpeg-version }}"
         export FFMPEG_ROOT="${PWD}/ffmpeg"


.github/workflows/build_ffmpeg.yaml
```diff
@@ -10,7 +10,9 @@
 name: Build non-GPL FFmpeg from source
 
 on:
-  workflow_dispatch:
+  pull_request:
```
NicolasHug, line 13: I think we still need the `workflow_dispatch` stuff for us to trigger this workflow manually

.github/workflows/build_ffmpeg.yaml
```diff
@@ -46,13 +48,12 @@ jobs:
...
       matrix:
         ffmpeg-version: ["4.4.4", "5.1.4", "6.1.1", "7.0.1"]
-        runner: ["macos-m1-stable"]
     uses: pytorch/test-infra/.github/workflows/macos_job.yml@main
     with:
       job-name: Build
       upload-artifact: ffmpeg-lgpl
       repository: pytorch/torchcodec
-      runner: "${{ matrix.runner }}"
+      runner: macos-14-xlarge
```
NicolasHug, line 56: Q - what's the reason for changing the worker?

NicolasHug (approved):
Approving to unblock but let's make sure we can still trigger it manually (we can merge the PR now and then verify and fix later if needed)

.github/workflows/build_ffmpeg.yaml
scotts: Ah! I misread the docs. I thought we only needed that if we wanted to have inputs. Let's fix now.

.github/workflows/build_ffmpeg.yaml
scotts: To avoid the long macos-m1-stable queue. This is a GitHub hosted runner that seems to get a worker faster. I have a lot of difficulty getting a straight-forward answer to what Mac runners are arm64 versus x86, but this blog post indicates that macos-14-xlarge is arm64 (M1): https://github.blog/news-insights/product-news/introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/

PR #315, opened 2024-10-28
Added benchmark chart to README.md

ahmadsharif1:
This PR:

1. Adds some code to generate synthetic videos.
2. Adds a script that the user can run to generate a png file with benchmark results comparing torchcodec with torchvision and torchaudio. It also compares it to torchcodec with 1 thread.
The benchmark is run on 3 seeking patterns -- just getting the first frame, getting 10 consecutive frames and seeking uniformly and decoding a frame after seeking.
3. Adds this figure to the readme.


scotts:
This is great! Are we also okay with comparing to Decord?

NicolasHug:
Should we also compare a call to `.get_frames_at_indices` (or pts)? I suspect torchcodec will shine compared to (at least) torchvision?

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -299,6 +301,55 @@ def create_torchcodec_decoder_from_file(video_file):
...
+                                pix_fmt,
+                                outfile,
+                            ]
+                            executor.submit(generate_video, command)
+                            video_count += 1
+
+    executor.shutdown(wait=True)
+    print(f"Generated {video_count} videos")
+
+
```
NicolasHug, line 352: On the plot below the "decoder" legend is repeated: on the left, and on every single plot. It might be best to only keep one so as not to overload the plots

scotts: Agreed it would look best with a single legend outside of all plots. I didn't say anything before because I wasn't sure how, but digging through old matplotlib code from an old research project, I think I found an example: [experiments/process_bulk_evict.py](https://github.com/IBM/sliding-window-aggregators/blob/566b7222c205de242b87ce65013949c35e518a8b/experiments/process_bulk_evict.py#L144). We have to create subplots, and put the legend on the graph object, not the axes.

scotts (approved):

scotts:
I'm happy to commit this version and work on follow-ups later. I think we'll probably spend a bunch of time tweaking the presentation of the graph, and iterating on the exact experiments we run.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -299,6 +301,55 @@ def create_torchcodec_decoder_from_file(video_file):
...
+    output_dir,
+):
+    executor = ThreadPoolExecutor(max_workers=20)
+    video_count = 0
+    for resolution in resolutions:
+        for duration in durations:
+            for fps in fpses:
+                for gop_size in gop_sizes:
+                    for encoding in encodings:
+                        for pix_fmt in pix_fmts:
```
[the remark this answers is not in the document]
ahmadsharif1: Done

ahmadsharif1:
I merged this as is.

I'll create an issue for follow-up work on benchmarking -- throughput benchmarks, sampler benchmarks, better video selection, etc.

PR #317, opened 2024-10-29
Refac: Straightforward output shape permutation

NicolasHug:
This PR is about where and when we call `MaybePermuteHWC2CHW()`. It's **not** about tensor allocation (this will come, later).
 
-----

At a high-level, this PR changes all conditional call patterns like:

```C++
if (cond) {
  output.frames = MaybePermuteHWC2CHW(output.frames)
}
```

to a plain, unconditional

```C++
output.frames = MaybePermuteHWC2CHW(output.frames)
```

This makes it a lot simpler to reason about our output  shape  permutation. In `main`, `cond` is typically input-dependent (but really, *caller*-dependent), and it leads to a state that's hard to reason about.

Another benefit of this PR is that now **all** low-level decoding routines (like `convertAVFrameToDecodedOutputOnCPU()`) have a simpler interface: they only ever take and return HWC tensors.

----

At a lower level, the following changes were made:

- `MaybePermuteHWC2CHW()` is now a method so we can pass a `streamIndex`. It makes its interface slightly simpler.
- It's now up to every high-level decoding function to call `MaybePermuteHWC2CHW()`.
- Some methods like `getFrameAtIndex()` and `getNextDecodedOutputNoDemux()` were used both as a high-level decoding entry-point **and** as a low-level subroutine of other entry-points. I split those into `getFrameAtIndex()/getFrameAtIndexInternal()`  and `getNextFrame()/getNextDecodedOutputNoDemux()` to clearly distinguish between the public entry point and the underlying private helper. Note that this isn't just a "nice-to-have" or a nit-pick, it's a necessary change for the goal of this PR.
- `getNextFrame()` is the new public entry point, `getNextDecodedOutputNoDemux()` is now private.

-----

A follow-up of this PR will be to unify the tensor allocation. I think it'd make sense for tensors to *always* be pre-allocated by the high-level decoding entry points. It will allow us to unify the allocation logic in a single place.

### DIFF
--- a/packaging/check_glibcxx.py
+++ b/packaging/check_glibcxx.py
@@ -46,7 +46,9 @@ for line in symbol_matches:
         all_symbols.add(match.group(0))
 
 if not all_symbols:
-    raise ValueError(f"No GLIBCXX symbols found in {symbol_matches}. Something is wrong.")
+    raise ValueError(
+        f"No GLIBCXX symbols found in {symbol_matches}. Something is wrong."
+    )
 
 all_versions = (symbol.split("_")[1].split(".") for symbol in all_symbols)
 all_versions = (tuple(int(v) for v in version) for version in all_versions)
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1267,7 +1267,7 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getNextRawDecodedOutputNoDemux() {
   return rawOutput;
 }
 
-VideoDecoder::DecodedOutput VideoDecoder::getNextFrame() {
+VideoDecoder::DecodedOutput VideoDecoder::getNextFrameNoDemux() {
   auto output = getNextDecodedOutputNoDemux();
   output.frame = MaybePermuteHWC2CHW(output.streamIndex, output.frame);
   return output;
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1260,6 +1260,12 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getNextRawDecodedOutputNoDemux() {
   return rawOutput;
 }
 
+VideoDecoder::DecodedOutput VideoDecoder::getNextFrame() {
+  auto output = getNextDecodedOutputNoDemux();
+  output.frame = MaybePermuteHWC2CHW(output.streamIndex, output.frame);
+  return output;
+}
+
 VideoDecoder::DecodedOutput VideoDecoder::getNextDecodedOutputNoDemux(
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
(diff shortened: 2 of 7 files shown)


packaging/check_glibcxx.py
```diff
@@ -46,7 +46,9 @@
         all_symbols.add(match.group(0))
 
 if not all_symbols:
-    raise ValueError(f"No GLIBCXX symbols found in {symbol_matches}. Something is wrong.")
+    raise ValueError(
+        f"No GLIBCXX symbols found in {symbol_matches}. Something is wrong."
+    )
```
NicolasHug, line 51: Not sure why, my linter started to want to format this. If the linter on our CI job is OK with it, can we let this in? Otherwise I have to manually revert it on all my commits.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -929,14 +929,6 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
           std::to_string(static_cast<int>(streamInfo.colorConversionLibrary)));
     }
-    if (!preAllocatedOutputTensor.has_value()) {
-      // We only convert to CHW if a pre-allocated tensor wasn't passed. When a
-      // pre-allocated tensor is passed, it's up to the caller (typically a
-      // batch API) to do the conversion. This is more efficient as it allows
-      // batch NHWC tensors to be permuted only once, instead of permuting HWC
-      // tensors N times.
-      output.frame = MaybePermuteHWC2CHW(streamInfo.options, output.frame);
-    }
```
NicolasHug, line 939: This removal is actually the main change. Everything else is just (sensible) patching until tests work

packaging/check_glibcxx.py
scotts: Linters gonna lint.

scotts (approved):

scotts:
Looks good to me. @ahmadsharif1 should also review.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1261,6 +1267,12 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getNextRawDecodedOutputNoDemux() {
   return rawOutput;
 }
 
+VideoDecoder::DecodedOutput VideoDecoder::getNextFrame() {
```
ahmadsharif1, line 1270: For consistency this should have a NoDemux suffix

ahmadsharif1 (approved):

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -386,6 +387,13 @@ class VideoDecoder {
       DecodedOutput& output,
       std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt);
 
+  DecodedOutput getFrameAtIndexInternal(
```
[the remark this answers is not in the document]
NicolasHug: Yes definitely, ultimately I want to label all functions in the decoding stack with expected input and output shapes. I'll follow-up with that, I think it'll be part of a sequence of PRs, after the one about up-leveling the tensor allocation

PR #319, opened 2024-10-29
Add batch decoding support to CUDA

ahmadsharif1:
1. Allocate a batch tensor on the correct device. When cuda is passed in it uses that now.
2. Pass in the batch tensor's view to the color conversion function `convertAVFrameToDecodedOutputOnCuda()`.
3. Add a test to test frame contents.
4. Added a TODO to eventually merge preAllocatedOutputTesnor into RawDecodedOutput because it doesn't make sense to pass in two output data pointers.
5. Add device to VideoDecoder class
6. Update sampler benchmark to take in device and video arguments from the commandline

Sampler benchmark results:

```
CPU:
python benchmarks/samplers/benchmark_samplers.py --device=cpu
----------
num_clips = 1
clips_at_random_indices     med = 23.16ms +- 16.18  med fps = 431.8
clips_at_regular_indices    med = 5.67ms +- 0.43  med fps = 1764.3
clips_at_random_timestamps  med = 22.54ms +- 16.21  med fps = 443.7
clips_at_regular_timestamps med = 7.46ms +- 5.66  med fps = 1339.7
----------
num_clips = 50
clips_at_random_indices     med = 2400.86ms +- 803.05  med fps = 208.3
clips_at_regular_indices    med = 1343.50ms +- 288.18  med fps = 372.2
clips_at_random_timestamps  med = 1170.24ms +- 727.77  med fps = 427.3
clips_at_regular_timestamps med = 950.92ms +- 294.30  med fps = 515.3

CUDA:
python benchmarks/samplers/benchmark_samplers.py --device=cuda:0
----------
num_clips = 1
[AVHWDeviceContext @ 0x8793680] Using current CUDA context.
clips_at_random_indices     med = 245.46ms +- 116.64  med fps = 40.7
clips_at_regular_indices    med = 284.49ms +- 39.86  med fps = 35.2
clips_at_random_timestamps  med = 264.93ms +- 115.74  med fps = 37.7
clips_at_regular_timestamps med = 283.26ms +- 9.99  med fps = 35.3
----------
num_clips = 50
[AVHWDeviceContext @ 0x8d0d680] Using current CUDA context.
clips_at_random_indices     med = 308.00ms +- 104.52  med fps = 1623.4
clips_at_regular_indices    med = 286.54ms +- 12.69  med fps = 1744.9
clips_at_random_timestamps  med = 368.12ms +- 105.73  med fps = 1358.3
clips_at_regular_timestamps med = 285.32ms +- 13.19  med fps = 1717.4
```

CUDA is only worth it for lots of decoding (and could win at throughput) and potentially for higher resolution videos.

Also interestingly enough the variability in CUDA is quite low.

### DIFF
--- a/benchmarks/samplers/benchmark_samplers.py
+++ b/benchmarks/samplers/benchmark_samplers.py
@@ -1,3 +1,4 @@
+import argparse
 from pathlib import Path
 from time import perf_counter_ns
 
@@ -45,8 +46,7 @@ def report_stats(times, num_frames, unit="ms"):
     return med, fps
 
 
-def sample(sampler, **kwargs):
-    decoder = VideoDecoder(VIDEO_PATH)
+def sample(decoder, sampler, **kwargs):
     return sampler(
         decoder,
         num_frames_per_clip=10,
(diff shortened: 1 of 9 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -859,13 +859,14 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
     convertAVFrameToDecodedOutputOnCPU(
         rawOutput, output, preAllocatedOutputTensor);
   } else if (streamInfo.options.device.type() == torch::kCUDA) {
-    // TODO: handle pre-allocated output tensor
+    // TODO: we should fold preAllocatedOutputTensor into RawDecodedOutput.
```
NicolasHug, line 862: Nit: move this TODO outside of this else/if block (just on top of it?), because this applies to CPU as well, not just to the CUDA branch. We may also want to open an issue?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -41,6 +41,8 @@ class VideoDecoder:
             instances of ``VideoDecoder`` in parallel. Use a higher number for multi-threaded
             decoding which is best if you are running a single instance of ``VideoDecoder``.
             Default: 1.
+        device (str or torch.device, optional): The device to use for decoding.
```
NicolasHug, line 44:
```suggestion
        device (str or torch.device, optional): The device to use for decoding. Default: "cpu".
```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -41,6 +41,8 @@ class VideoDecoder:
...
+
```
NicolasHug, line 45: This wasn't introduced in this PR, but we might as well fix it here: the `.. note::` below should be part of the parameter description of the `dimension_order` parameter. Do you mind moving it back up?

test/utils.py
```diff
@@ -33,6 +33,12 @@ def assert_tensor_equal(*args, **kwargs):
     torch.testing.assert_close(*args, **kwargs, atol=absolute_tolerance, rtol=0)
 
 
+# Asserts that at most percentage of the elements are different by more than abs_tolerance.
+def assert_tensor_nearly_equal(frame1, frame2, percentage=0.3, abs_tolerance=20):
```
NicolasHug, line 37: Nit regarding the name: we already have `assert_tensor_close`, which semantically convey the same meaning as "nearly equal" to me. So the distinction between these isn't obvious. Maybe `assert_tensor_close_on_at_least(...)`?

test/decoders/test_video_decoder_ops.py
```diff
@@ -198,6 +222,40 @@ def test_get_frames_by_pts(self):
         with pytest.raises(AssertionError):
             assert_tensor_equal(frames[0], frames[-1])
 
+    # TODO: Figure out how to parameterize this test to run on both CPU and CUDA.abs
+    # The question is how to have the @needs_cuda decorator with the pytest.mark.parametrize
+    # decorator on the same test.
```
NicolasHug, line 227: It's simple!

We just need to define this new util

https://github.com/pytorch/vision/blob/e9a3213524a0abd609ac7330cf170b9e19917d39/test/common_utils.py#L122-L125

and it can be used like this

https://github.com/pytorch/vision/blob/e9a3213524a0abd609ac7330cf170b9e19917d39/test/test_utils.py#L221


If you want, we can merge this PR as-is and follow-up with that

test/decoders/test_video_decoder_ops.py
```diff
@@ -137,6 +143,24 @@ def test_get_frames_at_indices(self):
         assert_tensor_equal(frames0and180[0], reference_frame0)
         assert_tensor_equal(frames0and180[1], reference_frame180)
 
+    @needs_cuda
+    def test_get_frames_at_indices_with_cuda(self):
```
NicolasHug, line 147: We'll also want to test `get_frames_in_range`, and all the batch-APIs?
I feel like we should be parametrizing a fair amount of our tests. But this can be done as a follow-up.

NicolasHug (approved):
Thank you @ahmadsharif1 . Only minor suggestions from me.

This is not immediately related to this PR, but now that we publicly expose CUDA, we'll want to beef-up our CUDA tests. They're pretty minimal right now. The test utils that I linked-to below will be useful. Let's follow-up with that in a separate PR (happy to help).

benchmarks/samplers/benchmark_samplers.py
```diff
@@ -45,51 +46,72 @@ def report_stats(times, num_frames, unit="ms"):
...
+            clips_at_regular_timestamps,
+            seconds_between_clip_starts=seconds_between_clip_starts,
+            num_exp=NUM_EXP,
+            warmup=2,
+        )
+        report_stats(times, num_frames, unit="ms")
+
+
+if __name__ == "__main__":
+    DEFAULT_VIDEO_PATH = Path(__file__).parent / "../../test/resources/nasa_13013.mp4"
```
scotts, line 112: Ideally this would be moved inside of `main()`. If we want a function that's only about executing the experiments, then we could move what is in main into another function. Basically, we want to avoid execution at global scope. We can deal with that later in a refactor, though.

test/utils.py
scotts: Agreed. Even better if we can use the same utility function - I suspect that the logic we're doing in this function is quite similar to what `torch.testing.assert_close()` is already doing.

The answer also might be that we just eliminate both `assert_tensor_close()` and `assert_tensor_nearly_equal()`, and just use plain `torch.testing.assert_close()` with scenario-specific tolerances.

scotts (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Done

src/torchcodec/decoders/_video_decoder.py
ahmadsharif1: done

test/decoders/test_video_decoder_ops.py
ahmadsharif1: I'll do that as a follow-up

benchmarks/samplers/benchmark_samplers.py
ahmadsharif1: I'll do that as a follow-up

PR #321, opened 2024-10-30
Unify output tensor allocation across all callers and devices

NicolasHug:
This PR unifies the output tensor frame allocation within a single util: `allocateEmptyHWCTensorForStream`.

All high-level decoding entry-points are now responsible for allocating that output tensor.

All lower-level decoding entry-points now accept **and expect** this pre-allocated tensor to be passed. These low-level entry-points **do not** allocate anything that gets returned.

TODO, as a follow-up: try to fold the `preAllocatedOutputTensor` parameter within the `rawOutput` struct.


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -187,18 +187,34 @@ VideoDecoder::VideoStreamDecoderOptions::VideoStreamDecoderOptions(
...
-    : frames(torch::empty(
-          {numFrames,
-           options.height.value_or(*metadata.height),
-           options.width.value_or(*metadata.width),
-           3},
-          at::TensorOptions(options.device).dtype(torch::kUInt8))),
-      ptsSeconds(torch::empty({numFrames}, {torch::kFloat64})),
-      durationSeconds(torch::empty({numFrames}, {torch::kFloat64})) {}
+torch::Tensor VideoDecoder::allocateEmptyHWCTensorForStream(
+    int streamIndex,
```
NicolasHug, line 191: Please double check this:

To allocate a tensor, we need its height and width. To get those, we need a streamIndex[1]. To get the streamIndex, we need either:
- it to be passed as a parameter
- for the "NoDemux" functions, we need at least a `RawOutput`.


[1] well, not strictly true, there are some cases where it's not needed, but the whole point of this PR is to unify that logic.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -187,18 +187,34 @@ VideoDecoder::VideoStreamDecoderOptions::VideoStreamDecoderOptions(
...
+                           .layout(torch::kStrided)
+                           .device(options.device.type());
+  if (numFrames.has_value()) {
+    return torch::empty({numFrames.value(), height, width, 3}, tensorOptions);
+  } else {
+    return torch::empty({height, width, 3}, tensorOptions);
+  }
+}
+
+VideoDecoder::BatchDecodedOutput VideoDecoder::allocateBatchDecodedOutput(
```
NicolasHug, line 209: The previous `BatchDecodedOutput` constructor wasn't a member of `VideoDecoder`, and so I got compilation errors because we're calling `allocateEmptyHWCTensorForStream` which itself is a member method (because it accesses attributes blahblahblah).

So I had to make the "constructor" a member itself.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: In the most generic case this isn't true because a single stream can have frames with different heights and widths:

https://github.com/pytorch/torchcodec/issues/312

But ignoring that fact, to get the frame dimensions before decoding you do need the stream_index

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: I understand that but by making it a member we are letting it access a lot more information than it strictly needs. It only needs height and width, and not all the other stuff in VideoDecoder. What's worse is it can accidentally change the state of the VideoDecoder because it's not `const`

I think it may be better to pass in the height and width after looking it up in the caller.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -971,8 +962,11 @@ VideoDecoder::DecodedOutput VideoDecoder::getFramePlayedAtTimestampNoDemux(
         return seconds >= frameStartTime && seconds < frameEndTime;
       });
   // Convert the frame to tensor.
-  auto output = convertAVFrameToDecodedOutput(rawOutput);
-  output.frame = MaybePermuteHWC2CHW(output.streamIndex, output.frame);
+  auto streamIndex = rawOutput.streamIndex;
+  auto preAllocatedOutputTensor = allocateEmptyHWCTensorForStream(streamIndex);
```
ahmadsharif1, line 966: In the light of this: https://github.com/pytorch/torchcodec/issues/312, it would be more correct to use rawOutput.frame's dimensions, and not the stream's dimensions

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1267,13 +1263,17 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getNextRawDecodedOutputNoDemux() {
 }
 
 VideoDecoder::DecodedOutput VideoDecoder::getNextFrameNoDemux() {
-  auto output = getNextFrameOutputNoDemuxInternal();
-  output.frame = MaybePermuteHWC2CHW(output.streamIndex, output.frame);
+  auto rawOutput = getNextRawDecodedOutputNoDemux();
+  auto streamIndex = rawOutput.streamIndex;
+  auto preAllocatedOutputTensor = allocateEmptyHWCTensorForStream(streamIndex);
```
ahmadsharif1, line 1268: Ditto

ahmadsharif1 (commented):
https://github.com/pytorch/torchcodec/issues/312 is closely related to this PR, hence I am bringing it up. You can probably address that next.

For this PR, for non-batch functions we should allocate the tensor using the frame's own width and height. That way users can get all the frames in a video whose stream has varying dimensions by calling non-batch functions.

For batch functions we can use streamIndex and if the user wants to get frames that are of different sizes, we can either return an error (for now) or resize them later.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: A general principle I like to apply is that classes should know how to create correct objects, and ideally, constructed objects should be valid. That means classes should have constructors, and the construtor is responsible for making the object valid. So I think we should keep the `BatchDecodedOutput` explicit constructor. If `VideoDecoder` then needs to construct `BatchDecodedOutput` objects using some of its own internal state, then `VideoDecoder` should figure out what's relevant to `BatchDecodedOutput` and pass it in.

We may need a `VideoDecoder::allocatedBatchDecodedOutput()` method, but if we do, it should call `BatchDecodedOutput`'s explicit constructor. What this could potentially look like:
```
VideoDecoder::BatchDecodedOutput::BatchDecodedOutput(
  int64_t numFrames,
  torch::Tensor frames)
  : frames(frames),
  ptsSeconds(torch::empty({numFrames}, {torch::kFloat64})),
  durationSeconds(torch::empty({numFrames}, {torch::kFloat64})) {}

VideoDecoder::BatchDecodedOutput VideoDecoder::allocateBatchDecodedOutput(
  int streamIndex,
  int64_t numFrames) {
  return BatchDecodedOutput(
    numFrames,
    allocateEmptyHWCTensorForStream(streamIndex, numFrames)
  );
}
```
We can take advantage of tensors have reasonable move-like semantics.

NicolasHug:
I've discussed this PR with both of you @ahmadsharif1 @scotts and the conclusion is that bubbling-up the tensor allocation might not be the best move at this time, because it would change the source of where we retrieve height and width from.

I'll follow-up soon with some alternative improvements.

PR #322, opened 2024-10-30
Remove preAllocatedTensor from getFrameAtIndex

NicolasHug:
This is a small clean-up that I forgot to do in https://github.com/pytorch/torchcodec/pull/317

`getFrameAtIndex()` was / is never called with its optional `preAllocatedTensorOutput` parameter, so this PR removes it. This makes `getFrameAtIndex()` consistent with the other high-level entry-points, which do not accept `preAllocatedTensorOutput` either.

In the future, we'll probably want to put back this parameter for users to pass their own pre-allocated tensors, but we should add that consistently across all entry-points.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.h
+++ b/src/torchcodec/decoders/_core/VideoDecoder.h
@@ -227,8 +227,7 @@ class VideoDecoder {
 
   DecodedOutput getFrameAtIndex(
       int streamIndex,
-      int64_t frameIndex,
-      std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt);
+      int64_t frameIndex);
   struct BatchDecodedOutput {
     torch::Tensor frames;
     torch::Tensor ptsSeconds;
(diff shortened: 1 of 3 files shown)


src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -225,7 +225,8 @@ class VideoDecoder {
   // seconds=5.999, etc.
   DecodedOutput getFramePlayedAtTimestampNoDemux(double seconds);
 
-  DecodedOutput getFrameAtIndex(
+  DecodedOutput getFrameAtIndex(int streamIndex, int64_t frameIndex);
+  DecodedOutput getFrameAtIndexInternal(
```
NicolasHug, line 229: I had to move this one back to public unfortunately, because we must now rely on this one for our C++ tests. Eventually once we accept pre-allocated tensors in all high-level entrypoints, we'll be able to move it back to private.

scotts: We could also use the [friend](https://en.cppreference.com/w/cpp/language/friend) declaration, but I'd prefer not to even mention the testing class in the implementation. I think this is fine.

ahmadsharif1 (approved):

PR #326, opened 2024-10-30
Add more cuda tests

ahmadsharif1:
This addresses comments:

https://github.com/pytorch/torchcodec/pull/319#discussion_r1822348889
https://github.com/pytorch/torchcodec/pull/319#discussion_r1822353651

1. Add a mark in conftest.py
2. Add a decorator that passes in cpu and cuda devices
3. Adds more tests for cuda, especially batch frame extraction tests
4. Update the tensor_close function to be more tolerant of differences. The default is now 90% of the values should be within 20 of each other.

### DIFF
--- a/test/decoders/test_video_decoder_ops.py
+++ b/test/decoders/test_video_decoder_ops.py
@@ -39,6 +39,8 @@ from torchcodec.decoders._core import (
 from ..utils import (
     assert_tensor_close_on_at_least,
     assert_tensor_equal,
+    cpu_and_cuda,
+    get_tensor_compare_function,
     NASA_AUDIO,
     NASA_VIDEO,
     needs_cuda,
@@ -129,10 +131,12 @@ class TestOps:
         assert pts.item() == pytest.approx(6.006, rel=1e-3)
         assert duration.item() == pytest.approx(0.03337, rel=1e-3)
 
-    def test_get_frames_at_indices(self):
+    @pytest.mark.parametrize("device", cpu_and_cuda())
+    def test_get_frames_at_indices(self, device):
+        tensor_compare_function = get_tensor_compare_function(device)
         decoder = create_from_file(str(NASA_VIDEO.path))
         scan_all_streams_to_update_metadata(decoder)
-        add_video_stream(decoder)
+        add_video_stream(decoder, device=device)
         frames0and180, *_ = get_frames_at_indices(
             decoder, stream_index=3, frame_indices=[0, 180]
         )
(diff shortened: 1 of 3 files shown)


test/utils.py
```diff
@@ -34,10 +50,17 @@ def assert_tensor_equal(*args, **kwargs):
 
 
 # Asserts that at least `percentage`% of the values are within the absolute tolerance.
-def assert_tensor_close_on_at_least(frame1, frame2, percentage=99.7, abs_tolerance=20):
+def assert_tensor_close_on_at_least(frame1, frame2, percentage=90, abs_tolerance=20):
```
NicolasHug, line 53: Since the default tolerance has changed, did that implicitly changed the tolerance of the existing calls as well? E.g. the ones introduced in https://github.com/pytorch/torchcodec/pull/319/files has a stricter tolerance IIUC - we may need to update those?

test/utils.py
```diff
@@ -21,6 +21,22 @@ def needs_cuda(test_item):
...
+def cpu_and_cuda():
+    # TODO: Figure out how to dedupe this logic in needs_cuda() and here.
+    if not torch.cuda.is_available():
+        if os.environ.get("FAIL_WITHOUT_CUDA") == "1":
+            raise RuntimeError("CUDA is required for this test")
+        return ("cpu",)
+    return ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
+
+
+def get_tensor_compare_function(device):
```
NicolasHug, line 33: I'm a bit uncomfortable with naming this `get_tensor_compare_function` because it should be frame-specific, not tensor specific. E.g. for tensors like `pts_seconds`, we'd want to use a stricter equality check both on CPU and GPU.

Basically I think this should be `get_frame_compare_function()`.

test/utils.py
```diff
@@ -21,6 +21,22 @@ def needs_cuda(test_item):
     return test_item
 
 
+def cpu_and_cuda():
+    # TODO: Figure out how to dedupe this logic in needs_cuda() and here.
```
NicolasHug, line 25: I think we should make `cpu_and_cuda()` and `needs_cuda()` be exactly like they are in torchvision, and just handle the skip and `FAIL_WITHOUT_CUDA` within https://github.com/pytorch/vision/blob/80edd48291aa83eb57d90a396e57e036f6bb7fe8/test/conftest.py#L25C5-L25C34. Our version of that would be a lot simpler than that of torchvision though, don't be scared.


I can take care of it if you want (but I'd prefer doing this change within this PR)

test/utils.py
ahmadsharif1: I am fine using higher tolerances for those too -- as they may have different values for different GPU models anyway

test/decoders/test_video_decoder_ops.py
```diff
@@ -129,41 +131,26 @@ def test_get_frame_with_info_at_index(self):
         assert pts.item() == pytest.approx(6.006, rel=1e-3)
         assert duration.item() == pytest.approx(0.03337, rel=1e-3)
 
-    def test_get_frames_at_indices(self):
+    @pytest.mark.parametrize("device", cpu_and_cuda())
+    def test_get_frames_at_indices(self, device):
+        tensor_compare_function = get_frame_compare_function(device)
```
NicolasHug, line 136: Nit: here and everywhere below, call this `frame_compare_function`

NicolasHug (approved):
Thanks @ahmadsharif1 

IIUC this PR mostly tests `get_frames_at_indices`, `get_frames_displayed_at` and `get_frames_in_range`.

But for coverage we probably want to also parametrize the existing tests for `get_frame_at`, `get_frame_displayed_at`, `get_next_frame()`, etc.

And we'll also want to add tests to the public decoder class.

I can address those in a follow-up if you'd like

ahmadsharif1:
Thanks. Actually I just realized that in doing all the test refactors I missed a very important check -- we don't actually check that the frame is on the correct device.

@NicolasHug do you have any opinions on how it should be done.

I think we can even have a function like `test_frame_against_reference(frame1, frame2, expected_device)`

Within that function we can 1) assert frame is on device and 2) use the correct comparison function -- i.e. approximately equal for cuda and exactly equal for CPU

Thoughts?

As far as more tests are concerned, you (or I) can address those later. But let's align on the overall testing flow for cuda tests including testing the device and tensor values both.

PR #327, opened 2024-10-31
Generalize README benchmark

scotts:
I could not run the benchmarks at first; these are the changes I had to make to get them to run fresh.

### DIFF
--- a/benchmarks/decoders/generate_readme_chart.py
+++ b/benchmarks/decoders/generate_readme_chart.py
@@ -42,10 +42,10 @@ def main() -> None:
     gop_sizes = [600]
     durations = [10]
     pix_fmts = ["yuv420p"]
-    ffmpeg_path = "/usr/local/bin/ffmpeg"
-    videos_path = "/tmp/videos"
-    shutil.rmtree(videos_path)
-    os.makedirs(videos_path)
+    ffmpeg_path = "ffmpeg"
+    videos_dir_path = "/tmp/torchcodec_benchmarking_videos"
+    shutil.rmtree(videos_dir_path, ignore_errors=True)
+    os.makedirs(videos_dir_path)
     generate_videos(
         resolutions,
         encodings,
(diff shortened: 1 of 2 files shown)


benchmarks/decoders/generate_readme_chart.py
```diff
@@ -42,9 +42,9 @@ def main() -> None:
     gop_sizes = [600]
     durations = [10]
     pix_fmts = ["yuv420p"]
-    ffmpeg_path = "/usr/local/bin/ffmpeg"
-    videos_path = "/tmp/videos"
-    shutil.rmtree(videos_path)
+    ffmpeg_path = "ffmpeg"
+    videos_path = "/tmp/torchcodec_benchmarking_videos"
+    shutil.rmtree(videos_path, ignore_errors=True)
```
scotts, line 47: We don't want to fail if the tmp directory does not exist already.

benchmarks/decoders/generate_readme_chart.py
```diff
@@ -56,7 +56,7 @@ def main() -> None:
         ffmpeg_path,
         videos_path,
     )
-    video_paths = glob.glob(f"{videos_path}/*.mp4")
+    video_files = glob.glob(f"{videos_path}/*.mp4")
```
scotts, line 59: I was confused by how close the old variable names were to each other.

benchmarks/decoders/generate_readme_chart.py
```diff
@@ -42,9 +42,9 @@ def main() -> None:
     gop_sizes = [600]
     durations = [10]
     pix_fmts = ["yuv420p"]
-    ffmpeg_path = "/usr/local/bin/ffmpeg"
-    videos_path = "/tmp/videos"
-    shutil.rmtree(videos_path)
+    ffmpeg_path = "ffmpeg"
+    videos_path = "/tmp/torchcodec_benchmarking_videos"
```
ahmadsharif1, line 46: Maybe this should be renamed to videos_dir_path

I like path over files because it tells the user it's a full path and not the filename only

benchmarks/decoders/generate_readme_chart.py
ahmadsharif1: See comment above about rename

ahmadsharif1 (approved):
Minor comment about renames

ahmadsharif1 (approved):
Minor comment about renames

benchmarks/decoders/generate_readme_chart.py
scotts: Yeah, good point, I'll make one use `dir` and the other `file`.

PR #328, opened 2024-10-31
Seperate out readme experiments from chart generation

scotts:
Makes the following changes:

1. Separates the _running_ of the experiments from the generation of the pretty graph we show to users. We want to do this so we can tweak the presentation without having to re-run experiments. We achieve this by:
    a. Splitting the current script into two: `generate_readme_data.py` and `generate_readme_chart.py`.
    b. Using a JSON file as the intermediary between the two.
2. Checks-in a new chart. It's very similar to the old one.
3. Checks-in the raw data backing up that chart as a JSON file - I think it's important to have the raw data around, and that we want the raw data and the graphs to match.


benchmarks/decoders/generate_readme_data.py
```diff
@@ -0,0 +1,73 @@
...
+        decoder_dict,
+        video_files_paths,
+        num_uniform_samples,
+        10,
+        False,
+    )
+
+    data_json = Path(__file__).parent / "benchmark_readme_data.json"
+    with open(data_json, "w") as write_file:
+        json.dump(df_data, write_file, sort_keys=True, indent=4)
```
ahmadsharif1, line 69: Is it worth dumping the list of results in a "results" key and separately another key with "metadata" that has metadata like machine info, etc.?

ahmadsharif1 (approved):
You can submit as it, but if it's decoupled like that we may want to have another key for non-test data like machine info, # of cpus, gpus, python version, etc.

benchmarks/decoders/generate_readme_data.py
scotts: Oh, yes, great idea. Let me do that now.

scotts: Done. Ideally we'd also get the model name like you can get from `cat /proc/cpuinfo`. It seems like `platform.processor()` used to return that, but now it seems to just be the same as `platform.machine()`. If we really want that, we can either use the [cpuinfo](https://pypi.org/project/py-cpuinfo/) library (which we'd have to install), or just read the proc filesystem ourselves.

scotts:
@ahmadsharif1, I added the machine metadata into the data file. We can improve it over time if we want to - I wish we had the processor model name and speed.

PR #329, opened 2024-11-01
Benchmark graph improvements

scotts:
Shared legend across all subplots; removes y-axis label when redundant.


ahmadsharif1:
The legend seems like it's out of the chart a bit. But you can iterate on it.

PR #331, opened 2024-11-04
Ensure we don't read invalid memory when seeking to pts before the first frame

### LINKED ISSUE #307: We could potentially access invalid memory while seeking
We call this function to get us the index of the highest keyframe that is <= the target pts:

https://github.com/pytorch/torchcodec/blob/59af4b77120e53c4fb8cf32e1c459ff5c3a10f12/src/torchcodec/decoders/_core/VideoDecoder.cpp#L698-L699

However this can return -1 if we are trying to seek to some negative value before the first keyframe's pts, and that would be invalid memory.

i.e. desiredKeyFrameIndex could be -1.

We probably should have an std::max(desiredKeyFrameIndex, 0) before looking up the desiredPts.

ahmadsharif1:
This PR does:

1. Ensures we don't read keyFrames[-1] when trying to seek before the first keyFrame. Note that our function returns the last keyframe whose pts is <= to the desired pts. And it returns -1 if no such keyframe is found.
2. We also use a single function to convert seconds to pts. Here we do a call to `std::round()` to ensure that we go back to int64 as faithfully as possible. Previously we were truncating and that could lead to off-by-one errors.

This fixes #307 (in particular, (1) fixes that)

I added a test for (1) but adding a test for (2) requires a lot of precise video editing that I haven't done.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -34,6 +34,10 @@ double ptsToSeconds(int64_t pts, const AVRational& timeBase) {
   return ptsToSeconds(pts, timeBase.den);
 }
 
+int64_t secondsToClosestPts(double seconds, const AVRational& timeBase) {
+  return static_cast<int64_t>(std::round(seconds * timeBase.den));
+}
+
 struct AVInput {
   UniqueAVFormatContext formatContext;
   std::unique_ptr<AVIOBytesContext> ioBytesContext;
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -696,6 +701,7 @@ void VideoDecoder::maybeSeekToBeforeDesiredPts() {
   if (!firstStreamInfo.keyFrames.empty()) {
     int desiredKeyFrameIndex =
         getKeyFrameIndexForPts(firstStreamInfo, desiredPts);
+    desiredKeyFrameIndex = std::max(desiredKeyFrameIndex, 0);
```
scotts, line 704: When tracing through this logic, I realized that we're duplicating the check that's inside of `getKeyFrameIndexForPts`. In that function, we'll always call `getKeyFrameIndexForPtsUsingScannedIndex`. It might be better to just call that function directly, as it also matches with the comment above.

test/decoders/test_video_decoder_ops.py
```diff
@@ -84,6 +84,21 @@ def test_seek_and_next(self, device):
         )
         frame_compare_function(frame_time6, reference_frame_time6.to(device))
 
+    @pytest.mark.parametrize("device", cpu_and_cuda())
+    def test_seek_to_negative_pts(self, device):
+        device = "cpu"
```
scotts, line 89: Doesn't think override the parameterization of the test?

ahmadsharif1: Good catch. Yeah, forgot to remove it before pushing to my fork

scotts (approved):

PR #332, opened 2024-11-05
WIP

NicolasHug:


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1382,7 +1380,11 @@ torch::Tensor VideoDecoder::convertFrameToTensorUsingFilterGraph(
   ffmpegStatus =
       av_buffersink_get_frame(filterState.sinkContext, filteredFrame.get());
   TORCH_CHECK_EQ(filteredFrame->format, AV_PIX_FMT_RGB24);
-  std::vector<int64_t> shape = {filteredFrame->height, filteredFrame->width, 3};
+  int height = 0, width = 0;
+  std::tie(height, width) = getHeightAndWidthFromOptionsOrAVFrame(
+      streams_[streamIndex].options, filteredFrame.get());
+  std::vector<int64_t> shape = {height, width, 3};
```
NicolasHug, line 1386: So this is the only place where the logic is slightly changed (but I think the behavior is the same): we go through `options` first. I *think* this is correct, because the `filteredFrame` should respect `options` itself?
In any case, this is what is done for the swscale case as well. If that's incorrect or a change of behavior, LMK.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -195,14 +195,13 @@ VideoDecoder::BatchDecodedOutput::BatchDecodedOutput(
...
-          {numFrames,
-           options.height.value_or(*metadata.height),
-           options.width.value_or(*metadata.width),
-           3},
-          at::TensorOptions(options.device).dtype(torch::kUInt8))),
-      ptsSeconds(torch::empty({numFrames}, {torch::kFloat64})),
-      durationSeconds(torch::empty({numFrames}, {torch::kFloat64})) {}
+    : ptsSeconds(torch::empty({numFrames}, {torch::kFloat64})),
+      durationSeconds(torch::empty({numFrames}, {torch::kFloat64})) {
+  int height = 0, width = 0;
```
scotts, line 200: Same nit, and there's a few other places.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1406,6 +1408,38 @@ VideoDecoder::~VideoDecoder() {
...
+}
+
+std::tuple<int, int> getHeightAndWidthFromOptionsOrAVFrame(
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    AVFrame* avFrame) {
+  return std::make_tuple(
+      options.height.value_or(avFrame->height),
+      options.width.value_or(avFrame->width));
+}
+
```
scotts, line 1426: I'd actually prefer that we just call both of these functions `getHeightAndWidth` and allow C++ function overloading to determine which one to call. In C++, because the types of parameters are formally a part of the function signature, it's less common to encode the type of one of the parameters in the name. This, however, is a question of style, and I know @ahmadsharif1 may feel differently.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1406,6 +1408,38 @@ VideoDecoder::~VideoDecoder() {
...
+torch::Tensor allocateEmptyHWCTensor(
+    int height,
+    int width,
+    torch::Device device,
+    std::optional<int> numFrames) {
+  auto tensorOptions = torch::TensorOptions()
+                           .dtype(torch::kUInt8)
+                           .layout(torch::kStrided)
+                           .device(device);
```
scotts, line 1435: Not critical or blocking, but this may be a good time to put in `TORCH_CHECK`s to make sure the height and width are not negative. I just looked it up, and I think the tensor API can handle that, but that's probably not going to be what we expect. (I know you plan on a follow-up PR with more error-checking, so this sort of thing may be more appropriate there.)

scotts (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Looks correct to me, but @ahmadsharif1 should also reason through it.

scotts:
Almost forgot to mention: great refactoring, and thanks for spending the time to reason through all of this and make sure what we're doing is sensible!

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -209,8 +197,9 @@ void convertAVFrameToDecodedOutputOnCuda(
       src->format == AV_PIX_FMT_CUDA,
       "Expected format to be AV_PIX_FMT_CUDA, got " +
           std::string(av_get_pix_fmt_name((AVPixelFormat)src->format)));
-  int width = options.width.value_or(codecContext->width);
-  int height = options.height.value_or(codecContext->height);
+  int height = 0, width = 0;
+  std::tie(height, width) =
```
ahmadsharif1, line 201: IMO this is less maintainable/safe than returning a struct with named members.

Someone could write:

`std::tie(width, height)` instead of `std::tie(height, width)` and it would be hard to notice.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -413,6 +414,63 @@ class VideoDecoder {
...
+// it is very important to check the height and width assumptions where the
+// tensors memory is used/filled in order to avoid segfaults.
+
+std::tuple<int, int> getHeightAndWidthFromOptionsOrMetadata(
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    const VideoDecoder::StreamMetadata& metadata);
+
+std::tuple<int, int> getHeightAndWidthFromOptionsOrAVFrame(
+    const VideoDecoder::VideoStreamDecoderOptions& options,
+    AVFrame* avFrame);
```
ahmadsharif1, line 466: This should be const too

const & would be preferred since this is a mandatory parameter and should never be null

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Maybe add a TORCH_CHECK to make sure the filteredFrame has the expected dimensions?

ahmadsharif1 (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: I'll add the validity checks as follow up!

PR #333, opened 2024-11-05
Fix Lint job

NicolasHug:
Our lint job is broken on main: see latest commit with a failing pre-commit job https://github.com/pytorch/torchcodec/actions/runs/11672509149/job/32501275316

It's failing, but *green*. This is because we have deactivated the ability for the job to properly fail

This PR:

- let the pre-commit check fail when needed: see that it fails, as expected, in [`736bdc4` (#333)](https://github.com/pytorch/torchcodec/pull/333/commits/736bdc4db50dcd2a8c8f8030422c6f927f90d6fc)
- fixes the code formatting accordingly.


.github/workflows/lint.yaml
```diff
@@ -36,8 +36,6 @@ jobs:
         run: |
           python -m pip install pre-commit
       - name: Run pre-commit checks
-        # Continue on error to display the formatting diff in case lint fails.
-        continue-on-error: true
```
NicolasHug, line 40: This is what prevented the CI to be red when the pre-commit job failed.

We'll have to find a better way to display the formatting diff.

ahmadsharif1 (approved):

PR #335, opened 2024-11-05
Added doc for nvdec

ahmadsharif1:
1. Add documentation for using cuda decoding.
2. Updated the docs workflow to actually build a cuda wheel so it can be used later to generate the docs.
3. The documentation contains instructions on how to install the cuda wheel and scenarios where we expect users to benefit from using CUDA decoding.

Result screenshot:

image


examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+CUDA Decoding can be faster than CPU Decoding for the actual decoding step and for
+subsequent transform steps like scaling, cropping or rotating. This is because the decode step leaves
+the decoded tensor in GPU memory so the GPU doesn't have to fetch from main memory before
+running the transform steps. Encoded packets are often much smaller than decoded frames so
+CUDA decoding also uses less PCI-e bandwidth.
+
+CUDA Decoding can offer speed-up over CPU Decoding in a few scenarios:
+
+#. You are decoding a large resolution video
+#. You are decoding a large batch of videos that's saturting the CPU
```
scotts, line 24: "saturting" -> "saturating"

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+#. You want to do whole-image transforms like scaling or convolutions on the decoded tensors
+   after decoding
+#. Your CPU is saturated and you want to free it up for other work
+
+In some scenarios CUDA Decoding can be slower than CPU Decoding, example:
+
+#. If your GPU is already busy and CPU is not
+#. If you have small resolution videos and the PCI-e transfer latency is large
+#. You want bit-exact results compared to CPU Decoding
```
scotts, line 33: The last item is not really a reason that CUDA Decoding may be slower, but a general reason not to use CUDA Decoding. I think the framing of the list may be better as "Here are situations where CUDA Decoding may not make sense."

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+
+It's best to experiment with CUDA Decoding to see if it improves your use-case. With
+TorchCodec you can simply pass in a device parameter to the VideoDecoder class to
+use CUDA Decoding.
```
scotts, line 37: Let's make `VideoDecoder` a proper reference so users can follow the link.

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+In order use CUDA Decoding will need the following installed in your environment:
+
+#. CUDA-enabled pytorch
+#. FFMPEG binaries that support NVDEC-enabled codecs
+#. libnpp and nvrtc (these are usually installed when you install the full cuda-toolkit)
+
+
+FFMPEG versions 5, 6 and 7 from conda-forge are built with NVDEC support and
+you can install them by running (for example to install ffmpeg version 7):
+
```
scotts, line 48: I'd rephrase as:
> FFmpeg versions 5, 6 and 7 from conda-forge are built with NVDEC support and you can install them with conda. For example, to install FFmpeg version 7:

.github/workflows/docs.yaml
```diff
@@ -26,16 +26,14 @@ jobs:
...
+      - name: Install torchcodec from nightly
         run: |
-          # TODO: torchvision and torchaudio shouldn't be needed. They were only added
-          #  to silence an error as seen in https://github.com/pytorch/torchcodec/issues/203
-          python -m pip install --pre torch torchvision torchaudio --index-url https://download.pytorch.org/whl/nightly/cpu
-          conda install "ffmpeg=7.0.1" pkg-config -c conda-forge
-          ffmpeg -version
-      - name: Build and install torchcodec
+          pip3 install --pre torch torchvision torchaudio torchcodec --index-url https://download.pytorch.org/whl/nightly/cu124
+      - name: Install FFMPEG and other deps
```
NicolasHug, line 32: Before committing this, can  you double check that the doc build can pick up a change that we make to the code? e.g. modify a docstring within the repo, and maybe `print()` something within the decoder, and make sure these 2 changes are reflected within the doc build.
I *think* it will work but I'm not 100% sure.

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Accelerated video decoding with NVDEC
+=====================================
+
+.. _nvdec_tutorial:
+
+**Author**: `Ahmad Sharif <ahmads@meta.com>`__
```
NicolasHug, line 12: I don't recommend adding authorship, we don't do it in the other tutorials. Over time the docs are written and re-written by many contributors, so this usually isn't accurate.

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+"""
+Accelerated video decoding with NVDEC
```
NicolasHug, line 7: Users who don't know what NVDEC is (i.e. 99% of them) won't know that this is about CUDA GPUs. I think the title should include the word "CUDA" and / or "GPU". Maybe

> Accelerated video decoding on CUDA GPUs with NVDEC

?

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+#. You want to do whole-image transforms like scaling or convolutions on the decoded tensors
+   after decoding
+#. Your CPU is saturated and you want to free it up for other work
+
+
+Here are situations where CUDA Decoding may not make sense:
+
+#. You want bit-exact results compared to CPU Decoding
+#. If you have small resolution videos and the PCI-e transfer latency is large
+#. If your GPU is already busy and CPU is not
```
NicolasHug, line 34: super nit: remove "if" for consistency with other list items

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+
+It's best to experiment with CUDA Decoding to see if it improves your use-case. With
+TorchCodec you can simply pass in a device parameter to the
+:class:`~torchcodec.decoders.VideoDecoder` class to use CUDA Decoding.
+
+
+In order use CUDA Decoding will need the following installed in your environment:
```
NicolasHug, line 41:
```suggestion
In order to use CUDA Decoding will need the following installed in your environment:
```

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+
+#. CUDA-enabled pytorch
```
NicolasHug, line 43: Let's link to https://pytorch.org/get-started/locally/

```suggestion
#. `CUDA-enabled pytorch <https://pytorch.org/get-started/locally/>`_
```

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+.. code-block:: bash
+   # These libraries are needed for CUDA decoding with TorchCodec
+   conda install ffmpeg=7 -c conda-forge
+   conda install libnpp cuda-nvrtc -c nvidia
+"""
+
+# %%
+######################################################################
+# Checking if Pytorch has CUDA enabled
+######################################################################
```
NicolasHug, line 60: Nit: in other examples of the type we have used underlines https://github.com/pytorch/torchcodec/blob/c2bea4be0755aec1d9ca0813b2125c6f90f0bd8a/examples/basic_example.py#L115-L116

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+# %%
+######################################################################
+# CUDA Decoding using VideoDecoder
+######################################################################
+#
+# To use CUDA decoder, you need to pass in a cuda device to the decoder.
+#
+from torchcodec.decoders import VideoDecoder
+
+vd = VideoDecoder(video_file, device="cuda:0")
```
NicolasHug, line 109: I think we should just pass "cuda", this is what is more common in such tutorials, and what most users do. Those who have more than one GPU and need to specify the index would know how to do it.

Also let's name this `decoded` for consistency with the rest of our docs (and this example)

```suggestion
decoder = VideoDecoder(video_file, device="cuda")
```

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,179 @@
...
+    return [decoder.get_frame_played_at(ts) for ts in timestamps]
+
+
+def get_numpy_images(frames):
+    numpy_images = []
+    for frame in frames:
+        # We transfer to the CPU so they can be visualized by matplotlib.
+        numpy_image = frame.data.to("cpu").permute(1, 2, 0).numpy()
+        numpy_images.append(numpy_image)
+    return numpy_images
```
NicolasHug, line 146: Consider using the same logic we have in our 2 existing examples for plotting images (and for plot_cpu_and_cuda_images).

https://github.com/pytorch/torchcodec/blob/c2bea4be0755aec1d9ca0813b2125c6f90f0bd8a/examples/basic_example.py#L36-L51

It'll be easier to clean up / factorize into a util later. You typically don't need to directly convert to numpy for this.

It will also make the example easier to follow because right now there are plotting/conversion utilities mixed with decoding code. It's best to keep those separate.

NicolasHug (commented):
Great tutorial! It's great to see that the UX for installing gpu-compatible wheels is fairly straightforward. I left a few comments below.


Doc job is failing with a minor build issue: https://github.com/pytorch/torchcodec/actions/runs/11734305974/job/32690104327?pr=335

As a result I haven't been able to look at the rendered docs, please let me do another round before landing :)

NicolasHug (commented):
Great tutorial! It's great to see that the UX for installing gpu-compatible wheels is fairly straightforward. I left a few comments below.


Doc job is failing with a minor build issue: https://github.com/pytorch/torchcodec/actions/runs/11734305974/job/32690104327?pr=335

As a result I haven't been able to look at the rendered docs, please let me do another round before landing :)

.github/workflows/docs.yaml
ahmadsharif1: I can test but I don't think it will work because this is picking up the nightly from download.pytorch.org

examples/basic_cuda_example.py
ahmadsharif1: I updated this now

examples/basic_cuda_example.py
ahmadsharif1: Good point. I copied this from audio and kept it. I deleted it now

examples/basic_cuda_example.py
ahmadsharif1: Done

examples/basic_cuda_example.py
ahmadsharif1: done

examples/basic_cuda_example.py
ahmadsharif1: done

examples/basic_cuda_example.py
scotts: Agreed that we want a title that people who are not familiar with NVDEC will know right away it's about GPUs and CUDA. My suggested phrasing:

> Accelerated video decoding on GPUs with CUDA and NVDEC

We also want to define NVDEC in our tutorial. We can potentially link to some NVIDIA docs for readers to learn more, such as: https://docs.nvidia.com/video-technologies/video-codec-sdk/12.1/nvdec-video-decoder-api-prog-guide/index.html.

examples/basic_cuda_example.py
ahmadsharif1: I actually tried this suggestion but the results look very ugly. Something (sphinx, torchvision or matplotlib) is compressing the images and it makes them look really low-resolution.

Example code that I used and output:

```
def plot(
    frames1: List[torch.Tensor],
    frames2: List[torch.Tensor],
    title1: Optional[str] = None,
    title2: Optional[str] = None,
):
    try:
        import matplotlib.pyplot as plt
        from torchvision.transforms.v2.functional import to_pil_image
        from torchvision.utils import make_grid
    except ImportError:
        print("Cannot plot, please run `pip install torchvision matplotlib`")
        return
    # Create a list of alternating frames from frames1 and frames2
    frames = [f for pair in zip(frames1, frames2) for f in pair]
    plt.rcParams["savefig.bbox"] = "tight"
    fig, ax = plt.subplots()
    ax.imshow(to_pil_image(make_grid(frames, nrow=2)))
    ax.set(xticklabels=[], yticklabels=[], xticks=[], yticks=[])
    plt.tight_layout()
```

Output:

image


Original output:

image


The original output is better so IMO we should submit as-is and iterate on this later (without compromising the image quality).

.github/workflows/docs.yaml
```diff
@@ -5,45 +5,113 @@ on:
...
+  build-docs:
+    runs-on: linux.4xlarge.nvidia.gpu
     strategy:
       fail-fast: false
+      matrix:
+          # 3.9 corresponds to the minimum python version for which we build
+          # the wheel unless the label cliflow/binaries/all is present in the
+          # PR.
+          # For the actual release we should add that label and change this to
+          # include more python versions.
```
NicolasHug, line 57: This comment doesn't seem relevant for the doc job, let's remove it

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+from torchcodec.decoders import VideoDecoder
+
+decoder = VideoDecoder(video_file, device="cuda")
+frame = decoder[0]
+
+# %%
+#
+# The video frames are decoded and returned as tensor of NCHW format.
+
+print(frame.data.shape, frame.data.dtype)
```
NicolasHug, line 120: Here and below, `frame` is already a pure tensor. No need to access its `data` attribute.

NicolasHug (commented):
Some more comments, I haven't been able to check the rendered docs yet

.github/workflows/docs.yaml
```diff
@@ -5,45 +5,113 @@ on:
...
+          ${CONDA_RUN} python -m pip install $wheel_path -vvv
+
+      - name: Check out repo
+        uses: actions/checkout@v3
+
       - name: Install doc dependencies
         run: |
           cd docs
-          python -m pip install -r requirements.txt
+          ${CONDA_RUN} python -m pip install -r requirements.txt
```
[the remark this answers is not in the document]
ahmadsharif1: Everything is installed in a conda env. Without CONDA_RUN the `pip` that's used is the outside pip

ahmadsharif1 (commented):
Thanks for the review

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,174 @@
...
+    return [decoder.get_frame_played_at(ts).data for ts in timestamps]
+
+
+timestamps = [12, 19, 45, 131, 180]
+cpu_frames = get_frames(timestamps, device="cpu")
+cuda_frames = get_frames(timestamps, device="cuda:0")
+
+
+def plot_cpu_and_cuda_frames(
+    cpu_frames: list[torch.Tensor], cuda_frames: list[torch.Tensor]
```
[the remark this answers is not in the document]
ahmadsharif1: Great point. Done

examples/basic_cuda_example.py
```diff
@@ -0,0 +1,172 @@
...
+
+# %%
+#
+# They look visually similar to the human eye but there may be subtle
+# differences because CUDA math is not bit-exact with respect to CPU math.
+#
+first_cpu_frame = cpu_frames[0].data.to("cpu")
+first_cuda_frame = cuda_frames[0].data.to("cpu")
+frames_equal = torch.equal(first_cpu_frame, first_cuda_frame)
+print(f"{frames_equal=}")
```
NicolasHug, line 172: Nit: Instead of this binary indicator, we may want to print max abs diff and mean abs diff? We could even do it across all frames.

NicolasHug (approved):
Thank you @ahmadsharif1 !

PR #338, opened 2024-11-06
Add docstrings for samplers

NicolasHug:
This PR adds docstring to the samplers. This is just docstrings, not the entire docs. Tutorials / examples / user guides will come later.

This PR is best reviewed by checking it out locally, building and visualizing the docs on your own machine. Alternatively you can download the built docs artifacts from the CI job.

You'll see that I went for a non-classical way of specifying the docs. Similar setups exist in repos like pytorch/pytorch. The main advantage is that it can centralize the docs, avoids repetition, and makes edits easier. The main drawback is that docs are harder to figure out when browsing the code - getting the docs require going online to see the html files.

I don't mind reverting to a more conventional setup if you prefer, but I'll ask to do this only later, when the docs are a bit more stable. I will be writing tutorials in follow-up PRs, and I suspect I'll have to update the docstrings. It will be a lot easier to do so in the current setup.

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -19,8 +19,8 @@ Glossary
        not involve decoding**, so it is a lot cheaper than decoding the file.
 
     clips
-        A clip is a sequence of frames, usually in presentation order. The
-        frames may not necessarily be consecutive. A clip is represented as a 4D
+        A clip is a sequence of frames, usually in :term:`pts` order. The frames
+        may not necessarily be consecutive. A clip is represented as a 4D
         :class:`~torchcodec.FrameBatch`. A group of clips, which is what the
         :ref:`samplers <samplers>` return, is represented as 5D
         :class:`~torchcodec.FrameBatch`.
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -17,3 +17,10 @@ Glossary
        A scan corresponds to an entire pass over a video file, with the purpose
        of retrieving metadata about the different streams and frames. **It does
        not involve decoding**, so it is a lot cheaper than decoding the file.
+
+    clips
+        A clip is a sequence of frames, usually in presentation order. The
+        frames may not necessarily be consecutive. A clip is represented as a 4D
+        :class:`~torchcodec.FrameBatch`. A group of clips, which is what the
+        :ref:`samplers <samplers>` return, is represented as 5D
+        :class:`~torchcodec.FrameBatch`.
--- a/src/torchcodec/samplers/_common.py
+++ b/src/torchcodec/samplers/_common.py
@@ -69,3 +69,13 @@ def _reshape_4d_framebatch_into_5d(
         pts_seconds=frames.pts_seconds.view(num_clips, num_frames_per_clip),
         duration_seconds=frames.duration_seconds.view(num_clips, num_frames_per_clip),
     )
+
+
+_FRAMEBATCH_RETURN_DOCS = """
+    Returns:
+        FrameBatch:
+            The sampled clips, as a 5D :class:`~torchcodec.FrameBatch`.
+            The shape of the ``data`` field is (``num_clips``,
+            ``num_frames_per_clips``, ...) where ... is (H, W, C) or (C, H, W)
+            depending on the decoder options.
+"""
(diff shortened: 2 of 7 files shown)


src/torchcodec/samplers/_index_based.py
```diff
@@ -227,3 +227,57 @@ def clips_at_regular_indices(
...
+        policy (str, optional): Defines how to construct clips that span beyond
+            the end of the video. This is best described with an example:
+            assuming the last valid index in a video is 99, and a clip was
+            sampled to start at index 95, with ``num_frames_per_clip=5`` and
+            ``num_indices_between_frames=2``, the indices of the frames in the
+            clip are supposed to be [95, 97, 99, 101, 103]. But 101 and 103 are
+            invalid indices, so the ``policy`` parameter defines how to replace
+            those frames, with valid indices:
+
+            - "repeat": repeats the last valid frame of the clip. We would get
```
scotts, line 264: Really minor, but the left quotation mark for "repeat", "wrap" and "error" is facing the wrong way in the rendered docs. I'm not sure how or even if we can fix it. Uses of quotes in the middle of text seem fine (like "dilation" on line 240).

src/torchcodec/samplers/_common.py
```diff
@@ -69,3 +69,15 @@ def _reshape_4d_framebatch_into_5d(
...
+
+_FRAMEBATCH_RETURN_DOCS = """
+    Returns:
+        FrameBatch:
+            The sampled :term:`clips`, as a 5D :class:`~torchcodec.FrameBatch`.
+            The shape of the ``data`` field is (``num_clips``,
+            ``num_frames_per_clips``, ...) where ... is (H, W, C) or (C, H, W)
+            depending on the decoder options. The shape of the ``pts_seconds``
+            and ``duration_seconds`` fields is (``num_clips``,
+            ``num_frames_per_clips``)
```
scotts, line 82: Missing period at the end of the sentence. :)

src/torchcodec/samplers/_index_based.py
NicolasHug: I noticed this too lol, no idea why. This is likely related to the pytorch-sphinx-theme css style.

src/torchcodec/samplers/_common.py
```diff
@@ -69,3 +69,15 @@ def _reshape_4d_framebatch_into_5d(
...
     )
+
+
+_FRAMEBATCH_RETURN_DOCS = """
+    Returns:
+        FrameBatch:
+            The sampled :term:`clips`, as a 5D :class:`~torchcodec.FrameBatch`.
+            The shape of the ``data`` field is (``num_clips``,
+            ``num_frames_per_clips``, ...) where ... is (H, W, C) or (C, H, W)
+            depending on the decoder options. The shape of the ``pts_seconds``
```
scotts, line 80: Can we link directly to the docs for `dimension_order` in `VideoDecoder` here? If we can't do that, we can explicitly mention `dimension_order` and link to the docs for `VideoDecoder`.

scotts (approved):

src/torchcodec/samplers/_index_based.py
scotts: ¯\_(ツ)_/¯

scotts:
This looks great, all of my comments are just minor improvements.

On the code organization: makes sense to me, let's go with this and see how it feels to live with it. One thought: should we put a comment (not a docstring!) at the top of the functions themselves saying that the official docs are in _common.py?

docs/source/glossary.rst
```diff
@@ -17,3 +17,10 @@ Glossary
        A scan corresponds to an entire pass over a video file, with the purpose
        of retrieving metadata about the different streams and frames. **It does
        not involve decoding**, so it is a lot cheaper than decoding the file.
+
+    clips
+        A clip is a sequence of frames, usually in presentation order. The
```
ahmadsharif1, line 22: presentation or presentation timestamp?

src/torchcodec/samplers/_index_based.py
```diff
@@ -227,3 +229,57 @@ def clips_at_regular_indices(
         sampling_range_end=sampling_range_end,
         policy=policy,
     )
+
+
+_COMMON_DOCS = f"""
+    Args:
+        decoder (VideoDecoder): The :class:`~torchcodec.decoders.VideoDecoder`
+            instance to sample clips from.
+        num_clips (int, optional): The number of clips to sample. Default: 1.
```
ahmadsharif1, line 238: to return instead of to sample may be better

ahmadsharif1 (approved):

PR #339, opened 2024-11-06
Validation of allocated output tensor shapes

NicolasHug:
This PR adds checks for the shape of the output tensors we allocate. This should help us avoid some potential memory corruption issues.

This PR does **not** change any existing logic (well, not really, see details in comments below). It only adds safer checks against existing assumptions.


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -890,41 +890,68 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
+    TORCH_CHECK(
+        (shape.size() == 3) && (shape[0] == expectedOutputHeight) &&
+            (shape[1] == expectedOutputWidth) && (shape[2] == 3),
+        "Expected pre-allocated tensor of shape ",
+        expectedOutputHeight,
+        "x",
+        expectedOutputWidth,
+        "x3, got ",
+        shape);
+  }
```
NicolasHug, line 910: The check above validates the shape of the pre-allocated output tensor. It was previously only done for swscale, but it should also be done for filtergraph, hence why I moved it up.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1303,24 +1330,23 @@ double VideoDecoder::getPtsSecondsForFrame(
...
 }
 
-void VideoDecoder::convertFrameToBufferUsingSwsScale(
-    RawDecodedOutput& rawOutput) {
-  AVFrame* frame = rawOutput.frame.get();
-  int streamIndex = rawOutput.streamIndex;
+int VideoDecoder::convertFrameToBufferUsingSwsScale(
+    int streamIndex,
+    const AVFrame* frame,
+    torch::Tensor& outputTensor) {
```
NicolasHug, line 1336: I have updated the signature of `convertFrameToBufferUsingSwsScale` for it to be very similar to `convertFrameToTensorUsingFilterGraph`. The signatures are now:

```
int VideoDecoder::convertFrameToBufferUsingSwsScale(
    int streamIndex,
    const AVFrame* frame,
    torch::Tensor& outputTensor);

torch::Tensor VideoDecoder::convertFrameToTensorUsingFilterGraph(
    int streamIndex,
    const AVFrame* frame);
```

Their signatures being so different previously was a constant source of confusion for me.

I realize that this change seemingly goes against our TODO of folding the pre-allocated tensor within `RawOutput`. But I think this TODO will in fact be easier to address with that change, as both APIs will be more closely aligned.

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -224,6 +224,9 @@ void convertAVFrameToDecodedOutputOnCuda(
 
   auto start = std::chrono::high_resolution_clock::now();
 
+  // TODO height and width info of output tensor comes from the metadata, which
+  // may not be accurate. How do we make sure we won't corrupt memory if the
+  // allocated tensor is too short/large?
```
NicolasHug, line 229: @ahmadsharif1 any idea on how to do this?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -890,41 +890,68 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
+      getHeightAndWidthFromOptionsOrAVFrame(streamInfo.options, *frame);
+  int expectedOutputHeight = frameDims.height;
+  int expectedOutputWidth = frameDims.width;
+
+  if (preAllocatedOutputTensor.has_value()) {
+    auto shape = preAllocatedOutputTensor.value().sizes();
+    TORCH_CHECK(
+        (shape.size() == 3) && (shape[0] == expectedOutputHeight) &&
+            (shape[1] == expectedOutputWidth) && (shape[2] == 3),
+        "Expected pre-allocated tensor of shape ",
```
scotts, line 904: You should also be able to do `(shape == {expectedOutputHeight, expectedOutputWidth, 3})` if you think that's clearer. I'm not sure.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1406,6 +1429,10 @@ VideoDecoder::~VideoDecoder() {
   }
 }
 
+FrameDims getHeightAndWidthFromResizedAVFrame(const AVFrame& resizedAVFrame) {
```
scotts, line 1432: Because the function itself has no opinion on what kind of frame it is (resized or not), I think the name should just be `getHeightAndWidthFromAVFrame()`.

scotts (approved):

scotts:
The changes look good to me, but I think @ahmadsharif1 should also review.

src/torchcodec/decoders/_core/VideoDecoder.cpp
ahmadsharif1: Why is this function now returning the height?

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -426,30 +429,32 @@ class VideoDecoder {
...
 //
-// The source of truth for height and width really is the AVFrame: it's the
-// decoded ouptut from FFmpeg. The info from the metadata (i.e. from the
-// CodecContext) may not be as accurate. However, the AVFrame is only available
-// late in the call stack, when the frame is decoded, while the CodecContext is
-// available early when a stream is added. This is why we use the CodecContext
-// for pre-allocating batched output tensors (we could pre-allocate those only
-// once we decode the first frame to get the info frame the AVFrame, but that's
-// a more complex logic).
+// The source of truth for height and width really is the (resized) AVFrame:
```
ahmadsharif1, line 450: The decoded output from ffmpeg isn't resized. The resize happens later

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -463,6 +468,10 @@ struct FrameDims {
   FrameDims(int h, int w) : height(h), width(w) {}
 };
 
+// There's nothing preventing you from calling this on a non-resized frame, but
+// please don't.
+FrameDims getHeightAndWidthFromResizedAVFrame(const AVFrame& resizedAVFrame);
```
ahmadsharif1, line 473: getHeightAndWidth can just be called getFrameDims

src/torchcodec/decoders/_core/CudaDevice.cpp
ahmadsharif1: You can look at rawOutput.frame's dimensions and use those, right?

NicolasHug: Are you suggesting to get height and width from the AVFrame instead of from the Metadata then? I don't mind but that's a change of logic. Do you know why it wasn't done like that before?

NicolasHug: discussed offline, I'll submit a follow-up to do that.

PR #341, opened 2024-11-07
Samplers tutorial

NicolasHug:
This PR adds a notebook-like example for the sampling APIs.

### DIFF
--- a/examples/sampling.py
+++ b/examples/sampling.py
@@ -1,0 +1,241 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+=========================
+How to sample video clips
+=========================
+
+In this example, we'll learn how to sample video :term:`clips` from a video. A
+clip generally denotes a sequence or batch of frames, and is typically passed as
+input to video models.
+"""
+
+# %%
+# First, a bit of boilerplate: we'll download a video from the web, and define a
+# plotting utility. You can ignore that part and jump right below to
+# :ref:`sampling_tuto_start`.
+
+from typing import Optional
+import torch
+import requests
+
+
+# Video source: https://www.pexels.com/video/dog-eating-854132/
+# License: CC0. Author: Coverr.
+url = "https://videos.pexels.com/video-files/854132/854132-sd_640_360_25fps.mp4"
+response = requests.get(url, headers={"User-Agent": ""})
+if response.status_code != 200:
+    raise RuntimeError(f"Failed to download video. {response.status_code = }.")
+
+raw_video_bytes = response.content
+
+
+def plot(frames: torch.Tensor, title : Optional[str] = None):
+    try:
+        from torchvision.utils import make_grid
+        from torchvision.transforms.v2.functional import to_pil_image
+        import matplotlib.pyplot as plt
+    except ImportError:
+        print("Cannot plot, please run `pip install torchvision matplotlib`")
+        return
+
+    plt.rcParams["savefig.bbox"] = 'tight'
+    fig, ax = plt.subplots()
+    ax.imshow(to_pil_image(make_grid(frames)))
+    ax.set(xticklabels=[], yticklabels=[], xticks=[], yticks=[])
+    if title is not None:
+        ax.set_title(title)
+    plt.tight_layout()
+
+
+# %%
+# .. _sampling_tuto_start:
+#
+# Creating a decoder
+# ------------------
+#
+# Sampling clips from a video always starts by creating a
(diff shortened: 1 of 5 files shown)


examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+# ---------------
+#
+# We can now use our decoder to sample clips. Let's first look at a simple
+# example: all other samplers follow similar APIs and principles. We'll use
+# :func:`~torchcodec.samplers.clips_at_random_indices` to sample clips that
+# start at random indices.
+
+from torchcodec.samplers import clips_at_random_indices
+
+torch.manual_seed(0)  # The samplers RNG is controlled by pytorch's RNG
```
scotts, line 81: Relevant to all of our recent discussion, I think we should have a more substantial comment here. Something like:
```
# We're setting the seed to a fixed value here because we want to get reproducible results
# everytime we run the code in this tutorial. In general, you should do the same if you also 
# want reproducible results, but it is not an actual requirement for using our API.
```

examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+clips = clips_at_random_indices(
+    decoder,
+    num_clips=5,
+    num_frames_per_clip=4,
+    num_indices_between_frames=3,
+)
+clips
```
scotts, line 88: Stray variable reference. :)

examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+#   :class:`~torchcodec.VideoDecoder`. This is typically what would get passed
+#   to the model.
+# - ``pts_seconds``: a 2D float tensor of shape (num_clips, num_frames_per_clip)
+#   giving the starting timestamps of each frame within each clip, in seconds.
+# - ``duration_seconds``: a 2D float tensor of shape (num_clips,
+#   num_frames_per_clip) giving the duration of each frame within each clip, in
+#   seconds.
+
+plot(clips[0].data)
+
```
scotts, line 107: This doesn't render an image in the docs - I assume it's supposed to?

examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+# %%
+# Indexing and manipulating clips
+# -------------------------------
+#
+# Clips are :class:`~torchcodec.FrameBatch` objects, and they support native
+# pytorch indexing semantics (including fancy indexing). This makes it easy to
+# filter clips based on a given criteria. For example, from the clips above we
+# can easily filter out those who start *after* a specific timestamp:
+clip_starts = clips.pts_seconds[:, 0]
```
scotts, line 116: I actually don't fully understand what we're doing here - I clearly have a lot to learn about PyTorch tensors. :) Are we making a view (copy?) of all `clip.pts_seconds`? Can you explain (to me, not necessarily to the reader) the mechanics of what's going on?

examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+clip_starts
+
+# %%
+clips_starting_after_five_seconds = clips[clip_starts > 5]
```
scotts, line 120: This is super cool!

examples/sampling.py
NicolasHug: I meant it to be there (you'll find other instances). It is shown as output in the rendered example.

examples/sampling.py
```diff
@@ -0,0 +1,247 @@
...
+# %%
+# Indexing and manipulating clips
+# -------------------------------
+#
+# Clips are :class:`~torchcodec.FrameBatch` objects, and they support native
+# pytorch indexing semantics (including fancy indexing). This makes it easy to
+# filter clips based on a given criteria. For example, from the clips above we
+# can easily filter out those who start *after* a specific timestamp:
+clip_starts = clips.pts_seconds[:, 0]
+clip_starts
```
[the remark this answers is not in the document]
NicolasHug: it is intended :)

scotts:
Great tutorial! There's at least a rendering issue with our docs job, but nothing blocking.

scotts (approved):

examples/sampling.py
NicolasHug: I think pytorch is able to create a view, because with that index the strides are regular. I don't know if there's any such guarantee in general though.

`[:, i]`  is a typical indexing strat that means "select the i'th elements along the 2nd dim, don't index the first dim". We end up with a tensor with the same length as the first dim. There can be as many `:` as there are dimension. PyTorch largely follows numpy's indexing semantics, and they do a much better job that I will at explaning all this: https://numpy.org/doc/stable/user/basics.indexing.html#advanced-indexing

examples/sampling.py
NicolasHug: great catch, thanks for checking. I didn't realize the example wasn't properly built on the CI because I was using a slightly different build in my machine. It should be fixed now, I will double check before merging

PR #347, opened 2024-11-08
Use AVFrame info for height and width in GPU APIs

NicolasHug:
instead of relying on metadata, like what we previously did.

Addresses https://github.com/pytorch/torchcodec/pull/339#discussion_r1832919983

### DIFF
--- a/src/torchcodec/decoders/_core/CudaDevice.cpp
+++ b/src/torchcodec/decoders/_core/CudaDevice.cpp
@@ -187,7 +187,6 @@ void initializeContextOnCuda(
 void convertAVFrameToDecodedOutputOnCuda(
     const torch::Device& device,
     const VideoDecoder::VideoStreamDecoderOptions& options,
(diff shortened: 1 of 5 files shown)


src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -222,11 +219,10 @@ void convertAVFrameToDecodedOutputOnCuda(
   // Use the user-requested GPU for running the NPP kernel.
   c10::cuda::CUDAGuard deviceGuard(device);
 
-  auto start = std::chrono::high_resolution_clock::now();
+  NppiSize oSizeROI = {width, height};
+  Npp8u* input[2] = {src->data[0], src->data[1]};
```
NicolasHug, line 223: Just a drive-by, I moved this down closer to where they're used

ahmadsharif1 (approved):

PR #348, opened 2024-11-08
Indicate `num_ffmpeg_threads` is only relevant on CPU

NicolasHug:
I could be wrong!
But if I'm not, I think we should just indicate that it's ignored on CPU. I don't think it's worth raising an error if users pass `num_ffmpeg_threads=3, device="cuda"` because it makes it really annoying to run "generic" loop and comparisons, e.g. stuff like

```py
for device in ("cuda", "cpu"):
    decoder = VideoDecoder(..., num_ffmpeg_threads=1, device=device)  # no point errorring here
```

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -43,11 +43,11 @@ class VideoDecoder:
                 cheap no-copy operation that allows these frames to be
                 transformed using the `torchvision transforms
                 <https://pytorch.org/vision/stable/transforms.html>`_.
-        num_ffmpeg_threads (int, optional): The number of threads to use for decoding.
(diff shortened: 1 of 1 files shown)


scotts:
Seems correct to me!

scotts (approved):

ahmadsharif1:
I don't think this is ignored for CUDA.

https://github.com/pytorch/torchcodec/blob/43ee8076e861a975a93fd019a95bf1860f827edf/src/torchcodec/decoders/_core/VideoDecoder.cpp#L438

It would be nice to run some experiments to see how it affects CUDA performance. And if we can confirm it doesn't affect performance, we can gate setting the threads on device=="cpu".

NicolasHug:
Thanks for your input @ahmadsharif1 , I'll close this and we can follow-up on https://github.com/pytorch/torchcodec/issues/353 then

PR #349, opened 2024-11-08
Test CUDA APIs of VideoDecoder

NicolasHug:
This PR adds tests for the public methods of `VideoDecoder` when `device="cuda"`.


ahmadsharif1 (approved):
Maybe add a comment above 265 to say that not all GPUs support it so we are intentionally running it just on the CPU

PR #350, opened 2024-11-08
Test x265 on CUDA

NicolasHug:


ahmadsharif1:
[AVHWDeviceContext @ 0x555ae5a3d280] Using current CUDA context.
[hevc @ 0x555ae51e4900] Hardware is lacking required capabilities
[hevc @ 0x555ae51e4900] Failed setup for format cuda: hwaccel initialisation returned error.

Not all GPUs support all encodings. I can add that to the docs.

https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new has the support matrix

Your job used a tesla m60 gpu

image

which doesn't have h265 support

PR #351, opened 2024-11-08
Use more interesting videos for README experiment

scotts:
Before this PR, our README experiments were using a video that was just a single blue frame for 30 seconds. This PR changes the README experiment to use two different videos:

1. A Mandelbrot video generated from FFmpeg.
2. The long version of the NASA video we often use for testing. It's got a lot of variety, and I think it's a great representative video.

I also experimented with a third video, using FFmpeg's [rgbtestsrc](https://ffmpeg.org/ffmpeg-filters.html#allrgb_002c-allyuv_002c-color_002c-colorchart_002c-colorspectrum_002c-haldclutsrc_002c-nullsrc_002c-pal75bars_002c-pal100bars_002c-rgbtestsrc_002c-smptebars_002c-smptehdbars_002c-testsrc_002c-testsrc2_002c-yuvtestsrc) option. The results were roughly the same as Mandelbrot - interesting for us, and some evidence that our existing videos are not outliers, but it doesn't tell our users much new, and it adds 30% to the size of the charts. I think even the current size is maybe too big.

~~The other big change is that our old charts were using the _median_ with error bars representing p25 and p75 percentile. I changed our charts to use the _mean_ and the error bars are the standard deviation. In my experience reporting performance in computer systems, I have always used mean and standard deviation rather than the median. I think including the outliers is important. You can have a great median and terrible p90 performance.~~

After discussing, it's easier to go with the median. We have a challenge that we're turning the times into a rate, and average that properly will take more thinking.


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -534,9 +526,13 @@ def run_benchmarks(
                 df_item["median"] = results[-1].median
                 df_item["iqr"] = results[-1].iqr
                 df_item["type"] = f"{kind}:seek()+next()"
-                df_item["fps"] = 1.0 * num_samples / results[-1].median
-                df_item["fps_p75"] = 1.0 * num_samples / results[-1]._p75
-                df_item["fps_p25"] = 1.0 * num_samples / results[-1]._p25
+                df_item["fps_mean"] = num_samples / results[-1].mean
```
ahmadsharif1, line 529: This is not the arithmetic mean fps. Because it's the arithmetic mean of 1/fps -- which means it's really the harmonic mean

ahmadsharif1 (commented):
I prefer median because it makes the default pytorch.utils.benchmark table results look identical to the charts. Mean may cause user confusion.

Also see my comment that it's not really the arith mean fps -- it's the harmonic mean

Using the higher res nasa video looks good to me

benchmarks/decoders/benchmark_decoders_library.py
scotts: Yes, agreed. That's due to us reporting fps instead of time. Do you think this is a problem?

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -419,16 +428,16 @@ def plot_data(df_data, plot_path):
             ax = axes[row, col]  # Select the appropriate axis
 
             # Set the title for the subplot
-            base_video = os.path.basename(video)
+            base_video = os.path.basename(video).removesuffix(".mp4")
```
NicolasHug, line 431: We should stop using `os.path`, I had tried to migrate awa from those already in https://github.com/pytorch/torchcodec/pull/249

I know this wasn't introduced in this PR but do you mind relyin on `pathlib` for this?

benchmarks/decoders/generate_readme_data.py
```diff
@@ -14,47 +14,57 @@
...
     decoder_dict["TorchCodec"] = TorchCodecPublic()
-    decoder_dict["TorchCodec[num_threads=1]"] = TorchCodecPublic(num_ffmpeg_threads=1)
-    decoder_dict["TorchVision[backend=video_reader]"] = TorchVision("video_reader")
+    decoder_dict["TorchVision[video_reader]"] = TorchVision("video_reader")
     decoder_dict["TorchAudio"] = TorchAudioDecoder()
     decoder_dict["Decord"] = DecordAccurateBatch()
 
     # These are the number of uniform seeks we do in the seek+decode benchmark.
     num_samples = 10
+    video_files_paths = glob.glob(f"{videos_dir_path}/*.mp4")
```
NicolasHug, line 67: We can use `pathlib.glob` here too instead of `glob.glob`

benchmarks/decoders/benchmark_decoders_library.py
NicolasHug: Wait, I'm not sure I understand @ahmadsharif1 's comment. Are we reporting `fps` or `1 / fps`?

NicolasHug (approved):

benchmarks/decoders/benchmark_decoders_library.py
scotts: Actually, I don't think this is the harmonic mean, but I think we do want the harmonic mean. In order to get it, I think we need to do:
```
df_item["fps_mean"] = num_samples / np.sum([1/t for t in results[-1].times])
```

README.md
```diff
@@ -124,6 +124,11 @@ The following was generated by running [our benchmark script](./benchmarks/decod
 
 ![benchmark_results](./benchmarks/decoders/benchmark_readme_chart.png)
 
+The top row is a [Mandelbrot](https://ffmpeg.org/ffmpeg-filters.html#mandelbrot) video
+generated from FFmpeg that has a resolution of 1280x720 at 60 fps and is 120 seconds long.
+The bottom row is [promotional video from NASA](https://download.pytorch.org/torchaudio/tutorial-assets/stream-api/NASAs_Most_Scientifically_Complex_Space_Observatory_Requires_Precision-MP4_small.mp4)
```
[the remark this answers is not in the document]
scotts: I'm going to merge as-is. Let's discuss later what we want long-term. I have a slight preference to show two videos, to give some indication that the NASA video isn't a fluke and to include a higher resolution vidoe.

PR #356, opened 2024-11-09
Fix trunk; no default C++ operators on our version of C++

scotts:
PR #354 broke the trunk because it took advantage of default C++ operators, and that is a C++20 feature. We're not on C++20 yet. I'm not sure why the tests did not run on that PR. This PR should fix the trunk.


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1340,7 +1352,11 @@ int VideoDecoder::convertFrameToBufferUsingSwsScale(
   int expectedOutputHeight = outputTensor.sizes()[0];
   int expectedOutputWidth = outputTensor.sizes()[1];
   auto curFrameSwsContextKey = SwsContextKey{
-      frame->width, frame->height, frameFormat, expectedOutputWidth, expectedOutputHeight};
+      frame->width,
+      frame->height,
+      frameFormat,
+      expectedOutputWidth,
+      expectedOutputHeight};
```
scotts, line 1359: This is a drive-by linting. I think it's easier to keep it in the PR.

ahmadsharif1 (approved):

PR #359, opened 2024-11-11
Add the ability to benchmark throughput using multiple threads

ahmadsharif1:
The batch mode is a new mode that decodes a batch of 40 copies of the decoder using 8 threads.

Tested:

```
video=/home/ahmads/personal/torchcodec/benchmarks/decoders/../../test/resources/nasa_13013.mp4, decoder=TorchCodecPublic
[---------------------------------------------------------------- video=/home/ahmads/personal/torchcodec/benchmarks/decoders/../../test/resources/nasa_13013.mp4 h264 480x270, 13.013s 29.97002997002997fps -----------------------------------------------------------------]
                        |  uniform 10 seek()+next()  |  batch uniform 10 seek()+next()  |  random 10 seek()+next()  |  batch random 10 seek()+next()  |  1 next()  |  batch 1 next()  |  10 next()  |  batch 10 next()  |  100 next()  |  batch 100 next()  |  create()+next()
1 threads: -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
      TorchCodecPublic  |            67.0            |              841.9               |            60.5           |              743.9              |    21.4    |      219.4       |     24.1    |       276.5       |     69.9     |       812.5        |                 
      TorchCodecCore    |                            |                                  |                           |                                 |            |                  |             |                   |              |                    |        18.5     

```


scotts:
I assume the removal of the other decoders is temporary while getting everything working?

On the chart generated by `generate_readme_*.py`, I think we want to be selective on what we add to it. I think we want no more than four experiments per row. This is in contrast to the output from `benchmark_decoders.py`, where we can have many experiments. I see `benchmark_decoders.py` as a perf development tool, and `generate_readme_*.py` as our external showcase.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -479,6 +479,30 @@ def get_metadata(video_file_path: str) -> VideoStreamMetadata:
     return VideoDecoder(video_file_path).metadata
 
 
+def run_batch_using_threads(function, *args, num_threads=10, batch_size=40):
+    executor = ThreadPoolExecutor(max_workers=10)
+    for _ in range(batch_size):
+        executor.submit(function, *args)
+    executor.shutdown(wait=True)
```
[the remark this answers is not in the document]
ahmadsharif1: Great point. Done

ahmadsharif1 (commented):
I actually don't want to update the chart or data in this PR.

I just want the cli tool to have the option to benchmark throughput for now.

CUDA and decord may not work in the same process (I haven't tested those). I just wanted the ability to benchmark throughput. I already got an interesting finding that the cuda decoder is slower on some videos.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -479,13 +480,50 @@ def get_metadata(video_file_path: str) -> VideoStreamMetadata:
     return VideoDecoder(video_file_path).metadata
 
 
+class BatchParameters(NamedTuple):
```
scotts, line 483: Is there a specific reason we're using `NamedTuple` and not a dataclass?

ahmadsharif1: I thought it would be lighterweight than dataclass, but I am not too sure. Do you have a preference?

scotts (approved):

benchmarks/decoders/benchmark_decoders_library.py
scotts: I prefer dataclasses. There are some instances where you need to use a namedtuple, but in general, I consider dataclasses to have supplanted namedtuples.

PR #361, opened 2024-11-12
Remove mentions of 2.4 in docs and error messages

NicolasHug:
Closes https://github.com/pytorch/torchcodec/issues/235

This PR also:

- adds a compatibility table to show which version of torchcodec is compatible with which version of torch
- removes the installation instructions we have in the docs. Instead, the docs now link to the README which is the single source of truth. With the upcoming CUDA installation instructions, I think duplicating the instructions in 2 separate locations (and formats!) becomes unmaintainable. The users still get the installation instructions when browsing the docs, it just sends them to the README
- Some minor drive-by (see comments below)


README.md
```diff
@@ -134,10 +145,6 @@ encoded with libx264 and yuv420p pixel format.
 
 We are actively working on the following features:
 
-- [Ship wheels for MacOS](https://github.com/pytorch/torchcodec/issues/111), so
-  that MacOS users can `pip install torchcodec`. For now this is only supported
-  on Linux, but MacOS users can [build from source](./CONTRIBUTING.md).
-- [GPU decoding](https://github.com/pytorch/torchcodec/pull/58)
```
NicolasHug, line 140: Drive-by

scotts (approved):

README.md
```diff
@@ -94,29 +94,40 @@ ffmpeg -f lavfi -i \
...
+   ```
+
+The following table indicates the compatibility between versions of
+`torchcodec`, `torch` and Python.
+
+| `torchcodec`       | `torch`            | Python              |
+| ------------------ | ------------------ | ------------------- |
+| `main` / `nightly` | `main` / `nightly` | `>=3.9`, `<=3.12`   |
+| not yet supported  | `2.5`              | `>=3.9`, `<=3.12`   |
+| `0.0.3`            | `2.4`              | `>=3.8`, `<=3.12`   |
```
ahmadsharif1, line 128: Do we want to have a torchcodec with cuda column here too?

NicolasHug: Not sure if I understand correctly. The CPU and CUDA version of torchcodec should have the same expected pytorch version?

PR #362, opened 2024-11-12
Add workflow tests for benchmarks

scotts:
We're not currently testing our benchmarks. This is a problem, because we tend to only run the benchmarks in phases when we're digging into performance. This means it's easy for our benchmarks to be silently broken when we make code changes. This PR adds a benchmark testing workflow, and makes `generate_readme_data.py` easy to run in a test mode.

We'll want to expand this workflow to test all benchmarks we care about. This means we'll probably need to move it to GPU instances, but we can iterate on that.


benchmarks/decoders/benchmark_decoders.py
```diff
@@ -13,7 +13,7 @@
 
 from benchmark_decoders_library import (
     AbstractDecoder,
-    BatchParameters,
+    # BatchParameters,
```
scotts, line 16: Ignore changes here; once we fix these failures on trunk, I'll pull those changes. These changes just let this run without error.

NicolasHug:
I suspect we won't move forward with this any time soon, as we might entirely revisit our benchmark strategy. I'll close this but happy to reconsider something similar!

PR #364, opened 2024-11-12
Use std::string_view instead of c10::string_view because the latter is being deprecated

ahmadsharif1:
Summary: `c10::string_view` is being removed, so we need to migrate.

Differential Revision: D65830279


scotts:
Addressed with #439.

PR #365, opened 2024-11-12
Improve CUDA documentation by adding TorchCodec installation instructions

ahmadsharif1:


scotts:
Since we've recently made our README contain the only installation instructions, I think it makes sense to put these instructions in the README. In the "Installing TorchCodec" section, we should probably have one for "CPU only" and "CUDA enabled" (or something close to that).

ahmadsharif1:
Ready for review again @scotts

scotts (approved):

PR #366, opened 2024-11-13
Let wheel names get the proper platform suffix

NicolasHug:


src/torchcodec/__init__.py
```diff
@@ -9,4 +9,10 @@
 from ._frame import Frame, FrameBatch  # usort:skip # noqa
 from . import decoders, samplers  # noqa
 
-__version__ = "0.0.4.dev"
+
+__version__ = "0.0.0"  # Fake version, just makes sure wheel building works
```
NicolasHug, line 13: TorchVision and TorchAudio don't need this because they still rely on `setup.py bdist_wheel` while we rely on the more modern `python -m build`.
The core of the issue here is the `test-infra` requirements of getting the version dynamically from an auto-generated `version.py` file within `setup.py`.

PR #369, opened 2024-11-13
Make workflows run on opened and edited PRs

scotts:


scotts:
Doesn't seem to work as expected. Abandoning in favor of #370.

PR #373, opened 2024-11-13
Release branch: rely on release/2.5 jobs from test-infra

NicolasHug:


scotts (approved):

NicolasHug:
CUDA builds are passing but installation is failing. Not sure why. Looks like potentially a difference in machine. We need to merge this anyway.

PR #378, opened 2024-11-15
Added a decode+resize benchmark and cuda decoder

ahmadsharif1:


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -239,6 +261,12 @@ def __init__(self, num_ffmpeg_threads=None, device="cpu"):
         )
         self._device = device
 
+        import torchvision  # noqa: F401
+        from torchvision.transforms import v2 as transforms_v2
```
scotts, line 265: We don't need to say `import torchvision`. The `from torchvision.transforms` will take care of loading the module.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -239,6 +261,12 @@ def __init__(self, num_ffmpeg_threads=None, device="cpu"):
...
+
+        self.torchvision = torchvision
```
scotts, line 267: I don't think we actually use `self.torchvision`? We should be able to remove it.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -535,6 +595,8 @@ def run_benchmarks(
     results = []
     df_data = []
     verbose = False
+    # TODO: change this back before landing.
+    min_runtime_seconds = 0.1
```
scotts, line 599: I have an unmerged change in PR #362 to make this sort of benchmark testing easier: https://github.com/pytorch/torchcodec/pull/362/files#diff-c378bd5d03e7daa116eaeeeb86921e8134f0feefb0fcdf9f020f872676c5c00dR31-R34

benchmarks/decoders/benchmark_readme_data.json
```diff
@@ -1,225 +1,201 @@
...
-        "iqr": 0.007216569967567921,
-        "median": 0.4879263397306204,
+        "iqr": 0.0,
+        "median": 0.9324539797380567,
         "type": "100 next()",
         "video": "/tmp/torchcodec_benchmarking_videos/nasa_960x540_206s_30fps_yuv420p.mp4"
     },
     {
-        "cpu_count": 56,
+        "cpu_count": 22,
```
scotts, line 195: In `generate_readme_data.py`, we should add something like:
```
  "is_cuda_available": str(torch.cuda.is_available()),
```
In the collection of the system stats.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -111,6 +121,18 @@ def get_consecutive_frames_from_video(self, video_file, numFramesToDecode):
...
+    def decode_and_transform(self, video_file, pts_list, height, width, device):
+        self.torchvision.set_video_backend(self._backend)
+        reader = self.torchvision.io.VideoReader(video_file, "video")
+        frames = []
+        for pts in pts_list:
+            reader.seek(pts)
+            frame = next(reader)
+            frames.append(frame["data"].permute(1, 2, 0))
+        frames = [frame.to(device) for frame in frames]
+        frames = self.transforms_v2.functional.resize(frames, (height, width))
```
scotts, line 133: Naive question that applies to all implementations that use the transformation: how do we ensure it's done on the GPU?

ahmadsharif1: Correct

scotts:
Ideally we would also update the README description, as once this PR is merged, the chart is live. But we can also do that on a follow-up. That is, a 22-core Linux system with whatever kind of GPU.

scotts (approved):

ahmadsharif1:
I tweaked the README.md file as well. Later on we can put the machine info in the chart itself so this doesn't have to be kept manually up-to-date.

PR #379, opened 2024-11-17
Fix off by one error in VideoClipSampler

romanmeta:
Summary:
When we want to sample `N` frames ending at frame with index `i`, we should be taking frames starting at index `(i+1)-N`, not `i-N`.

Example for N = 3:
i-2, i-1, i

Differential Revision: D66074072


src/torchcodec/_samplers/video_clip_sampler.py
```diff
@@ -212,7 +212,7 @@ def _get_clips_for_index_based_sampling(
         sample_start_index = max(0, index_based_sampler_args.sample_start_index)
         sample_end_index = (
             min(
-                index_based_sampler_args.sample_end_index,
+                index_based_sampler_args.sample_end_index + 1,
                 metadata_json["numFrames"],
```
ahmadsharif1, line 216: I am wondering if this needs to have numFrames - 1 here, because we call linspace on it below which can return a value equal to this value.

So that will also index out of bounds.

ahmadsharif1 (commented):
We are deprecating this sampler, so in general changes here are discouraged. However bug fixes may still be okay.

Do you have a test case? Does it work with my proposal?

src/torchcodec/_samplers/video_clip_sampler.py
romanmeta: No, the `numFrames` is actually the correct one. We subtract 1 in the next 2 lines `-index_based_sampler_args.video_frame_dilation * index_based_sampler_args.frames_per_clip`.

romanmeta:
> We are deprecating this sampler, so in general changes here are discouraged. However bug fixes may still be okay.
> 
> Do you have a test case? Does it work with my proposal?

Any 1 frame video currently crashes the sampler. As we pass start_index = 0 and end_index = 0. The end_index becomes -1 and causes exception inside VideoDecoder.

ahmadsharif1 (approved):

PR #382, opened 2024-11-18
[Cherry-pick for 0.0.4] Let wheel names get the proper platform suffix (#366)

NicolasHug:
CP of #366 + version change to 0.0.4


NicolasHug:
This seems to be working as expected, I'm seeing the CPU version as `torchcodec.__version__ = '0.0.4+cpu'` and the wheel names are correct too

PR #384, opened 2024-11-18
[Release] Test against all Python versions

NicolasHug:


ahmadsharif1:
This is not for actual merge, right?

NicolasHug:
No you're right, this is the PR we'll use to download the wheels from.

PR #385, opened 2024-11-18
[Release] Rely on release/2.5 branch - again

NicolasHug:
Forgot these in https://github.com/pytorch/torchcodec/pull/373


ahmadsharif1 (approved):

NicolasHug:
With https://github.com/pytorch/torchcodec/pull/385 in, all wheel building jobs are passing.

The CUDA install-and-test jobs are failing with the same error we see on main, but the CPU ones are passing too. Let's land and try to fix the rest.

PR #390, opened 2024-11-19
Fix Python version in CUDA wheel name

NicolasHug:
This is what is causing the existing errors like

```
ERROR: TorchCodec-0.0.4+cu118-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl is not a supported wheel on this platform.
```

in our release branch https://github.com/pytorch/torchcodec/actions/runs/11901772712/job/33165767469?pr=389

I'll land this on main and cherry-pick for the release branch


NicolasHug:
I'm seeing:

```
    ARTIFACT_NAME: pytorch_torchcodec__3.9_cu124_x86_64
```

So this seems to work as expected. `main` is broken already.

PR #405, opened 2024-11-20
Place a system metadata stamp on readme chart

scotts:
This also reorients the data file itself to be structured as:
```
{
  "experiments":
    [
      # list of individual experiments
    ]
  "system_metadata":
    {
      # key-value pairs of metadata
    }
}
```
Because this changes the data file, I'll wait until the current experiments are done, and then I'll merge things manually.


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -518,6 +518,14 @@ def plot_data(df_data, plot_path):
         for col in range(video_type_combinations[unique_videos[row]], max_combinations):
             fig.delaxes(axes[row, col])
 
+    plt.gcf().text(
```
ahmadsharif1, line 521: I would prefer top left instead of bottom left, but I'll leave it to you

ahmadsharif1 (approved):

benchmarks/decoders/benchmark_decoders_library.py
ahmadsharif1: Also, is_cuda_available isn't that useful. It would be better to get the GPU version

scotts: Done. Honestly, neither look "good". And I realized we need to do either top or bottom; that will be robust to changing the number of videos.

benchmarks/decoders/generate_readme_data.py
```diff
@@ -83,21 +83,24 @@ def main() -> None:
...
-                "cpu_count": os.cpu_count(),
-                "system": platform.system(),
-                "machine": platform.machine(),
-                "python_version": str(platform.python_version()),
-                "is_cuda_available": str(torch.cuda.is_available()),
-            }
-        }
-    )
+    data_for_writing = {
+        "experiments": results,
```
ahmadsharif1, line 87: nit: why not call these results?

scotts: When reading a data file, I find it more clear to see "experiments" and "system_metadata". But in the code, I think "results" is more clear that it's the return value of running the experiments.

ahmadsharif1 (approved):

PR #410, opened 2024-11-22
PR with benchmark results

ahmadsharif1:
Also use 0 explicitly for the number of torchvision threads.


scotts (approved):

ahmadsharif1:
Yeah, unfortunately that's the case. We can do a special case in a different PR

PR #411, opened 2024-11-24
Simpler experiment names

scotts:
1. Simplifies experiment names, naming them more after what they do rather than how they do it.
2. Renames the actual methods to match.
3. Unifies where those names come from. (Except for the concurrency experiment, which is already a special case.)
4. Moves the stamp to the bottom left. It's overlapping with text in the upper left.


ahmadsharif1 (approved):
IMO "decode" feels a bit more explicit/accurate compared to "sample"

scotts:
@ahmadsharif1, sure let's go with that.

PR #413, opened 2024-11-28
Improve GPU frames accuracy

NicolasHug:
We now call `nppiNV12ToRGB_709CSC_8u_P2C3R` for GPU frame color conversion, instead of `nppiNV12ToRGB_709HDTV_8u_P2C3R`, as suggested in https://github.com/pytorch/torchcodec/issues/412

This dramatically improves accuracy of decoded frames, as they are now much closer to the expected CPU ones, within `atol=2` (this is even stricter than our frame comparison for MacOS!).

Closes https://github.com/pytorch/torchcodec/issues/412

Thank you @fmassa for the original hint in https://github.com/pytorch/torchcodec/pull/372#discussion_r1840965468 and @pjs102793 for chasing this up!

### DIFF
--- a/src/torchcodec/decoders/_core/CudaDevice.cpp
+++ b/src/torchcodec/decoders/_core/CudaDevice.cpp
@@ -225,7 +225,6 @@ void convertAVFrameToDecodedOutputOnCuda(
   auto start = std::chrono::high_resolution_clock::now();
   NppStatus status;
   if (src->colorspace == AVColorSpace::AVCOL_SPC_BT709) {
-    // status = nppiNV12ToRGB_709HDTV_8u_P2C3R(
     status = nppiNV12ToRGB_709CSC_8u_P2C3R(
         input,
         src->linesize[0],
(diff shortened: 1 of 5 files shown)


NicolasHug:
CI is completely broken right now, but I can confirm locally that we can now lower all checks to `assert_close(atol=2)` ! So that's a *massive* accuracy improvement

scotts (approved):

scotts:
Wow! Great catch! Thanks @fmassa and @NicolasHug!

src/torchcodec/decoders/_core/CudaDevice.cpp
```diff
@@ -225,7 +225,7 @@ void convertAVFrameToDecodedOutputOnCuda(
   auto start = std::chrono::high_resolution_clock::now();
   NppStatus status;
   if (src->colorspace == AVColorSpace::AVCOL_SPC_BT709) {
-    status = nppiNV12ToRGB_709HDTV_8u_P2C3R(
+    status = nppiNV12ToRGB_709CSC_8u_P2C3R(
```
ahmadsharif1, line 228: Drive-by: this should probably be gated by AVFrame::color_range.

https://ffmpeg.org/doxygen/trunk/pixfmt_8h.html#a3da0bf691418bc22c4bcbe6583ad589a

MPEG has limited color-range while JPEG has full color range.

We were using full for all frames before -- now we are using limited for all. Colorspace and color-range are orthogonal so probably need nested ifs here.

PR #417, opened 2024-12-02
Improve/fix test assertion

NicolasHug:
The main thing this PR fixes is the fact that we were calling `assert_tensor_equal` when checking pts or duration values. That was wrong. The reason it was wrong is that `assert_tensor_equal` allows for `atol=2` or `atol=3` on CUDA or MACOS. Calling `assert_tensor_equal` for checking against a PTS thus means "assert the pts is this value, plu or minus 3 seconds". That's waaaay too loose. This PR changes those checks to something stricter, and renames `assert_tensor_equal` into `assert_frames_equal` so that we know to only call that with frames.

Detailed changes:

- rename `assert_tensor_equal` into `assert_frames_equal`
- fix loose pts and duration checks by calling plain `torch.testing.assert_close(...)`
- remove our shallow wrapper:  `assert_tensor_close` and replace its usage (5-6 cases only, and located in a single function only) with plain `torch.testing.assert_close(...)`.

### DIFF
--- a/test/samplers/test_samplers.py
+++ b/test/samplers/test_samplers.py
@@ -19,7 +19,7 @@ from torchcodec.samplers._common import _POLICY_FUNCTIONS
 from torchcodec.samplers._index_based import _build_all_clips_indices
 from torchcodec.samplers._time_based import _build_all_clips_timestamps
 
-from ..utils import assert_tensor_equal, NASA_VIDEO
+from ..utils import assert_frames_equal, NASA_VIDEO
 
 
 def _assert_output_type_and_shapes(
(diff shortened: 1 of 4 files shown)


test/samplers/test_samplers.py
```diff
@@ -436,15 +436,19 @@ def test_random_sampler_randomness(sampler):
...
-        assert_tensor_equal(clip_1.data, clip_2.data)
-        assert_tensor_equal(clip_1.pts_seconds, clip_2.pts_seconds)
-        assert_tensor_equal(clip_1.duration_seconds, clip_2.duration_seconds)
+        assert_frames_equal(clip_1.data, clip_2.data)
+        torch.testing.assert_close(
+            clip_1.pts_seconds, clip_2.pts_seconds, rtol=0, atol=0
+        )
+        torch.testing.assert_close(
+            clip_1.duration_seconds, clip_2.duration_seconds, rtol=0, atol=0
+        )
```
NicolasHug, line 445: This is an example of the kind of assertions that were way too loose.

scotts (approved):

PR #432, opened 2024-12-13
Lazily init filtergraph so it can respect raw decoded resolution

scotts:
We've encountered runtime errors with some videos where the resolution in the stream metadata disagrees with the resolution of the raw decoded frame. Because we have been initializing filtergraph so early, we've had to rely on the stream metadata. This PR does the initialization laziily, so we can use the resolution of the raw decoded frame.

Note that we still rely on the stream metadata for the batch APIs - they can run into the same problem. Dealing with that case will be more involved, since we'll need to do a sacrificial decode to discover what the size of the tensors should be. We should create an issue to track that case.

Note that I did consider just obeying whatever the stream metadata was. However, for swscale, we're also using the raw decoded frame to determine what the width should be. We should maintain the same behavior no matter which method we're using for the color space conversion.

This change seems to be performance neutral. Running:

```
python benchmarks/decoders/benchmark_decoders.py --decoders decord,decord_batch,torchcodec_public,torchcodec_core,torchaudio --min-run-seconds 20
```
I get:
```
[- video=/home/scottas/github/torchcodec/benchmarks/decoders/../../test/resources/nasa_13013.mp4 h264 480x270, 13.013s 29.97002997002997fps -]
1 threads: -----------------------------------------------------------------------------------------------------------------------------------
      DecordAccurate       |            54.6            |           121.3           |       13.0       |        18.5       |        65.8      
      TorchAudio           |           187.1            |           200.1           |       11.6       |        13.1       |        32.6      
      TorchCodecPublic     |            49.2            |            44.8           |       14.1       |        14.8       |        29.1      
      DecordAccurateBatch  |            52.6            |           118.1           |       13.2       |        16.2       |        48.3      
      TorchCodecCore       |            49.5            |            44.6           |       15.7       |        18.3       |        45.7      
                                                             
```
Which compares favorably with what I reported in #431.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -101,7 +101,7 @@ std::vector<std::string> splitStringWithDelimiters(
   return result;
 }
 
-VideoDecoder::ColorConversionLibrary getDefaultColorConversionLibraryForWidth(
+VideoDecoder::ColorConversionLibrary getDefaultColorConversionLibrary(
     int width) {
   VideoDecoder::ColorConversionLibrary library =
       VideoDecoder::ColorConversionLibrary::SWSCALE;
@@ -299,9 +299,10 @@ std::unique_ptr<VideoDecoder> VideoDecoder::createFromBuffer(
   return decoder;
 }
 
-void VideoDecoder::initializeFilterGraphForStream(
+void VideoDecoder::initializeFilterGraph(
     StreamInfo& streamInfo,
-    const AVFrame& frame) {
+    int expectedOutputHeight,
+    int expectedOutputWidth) {
   FilterState& filterState = streamInfo.filterState;
   if (filterState.filterGraph) {
     return;
@@ -378,10 +379,8 @@ void VideoDecoder::initializeFilterGraphForStream(
   inputs->pad_idx = 0;
   inputs->next = nullptr;
 
-  auto frameDims =
-      getHeightAndWidthFromOptionsOrAVFrame(streamInfo.options, frame);
   std::stringstream description;
-  description << "scale=" << frameDims.width << ":" << frameDims.height;
+  description << "scale=" << expectedOutputWidth << ":" << expectedOutputHeight;
   description << ":sws_flags=bilinear";
 
   AVFilterInOut* outputsTmp = outputs.release();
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -121,7 +121,7 @@ VideoDecoder::ColorConversionLibrary getDefaultColorConversionLibraryForWidth(
 // or 4D.
 // Calling permute() is guaranteed to return a view as per the docs:
 // https://pytorch.org/docs/stable/generated/torch.permute.html
-torch::Tensor VideoDecoder::MaybePermuteHWC2CHW(
+torch::Tensor VideoDecoder::maybePermuteHWC2CHW(
```
scotts, line 124: Drive-by refactor: functions should start with a lower case. Upper case camel case identifiers is reserved for type names.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -300,30 +300,30 @@ std::unique_ptr<VideoDecoder> VideoDecoder::createFromBuffer(
 }
 
 void VideoDecoder::initializeFilterGraphForStream(
-    int streamIndex,
-    const VideoStreamDecoderOptions& options) {
-  FilterState& filterState = streams_[streamIndex].filterState;
+    StreamInfo& streamInfo,
+    const AVFrame& frame) {
```
NicolasHug, line 304: (Not critical for merging)

Instead of passing the entire AVFrame object, should we just pass the height and width? It would make it more obvious what properties from the AVFrame are really needed for initialization, and we could directly pass `expectedOutputHeight` and `expectedOutputWidth` from within `convertAVFrameToDecodedOutputOnCPU`?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -478,7 +478,6 @@ void VideoDecoder::addVideoStreamDecoder(
       options.colorConversionLibrary.value_or(defaultColorConversionLibrary);
 
   if (colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH) {
-    initializeFilterGraphForStream(streamNumber, options);
```
NicolasHug, line 481: (Not critical for merging)

Seems like lines 478 to 492 can now be collapsed into just a one-liner?

```C
streamInfo.colorConversionLibrary = options.colorConversionLibrary.value_or(defaultColorConversionLibrary);
```



In particular I wonder if the `else` clause where we `throw` is/was needed at all?

NicolasHug (approved):
Nice catch and great fix, thank you!

NicolasHug:
> Note that we still rely on the stream metadata for the batch APIs - they can run into the same problem

Are you able to confirm that calling e.g. `decoder.get_frames_at([3, 4])` on that same problematic video hits this exact check:

https://github.com/pytorch/torchcodec/blob/85c1ac0608add7ec6b2a5cf7a9b304e340b78eb2/src/torchcodec/decoders/_core/VideoDecoder.cpp#L905-L916

NicolasHug:
> Dealing with that case will be more involved, since we'll need to do a sacrificial decode to discover what the size of the tensors should be. We should create an issue to track that case.

I remember now that I had documented that potential workaround in code comments.  I opened https://github.com/pytorch/torchcodec/issues/433 to keep track of it.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: I thought about that too, and decided against it so I could use the `getHeightAndWidthFromOptionsOrAVFrame()` convenience function. But let me take another look at the code, because I agree that I don't like passing around the AVFrame pointer more than we have to.

scotts:
> Are you able to confirm that calling e.g. `decoder.get_frames_at([3, 4])` on that same problematic video hits this exact check:
> 
> https://github.com/pytorch/torchcodec/blob/85c1ac0608add7ec6b2a5cf7a9b304e340b78eb2/src/torchcodec/decoders/_core/VideoDecoder.cpp#L905-L916

Yup, confirmed.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -938,6 +937,16 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
     } else if (
         streamInfo.colorConversionLibrary ==
         ColorConversionLibrary::FILTERGRAPH) {
+      // Note that is is a lazy init; we initialize filtergraph the first time
```
NicolasHug, line 940: typo nit

```suggestion
      // Note that is a lazy init; we initialize filtergraph the first time
```

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Oh, duh, there's no need to even call the convenience function inside `getHeightAndWidthFromOptionsOrAVFrame()`. We call it outside, pass the values in, no need to check options again.

PR #434, opened 2024-12-14
Expose constructors in C++

scotts:
Prior to this PR, the `VideoDecoder` class in C++ had a single private constructor that did not do anything. The static create family of functions were effectively the constructors because they would set up the fields and do the initialization. We also had an empty `DecoderOptions` struct that was intentioned to be filled with parameters we would need at construction time. This PR:

1. Adds two public, explicit constructors.
2. Implements the create family of functions by calling the constructors.
3. Removes the empty options struct.

The value of doing this:

1. C++ has rich semantics and features around object construction. We were not taking advantage of it, and instead on a path to replicate some it by hand.
2. Before this change, no one could instantiate a `VideoDecoder` object on the stack. There's no need to prevent that use.
3. It's easier to reason about C++ objects when they obey the principle that the object is valid if the constructor has succeeded.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -219,14 +219,17 @@ bool VideoDecoder::SwsContextKey::operator!=(
 VideoDecoder::VideoDecoder(const std::string& videoFilePath) {
   AVInput input = createAVFormatContextFromFilePath(videoFilePath);
   formatContext_ = std::move(input.formatContext);
+
   initializeDecoder();
 }
 
 VideoDecoder::VideoDecoder(const void* buffer, size_t length) {
   TORCH_CHECK(buffer != nullptr, "Video buffer cannot be nullptr!");
+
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -275,28 +290,14 @@ void VideoDecoder::initializeDecoder() {
...
-  return decoder;
+    const std::string& videoFilePath) {
+  return std::unique_ptr<VideoDecoder>(new VideoDecoder(videoFilePath));
 }
 
 std::unique_ptr<VideoDecoder> VideoDecoder::createFromBuffer(
     const void* buffer,
-    size_t length,
-    const VideoDecoder::DecoderOptions& options) {
-  TORCH_CHECK(buffer != nullptr, "Video buffer cannot be nullptr!");
```
scotts, line 292: Note that this `TORCH_CHECK` was almost certainly optimized out in optimized builds. Because we already used `buffer`, it being null at this point would be undefined behavior, and the C++ compiler is allowed to assume it won't happen. Null checks must always be before use.

PR #435, opened 2024-12-16
Refactor VideoDecoder C++ initialization

scotts:
In the C++ layer, in `VideoDeocder::initializeDecoder()`, we were calling `std::vector::resize()`, adding 1 each time, and then calling `std::vector::back()` to get a reference to the last element and fill it out. That is not how vectors are designed to be used. Instead, in the loop, we just create a struct on the stack, fill it with values, and call `std::vector::push_back()`.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -234,36 +234,41 @@ VideoDecoder::VideoDecoder(const void* buffer, size_t length) {
 }
 
 void VideoDecoder::initializeDecoder() {
-  // Some formats don't store enough info in the header so we read/decode a few
-  // frames to grab that. This is needed for the filter graph. Note: If this
-  // takes a long time, consider initializing the filter graph after the first
-  // frame decode.
+  TORCH_CHECK(!initialized_, "Attempted double initialization.");
+
   int ffmpegStatus = avformat_find_stream_info(formatContext_.get(), nullptr);
   if (ffmpegStatus < 0) {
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -234,59 +234,72 @@ VideoDecoder::VideoDecoder(const void* buffer, size_t length) {
 }
 
 void VideoDecoder::initializeDecoder() {
-  // Some formats don't store enough info in the header so we read/decode a few
-  // frames to grab that. This is needed for the filter graph. Note: If this
-  // takes a long time, consider initializing the filter graph after the first
-  // frame decode.
```
NicolasHug, line 240: Curious if we should keep that comment, or what was the reason to remove it? It might be helpful to keep (especially the "Note" part), particularly so that the reader knows that some frame decoding does happen within `avformat_find_stream_info()`?

I'm finding this from the [FFmpeg docs](https://ffmpeg.org/doxygen/6.1/group__lavf__decoding.html)

> Some formats do not have a header or do not store enough information there, so it is recommended that you call the [avformat_find_stream_info()](https://ffmpeg.org/doxygen/6.1/group__lavf__decoding.html#gad42172e27cddafb81096939783b157bb) function which tries to read and decode a few frames to find missing **information.**

scotts: Good catch. I removed the comment because of the latter part about lazy initialization. But the other stuff is both relevant and informative. I'll submit a PR which revives the relevant stuff.

PR #451, opened 2025-01-10
Test if h265 can work on other GPU runners

scotts:


NicolasHug:
I think we can close this, we do test on h265 with CUDA now, for both the FFmpeg and Beta interfaces.

PR #452, opened 2025-01-16
Make C++ compilation strict, treat warnings as errors

scotts:
We still sometimes get C++ code that compiles externally that does not compile internally. I think one way to make this less likely is for our CMake builds to be more restrictive. This means compiling with:
```
-Wall -Wextra -pedantic -Werror
```
However, turning that on means we need to fixup a bunch of integer comparisons, so there are a lot of small changes in this PR. It's also worth noting that I still could not reproduce the errors I get on internal builds. I think there will always be the possibility of disagreement because of different compilers.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -112,7 +112,7 @@ jobs:
         run: |
           cd docs
           ${CONDA_RUN} make html
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: Built-Docs
           path: docs/build/html/
--- a/src/torchcodec/decoders/_core/CMakeLists.txt
+++ b/src/torchcodec/decoders/_core/CMakeLists.txt
@@ -4,7 +4,7 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 find_package(Torch REQUIRED)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS} -Wall -Wextra -pedantic -Werror")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -pedantic -Werror ${TORCH_CXX_FLAGS}")
 find_package(Python3 ${PYTHON_VERSION} EXACT COMPONENTS Development)
 
 function(make_torchcodec_library library_name ffmpeg_target)
--- a/src/torchcodec/decoders/_core/CMakeLists.txt
+++ b/src/torchcodec/decoders/_core/CMakeLists.txt
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.18)
 project(TorchCodec)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
-add_compile_options(-Wall -Wextra -pedantic -Werror -Wparentheses)
+add_compile_options(-Wall -Wextra -pedantic -Werror)
 
 find_package(Torch REQUIRED)
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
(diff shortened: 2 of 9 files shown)


src/torchcodec/decoders/_core/FFMPEGCommon.cpp
```diff
@@ -21,7 +21,7 @@ int64_t getDuration(const UniqueAVFrame& frame) {
 }
 
 int64_t getDuration(const AVFrame* frame) {
-#if LIBAVUTIL_VERSION_MAJOR < 59
+#if LIBAVUTIL_VERSION_MAJOR < 58
```
scotts, line 24: I think we just had this wrong. It was deprecated in 58.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -206,7 +206,8 @@ VideoDecoder::BatchDecodedOutput::BatchDecodedOutput(
 
 bool VideoDecoder::DecodedFrameContext::operator==(
     const VideoDecoder::DecodedFrameContext& other) {
-  return decodedWidth == other.decodedWidth && decodedHeight == decodedHeight &&
```
scotts, line 209: Note that the stricter compilation revealed a real logic error here, as we had the code `decodedHeight == decodedHeight`.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.18)
 project(TorchCodec)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+add_compile_options(-Wall -Wextra -pedantic -Werror)
```
NicolasHug, line 5: I did a brief google investigation and it looks like `add_compile_options` is similar to setting `CMAKE_CXX_FLAGS` (as we do below line 7-8), with a different scope.
While [some sources](https://www.reddit.com/r/cpp_questions/comments/xdv88y/difference_between_cmake_ways_to_add_compile_flags/) suggest that `add_compile_options` is better scoped, I feel like we should stick to setting `CMAKE_CXX_FLAGS` here. Because a) this is what we already do, so it's best to only have one way to set flags, and b) we want to set the flags of the entire project anyway, for all targets, which is what `CMAKE_CXX_FLAGS` does.

NicolasHug (commented):
Thanks for the fixes @scotts ! Looks like we have some compile failures. Not sure if those are from this PR or if `main` is failing too - do you mind checking?

scotts:
The compilation failures are almost certainly caused by this PR; I'll dig in. What this does open us up to is slight differences in compilers can now cause compilation failure. For example, everything compiled clean on my system, but then I've had to deal with at least two problems in CI.

On the one hand, that's kind of the point of this PR: let's be more strict with the hopes we will fail less internally. On the other hand, it does mean we (and other open source devs) will run into this whack-a-mole with CI systems may be more strict than local. Hopefully that calms down after this PR that changes a bunch at once.

.github/workflows/docs.yaml
```diff
@@ -112,7 +112,7 @@ jobs:
         run: |
           cd docs
           ${CONDA_RUN} make html
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
```
scotts, line 115: Missed doing this one last week, and it only just now started failing.

scotts:
Everything works now. And I have more confidence this is the right thing to do, because the failing checks were in CUDA code. My dev environment is non-CUDA, hence why I didn't catch it locally.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 find_package(Torch REQUIRED)
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS}")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${TORCH_CXX_FLAGS} -Wall -Wextra -pedantic -Werror")
```
NicolasHug, line 7: Nit: I wonder if we should add those **before** `${TORCH_CXX_FLAGS}` so that they can take precedence?

NicolasHug (approved):
Thank you @scotts !

PR #454, opened 2025-01-20
Better variable names for `scanFileAndUpdateMetadataAndIndex`

NicolasHug:
In `scanFileAndUpdateMetadataAndIndex`, we use the variable name `stream` to denote 3 different kinds of objects:

- `StreamMetadata`
- `AVStream`
- `StreamInfo`

I personally find this particularly confusing when reading the code, because this forces me to look at the variable declaration (in the best case), or at the function declaration (when `auto` is used) to figure out what the type is.

This PR renames the single `stream` name into:

- `streamMetadata`
- `avStream`
- `streamInfo`

I propose to align the entire code-base to these names (for future PRs).

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -562,13 +562,14 @@ void VideoDecoder::scanFileAndUpdateMetadataAndIndex() {
     if (packet->flags & AV_PKT_FLAG_DISCARD) {
       continue;
     }
-    auto& stream = containerMetadata_.streams[streamIndex];
-    stream.minPtsFromScan =
-        std::min(stream.minPtsFromScan.value_or(INT64_MAX), packet->pts);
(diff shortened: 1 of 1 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -578,16 +579,17 @@ void VideoDecoder::scanFileAndUpdateMetadataAndIndex() {
     }
     streams_[streamIndex].allFrames.push_back(frameInfo);
   }
-  for (size_t i = 0; i < containerMetadata_.streams.size(); ++i) {
-    auto& streamMetadata = containerMetadata_.streams[i];
-    auto stream = formatContext_->streams[i];
+  for (size_t streamIndex = 0; streamIndex < containerMetadata_.streams.size();
```
NicolasHug, line 582: also renamed `i` to streamIndex since `streamIndex` is already (and properly) used just above, and just below.

scotts (approved):

PR #456, opened 2025-01-20
Remove unused fields from `RawDecodedOutput` struct

NicolasHug:
These don'tseem to be used anywhere. If tests are green, we may want to consider renaming `RawDecodedOutput` to something more descriptive. At this point it's just an AVFrame and a stream index.


scotts:
Agreed with the principle of simplification and better naming.

scotts (approved):

PR #457, opened 2025-01-20
Remove `streamType` field from `DecodedOutput`

NicolasHug:
We pretend to support other types than `AVMEDIA_TYPE_VIDEO` with this field, but this is a lie. We have strong baked-in assumptions everywhere that we currently only support video frames. The main indicator is that we don't allow to add a stream that's not a video stream.

Since we'll be tackling audio support soon, I think it's best to remove any relics from the past, since the code-base has evolved in ways that may now be incompatible with how we projected things at the time. Or, said differently, we don't want our design of audio support to be influenced by code we wrote 6 months ago.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -869,7 +869,6 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
   AVFrame* frame = rawOutput.frame.get();
   output.streamIndex = streamIndex;
   auto& streamInfo = streams_[streamIndex];
-  output.streamType = streams_[streamIndex].stream->codecpar->codec_type;
   output.pts = frame->pts;
   output.ptsSeconds =
       ptsToSeconds(frame->pts, formatContext_->streams[streamIndex]->time_base);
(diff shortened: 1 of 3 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -930,86 +929,78 @@ void VideoDecoder::convertAVFrameToDecodedOutputOnCPU(
...
-    // conversion objects as much as possible for performance reasons.
-    enum AVPixelFormat frameFormat =
-        static_cast<enum AVPixelFormat>(frame->format);
-    auto frameContext = DecodedFrameContext{
-        frame->width,
-        frame->height,
-        frameFormat,
-        expectedOutputWidth,
-        expectedOutputHeight};
+  // We need to compare the current frame context with our previous frame
```
NicolasHug, line 932: Diff looks bigger than it actually is because this block got indented to the left.

scotts (approved):

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -869,7 +869,6 @@ VideoDecoder::DecodedOutput VideoDecoder::convertAVFrameToDecodedOutput(
   AVFrame* frame = rawOutput.frame.get();
   output.streamIndex = streamIndex;
   auto& streamInfo = streams_[streamIndex];
-  output.streamType = streams_[streamIndex].stream->codecpar->codec_type;
```
scotts, line 872: I do think there is some value in doing a `TORCH_CHECK(streamInfo.stream->codecpar->codec_type == AVMEDIA_TYPE_VIDEO)` here or close by to make our assumptions explicit.

PR #461, opened 2025-01-21
Move packet allocation out of decoding loop

NicolasHug:
### What's in main

We allocate AVPacket objects in 2 places. In `scanFileAndUpdateMetadataAndIndex`:https://github.com/pytorch/torchcodec/blob/b1719e77a8b20022fd92bf24ac58b4dfac6d4e69/src/torchcodec/decoders/_core/VideoDecoder.cpp#L550

and in `getDecodedOutputWithFilter()`, the main decoding loop:

https://github.com/pytorch/torchcodec/blob/b1719e77a8b20022fd92bf24ac58b4dfac6d4e69/src/torchcodec/decoders/_core/VideoDecoder.cpp#L802

The `UniqueAVPacket` calls `av_packet_free` when it goes out of scope:

https://github.com/pytorch/torchcodec/blob/b1719e77a8b20022fd92bf24ac58b4dfac6d4e69/src/torchcodec/decoders/_core/FFMPEGCommon.h#L60-L61

In both cases, we allocate these objects *within* the decoding loop. That means that for each packet, we allocate a new `AVPacket` object.

### What's in other FFmpeg examples

 Now, [some resources](https://github.com/leandromoreira/ffmpeg-libav-tutorial/blob/master/0_hello_world.c#L166) I've seen, including [FFmpeg's tutorials](https://ffmpeg.org/doxygen/7.0/demux_decode_8c-example.html), don't do that. They allocate one **single** `AVPacket` object outside of the loop, and call [`av_packet_unref()`](https://ffmpeg.org/doxygen/7.0/group__lavc__packet.html#ga63d5a489b419bd5d45cfd09091cbcbc2) manually. (Note that `av_packet_free`, which is what we call in `main`, [also calls](https://ffmpeg.org/doxygen/6.0/avpacket_8c_source.html#l00073) `av_packet_unref()`)

### What this PR does

This PR does what the other examples do: move the `AVPacket` allocation outside of the loop, and manually call `av_packet_unref()` when needed.

### Do we want this?

I don't know how I feel about this. Sure, this can save a fair amount `malloc` calls, but whether this is noticeable is to be benchmarked. Importantly, we now run the risk of forgetting to call `av_packet_unref()` when needed, and leaking memory.
Perhaps there's a smart C++ construct we can use to make sure `av_packet_unref()` is called when needed, while still calling `av_packet_alloc()` only once?

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -763,6 +763,7 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
   }
   // Need to get the next frame or error from PopFrame.
   UniqueAVFrame frame(av_frame_alloc());
+  UniqueAVPacket packet(av_packet_alloc());
   int ffmpegStatus = AVSUCCESS;
   bool reachedEOF = false;
   int frameStreamIndex = -1;
@@ -802,7 +803,6 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
       // pulling frames from its internal buffers.
       continue;
     }
-    UniqueAVPacket packet(av_packet_alloc());
     ffmpegStatus = av_read_frame(formatContext_.get(), packet.get());
     decodeStats_.numPacketsRead++;
     if (ffmpegStatus == AVERROR_EOF) {
@@ -820,6 +820,7 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
         }
       }
       reachedEOF = true;
+      av_packet_unref(packet.get());
       continue;
     }
     if (ffmpegStatus < AVSUCCESS) {
@@ -829,16 +830,18 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
     }
     if (activeStreamIndices_.count(packet->stream_index) == 0) {
       // This packet is not for any of the active streams.
+      av_packet_unref(packet.get());
       continue;
     }
     ffmpegStatus = avcodec_send_packet(
         streams_[packet->stream_index].codecContext.get(), packet.get());
-    decodeStats_.numPacketsSentToDecoder++;
     if (ffmpegStatus < AVSUCCESS) {
       throw std::runtime_error(
           "Could not push packet to decoder: " +
           getFFMPEGErrorStringFromErrorCode(ffmpegStatus));
     }
+    decodeStats_.numPacketsSentToDecoder++;
+    av_packet_unref(packet.get());
   }
   if (ffmpegStatus < AVSUCCESS) {
     if (reachedEOF || ffmpegStatus == AVERROR_EOF) {
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -573,9 +573,9 @@ void VideoDecoder::scanFileAndUpdateMetadataAndIndex() {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


scotts:
We can do this in a very C++ way, but I understand it might seem too cute or just plain overkill. First, `UniqueAVPacket` becomes (we're assuming the existing `Deleter` utility we have defined):
```
class SharedAVPacket {
  friend class ReferenceAVPacket;
  private:
    std::shared_ptr<AVPacket, Deleter<AVPacket, void, av_packet_free>> packet;
  public:
    SharedAVPacket(): packet(av_packet_alloc()) {}
};
```
That is, we wrap a shared pointer but we do not create a getter. However, `ReferenceAVPacket` has full access to the internals of `SharedAVPacket`.

And then we define:
```
class ReferenceAVPacket {
  private:
    std::shared_ptr<AVPacket, Deleter<AVPacket, void, av_packet_free>> packet;
  public:
    ReferenceAVPacket(SharedAVPacket& shared): packet(shared.packet) {}
    ~ReferenceAVPacket() { av_packet_unref(packet.get()) }    
    AVPacket& operator* { return *packet.get(); }
    AVPacket* operator->  { return packet.get(); }
};
```
Because `SharedAVPacket` has no getter, we can't actually use it. We have to first pass it to a `ReferenceAVPacket` to access it. And `ReferenceAVPacket` will appropriately call `av_packet_unref` when it goes out of scope. Note that I'm not creating an alias of for the `shared_ptr` type; we actually *don't* want others using it.

Example usage would look like:
```
SharedAVPacket packet; // note we moved the actual allocation inside constructor
while (true) {
  ReferenceAVPacket packetRef(packet);
  int ret = av_read_frame(formatContext_.get(), *packet);
  // no matter how the loop ends, we'll always call av_packet_unref correctly
}
// when this outer scope ends, we'll always call av_packet_free correctly
```

scotts:
Another thought: it may be simpler if we just don't use the `shared_ptr`, and just store naked pointers in the objects, and then `SharedAVPacket` would call the free function explicitly in its destructor. The reference counting that `shared_ptr` does may be overkill for our needs.

With the `shared_ptr`, it's okay if the `ReferenceAVPacket` object outlives the `SharedAVPacket` object. The shared_ptr handles that for us. But if we decided we don't need that, we can not use `shared_ptr`, and then require that the `SharedAVPacket` always outlives the `ReferenceAVPacket`.

Yet another wrinkle: I just realized that the act that the action that bumps FFmpeg's reference count on the real packet is the call to `av_read_frame()`. That would imply we really want that call involved in the creation of a `ReferenceAVPacket` somehow, but I'm not sure how.

NicolasHug:
Thanks for the suggestion @scotts ! I ended up implementing the plain pointer version, with slightly different names. Let me know what you think!

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -874,12 +877,12 @@ VideoDecoder::RawDecodedOutput VideoDecoder::getDecodedOutputWithFilter(
     }
     ffmpegStatus = avcodec_send_packet(
         streams_[packet->stream_index].codecContext.get(), packet.get());
-    decodeStats_.numPacketsSentToDecoder++;
     if (ffmpegStatus < AVSUCCESS) {
       throw std::runtime_error(
           "Could not push packet to decoder: " +
           getFFMPEGErrorStringFromErrorCode(ffmpegStatus));
     }
+    decodeStats_.numPacketsSentToDecoder++;
```
scotts, line 885: Ha! Good catch.

scotts (approved):

scotts:
I love it! Very elegant. I do have name suggestions, but if you feel strongly about the names you go, we can merge. This is a clear improvement.

PR #462, opened 2025-01-22
Align variable and type names across the codebase

NicolasHug:
Needs https://github.com/pytorch/torchcodec/pull/440 to be merged first.

This is the continuation of https://github.com/pytorch/torchcodec/pull/454, extending it to the whole codebase. This PR renames types and variable names in order to make it more obvious what they refer to. Specifically:

- use `streamInfo` for `StreamInfo` structs
- use `avStream` for `AVStream` structs
- use `streamIndex` for stream indices (instead of `streamNumber`)
- use `avFrame` for `AVFrame` structs
- use `streamMetadata` for `StreamMetadata` objects
- rename the `streams_` field of `ContainerMetadata` into `streamMetadatas`. The `s` at the end feels a bit weird. Maybe this should be `streamsMetadata`, but that would be inconsistent with other similar vec fields, like `StreamInfos` (where the s is at the end). No strong opinion.
- renamed `VideoStreamDecoderOptions` into `VideoStreamOptions`
- use `videoStreamOptions` for `VideoStreamOptions` structs


scotts:
> rename the streams_ field of ContainerMetadata into streamMetadatas. The s at the end feels a bit weird. Maybe this should be streamsMetadata, but that would be inconsistent with other similar vec fields, like StreamInfos (where the s is at the end). No strong opinion.

I prefer `streamsMetadata` to `streamMetadatas`. Yeah, it's inconsistent with `streamInfos`, but "datas" just read wrong to me. But maybe `allStreamMetadata` is better than both.

scotts (approved):

NicolasHug:
CPU jobs are failling with an unrelated dep-related issue. All GPU jobs are compiling and running the tests, so this is good to land. Thanks for the review!

PR #469, opened 2025-01-23
More renaming, align output structs with Python code

NicolasHug:
This PR does a bunch of renamings towards aligning the C++ with the Python one.

- `DecodedOutput` is the moral equivalent of Python's `Frame`. It is now `FrameOutput`. Not `Frame`, because it feels to generic (there are many notions of "frame" in our C++ code, so the Output suffix helps disambiguate, and indicates this is what the decoder **returns**). Variables of that types are named `frameOutput`
- `BatchDecodedOutput` is the moral equivalent of Python's `FrameBatch`. It is now `FrameBatchOutput`. Variables of that type are named `frameBatchOutput`.
- The `.frame` and `.frames` tensor fields are now name `.data` in both structs, just like in Python.
- `RawDecodedOutput` becomes `AVFrameWithStreamIndex`. Variables of that type become `avFrameWithStreamIndex`. It's a bit of a mouthful. Happy to consider something else. I was tempted to call those variables just `avFrame`, but that would lead to awkward situations when we do e.g. `AVFrame* avFrame = avFrameWithStreamIndex.avFrame.get();`.


scotts:
I'm a big fan of all of this, with the exception of `AVFrameWithStreamIndex`. The problem here is that the `AVPacket` has the stream index, but when we decode that packet into an `AVFrame`, the resulting `AVFrame` has no stream index. That's really unfortunate.

I propose we call it `AVFrameStream`. My thinking is that this is the `AVFrame` for a particular stream - that we know the particular stream through the stream's index is an implementation detail. That's a bit less of a mouthful, and the standard variable name would be `avFrameStream` which is a bit better, too. Since this type is on the critical path in a lot of places, I think making it more readable is important.

scotts (approved):

PR #471, opened 2025-01-23
Rename `maybeDesiredPts_` into `desiredPts_`

NicolasHug:
I might be missing something, but I'm not sure why there is a `maybe` prefix. This field **is** the desired Pts we want to seek to. It's not conditional on anything.

The `maybe` prefix seems to be related to the logic of `maybeSeekToBeforeDesiredPts()`: it makes sense there, since we have a heuristic to decide whether or not we'll call `avformat_seek_file()`. We may or may not seek, hence the `maybe`.

But, regardless of whether a seek was done, we *will* end up at `maybeDesiredPts`. So, to me, this should simply be `desiredPts_`.


scotts:
The "maybe" is to indicate it's an optional. Agreed it's not needed on the value. However, I think we should rename it to `desiredPtsSeconds_`. As named, it would imply an int64 pts timestamp, but it's actually a double that represents time in seconds.

scotts:
Optimistically approving because I assume we'll both agree on the final name. :)

scotts (approved):

PR #477, opened 2025-01-24
Reorganize public APIs in `videoDecoder.h`

NicolasHug:
This PR re-organizes the public-facing part of  `videoDecoder.h`. Some APIs were grouped together when it didn't really make sense anymore. I also deleted outdated comments (or at risk of being outdated), and expanded on some.

Notably, this PR makes the constructor private. I think it makes sense since the decoder cannot be used without calling `initializeDecoder()`, which is already private.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.h
+++ b/src/torchcodec/decoders/_core/VideoDecoder.h
@@ -16,30 +16,7 @@
 
 namespace facebook::torchcodec {
 
-/*
-The VideoDecoder class can be used to decode video frames to Tensors.
-
-Example usage of this class:
-std::string video_file_path = "/path/to/video.mp4";
-VideoDecoder video_decoder = VideoDecoder::createFromFilePath(video_file_path);
-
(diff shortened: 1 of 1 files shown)


src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -288,9 +274,11 @@ class VideoDecoder {
   DecodeStats getDecodeStats() const;
   void resetDecodeStats();
 
-  double getPtsSecondsForFrame(int streamIndex, int64_t frameIndex);
-
  private:
+  explicit VideoDecoder(const std::string& videoFilePath, SeekMode seekMode);
```
scotts, line 278: The constructors have to remain public. Yes, `initializeDecoder()` is private, but the constructors call that. I'd like to eventually remove the static functions `createFromFilePath()` and `createFromBuffer()` and just have all callers use the constructors.

scotts (approved):

PR #479, opened 2025-01-24
Reorganize private APIs in videoDecoder.h

NicolasHug:
Needs https://github.com/pytorch/torchcodec/pull/478 first.

Same intent as https://github.com/pytorch/torchcodec/pull/477, but for private parts this time.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.h
+++ b/src/torchcodec/decoders/_core/VideoDecoder.h
@@ -16,30 +16,7 @@
 
 namespace facebook::torchcodec {
 
-/*
-The VideoDecoder class can be used to decode video frames to Tensors.
-
-Example usage of this class:
-std::string video_file_path = "/path/to/video.mp4";
-VideoDecoder video_decoder = VideoDecoder::createFromFilePath(video_file_path);
-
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -317,73 +304,117 @@ class VideoDecoder {
...
-      const std::vector<VideoDecoder::FrameInfo>& keyFrames,
-      int64_t pts) const;
-  int getKeyFrameIndexForPts(const StreamInfo& stream, int64_t pts) const;
+  // --------------------------------------------------------------------------
+  // CONSTRUCTORS AND INITIALIZERS
+  // --------------------------------------------------------------------------
+  // Don't use those, use the static methods to create a decoder object.
+
+  explicit VideoDecoder(const std::string& videoFilePath, SeekMode seekMode);
+  explicit VideoDecoder(const void* buffer, size_t length, SeekMode seekMode);
```
scotts, line 346: Flagging that per #477, we want to keep these public. We need to make sure that stays the case when we merge this PR with the results of #477 and #478.

scotts (approved):

PR #483, opened 2025-01-27
Remove multi-stream related code

NicolasHug:
Towards https://github.com/pytorch/torchcodec/issues/476

The main point of this PR is to raise an error when the usert tries to add more than one stream. This is a bug fix: previously, we weren't erroring, but we were returning silently wrong results as described in https://github.com/pytorch/torchcodec/issues/476.

In addition, we remove multi-stream logic in this PR: we don't support multiple streams or multi-plexing, so there's no point pretending. Effectively, this removes the `activeStreamIndices_` set and replaces it with `int activeStreamIndex_`. We keep the `streamInfos_` map and it is unchanged, as its logic is relevant for metadata.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -449,11 +449,7 @@ int VideoDecoder::getBestStreamIndex(AVMediaType mediaType) {
 void VideoDecoder::addVideoStreamDecoder(
     int preferredStreamIndex,
     const VideoStreamOptions& videoStreamOptions) {
-  if (activeStreamIndices_.count(preferredStreamIndex) > 0) {
-    throw std::invalid_argument(
-        "Stream with index " + std::to_string(preferredStreamIndex) +
-        " is already active.");
-  }
+  TORCH_CHECK(activeStreamIndex_ == -1, "Can only add one single stream.");
   TORCH_CHECK(formatContext_.get() != nullptr);
 
   AVCodecOnlyUseForCallingAVFindBestStream avCodec = nullptr;
@@ -520,7 +516,7 @@ void VideoDecoder::addVideoStreamDecoder(
   }
 
   codecContext->time_base = streamInfo.stream->time_base;
-  activeStreamIndices_.insert(streamIndex);
+  activeStreamIndex_ = streamIndex;
   updateMetadataWithCodecContext(streamInfo.streamIndex, codecContext);
   streamInfo.videoStreamOptions = videoStreamOptions;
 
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -941,11 +897,10 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
...
   // av_receive_frame() or the user will have seeked to a different location in
   // the file and that will flush the decoder.
-  StreamInfo& activeStreamInfo = streamInfos_[frameStreamIndex];
-  activeStreamInfo.currentPts = avFrame->pts;
-  activeStreamInfo.currentDuration = getDuration(avFrame);
+  streamInfo.currentPts = avFrame->pts;
+  streamInfo.currentDuration = getDuration(avFrame);
 
-  return AVFrameStream(std::move(avFrame), frameStreamIndex);
+  return AVFrameStream(std::move(avFrame), activeStreamIndex_);
```
NicolasHug, line 903: We don't really need this `AVFrameStream` struct anymore, since we know the streamIndex is always `activeStreamIndex_`. We could remove it, but I guess this is better done in another PR.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -907,15 +864,14 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
           getFFMPEGErrorStringFromErrorCode(ffmpegStatus));
     }
 
-    if (activeStreamIndices_.count(packet->stream_index) == 0) {
-      // This packet is not for any of the active streams.
+    if (packet->stream_index != activeStreamIndex_) {
```
NicolasHug, line 867: BTW, this is the "demux" part now. And I think we should be calling `av_read_frame` within its own `while` loop, for as long as the received packet isn't of the target stream.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -472,9 +471,7 @@ class VideoDecoder {
   ContainerMetadata containerMetadata_;
   UniqueAVFormatContext formatContext_;
   std::map<int, StreamInfo> streamInfos_;
-  // Stores the stream indices of the active streams, i.e. the streams we are
-  // decoding and returning to the user.
-  std::set<int> activeStreamIndices_;
+  int activeStreamIndex_ = -1;
```
scotts, line 474: I'd prefer to both make this a named constant, and to use something other than -1. Perhaps:
```
const int NO_ACTIVE_STREAM = -2;
```
Obviously 0 and above are potential active stream numbers, so they can't be used as such an indicator. But I'm worried that -1 can get confusing, because that is used to ask FFmpeg to find the best stream in some API calls. So I'd like to also treat -1 as a special stream value.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Yeah, what we're currently doing is inefficient, because if the packet is not the right stream, we first have to make a call to `avcodec_receive_frame()` that we know will fail. We're still doing the correct thing (I think), it's just inefficient.

scotts (approved):

PR #484, opened 2025-01-27
Rough implementation of getting key frame indices

scotts:
Only works in exact mode. Note that we had to start explicitly tracking `frameIndex` in our `FrameInfo` struct. That allows us to easily know the overall index for a key frame.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoderOps.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoderOps.cpp
@@ -48,7 +48,7 @@ TORCH_LIBRARY(torchcodec_ns, m) {
       "get_frames_by_pts_in_range(Tensor(a!) decoder, *, int stream_index, float start_seconds, float stop_seconds) -> (Tensor, Tensor, Tensor)");
   m.def(
       "get_frames_by_pts(Tensor(a!) decoder, *, int stream_index, float[] timestamps) -> (Tensor, Tensor, Tensor)");
-  m.def("get_key_frame_indices(Tensor(a!) decoder, int stream_index) -> int[]");
+  m.def("_get_key_frame_indices(Tensor(a!) decoder, int stream_index) -> int[]");
   m.def("get_json_metadata(Tensor(a!) decoder) -> str");
   m.def("get_container_json_metadata(Tensor(a!) decoder) -> str");
   m.def(
@@ -335,7 +335,7 @@ bool _test_frame_pts_equality(
       videoDecoder->getPtsSecondsForFrame(stream_index, frame_index);
 }
 
-std::vector<int64_t> get_key_frame_indices(
+std::vector<int64_t> _get_key_frame_indices(
     at::Tensor& decoder,
     int64_t stream_index) {
   auto videoDecoder = unwrapTensorToGetDecoder(decoder);
@@ -534,7 +534,7 @@ TORCH_LIBRARY_IMPL(torchcodec_ns, CPU, m) {
   m.impl("add_video_stream", &add_video_stream);
   m.impl("_add_video_stream", &_add_video_stream);
   m.impl("get_next_frame", &get_next_frame);
-  m.impl("get_key_frame_indices", &get_key_frame_indices);
+  m.impl("_get_key_frame_indices", &_get_key_frame_indices);
   m.impl("get_json_metadata", &get_json_metadata);
   m.impl("get_container_json_metadata", &get_container_json_metadata);
   m.impl("get_stream_json_metadata", &get_stream_json_metadata);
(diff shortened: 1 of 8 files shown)


src/torchcodec/decoders/_core/VideoDecoderOps.cpp
```diff
@@ -48,6 +48,7 @@ TORCH_LIBRARY(torchcodec_ns, m) {
       "get_frames_by_pts_in_range(Tensor(a!) decoder, *, int stream_index, float start_seconds, float stop_seconds) -> (Tensor, Tensor, Tensor)");
   m.def(
       "get_frames_by_pts(Tensor(a!) decoder, *, int stream_index, float[] timestamps) -> (Tensor, Tensor, Tensor)");
+  m.def("get_key_frame_indices(Tensor(a!) decoder, int stream_index) -> int[]");
```
NicolasHug, line 51: I think we can be conservative and expose it *privately* in the Python core API, for now? I.e. as `_get_key_frame_indices` instead of `get_key_frame_indices`

src/torchcodec/decoders/_core/video_decoder_ops.py
```diff
@@ -255,6 +256,13 @@ def get_frames_by_pts_in_range_abstract(
     )
 
 
+@register_fake("torchcodec_ns::get_key_frame_indices")
+def get_key_frame_indices_abstract(
+    decoder: torch.Tensor, *, stream_index: int
+) -> List[int]:
```
NicolasHug, line 262: I suspect this will be more useful if we returned a tensor instead of a list. Because tensor are a lot easier to  sub-sample with fancy indexing than lists are, and this function will typically be used in the context of a sampler. E.g.

```py
all_key_frame_indices = decoder._get_key_frame_indices()
NUM_KEY_FRAMES_TO_GET = 10
subset = torch.randint(0, len(all_key_frame_indices), size=(NUM_KEY_FRAMES_TO_GET,))
key_frames_indices = all_key_frame_indices[subset]

```



(Also, I'm assuming that the `vec<int> -> tensor` binding is as cheap or cheaper than the `vec<int> -> list[int]` conversion)

NicolasHug (approved):
Some non-blocking suggestions, approving to unblock

test/decoders/test_video_decoder.py
```diff
@@ -831,6 +831,14 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
         with pytest.raises(ValueError, match="Invalid stop seconds"):
             frame = decoder.get_frames_played_in_range(0, 23)  # noqa
 
+    @pytest.mark.parametrize("device", cpu_and_cuda())
+    def test_get_key_frame_indices(self, device):
+        decoder = VideoDecoder(NASA_VIDEO.path, device=device, seek_mode="exact")
+        key_frame_indices = decoder._get_key_frame_indices()
+        size = key_frame_indices.size()
+        assert size[0] > 0
+        assert len(size) == 1
```
scotts, line 840: @NicolasHug, is there a more PyTorch-y way to assert this?

NicolasHug: I can't find of an obvious one. In this specific case, maybe we can just hard-code `assert key_frame_indices.shape == (42,)`, but of course that's only working for this video.

On another note, I wonder if we should somewhat check the correctness of the returned values (still non-blocking)?

src/torchcodec/decoders/_core/video_decoder_ops.py
```diff
@@ -255,6 +256,13 @@ def get_frames_by_pts_in_range_abstract(
     )
 
 
+@register_fake("torchcodec_ns::_get_key_frame_indices")
+def get_key_frame_indices_abstract(
+    decoder: torch.Tensor, *, stream_index: int
+) -> List[int]:
```
NicolasHug, line 262: Nit: the type annotation says list but I think you changed it to tensor?

test/decoders/test_video_decoder.py
```diff
@@ -831,6 +831,28 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+
+        # The key frame indices were generated from the following command:
+        #   $ ffprobe -v error -hide_banner -select_streams v:1 -show_frames -of csv test/resources/nasa_13013.mp4 | grep -n ",I," | cut -d ':' -f 1 > key_frames.txt
+        # What it's doing:
+        #   1. Calling ffprobe on the second video stream, which is absolute stream index 3.
+        #   2. Showing all frames for that stream.
+        #   3. Using grep to find and count the "I" frames, which are the key frames.
+        #   4. Using cut to extract just the count for the frame.
+        # Finally, because the above produces a count, which is index + 1, we subtract
+        # one from all values manually to arrive at the values below.
```
NicolasHug, line 847: I suppose you meant the same thing, but my understanding is that `-n` will output the line number (not the count), and the `cut` part will select that line number (which is 1-based)?

scotts: Yeah, I meant the same thing. I can try to clarify.

PR #485, opened 2025-01-28
Remove createFrom functions in C++

scotts:
Completes the refactoring started in #434.

The `createFrom()` functions don't add any values, because the [`std::make_unique`](https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) function already does a similar thing.


NicolasHug (approved):
Pre-emptively approving, as the only blocker probably is just the CI

PR #502, opened 2025-02-06
Avoid unnecessary calls to `avcodec_receive_frame` in decoding loop

NicolasHug:
This PR improves our main decoding loop which, in `main`, is calling `avcodec_receive_frame` tons and tons of time when it doesn't need to.

This PR saves `643` calls on the short `nasa_13013.mp4` (1034 in main, 391 now).

See comments below it's easier to explain with the code.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -951,9 +951,6 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
   outerLoopStart:
     ffmpegStatus =
         avcodec_receive_frame(streamInfo.codecContext.get(), avFrame.get());
-    if (ffmpegStatus == AVERROR(EAGAIN)) {
-      printf("Yes\n");
-    }
 
     if (ffmpegStatus != AVSUCCESS && ffmpegStatus != AVERROR(EAGAIN)) {
       // Non-retriable error
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -948,8 +948,12 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
   int ffmpegStatus = AVSUCCESS;
   bool reachedEOF = false;
   while (true) {
+  outerLoopStart:
     ffmpegStatus =
         avcodec_receive_frame(streamInfo.codecContext.get(), avFrame.get());
+    if (ffmpegStatus == AVERROR(EAGAIN)) {
+      printf("Yes\n");
+    }
 
     if (ffmpegStatus != AVSUCCESS && ffmpegStatus != AVERROR(EAGAIN)) {
       // Non-retriable error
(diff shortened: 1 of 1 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -979,36 +980,36 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
     // We still haven't found the frame we're looking for. So let's read more
     // packets and send them to the decoder.
     ReferenceAVPacket packet(autoAVPacket);
-    ffmpegStatus = av_read_frame(formatContext_.get(), packet.get());
-    decodeStats_.numPacketsRead++;
+    bool foundPacketForStream = false;
```
NicolasHug, line 983: What happens in the loop below is barely changed. I only updated a `continue` into a `goto`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -979,36 +980,36 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
...
-    }
-
-    if (ffmpegStatus < AVSUCCESS) {
-      throw std::runtime_error(
-          "Could not read frame from input file: " +
-          getFFMPEGErrorStringFromErrorCode(ffmpegStatus));
-    }
-
-    if (packet->stream_index != activeStreamIndex_) {
-      continue;
```
NicolasHug, line 1011: So, in main, we just called `av_read_frame` to receive a packet, and now that we realize this isn't a packet for the target stream, we `continue;` which brings us back to the beginning of the main loop... which starts by calling `avcodec_receive_frame()`.

But there is no point `avcodec_receive_frame()`: it's bound to return `EAGAIN`, because we haven't sent any new packet to the decoder!

So that's what this PR is doing: we don't `continue;` and go back to the loop. Instead, we just call `av_read_frame` again until we get a packet for the target stream.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -979,36 +980,36 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
...
+      foundPacketForStream = packet->stream_index == activeStreamIndex_;
```
scotts, line 1012: Can we just make this part of the while loop condition? That is:
```
while (packet->stream_index != activeStreamIndex_) {
  //
}
```
I find introducing a boolean harder to read than just stating the condition.

scotts: Actually, I think we can make the looping conditions more explicit and not use a goto if we do:
```
while (packet->stream_index != activeStreamIndex_ && !reachedEOF) {
      ffmpegStatus = av_read_frame(formatContext_.get(), packet.get());
      decodeStats_.numPacketsRead++;

      if (ffmpegStatus == AVERROR_EOF) {
        // End of file reached. We must drain the codec by sending a nullptr
        // packet.
        ffmpegStatus = avcodec_send_packet(
            streamInfo.codecContext.get(),
            /*avpkt=*/nullptr);
        if (ffmpegStatus < AVSUCCESS) {
          throw std::runtime_error(
              "Could not flush decoder: " +
        }
        reachedEOF = true;
      }
}

// We've reached the end of file so we can't read any more packets from
// it, but the decoder may still have frames to read in its buffer.
// Continue iterating to try reading frames.
if (reachedEOF) {
  continue;
}
```

scotts:
Great catch! We should definitely merge some version of this improved logic. Do my suggestions work? Are they more clear?

NicolasHug:
Thanks for the feedback @scotts 

Regarding using this condition:

```while (packet->stream_index != activeStreamIndex_) ```

I'm not sure we can do it safely, the packet just got allocated when we enter this loop, and I don't know if we can rely on FFmpeg to initialize `stream_index` to a correct value (like -2). So I changed it into a `do{} while()` loop. Not sure how we feel about this?

 > Actually, I think we can make the looping conditions more explicit and not use a goto if we do:


The suggestion kinda works but there is more we need to do: we still have to `break` out of the inner loop, and we end up with *two* of these blocks

```C
if (reachedEOF) {
  continue;
}
```

one before the inner loop, and one after.

I pushed the changed in [`f20d6f2` (#502)](https://github.com/pytorch/torchcodec/pull/502/commits/f20d6f20b97264ce4cd02e461c31277b3d508cd4), LMK what you think?

scotts (approved):

scotts:
@NicolasHug, I think it looks great! It's a little surprising that we have the two EOF checks, but I understand why.

NicolasHug:
C++ test failure is unrelated, merging.

PR #503, opened 2025-02-07
Avoid receiving undesired packets

NicolasHug:
This avoids receiving undesired packets when calling `av_read_frame()`. This saves 643 useless packets from being received on our nasa video. 

I was hoping we could remove/simplify a bit our `do/while` 

https://github.com/pytorch/torchcodec/blob/b87f00d587843f02941fc12779e72a6fef094e19/src/torchcodec/decoders/_core/VideoDecoder.cpp#L1007

and just do 

```
TORCH_CHECK(packet->stream_index == activeStreamIndex_);'
```

but while most tests are passing, some still fail. I suspect FFmpeg just doesn't fully respect the DISCARD flag under some specific conditions (that I haven't investigated).

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -222,6 +222,13 @@ void VideoDecoder::scanFileAndUpdateMetadataAndIndex() {
     return;
   }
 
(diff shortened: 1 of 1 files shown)


scotts (approved):

scotts:
Great catch again!

PR #506, opened 2025-02-09
Remove streamIndex  parameter and multi-stream related logic

NicolasHug:
Towards https://github.com/pytorch/torchcodec/issues/476. Our VideoDecoder is inherently single-stream and almost all `streamIndex` parameters aren't used, so this PR removes them. This also removes most but not all multi-stream related logic (there is some left in the scan for example).

I haven't updates the custom ops APIs nor the core APIs because this may create conflicts internally and I want to deal with that separately as a follow-up PR.

Other clean-ups we could do afterwards:

- Update Python core APIs and ops APIs to remove the stream_index parameter.
- Remove the AVFrameStream struct and just use AVFrame. We know the stream, we don't need to keep track of it.
- Remove the `StreamInfos_` vec and only keep a single `StreamInfo` for the active stream.
- Simplify the scan to ignore inactive streams (at the very least we don't need to set the StreamInfo_ for inactive streams).

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -349,7 +349,7 @@ VideoDecoder::ContainerMetadata VideoDecoder::getContainerMetadata() const {
 }
 
 torch::Tensor VideoDecoder::getKeyFrameIndices(int streamIndex) {
-  validateUserProvidedStreamIndex(streamIndex);
+  validateActiveStream();
   validateScannedAllStreams("getKeyFrameIndices");
 
   const std::vector<FrameInfo>& keyFrames = streamInfos_[streamIndex].keyFrames;
@@ -565,7 +565,7 @@ VideoDecoder::FrameOutput VideoDecoder::getFrameAtIndexInternal(
     int streamIndex,
     int64_t frameIndex,
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
-  validateUserProvidedStreamIndex(streamIndex);
+  validateActiveStream();
 
   const auto& streamInfo = streamInfos_[streamIndex];
   const auto& streamMetadata =
@@ -580,7 +580,7 @@ VideoDecoder::FrameOutput VideoDecoder::getFrameAtIndexInternal(
 VideoDecoder::FrameBatchOutput VideoDecoder::getFramesAtIndices(
     int streamIndex,
     const std::vector<int64_t>& frameIndices) {
-  validateUserProvidedStreamIndex(streamIndex);
+  validateActiveStream();
 
   auto indicesAreSorted =
       std::is_sorted(frameIndices.begin(), frameIndices.end());
@@ -643,7 +643,7 @@ VideoDecoder::FrameBatchOutput VideoDecoder::getFramesInRange(
     int64_t start,
     int64_t stop,
     int64_t step) {
-  validateUserProvidedStreamIndex(streamIndex);
+  validateActiveStream();
 
   const auto& streamMetadata =
       containerMetadata_.allStreamMetadata[streamIndex];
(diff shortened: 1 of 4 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -523,28 +527,17 @@ void VideoDecoder::addVideoStreamDecoder(
       videoStreamOptions.colorConversionLibrary.value_or(defaultLibrary);
 }
 
-void VideoDecoder::updateMetadataWithCodecContext(
```
NicolasHug, line 526: I inlined this above, as it became hard to justify this being a single separate method with only one call-site.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -523,28 +527,17 @@ void VideoDecoder::addVideoStreamDecoder(
...
-}
-
 // --------------------------------------------------------------------------
 // HIGH-LEVEL DECODING ENTRY-POINTS
 // --------------------------------------------------------------------------
 
-VideoDecoder::FrameOutput VideoDecoder::getNextFrameNoDemux() {
-  auto output = getNextFrameNoDemuxInternal();
-  output.data = maybePermuteHWC2CHW(output.streamIndex, output.data);
+VideoDecoder::FrameOutput VideoDecoder::getNextFrame() {
```
NicolasHug, line 534: Removed all the `NoDemux` suffixes - they weren't informative

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -665,29 +650,24 @@ VideoDecoder::FrameBatchOutput VideoDecoder::getFramesInRange(
...
-        streamInfo.currentPts + streamInfo.currentDuration,
-        streamInfo.timeBase);
-    if (seconds >= frameStartTime && seconds < frameEndTime) {
-      // We are in the same frame as the one we just returned. However, since we
-      // don't cache it locally, we have to rewind back.
-      seconds = frameStartTime;
-      break;
-    }
+VideoDecoder::FrameOutput VideoDecoder::getFramePlayedAt(double seconds) {
+  StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
```
NicolasHug, line 662: removed the `for` loop over StreamInfos_ and just indented to the left, using `activeStreamIndex_`. This is probably a minor fix.

src/torchcodec/decoders/_core/VideoDecoderOps.cpp
```diff
@@ -247,60 +247,59 @@ OpsFrameOutput get_next_frame(at::Tensor& decoder) {
...
   auto videoDecoder = unwrapTensorToGetDecoder(decoder);
-  auto result = videoDecoder->getFramePlayedAtNoDemux(seconds);
+  auto result = videoDecoder->getFramePlayedAt(seconds);
   return makeOpsFrameOutput(result);
 }
 
 OpsFrameOutput get_frame_at_index(
     at::Tensor& decoder,
-    int64_t stream_index,
+    [[maybe_unused]] int64_t stream_index,
```
NicolasHug, line 256: As mentioned in the PR description, I chose not to update the APIs of the ops and core APIs. Will deal with that in a follow-up.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1507,9 +1471,8 @@ void VideoDecoder::createSwsContext(
 // PTS <-> INDEX CONVERSIONS
 // --------------------------------------------------------------------------
 
-int VideoDecoder::getKeyFrameIndexForPts(
-    const StreamInfo& streamInfo,
-    int64_t pts) const {
+int VideoDecoder::getKeyFrameIndexForPts(int64_t pts) const {
+  const StreamInfo& streamInfo = streamInfos_.at(activeStreamIndex_);
```
NicolasHug, line 1475: Using `.at()` instead of `[]` to preserve the `const` qualifier of the method. I would otherwise get:


````
  /home/nicolashug/dev/torchcodec/src/torchcodec/decoders/_core/VideoDecoder.cpp:1475:65: error: passing ‘const std::map<int, facebook::torchcodec::VideoDecoder::StreamInfo>’ as ‘this’ argument discards qualifiers [-fpermissive]
   1475 |   const StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
        |                                                                 ^
```

@scotts is there a more idiomatic way of dealing with this?

scotts (approved):

PR #507, opened 2025-02-10
Add audio decoding support to all core APIs

NicolasHug:
This PR is not ready to be merged yet, but it's getting complex so we should start discussing it.

-----

This PR implements audio stream decoding for the C++ APIs and the core ops (I'll just refer to those as "core APIs" below).

It doesn't address any public API. The design of the public APIs is still very much open and *not* in scope for this PR.

### Design decisions

- Keep the core APIs close to the FFmpeg structures: FFmpeg handles / returns audio *frames*, so that's what the core do as well.
- The core APIs are unchanged: user can call `get_frame_at()`, `get_frames_played_in_range()`, etc., just like for video.
- When decoding audio, users get **frames**, just like in videos. For audio, a frame is a sequence of samples of shape (num_channels, num_samples).
- For batch APIs, we return **stacked** frames: the output tensor is of shape (num_frames, num_channels, num_samples). This drastically simplifies the C++ code, because it means the number of frames is the first dimension of the batch, and allows us to use the exact same code for our high-level batch-decoding APIs. It also significantly simplifies the test. Note that these 3D stacked frames are probably *not* what we'll want to return to users in our public APIs: users will want 2D contiguous audio samples of shape (num_channels, num_samples * num_frames). In any case, this isn't in scope for this PR, and an important point is that we can convert a frame from concatenated-2D to stacked-3D with O(1) no-copy operations.

Some design principles that aren't necessarily reflected in the PR yet, but that I think we should follow, and which should guide the decisions we make for the other points below:

- We should optimize for the use-case that is (roughly): `get_frames_played_in_range(start, stop)` where `start` and `stop` are user-defined. That is, decode contiguous audio samples from start to stop. This should be, by far, the most common decoding pattern. I'm saying *roughly* because I don't think the public API should expose the concept of frame, but that's fine for core APIs.
- We should optimize for cases where a scan is *not* made. I suspect we'll want *not* to scan by default for audio, because a scan is costly and can amount to 10+% of the total decoding time on a small audio file. And because most calls will look like `get_frames_played_in_range(start, stop)`, a scan is likely overkill anyway. Note that I'm talking about scan and not about `seek_mode`, because I'm not sure the concept of `seek_mode` applies to audio the same way it does for video.

### Wrinkles and blockers

This PR is currently facing the following limitations:

- We only support `fltp` audio formats (and return it as float). We'll want to support all existing audio formats. At this point, we'll need to decide what to return: always float32 tensors and handle the conversion internally, always native, or allow the user to choose. TorchAudio lets the user choose. I think that for a first iteration, we just need to support all formats and always return float32, because this is what the transforms expect. In any case: this is probably a bit of work, but I don't see anything blocking - it should be reasonably self-contained.
- We currently enforce that the `frame_size` field of the codec exists in order to run the approximate mode. This is a strong limitation because we shouldn't force the user to scan. See: TODO_FRAME_SIZE_APPROXIMATE_MODE
- We currently enforce that the `frame_size` field of the codec exists so that we can pre-allocate batch tensors in batch APIs. See TODO_FRAME_SIZE_BATCH_TENSOR_ALLOCATION. I think addressing this may might require some heavy changes, and because frames may have a non-constant number of samples in them, it may mean we can't return stacked frames anymore.
- See the note named `Note [Seek offset for audio]` and prepare for a headache.


src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -361,7 +365,8 @@ class VideoDecoder {
   // DECODING APIS AND RELATED UTILS
   // --------------------------------------------------------------------------
 
-  bool canWeAvoidSeeking(int64_t targetPts) const;
+  bool canWeAvoidSeekingVideo(int64_t targetPts) const;
+  bool canWeAvoidSeekingAudio(double desiredPtsSeconds) const;
```
scotts, line 369: Curious that we have to have different logic for both - and that they don't take the same kind of value. (I'm reading the .h first to get a survey of what changed; this is my note to dig into this deeper when I see the implementation.)

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -424,6 +437,11 @@ class VideoDecoder {
   // STREAM AND METADATA APIS
   // --------------------------------------------------------------------------
 
+  void addStream(
+      int streamIndex,
+      AVMediaType mediaType,
+      const VideoStreamOptions& videoStreamOptions = VideoStreamOptions());
+
```
scotts, line 444: We should probably make `addStream()` a private member function, and force callers to use the named ones. (Apologies if that is already the case; I have difficulty tracking where the public and private modifiers are in the PR.)

scotts:
Some questions (I'm going to edit this as I read):

* Is `fltp` the most common type of audio format? Do we have an understanding of how often folks will have to convert?
* Do we have a feel for how common it is for `frame_size` to be missing from the codec metadata?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -817,6 +869,30 @@ void VideoDecoder::setCursorPtsInSeconds(double seconds) {
   desiredPtsSeconds_ = seconds;
 }
 
+bool VideoDecoder::canWeAvoidSeekingAudio(double desiredPtsSeconds) const {
+  const StreamInfo& streamInfo = streamInfos_.at(activeStreamIndex_);
+  int64_t targetPts = *desiredPtsSeconds_ * streamInfo.timeBase.den;
```
scotts, line 874: Nit: we should probably use `secondsToClosestPts()` unless there's an explicit reason we don't want to use `std::round()`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -877,7 +953,17 @@ void VideoDecoder::maybeSeekToBeforeDesiredPts() {
   streamInfo.discardFramesBeforePts = desiredPts;
 
   decodeStats_.numSeeksAttempted++;
-  if (canWeAvoidSeeking(desiredPts)) {
+
+  // TODO_CODE_QUALITY The different signature is unfortunate
```
scotts, line 957: Why is the signature different? Don't we do the same computation in `canWeAvoidSeekingAudio()`?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -887,13 +973,85 @@ void VideoDecoder::maybeSeekToBeforeDesiredPts() {
...
   // See https://trac.ffmpeg.org/ticket/11137 for the underlying ffmpeg bug.
-  if (!streamInfo.keyFrames.empty()) {
+  if (avMediaType == AVMEDIA_TYPE_VIDEO && !streamInfo.keyFrames.empty()) {
     int desiredKeyFrameIndex = getKeyFrameIndexForPtsUsingScannedIndex(
         streamInfo.keyFrames, desiredPts);
     desiredKeyFrameIndex = std::max(desiredKeyFrameIndex, 0);
     desiredPts = streamInfo.keyFrames[desiredKeyFrameIndex].pts;
   }
 
+  if (avMediaType == AVMEDIA_TYPE_AUDIO) {
```
scotts, line 983: Let's do an `else if` here and throw an exception on `else` with unsupported media type.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -887,13 +973,85 @@ void VideoDecoder::maybeSeekToBeforeDesiredPts() {
...
+    // - This weird FFmpeg behavior is observable not just in Torchcodec, it
+    //   really seems to be an FFmpeg thing. Other decoders have the same
+    //   problem, like the ones in TorchVision. Those who do not exhibit this
+    //   behavior are solving it in inefficient ways: Decord effectively decodes
+    //   and caches the *entire* file when it is created, thus resolving the
+    //   data dependency. Similarly, TorchAudio effectively always decodes all
+    //   frames up to frame `i`, even after seeking to frame `i`, because it
+    //   sets the 'backwards' flag when it calls `av_seek_frame`: it actually
+    //   always seeks back to the beginning.
+  }
```
scotts, line 1053: Excellent analysis.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1047,13 +1205,14 @@ VideoDecoder::FrameOutput VideoDecoder::convertAVFrameToFrameOutput(
...
   frameOutput.ptsSeconds = ptsToSeconds(
       avFrame->pts, formatContext_->streams[streamIndex]->time_base);
   frameOutput.durationSeconds = ptsToSeconds(
       getDuration(avFrame), formatContext_->streams[streamIndex]->time_base);
-  // TODO: we should fold preAllocatedOutputTensor into AVFrameStream.
-  if (streamInfo.videoStreamOptions.device.type() == torch::kCPU) {
+  if (streamInfo.avMediaType == AVMEDIA_TYPE_AUDIO) {
+    convertAudioAVFrameToFrameOutputOnCPU(
+        avFrameStream, frameOutput, preAllocatedOutputTensor);
+  } else if (streamInfo.videoStreamOptions.device.type() == torch::kCPU) {
```
scotts, line 1215: We might want to still do an explicit check for `AVMEDIA_TYPE_VIDEO` here. The `device` field in `VideoStreamOptions` is not an optional, and it defaults to `kCPU`.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -139,9 +139,7 @@ class VideoDecoder {
   void addVideoStream(
       int streamIndex,
       const VideoStreamOptions& videoStreamOptions = VideoStreamOptions());
-  void addAudioStreamDecoder(
-      int streamIndex,
-      const AudioStreamOptions& audioStreamOptions = AudioStreamOptions());
+  void addAudioStream(int streamIndex);
```
scotts, line 142: I only just now realized that we've had `addAudioStreamDecoder()` declared in the .h all this time, with no implementation. 🤯

src/torchcodec/decoders/_core/VideoDecoder.h
scotts: Revisiting this comment after reading the implementation - I think it may be possible to get them to the same signature, and hide them behind a general dispatch.

test/decoders/test_video_decoder_ops.py
```diff
@@ -209,19 +333,24 @@ def test_get_frames_by_pts(self, device):
...
+            pytest.skip(reason="CUDA decoding not supported for audio")
+        decoder = create_from_file(str(test_ref.path))
+        _add_stream(decoder=decoder, test_ref=test_ref, device=device)
 
         metadata = get_json_metadata(decoder)
         metadata_dict = json.loads(metadata)
         num_frames = metadata_dict["numFrames"]
         assert num_frames == 390
+        if test_ref is NASA_AUDIO:
+            num_frames = 204
```
scotts, line 353: I'm confused here - does the assertion on line 351 pass for audio? If yes, why do we need to do a fixup after? I think this does call out that we'll need to reconcile our current metadata with audio.

test/decoders/test_video_decoder_ops.py
```diff
@@ -264,63 +393,81 @@ def test_pts_apis_against_index_ref(self, device):
...
         assert_frames_equal(bulk_frame0, ref_frame0.to(device))
 
-        ref_frame1 = NASA_VIDEO.get_frame_data_by_range(1, 2)
+        ref_frame1 = test_ref.get_frame_data_by_range(1, 2)
         bulk_frame1, *_ = get_frames_in_range(decoder, start=1, stop=2)
         assert_frames_equal(bulk_frame1, ref_frame1.to(device))
 
-        ref_frame389 = NASA_VIDEO.get_frame_data_by_range(389, 390)
-        bulk_frame389, *_ = get_frames_in_range(decoder, start=389, stop=390)
+        last_index = 389 if test_ref is NASA_VIDEO else 203  # TODO ew
```
scotts, line 414: Yeah, I think we should incorporate some concept of number of frames into the `TestContainerFile` util class. That should allow us to clean this and some other hard-coded things up.

test/decoders/test_video_decoder_ops.py
```diff
@@ -426,11 +573,12 @@ def test_video_get_json_metadata_with_stream(self):
         assert metadata_dict["minPtsSecondsFromScan"] == 0
         assert metadata_dict["maxPtsSecondsFromScan"] == 13.013
 
+    # TODO: Not sure whether this test still makes a lot of sense
```
scotts, line 576: Probably not - certainly don't do any work to keep it around. As I pointed out above, we'll need a re-think on metadata in general to make everything fit together. I _think_ we left space for audio in what we have, but we'll find out soon enough.

scotts:
I completed a whole pass - this looks great! I made a bunch of code comments, but nothing on the design or approach.

What is remaining to get this PR into a mergable state? I know that we want to support all audio formats on our first release with audio, but I think there's value in merging just fltp first, and then iteratively adding more. Keeping this PR in sync with main will become a significant burden.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: I think we could pass `desiredPtsSeconds_` to `canWeAvoidSeekingVideo`, that would lead to some minor redundancy but probably still cleaner

test/decoders/test_video_decoder_ops.py
NicolasHug: The assertion passes, yes. This PR doesn't address the audio metadata so all we get is wrong / related to videos.

We need to do this fixup because the audio stream only contains 204 frames, not 390. We would be trying to decode frames beyond 204 in the test, leading to a failure.

test/decoders/test_video_decoder_ops.py
NicolasHug: Yes, we should try to refactor a few things for these tests. I'm leaving that kind of stuff out of scope for the initial PR, so that we can focus on the high-level design / features first.

test/utils.py
```diff
@@ -23,12 +23,25 @@ def cpu_and_cuda():
     return ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
 
 
+def assert_frames_equal(*args, **kwargs):
```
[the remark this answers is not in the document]
NicolasHug: That would require changing all existing calls to `assert_frames_equal()`, and there are *a lot* of those.

There are alternatives, like automatically invoking a context manager depending on the test parametrization (when using NASA_VIDEO vs NASA_AUDIO), and all sorts of crazy things.

There are also more robust heuristics like checking the frame's number of dimensions and shape. We'll address this eventually.

NicolasHug:
I'll close this PR and freeze its state so that we can refer to it in the future - particularly the `Wrinkles and blockers` section and the `Note [Seek offset for audio]`. These will be pre-read requirements that inform the design I'll implement in the next few PRs.

jpc:
Hey, I was looking into the seeking performance regression for TorchCodec compared to the StreamDecoder in torchaudio. I cannot find `Note [Seek offset for audio]` anywhere? Is this an internal Meta document?

I probably went though a similar investigation myself a few months ago and ended up finding what I suspect is a bug in ffmpeg's `mp3` support. If yo uare loading from the start of the file, ffmpeg will apply an mp3 encoder library dependent offset to the sample index and it loads the correct audio. But when you are seeking anywhere by `start == 0` it does not do that. It seems this can be compensated for quite easily and afterwards I started getting correct sample buffers when seeking into the middle of even very long audio files. (+-1 sample)

The problem is partially explained here: http://mp3decoders.mp3-tech.org/decoders_lame.html (look for the "Encoder Delays" note – this shift is also hardcoded in the ffmpeg source code but I don't have the link)
```
reader = torchaudio.io.StreamReader(src=to_filelike(self.src))
info = reader.get_src_stream_info(0)

# mp3 has encoder delays that are not handled well when seeking (http://mp3decoders.mp3-tech.org/decoders_lame.html)
if info.codec == 'mp3':
    self.skip_samples = 1105
    
...

reader, sample_rate = self.get_reader(sample_rate)
# rought seek
reader.seek(max(0, start - 1), 'key')
reader.fill_buffer()
chunk, = reader.pop_chunks()
# tight crop (seems accurate down to 1 sample in my tests)
prefix = int((start - chunk.pts) * sample_rate) + self.skip_samples if start > 0 else 0
assert prefix >= 0
samples = chunk[prefix:prefix + int((end - start) * sample_rate)].mT
# clear out any remaining data
while chunk is not None:
    chunk, = reader.pop_chunks()
```

I could try to get a PR for something like this into torchcodec if you'd be willing to merge it. We could also push this onto the user but have a flag that enables seeking?

The current performance makes torchcodec unfortunately unusable in many of our use cases (1 - 10s chunk loading time instead of .1s).

NicolasHug:
Hi @jpc 

Thanks for reaching out.

The main note explaining the current situation is: https://github.com/meta-pytorch/torchcodec/blob/dc872282ed7e019422214a1e53681a87d7b37f05/src/torchcodec/_core/SingleStreamDecoder.cpp#L945-L966 

and the `Note [Seek offset for audio]` is [here](https://github.com/meta-pytorch/torchcodec/pull/507/files#diff-c7e9bb49689c1d47b393890217f3c999483406b81cf4298278451305d19034a8R985) in `VideoDecoder.cpp` which, I think, GitHub is hiding by default so you have to expand it.

----

These notes explain my understanding of the issue at different points in time. It is very likely that I am partially (or completely) wrong. When I was implementing the `AudioDecoder` I was indeed working mostly with mp3 files when testing, so it does seem very related to the offset samples that you're mentioning. 

I'd love to consider ways to improve the seeking behavior, thanks for offering support. If really the observed issue is only relevant for mp3 files, the first thing we could do is to enable proper seeking for all other formats.

> The current performance makes torchcodec unfortunately unusable in many of our use cases (1 - 10s chunk loading time instead of .1s).

Can you share more about this? What seeking patterns are you currently using?

jpc:
Hey, sorry for not replying sooner but I was swamped with work so I had stuck to workarounds and leave the proper solution for later.

We open-sourced the code we are using (it’s still pretty rough) here https://github.com/HumeAI/wsds

The main idea is to store audio data in original format. For processing we extract whole files and split them up into shorter chunks for processing. After we have processed we end up with a list of timestamp ranges and related transcriptions, classifier outputs and such. This is for audio but I think a similar approach should work for video.

When we want to use the results it helps a lot to be able to extract these segments at random and fetch the corresponding audio fragment. Seeking is crucial to make this step perform reasonably well since the source audio files tend to be many hours long.

I gathered a large set of files in different formats and encoders. Maybe it would let us test the seeking on enough audio in the wild to be confident it works!

You are not gonna be at NeurIPS by any chance?

PR #521, opened 2025-02-26
Drop 12.4 support

NicolasHug:
See https://github.com/pytorch/pytorch/issues/145544 and https://fburl.com/workplace/sqyr6fci (internal link).

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,14 +146,12 @@ First, make sure you have a GPU that has NVDEC hardware that can decode the
 format you want. Refer to Nvidia's GPU support matrix for more details
 [here](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new).
 
-1. Install CUDA Toolkit. Pytorch and TorchCodec supports CUDA Toolkit
-   versions 11.8, 12.1 or 12.4. In particular TorchCodec depends on
-   CUDA libraries libnpp and libnvrtc (which are part of CUDA Toolkit).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -146,14 +146,12 @@ First, make sure you have a GPU that has NVDEC hardware that can decode the
 format you want. Refer to Nvidia's GPU support matrix for more details
 [here](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new).
 
-1. Install CUDA Toolkit. Pytorch and TorchCodec supports CUDA Toolkit
-   versions 11.8, 12.1 or 12.4. In particular TorchCodec depends on
-   CUDA libraries libnpp and libnvrtc (which are part of CUDA Toolkit).
+1. Install Pytorch corresponding to your CUDA Toolkit using the
+   [official instructions](https://pytorch.org/get-started/locally/). You'll
+   need the `libnpp` and `libnvrtc` CUDA libraries, which are usually part of
+   the CUDA Toolkit.
```
NicolasHug, line 152: I removed the previously-first entry, because I don't think we should be in the business of telling people how to install the CUDA toolkit. We were just saying "you should install it" without providing any guidance, so it wasn't that useful anyway. 

And I fear it may raise the barrier. In the majority of cases, users who have a working pytorch installation with cuda won't need to do anything special, because all the necessary libraries are already installed.

scotts (approved):

PR #524, opened 2025-02-26
Audio encoding - part 1 of N

NicolasHug:
Expect N to be large :)

This PR implements a basic and not feature complete audio encoder, which seems to work OK in some limited scenarios.  There are *a lot* of TODOs left in the code.

This is only C++ and core APIs, nothing is public yet. The current design is to pass all the necessary parameters to the constructor. Namely, with the core API:

```py
encoder = create_encoder(wf=samples, sample_rate=sample_rate, filename=output.mp3)
encode(encoder)
```

The reason is: all these parameters are required in order to initialize the `AVFormatContext` and the `AVCodecContext`. This may very well change, i.e. eventually we may decide that we don't even want to expose this as a C++ object but rather as a pure function? It'll be easier to decide later once we're more feature complete.


src/torchcodec/_core/CMakeLists.txt
```diff
@@ -61,6 +61,9 @@ function(make_torchcodec_libraries
         AVIOContextHolder.cpp
         FFMPEGCommon.cpp
         SingleStreamDecoder.cpp
+        # TODO: lib name should probably not be "*_decoder*" now that it also
+        # contains an encoder
+        Encoder.cpp
```
scotts, line 66: Maybe it should be `libtorchcodec_coreN.so`?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
+#include "src/torchcodec/_core/Encoder.h"
+#include "torch/types.h"
+
+extern "C" {
+#include <libavcodec/avcodec.h>
```
scotts, line 5: These are already included in `FFMPEGCommon.h`, so we shouldn't need to include them here again. (This inspired #615.)

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -381,6 +384,38 @@ OpsAudioFramesOutput get_frames_by_pts_in_range_audio(
...
+  std::unique_ptr<Encoder> uniqueEncoder =
+      std::make_unique<Encoder>(wf, static_cast<int>(sample_rate), file_name);
+  return wrapEncoderPointerToTensor(std::move(uniqueEncoder));
+}
+
+void encode(at::Tensor& encoder) {
+  auto encoder_ = unwrapTensorToGetEncoder(encoder);
+  encoder_->encode();
+}
+
```
scotts, line 418: Let's make these names `create_audio_encoder()` and `encode_audio()` here and in the Python layer as well. Same as the underlying C++ class, if we can make things generic for video later, we can consolidate. But I anticipate we'll probably need different entry points due to different kind of input parameters.

src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -50,9 +50,12 @@ struct Deleter {
 };
 
 // Unique pointers for FFMPEG structures.
-using UniqueAVFormatContext = std::unique_ptr<
+using UniqueAVFormatContextForDecoding = std::unique_ptr<
     AVFormatContext,
     Deleterp<AVFormatContext, void, avformat_close_input>>;
+using UniqueAVFormatContextForEncoding = std::unique_ptr<
+    AVFormatContext,
+    Deleter<AVFormatContext, void, avformat_free_context>>;
```
scotts, line 58: Let's omit the preposition and call them `UniqueDecodingAVFormatContext` and `UniqueEncodingAVFormatContext`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+      wf_.dtype());
+  TORCH_CHECK(
+      wf_.dim() == 2, "waveform must have 2 dimensions, got ", wf_.dim());
+  AVFormatContext* avFormatContext = nullptr;
+  avformat_alloc_output_context2(
+      &avFormatContext, nullptr, nullptr, fileName.data());
+  TORCH_CHECK(
+      avFormatContext != nullptr,
+      "Couldn't allocate AVFormatContext. ",
+      "Check the desired extension?");
```
scotts, line 32: We should check the return value of `avformat_alloc_output_context2()`; that should enable us to return a more useful error message. See https://ffmpeg.org/doxygen/3.3/avformat_8h.html#a6ddf3d982feb45fa5081420ee911f5d5.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+      "frame_size is ",
+      avCodecContext_->frame_size,
+      ". Cannot encode. This should probably never happen?");
+
+  // We're allocating the stream here. Streams are meant to be freed by
+  // avformat_free_context(avFormatContext), which we call in the
+  // avFormatContext_'s destructor.
+  avStream_ = avformat_new_stream(avFormatContext_.get(), nullptr);
+  TORCH_CHECK(avStream_ != nullptr, "Couldn't create new stream.");
+  avcodec_parameters_from_context(avStream_->codecpar, avCodecContext_.get());
```
scotts, line 101: We should check the return value of `avcodec_parameters_from_context()`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+  TORCH_CHECK(
+      avCodecContext_->frame_size > 0,
+      "frame_size is ",
+      avCodecContext_->frame_size,
+      ". Cannot encode. This should probably never happen?");
+
+  // We're allocating the stream here. Streams are meant to be freed by
+  // avformat_free_context(avFormatContext), which we call in the
+  // avFormatContext_'s destructor.
+  avStream_ = avformat_new_stream(avFormatContext_.get(), nullptr);
```
scotts, line 99: I want to make sure I understand the relationship here: `avformat_new_stream()` takes a pointer to an `AVFormatContext`. It creates a new stream, associates that stream with the provided `AVFormatContext` such that the `AVFormatContext` owns the stream, and returns a pointer to that newly created stream?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+      "Couldn't allocate avFrame's buffers: ",
+      getFFMPEGErrorStringFromErrorCode(status));
+
+  AutoAVPacket autoAVPacket;
+
+  uint8_t* pwf = static_cast<uint8_t*>(wf_.data_ptr());
+  auto numSamples = wf_.sizes()[1]; // per channel
+  auto numEncodedSamples = 0; // per channel
+  auto numSamplesPerFrame =
+      static_cast<long>(avCodecContext_->frame_size); // per channel
```
scotts, line 125: The docs have `frame_size` as an `int`, not a `long`: https://ffmpeg.org/doxygen/6.0/structAVCodecContext.html#aec57f0d859a6df8b479cd93ca3a44a33. Is that perhaps version specific, or picked up from some examples?

In general, I prefer to avoid introducing `long` unless I want to match against an exist C API. Otherwise, I think it's better to prefer the explicitly sized integer types such as `int64_t` if we have reason to believe that an `int` is not big enough.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+  auto numBytesPerSample = wf_.element_size();
+  auto numBytesPerChannel = numSamples * numBytesPerSample;
```
scotts, line 127: I think `numBytesPerSample` is actually becoming a `size_t`, which is defined as an unsigned integer that "can store the maximum size of a theoretically possible object of any type (including array)." See: https://en.cppreference.com/w/cpp/types/size_t. In general, `size_t` is awkward to work with, despite it being the default size type for all of the standard containers. (Stroustrup has even called that a mistake! See: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1428r0.pdf.)

Is that perhaps why you ended up casting `numSamplesPerFrame` to a larger size? Also note that `numEncodedSamplers` is going to be an `int`, as that's the default integer type for a numeric literal.

Because all of this is confusing, and integer conversion errors painful to deal with (see our recent problems with overflow with the bytes context), I think it's best to not use `auto` for integer values and be explicit.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+        "Couldn't make AVFrame writable: ",
+        getFFMPEGErrorStringFromErrorCode(status));
+
+    auto numSamplesToEncode = std::min(
+        numSamplesPerFrame, static_cast<long>(numSamples - numEncodedSamples));
+    auto numBytesToEncode = numSamplesToEncode * numBytesPerSample;
+
+    for (int ch = 0; ch < wf_.sizes()[0]; ch++) {
+      memcpy(
+          avFrame->data[ch], pwf + ch * numBytesPerChannel, numBytesToEncode);
```
scotts, line 148: Because you're saying `memcpy()` not `std::memcpy()`, technically, you're calling the C version of `memcpy()`: https://en.cppreference.com/w/c/string/byte/memcpy, rather than the C++ version in the standard library: https://en.cppreference.com/w/cpp/string/byte/memcpy.

src/torchcodec/_core/Encoder.h
```diff
@@ -0,0 +1,32 @@
...
+namespace facebook::torchcodec {
+class Encoder {
+ public:
+  ~Encoder();
+
+  Encoder(const torch::Tensor wf, int sampleRate, std::string_view fileName);
+  void encode();
+
+ private:
+  void encode_inner_loop(
```
scotts, line 14: Nit: camelCase instead of snake_case.

For the record, I prefer snake case, but I think being consistent is more important. If I could wave a magic wand and make our repo all snake case for variable and function names, I would. Class names should still be pascal case, I believe.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // wouldn't match the ffmpeg CLI.
+    avFrame->nb_samples = numSamplesToEncode;
+    encode_inner_loop(autoAVPacket, avFrame);
+
+    avFrame->pts += numSamplesToEncode;
+    numEncodedSamples += numSamplesToEncode;
+  }
+  TORCH_CHECK(numEncodedSamples == numSamples, "Hmmmmmm something went wrong.");
+
+  encode_inner_loop(autoAVPacket, UniqueAVFrame(nullptr)); // flush
```
scotts, line 165: Maybe define this as a helper function to make it more clear what's going on?

src/torchcodec/_core/Encoder.h
```diff
@@ -0,0 +1,32 @@
...
+      AutoAVPacket& autoAVPacket,
+      const UniqueAVFrame& avFrame);
+
+  UniqueAVFormatContextForEncoding avFormatContext_;
+  UniqueAVCodecContext avCodecContext_;
+  AVStream* avStream_;
```
scotts, line 20: We store the stream pointer, but we don't actually use the stream itself anymore, we only reference its index and timebase. It may be clearer what's going on if we just store those.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -381,6 +384,38 @@ OpsAudioFramesOutput get_frames_by_pts_in_range_audio(
   return makeOpsAudioFramesOutput(result);
 }
 
+at::Tensor wrapEncoderPointerToTensor(std::unique_ptr<Encoder> uniqueEncoder) {
+  Encoder* encoder = uniqueEncoder.release();
+
+  auto deleter = [encoder](void*) { delete encoder; };
+  at::Tensor tensor =
+      at::from_blob(encoder, {sizeof(Encoder)}, deleter, {at::kLong});
```
scotts, line 392: It should be `sizeof(Encoder*)` since we want space just to store the pointer, not the entire object.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -381,6 +384,38 @@ OpsAudioFramesOutput get_frames_by_pts_in_range_audio(
...
+  Encoder* encoder = static_cast<Encoder*>(buffer);
+  return encoder;
+}
+
+at::Tensor create_encoder(
+    const at::Tensor wf,
+    int64_t sample_rate,
+    std::string_view file_name) {
+  std::unique_ptr<Encoder> uniqueEncoder =
+      std::make_unique<Encoder>(wf, static_cast<int>(sample_rate), file_name);
```
scotts, line 410: We have to take in `sample_rate` as an `int64_t` because it's coming from Python and all Python ints become `int64_t`. And we have to turn it into an `int` because that's the type it is in the `AVFormatContext`, so that's all correct. But to be extra paranoid, we can do a `TORCH_CHECK(sample_rate <= std::numeric_limits<int>::max())`.

src/torchcodec/_core/custom_ops.cpp
NicolasHug: OK for `encode_audio()` op to be `encode_audio()`, but unless you feel strongly about I'll leave the `encode()` C++ method as `encode()`, since its class will be changed to `AudioEncoder`.

src/torchcodec/_core/Encoder.h
```diff
@@ -0,0 +1,32 @@
+#pragma once
+#include <torch/types.h>
+#include "src/torchcodec/_core/FFMPEGCommon.h"
+
+namespace facebook::torchcodec {
+class Encoder {
+ public:
```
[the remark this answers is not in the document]
NicolasHug: Sounds good. I'll keep the file name as `Encoder.cpp` because I  suspect we'll still implement both in the same files, and it will avoid potential internal migration headaches.

src/torchcodec/_core/Encoder.h
NicolasHug: I am more used to camel case too. If we really want to, surely there must exist some camelCase to smake_case auto converters?

src/torchcodec/_core/Encoder.cpp
NicolasHug: You're correct, or to be more precise: this is also what I understand from the [FFmpeg docs](https://ffmpeg.org/doxygen/5.1/group__lavf__core.html#gadcb0fd3e507d9b58fe78f61f8ad39827).

src/torchcodec/_core/Encoder.h
NicolasHug: I like to have the avStream_ field because it makes it obvious where `timebase` and `index` come from, without having to bubble up all the way to the constructor code. I'll switch it for `timeBase_` and `streamIndex_` fields for now, but we might revisit if we end up having to rely on more attributes from the AVStream?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+      "Couldn't allocate avFrame's buffers: ",
+      getFFMPEGErrorStringFromErrorCode(status));
+
+  AutoAVPacket autoAVPacket;
+
+  uint8_t* pwf = static_cast<uint8_t*>(wf_.data_ptr());
+  int numSamples = static_cast<int>(wf_.sizes()[1]); // per channel
+  int numEncodedSamples = 0; // per channel
+  int numSamplesPerFrame = avCodecContext_->frame_size; // per channel
+  int numBytesPerSample = wf_.element_size();
```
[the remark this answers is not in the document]
NicolasHug: Ugh yeah, I was looking at the Python doc which of course said "int"...

scotts (approved):

src/torchcodec/_core/Encoder.h
scotts: There probably are - I'm open to us doing a giant conversion at one point.

src/torchcodec/_core/Encoder.cpp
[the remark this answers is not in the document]
scotts: I'm a little concerned that our compiles let that through, even with -Werror and -pedantic. I _thought_ that would force us to be explicit about narrowing conversions.

scotts:
Looks great, thanks for tackling this feature! :)

PR #529, opened 2025-02-28
Remove compile test

scotts:
We're getting failures in some other places, and digging into why is not yet a priority. We can revisit when exporting and compilation become a priority.


NicolasHug (approved):
Approving to unblock but since these tests are passing on GitHub, I created https://github.com/pytorch/torchcodec/pull/530 as an alternative that only skips those tests on fbcode. But happy to go with either.

scotts:
Closed in favor of #530.

PR #534, opened 2025-03-04
Refactor ffmpegStatus to status

scotts:
In `VideoDecoder.cpp`, we've consolidate on `ffmpegStatus` as the `int` return variable from all FFMpeg calls. Using the same variable name everywhere is good, but I find the length of the name interferes with my own reading of the most important part of the statements, which is the actual call to the FFMpeg function.

This PR changes the name to `status`, as I find that easier to read. Let the bikeshedding commence. :)


NicolasHug (approved):
As much as I love to bikeshed, I'm afraid I like the changes as-is. Sorry.

PR #535, opened 2025-03-05
Audio metadata support

NicolasHug:
This PR adds metadata  support for audio streams. It does **not** add any audio decoding capabilities.

I'm not writing a design doc, because the proposed design is a natural generalization of what we already have for videos, and I hope the implementation is obvious enough. In the main lines:

- We add `AudioStreamMetadata`, the audio counterpart of existing `VideoStreamMetadata`. They share a bunch of fields, so I created a private base class `StreamMetadata` where the common fields are defined.
- `AudioStreamMetadata` does *not* expose any field that relates to the concept of frame (like `num_frames`, `average_fps`, etc.). This is in spite of "frame" being a valid audio concept within FFmpeg. Users care about audio samples, not frames. 
- The `VideoMetadata` class becomes `ContainerMetadata`. This is because we want to expose audio metadata on files that are pure audio (like mp3 files), not just videos files.
- Similar functions / parameters that where previously like "video metadata" now use the term "container" instead of "video". This is a tiiiiiny bit BC-breaking for core APIs, which were private anyway.
- Publicly, the audio metadata is exposed via `AudioDecoder(...).metadata`, just like for videos. The `AudioDecoder` class is currently private and does **not** support any decoding capabilities.

There are a bunch of TODOs marked as `TODO-AUDIO` which are items we'll need to address before the release, but that can be done in follow-up PRs.


src/torchcodec/decoders/_core/FFMPEGCommon.cpp
```diff
@@ -60,6 +60,17 @@ int64_t getDuration(const AVFrame* frame) {
 #endif
 }
 
+int64_t getNumChannels(const UniqueAVCodecContext& avCodecContext) {
+#if LIBAVFILTER_VERSION_MAJOR > 8 || \
+    (IBAVFILTER_VERSION_MAJOR == 8 && LIBAVFILTER_VERSION_MINOR >= 44)
```
NicolasHug, line 65: TODO: typo, missing `L`

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -497,6 +485,36 @@ void VideoDecoder::addVideoStream(
...
+    throw std::runtime_error(
+        "Seek mode is approximate, but stream " +
+        std::to_string(activeStreamIndex_) +
+        " does not have an average fps in its metadata.");
+  }
+
+  auto& streamInfo = streamInfos_[activeStreamIndex_];
+  streamInfo.videoStreamOptions = videoStreamOptions;
+  streamInfo.codecContext->thread_count =
+      videoStreamOptions.ffmpegThreadCount.value_or(0);
```
NicolasHug, line 514: TODO need to check before merging whether this should be put back within `addStream()`, before the call to `initializeContextOnCuda()`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -409,84 +408,73 @@ VideoDecoder::VideoStreamOptions::VideoStreamOptions(
...
 
   activeStreamIndex_ = av_find_best_stream(
-      formatContext_.get(), AVMEDIA_TYPE_VIDEO, streamIndex, -1, &avCodec, 0);
+      formatContext_.get(), mediaType, streamIndex, -1, &avCodec, 0);
+
   if (activeStreamIndex_ < 0) {
-    throw std::invalid_argument("No valid stream found in input file.");
+    throw std::invalid_argument(
+        "No valid stream found in input file. Is " +
+        std::to_string(streamIndex) + " of the desired media type?");
```
scotts, line 431: Minor, but since we're changing this, we could make it a `TORCH_CHECK`. We have a mix of `TORCH_CHECK` and throwing exceptions; I think we should make them all `TORCH_CHECK()`s. It's not a big enough deal to do everywhere, but might as well as we change code. Heads-up that it might require changing the C++ test case; it might expect `std::invalid_argument`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -497,6 +485,36 @@ void VideoDecoder::addVideoStream(
...
+  auto& streamMetadata =
+      containerMetadata_.allStreamMetadata[activeStreamIndex_];
+
+  if (seekMode_ == SeekMode::approximate &&
+      !streamMetadata.averageFps.has_value()) {
+    throw std::runtime_error(
+        "Seek mode is approximate, but stream " +
+        std::to_string(activeStreamIndex_) +
+        " does not have an average fps in its metadata.");
+  }
```
scotts, line 509: Ditto on `TORCH_CHECK`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1587,7 +1620,8 @@ double VideoDecoder::getMaxSeconds(const StreamMetadata& streamMetadata) {
 // VALIDATION UTILS
 // --------------------------------------------------------------------------
 
-void VideoDecoder::validateActiveStream() {
+void VideoDecoder::validateActiveStream(
+    std::optional<AVMediaType> avMediaType) {
```
scotts, line 1624: Is there ever a time when we will want to generically check the valid active stream but not know if we're audio or video? Is there any place we're doing it currently? If the answer to the second question is no, I'd prefer to make this a required field now. We can revisit when we encounter that scenario.

NicolasHug: Yes I should have commented on this: I made this optional because one the decoding PR (still WIP), there are calls to this function where we want to acept both audio and video streams (namely in getFramesPlayedInRange())

test/decoders/test_video_decoder.py
```diff
@@ -11,7 +11,9 @@
 import torch
 from torchcodec import FrameBatch
 
-from torchcodec.decoders import _core, VideoDecoder
+from torchcodec.decoders import _core, VideoDecoder, VideoStreamMetadata
+from torchcodec.decoders._audio_decoder import AudioDecoder
```
scotts, line 15: Since we're testing both video and audio in the same file, we should probably change this file name to `test_decoders.py` or something similar.

test/decoders/test_video_decoder_ops.py
```diff
@@ -5,6 +5,7 @@
 # LICENSE file in the root directory of this source tree.
 
 import os
+from functools import partial
```
scotts, line 8: Ditto for the ops tests - although we've been in this situation for a long time because we've had the audio metadata test.

scotts (approved):

scotts:
Previously, I had advocated to not have a `ContainerMetadata` class which was generic to both audio and video, but instead have a top-level `VideoMetadata` and `AudioMetadata` classes for each kind of file.

I still slightly prefer that approach, as I think we can model the difference between video and audio files in a way that I think will resonate more with our users. That is, the current way of modeling the metadata matches FFmpeg, which is convenient for the implementation; one generic class which both kinds of media can slot into. I agree that the incongruity is minor: if we have a separate top-level `AudioMetadata` class then we can model that it can only contain audio streams, while a `VideoMetadata` can contain audio, video and subtitles. I *think* audio files can have multiple streams, although it's rare and I can't get clarity on that.

With that said, I'm good with this approach, as users will mostly just see the values, not the types or even the class names.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: Agreed, let me open an issue for this, this would be a good first contribution

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: Going back to this:  `initializeContextOnCuda` doesn't seem to rely on `threadCount` at all, but `avcodec_open2` definitely does.
So I'll have to put this back, which is going to be ugly :/

NicolasHug:
Thanks for the review! I remember now we had a few discussions about the single "container" metadata vs having 2 separate video and audio classes for this. I don't feel too strongly about either at the moment, and now that we're introducing audio seems like *the* moment to figure that out for good. Let me merge this as-is so as to unblock the rest, but we'll make sure to explore alternative designs before making all this public

PR #538, opened 2025-03-06
Audio decoding support: range-based core API

NicolasHug:
This PR adds the `get_frames_by_pts_in_range_audio(start_seconds, stop_seconds=None) -> Tensor` **core** API.

- It returns a 2D tensor of shape `(num_channels, num_samples)`.
  - We don't return pts or duration. We might eventually do it, I'm open to it, but those are directly deductible from `start_seconds` and the sample rate, so I'm leaving it out. 
  - We're not returning something of shape (e.g.) `(num_frames, num_channels, num_samples_per_frame)`, and we never will, because audio frames generally contain variable number of samples. I found out the hard way.
- It is a frame-based API. That's OK for a core API. The public decoder method will be sample-based, but that's out of scope for now.
- It allows consecutive calls, but only if it doesn't require a backwards seek (that'll be implemented later). See the tests.
- `stop_seconds` is None by default, so that users can decode to the end of the file without knowing what the duration is. setting `stop_seconds=<some super high value>` doesn't raise an error either. (Note that IMHO we should extend this  to all range-based APIs: https://github.com/pytorch/torchcodec/issues/150)

We've discussed a lot offline already, so I won't be writing down everything that went into the design decisions here. But once everything is done, I'll make sure to write down a note in the code that documents why audio decoding is implemented the way it is.

----

Preliminary benchmarks against torchaudio (built from source) are very promising, even for this non-optimized first version:

```
Duration: 13s
torchcodec: med = 8.01ms +- 1.03
torchaudio: med = 11.90ms +- 0.60

Duration: 13s
torchcodec: med = 4.07ms +- 0.73
torchaudio: med = 7.22ms +- 0.63

Duration: 2m11s
torchcodec: med = 30.18ms +- 0.75
torchaudio: med = 45.17ms +- 1.85

Duration: 1h27m
torchcodec: med = 1060.43ms +- 23.78
torchaudio: med = 1746.49ms +- 22.55
```

Code:
<details>

```py

from torchcodec.decoders import _core as core
from torchaudio.io import StreamReader
import torch



    for _ in range(warmup):
        f(*args, **kwargs)

    times = []
    for _ in range(num_exp):
        start = perf_counter_ns()
        f(*args, **kwargs)
        end = perf_counter_ns()
        times.append(end - start)

def report_stats(times, unit="ms", prefix=""):
    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    print(f"{prefix}: {med = :.2f}{unit} +- {std:.2f}")
    return med



def codec(path, stream_index):
    decoder = core.create_from_file(path, seek_mode="approximate")
    core.add_audio_stream(decoder, stream_index=stream_index)

    core.get_frames_by_pts_in_range_audio(decoder, start_seconds=0, stop_seconds=None)

def audio(path, stream_index):
    reader = StreamReader(path)
    reader.add_audio_stream(frames_per_chunk=1024, stream_index=stream_index)
    for _ in reader.stream():
        pass

NUM_EXP = 30
WARMUP = 1
for path, stream_index, duration in (
    ("/home/nicolashug/dev/torchcodec/test/resources/nasa_13013.mp4", 4, "13s"),
    ("/home/nicolashug/dev/torchcodec/test/resources/nasa_13013.mp4.audio.mp3", 0, "13s"),
    ("/home/nicolashug/test_videos/long.mp4", 1, "2m11s"),
    ("/home/nicolashug/test_videos/output.mp4", 1, "1h27m"),
):
    print(f"Duration: {duration}")
    times = bench(codec, path, stream_index, num_exp=NUM_EXP, warmup=WARMUP)
    report_stats(times, prefix="torchcodec")

    times = bench(audio, path, stream_index, num_exp=NUM_EXP, warmup=WARMUP)
    report_stats(times, prefix="torchaudio")
    print()
```

</details>

----

There will be plenty of follow-ups, mainly:

- enable backwards seeks
- enable audio formats other than fltp
- enable a user-defined `sample_rate`
- expose a public method in `AudioDecoder`
- maybe, maybe not: something like the [`normalize` parameter of the torchaudio reader](https://pytorch.org/audio/main/generated/torchaudio.load.html), which allows users to specify whether they want a float tensor in [-1, 1], or a tensor with the same dtype as the audio format. We'll figure that out later. We'll probably always return float tensors by default anyway.
- perf: try to pre-allocate the output tensor and save copies


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -837,6 +839,66 @@ VideoDecoder::FrameBatchOutput VideoDecoder::getFramesPlayedInRange(
...
+    auto pts = secondsToClosestPts(seconds, streamInfo.timeBase);
+    return (
+        streamInfo.lastDecodedAvFramePts <= pts and
+        pts <= streamInfo.lastDecodedAvFramePts +
+                streamInfo.lastDecodedAvFrameDuration);
+  };
+
+  // TODO-AUDIO This essentially enforce that we don't need to seek (backwards).
+  // We should remove it and seek back to the stream's beginning when needed.
+  // See test_multiple_calls
```
NicolasHug, line 867: It'll be easier to implement the seek logic once https://github.com/pytorch/torchcodec/pull/543 is merged

test/decoders/test_decoders.py
```diff
@@ -939,11 +940,18 @@ def get_some_frames(decoder):
...
+        assert (
+            decoder.stream_index
+            == decoder.metadata.stream_index
+            == asset.default_stream_index
+        )
+        assert decoder.metadata.duration_seconds == pytest.approx(
+            asset.duration_seconds
+        )
+        assert decoder.metadata.sample_rate == asset.sample_rate
+        assert decoder.metadata.num_channels == asset.num_channels
```
NicolasHug, line 957: Changes here are mostly a drive-by. I'm extending this one because I removed `test_audio_get_json_metadata` which was outdated.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -837,6 +839,67 @@ VideoDecoder::FrameBatchOutput VideoDecoder::getFramesPlayedInRange(
...
+          decodeAVFrame([&streamInfo](AVFrame* avFrame) {
+            return (
+                streamInfo.discardFramesBeforePts <
+                avFrame->pts + getDuration(avFrame));
+          });
+      auto frameOutput = convertAVFrameToFrameOutput(avFrameStream);
+      tensors.push_back(frameOutput.data);
+    } catch (const EndOfFileException& e) {
+      reachedEOF = true;
+    }
```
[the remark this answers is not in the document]
scotts: Since we shouldn't ordinarily decode past the end of a file, I think it makes sense for us to throw exceptions when we reach the end of a file. Here, we're not really using an exception for control flow per se. That is, we're not *trying* to read past the end of the file, we just have to handle the case that we might.

With that said, I do find it more natural when the "normal" stop conditions are explicitly part of the while loop's condition, as opposed to setting a boolean inside the loop. But since you're depending on the internal state of the decoder to know the last decoded frame info, I don't know if that's possible. When I implemented something similar, I ended up using a [priming read](https://github.com/pytorch/torchcodec/pull/440/commits/081a5bb629e6624cd772ceea3354bbca253d0ce4#diff-c7e9bb49689c1d47b393890217f3c999483406b81cf4298278451305d19034a8R1465) to get around this problem.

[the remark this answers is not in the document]
scotts: Also, nit: for local variables used in a small space with a clear purpose, I prefer shorter names. So even `stop` as the boolean would make this easier for me to read.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -379,12 +384,18 @@ class VideoDecoder {
...
+      FrameOutput& frameOutput,
+      std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt);
+
   torch::Tensor convertAVFrameToTensorUsingFilterGraph(const AVFrame* avFrame);
 
   int convertAVFrameToTensorUsingSwsScale(
       const AVFrame* avFrame,
       torch::Tensor& outputTensor);
 
+  FrameBatchOutput makeFrameBatchOutput(int64_t numFrames);
```
scotts, line 398: There's no implementation for this member function? Is it a remnant of an old version of the code?

test/decoders/test_ops.py
```diff
@@ -628,22 +626,179 @@ def test_cuda_decoder(self):
...
+        "asset, expected_shape", ((NASA_AUDIO, (2, 1024)), (NASA_AUDIO_MP3, (2, 576)))
+    )
+    def test_decode_epsilon_range(self, asset, expected_shape):
+        decoder = create_from_file(str(asset.path), seek_mode="approximate")
+        add_audio_stream(decoder)
+
+        frames = get_frames_by_pts_in_range_audio(
+            decoder, start_seconds=5, stop_seconds=5 + 1e-5
+        )
+        assert frames.shape == expected_shape
```
scotts, line 710: Can we also assert the tensors are close?

test/decoders/test_ops.py
```diff
@@ -628,22 +626,179 @@ def test_cuda_decoder(self):
...
+        add_audio_stream(decoder)
+
+        frames = get_frames_by_pts_in_range_audio(
+            decoder, start_seconds=start_seconds, stop_seconds=stop_seconds
+        )
+
+        torch.testing.assert_close(frames, reference_frames)
+
+    @pytest.mark.parametrize(
+        "asset, expected_shape", ((NASA_AUDIO, (2, 1024)), (NASA_AUDIO_MP3, (2, 576)))
```
scotts, line 701: Can we use the number of channels and sample rate from the `TestContainer`? I know that for MP3, this is potentially a problem as the sample rate tends to vary, but noting that in a comment would help as well.

test/decoders/test_ops.py
```diff
@@ -628,22 +626,177 @@ def test_cuda_decoder(self):
...
+        # TODO-AUDIO We shouldn't error, we should just implement the seeking
+        # back to the beginning of the stream.
+
+        def get_reference_frames(start_seconds, stop_seconds):
+            decoder = create_from_file(str(asset.path), seek_mode="approximate")
+            add_audio_stream(decoder)
+
+            return get_frames_by_pts_in_range_audio(
+                decoder, start_seconds=start_seconds, stop_seconds=stop_seconds
+            )
```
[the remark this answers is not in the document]
scotts: Shouldn't this helper get its reference from the the test utils? Right now it's getting frames by decoding the file, which means we're not actually comparing against a reference - or am I missing something here?

test/utils.py
```diff
@@ -303,15 +316,114 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
+            # 0.13~, not 0.
+            return 0
+        try:
+            return next(
+                frame_index
+                for (frame_index, frame_info) in self.frames[stream_index].items()
+                if frame_info.pts_seconds
+                <= pts_seconds
+                < frame_info.pts_seconds + frame_info.duration_seconds
+            )
```
scotts, line 386: I have to admit I don't completely understand what's going on here. I get that we have a generator that does a linear walk through the frames, searching for the first frame that meets our conditions, but how does passing that generator to `next()` get us back the one index we want?

Also, it may be better to do this once, in `__post_init__()` and set up a mapping rather than doing it on every call.

test/decoders/test_ops.py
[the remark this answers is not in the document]
NicolasHug: You're correct. Relying on the references would mean converting all the timestamps into indices, and as mentioned in my comment just above, I wanted to avoid complicating this test further.

Eventually I think we will update this test (especially when we enable backwards seeking), at which point we could just rely on the reference frames.

If you prefer converting to indices straight away, I'm fine with it. Let me know.

test/utils.py
NicolasHug: Unfortunately we can't build a mapping for this because we are mapping contiguous timestamps to integers. I'll add a comment that `bisect` might make things faster if needed (although for such small arrays, the tests run very fast).

`next(it)` just returns the very first entry in `it`, whether `it` is a list, a tuple, a generator, etc. And then we build the generator in such a way that the first frame in the generator is the frame we want. What's going on is exactly the same as the snippet below:

```py
>>> gen = (i for i in range(10) if i > 5)
>>> next(gen)
6
```

The values in `[0, 5]` were never part of `gen`. The first entry in `gen` is `6`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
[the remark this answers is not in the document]
NicolasHug: OK. I'll use `finished` instead of `stop`, because we already have local variables named `stopPts` and `stopSeconds` in this function.

test/decoders/test_ops.py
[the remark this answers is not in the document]
scotts: Ohhhh, okay, I had misunderstood what you meant about avoiding the conversions. It's fine to leave as-is, but let's put in a comment explaining that we're comparing a decoder which only seeks once to a decoder which seeks multiple times along with a TODO to convert it to loading the reference frames from indices.

test/utils.py
scotts: Wow! I never thought about it that way, in that `next()` on a generator is effectively like `lst[0]` on a sequence.

scotts (approved):

PR #540, opened 2025-03-06
Better error message if metadata is missing in approximate mode

NicolasHug:
We haven't had this problem yet for videos, but as I'm developing audio support I started hitting some "bad optional access" errors. In my case, it was due to this bad access:

https://github.com/pytorch/torchcodec/blob/b7271fa66c0c1127144f2e4149c974a30ca75a3a/src/torchcodec/decoders/_core/VideoDecoder.cpp#L1594-L1595

We set the value here, when the decoder is created:

https://github.com/pytorch/torchcodec/blob/b7271fa66c0c1127144f2e4149c974a30ca75a3a/src/torchcodec/decoders/_core/VideoDecoder.cpp#L155-L158

Turns out that this value doesn't exist for our audio mp3 test file.

This PR changes the un-informative error message into something more informative. Note that we have to do the checks / enforcements at the very last moment instead of when the decoder is initiated (and when those fields are set), because that would prevent us from returning metadata in approximate mode.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1536,6 +1536,9 @@ int64_t VideoDecoder::secondsToIndexLowerBound(double seconds) {
     case SeekMode::approximate: {
       auto& streamMetadata =
           containerMetadata_.allStreamMetadata[activeStreamIndex_];
+      TORCH_CHECK(
+          streamMetadata.averageFps.has_value(),
+          "Cannot use approximate mode since we couldn't find the average fps from the metadata.");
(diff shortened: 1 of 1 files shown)


scotts:
> Note that we have to do the checks / enforcements at the very last moment instead of when the decoder is initiated (and when those fields are set), because that would prevent us from returning metadata in approximate mode.

Hmmm. That seems really unfortunate that we can't error as early as possible (construction time) in order to support a specific use-case. I originally called it "uncommon," but then I realized that I personally do it all the time. :) Let's merge this, and if users have difficulty that we're only erroring at decode time, we can revisit.

scotts (approved):

PR #541, opened 2025-03-07
Check in json files with test frame infos

NicolasHug:


test/utils.py
```diff
@@ -100,14 +101,64 @@ class TestFrameInfo:
...
+        # This will output the metadata for the frames of the second video
+        # stream (v:1). First audio stream would be a:0.
+        # Note that we are using the absolute stream index in the file. But
+        # ffprobe uses a relative stream for that media type.
+        for stream_index in self.stream_infos:
+            if stream_index in self.frames:
+                # .frames may be manually set: for some streams, we don't need
+                # the info for all frames. We don't need to load anything in
+                # this case
+                continue
```
NicolasHug, line 141: I don't feel too strongly about this but if we were to enforce the existence of a json file for every single asset, it would raise the barrier to the first test when e.g. adding a new test asset, which I think isn't desirable.

test/utils.py
```diff
@@ -253,55 +296,22 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
-            6: TestFrameInfo(pts_seconds=0.200200, duration_seconds=0.033367),
-            7: TestFrameInfo(pts_seconds=0.233567, duration_seconds=0.033367),
-            8: TestFrameInfo(pts_seconds=0.266933, duration_seconds=0.033367),
-            9: TestFrameInfo(pts_seconds=0.300300, duration_seconds=0.033367),
-            10: TestFrameInfo(pts_seconds=0.333667, duration_seconds=0.033367),
-            25: TestFrameInfo(pts_seconds=0.8342, duration_seconds=0.033367),
-            35: TestFrameInfo(pts_seconds=1.1678, duration_seconds=0.033367),
-        },
-    },
+    frames={},  # Automatically loaded from json file
```
NicolasHug, line 303: We have to create an empty `frames` field here. This field cannot have a default within the `TestContainerFile` definition, because [defaults aren't allowed for base dataclasses](https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses)

scotts (approved):

PR #542, opened 2025-03-07
Remove unused C++ decoder creation

scotts:
The C++ function `create_from_buffer()` is not actually used anywhere in our library. Let's remove it; personally, this makes it easier for me to navigate around and reason about our creation functions.


NicolasHug (approved):
image


image

PR #543, opened 2025-03-09
Simplify seeking and cursor logic

NicolasHug:
In main, we have 2 different fields to indicate where the user requested a seek:

- The `desiredPtsSeconds_` field of the decoder
- The `discardFramesBeforePts` field of active stream's `streamInfo`

They both morally represent the exact same thing. But one's a `std::optional<double>`, the other one's an `int_64_t`. One is set in `setCursorPtsInSeconds`, the other one is set in `decodeAVFrame()`.

This is all pretty confusing, so this PR simplify things: there is now only *one* single `cursor_` field, for the  decoder, and it's set in one single place.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -567,10 +567,8 @@ VideoDecoder::FrameOutput VideoDecoder::getNextFrame() {
 
 VideoDecoder::FrameOutput VideoDecoder::getNextFrameInternal(
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
-  AVFrameStream avFrameStream = decodeAVFrame([this](AVFrame* avFrame) {
-    StreamInfo& activeStreamInfo = streamInfos_[activeStreamIndex_];
-    return avFrame->pts >= activeStreamInfo.discardFramesBeforePts;
-  });
+  AVFrameStream avFrameStream = decodeAVFrame(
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -948,10 +946,9 @@ VideoDecoder::AVFrameStream VideoDecoder::decodeAVFrame(
 
   resetDecodeStats();
 
-  // Seek if needed.
-  if (desiredPtsSeconds_.has_value()) {
+  if (cursorWasJustSet_) {
     maybeSeekToBeforeDesiredPts();
-    desiredPtsSeconds_ = std::nullopt;
+    cursorWasJustSet_ = false;
```
NicolasHug, line 951: I'm not particularly fond of this logic, but don't shoot the messenger. This PR is only making an existing stateful logic a little bit more obvious. I.e. now we have an obvious `bool` field, while before we had a more confusing `std::optional` logic.

We could try to simplify this further, I gave it a quick try but our tests failed in some weird edge-cases, so I'm leaving this for later.

scotts:
This is great! I also tried simplifying this logic in PR #262, but I abandoned it because I there was concern it would undo the lazy behavior that we want. Now that we've simplified other stuff, and there's only one active stream, I think this is obviously maintaining the lazy behavior we want.

scotts (approved):

PR #550, opened 2025-03-12
Allow audio decoder to seek backwards

NicolasHug:
Towards https://github.com/pytorch/torchcodec/issues/549.

Changes are quite subtle, but I'm fairly confident they work as expected, since they pass our robust tests. I'll make sure to write some solid documentation around the mechanisms involved eventually.

Benchmarks results show no perf hit

```
Duration: 13s
torchcodec: med = 8.05ms +- 1.13
torchaudio: med = 12.27ms +- 0.51

Duration: 13s
torchcodec: med = 4.20ms +- 0.91
torchaudio: med = 7.21ms +- 0.58

Duration: 2m11s
torchcodec: med = 28.26ms +- 0.95
torchaudio: med = 45.72ms +- 0.89

Duration: 1h27m
torchcodec: med = 1046.49ms +- 66.00
torchaudio: med = 1746.49ms +- 22.55
```

Benchmark code is the same as in https://github.com/pytorch/torchcodec/pull/538, I'm not benchmarking the "backwards seeking" logic.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -877,10 +877,9 @@ torch::Tensor VideoDecoder::getFramesPlayedInRangeAudio(
   auto finished = false;
   while (!finished) {
     try {
-      AVFrameStream avFrameStream =
-          decodeAVFrame([this, startPts](AVFrame* avFrame) {
-            return startPts < avFrame->pts + getDuration(avFrame);
-          });
+      AVFrameStream avFrameStream = decodeAVFrame([startPts](AVFrame* avFrame) {
+        return startPts < avFrame->pts + getDuration(avFrame);
+      });
       auto frameOutput = convertAVFrameToFrameOutput(avFrameStream);
       tensors.push_back(frameOutput.data);
     } catch (const EndOfFileException& e) {
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -859,21 +859,20 @@ torch::Tensor VideoDecoder::getFramesPlayedInRangeAudio(
 
   StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
 
-  //   TORCH_CHECK(
-  //       streamInfo.lastDecodedAvFramePts +
-  //               streamInfo.lastDecodedAvFrameDuration <=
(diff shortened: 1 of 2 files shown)


test/decoders/test_ops.py
```diff
@@ -741,19 +741,18 @@ def test_decode_start_equal_stop(self, asset):
...
-            # Using this helper is OK for now: we're comparing a decoder which
-            # seeks multiple times with a decoder which seeks only once (the one
-            # here, treated as the reference)
+            # Usually we get the reference frames from the asset's methods, but
+            # for this specific test, this helper is more convenient, because
+            # relying on the asset would force us to convert all timestamps into
+            # indices.
+            # Ultimately, this test compares a "stateful decoder" which calls
+            # `get_frames_by_pts_in_range_audio()`` multiple times with a
+            # "stateless decoder" (the one here, treated as the reference)
```
NicolasHug, line 755: I've convinced myself that we should actually keep this helper instead of doing the conversion. Thoughts?

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -859,16 +859,14 @@ torch::Tensor VideoDecoder::getFramesPlayedInRangeAudio(
...
-
-  setCursorPtsInSeconds(startSeconds);
+  auto startPts = secondsToClosestPts(startSeconds, streamInfo.timeBase);
+  if (startPts < streamInfo.lastDecodedAvFramePts +
+          streamInfo.lastDecodedAvFrameDuration) {
+    // If we need to seek backwards, then we have to seek back to the beginning
+    // of the stream.
+    // TODO-AUDIO: document why this is needed in a big comment.
+    setCursorPtsInSeconds(INT64_MIN);
+  }
```
NicolasHug, line 869: Note that this is `INT64_MIN` and not 0, because some packets actually start before 0. In one of our assets the first packet is at `-1024`.
I noticed that passing an arbitrary low value like `-999999` makes FFmpeg unhappy and raise and error, but `INT64_MIN` seems to be understood and correct (although I haven't found docs on this).

test/decoders/test_ops.py
scotts: I think that's fine - then what we're testing is that decoder behaves the same when it seeks to a location "fresh" versus having to seek from some given location, including backwards. That seems reasonable.

scotts (approved):

PR #551, opened 2025-03-12
Allow `sample_rate` parameter to audio decoder

NicolasHug:

This PR allows the user to specify a a desired output sample rate. If the source stream isn't already in that sample rate, the sample are converted using swresample, which we already use for format conversion.

The PR addresses both core and public `AudioDecoder` APIs.

Converting sample rates isn't as trivial as converting formats, because this changes the number of output samples. And, importantly, we now need to account for libswresample's internal buffers:

> Note that the samples may get buffered in swr if you provide insufficient output space or if sample rate conversion is done, which requires "future" samples. [...] At the end of conversion the resampling buffer can be flushed by calling [swr_convert()](https://www.ffmpeg.org/doxygen/7.0/group__lswr.html#gad16a8308503be52a7e2d460983e81460) with NULL in and 0 in_count

(from [the docs](https://www.ffmpeg.org/doxygen/7.0/group__lswr.html))


test/decoders/test_decoders.py
```diff
@@ -1088,3 +1090,65 @@ def test_format_conversion(self):
...
+        atol = 1e-4 if start_seconds == 0 else 1e-2
+        rtol = 1e-6
+
+        # Upsample
+        decoder = AudioDecoder(SINE_MONO_S32_44100.path)
+        assert decoder.metadata.sample_rate == 44_100
+        frames_44100_native = decoder.get_samples_played_in_range(
+            start_seconds=start_seconds, stop_seconds=stop_seconds
+        )
+        assert frames_44100_native.sample_rate == 44_100
```
NicolasHug, line 1118: We're treating the above frames as the reference. We could check-in their corresponding `.pt` reference frames instead, but I prefer doing it this way because it really clearly illustrate that the reference have a sample rate of 44_100

src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -25,10 +25,13 @@ def __init__(
         source: Union[str, Path, bytes, Tensor],
         *,
         stream_index: Optional[int] = None,
+        sample_rate: Optional[int] = None,
```
NicolasHug, line 28: We could also consider exposing this as `desired_sample_rate` parameter? I don't have a strong opinion, the docs would make it clear what this means in any case.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -913,6 +917,11 @@ VideoDecoder::AudioFramesOutput VideoDecoder::getFramesPlayedInRangeAudio(
         (stopPts <= lastDecodedAvFrameEnd);
   }
 
+  torch::Tensor lastSamples = maybeFlushSwrBuffers();
+  if (lastSamples.numel() > 0) {
+    frames.push_back(lastSamples);
+  }
```
NicolasHug, line 923: Not particularly fond of the above. Maybe we *could* let `maybeFlushSwrBuffers` return a tensor of shape `(numChannels, 0)`, which could probably be `pushed_back()` unconditionally. Not sure that's better, there are probably nicer patterns I'm not seeing?

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -139,10 +139,18 @@ class VideoDecoder {
     torch::Device device = torch::kCPU;
   };
 
+  struct AudioStreamOptions {
+    AudioStreamOptions() {}
+
+    std::optional<int> sampleRate;
+  };
+
```
scotts, line 147: I don't think we need the indirection of having an options struct. I know it mirrors the pattern established on the video side, but I also don't think it's a good practice there, either. It's harder to get rid of the video options because we accept a string, and then we do a bunch of work parsing the string. Getting rid of `VideoStreamOptions` will mean updating a bunch of callers to pass real arguments instead of a string.

NicolasHug: I'm not a fan of the stringy video options either - that's why I didn't implement a string constructor  for audio options. I also opened https://github.com/pytorch/torchcodec/pull/577 to entirely remove string video options (it's not as hard as we thought).

I don't feel very strongly about this, but if we were to collapse both video options and audio options into the video decoder (or the StreamInfo), then we would have a lot of video-only and audio-only fields within the same struct. I personally find it cleaner to separate those into separate structs. Additionally, using option structs makes it very clear which fields/values come from user-specified parameter, in constrast to e.g. metadata or video properties.

LMK you thoughts, I'm fine with collapsing sampleRate as a StreamInfo field if you prefer, but we'd potentially be losing the 2 benefits mentioned above.

scotts: Yeah, that's a good point, we actually _store_ the options. I was thinking just from the `addAudioStream()` API perspective. Let's keep this then, and eventually get rid of the stringy options for video.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1419,55 +1433,107 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
+  torch::Tensor lastSamples = torch::empty(
+      {getNumChannels(streamInfo.codecContext), numRemainingSamples},
+      torch::kFloat32);
+  uint8_t* lastSamplesData = static_cast<uint8_t*>(lastSamples.data_ptr());
+
+  auto actualNumRemainingSamples = swr_convert(
+      streamInfo.swrContext.get(),
+      &lastSamplesData,
+      numRemainingSamples,
+      NULL,
```
scotts, line 1531: s/`NULL`/`nullptr`/g

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Everything I can think of that does things unconditionally in this function is way too cute (returning a vector of tensors; using `std::copy()`). It may be more clear about intent if `maybeFlushSwrBuffers()` returns an optional so that then we don't need to use an empty tensor to indicate nothing to do.

scotts (approved):

PR #552, opened 2025-03-12
Return pts of first frame in audio API

NicolasHug:
I thought we didn't need this, but we actually need it in order to implement the public sample-based API (which is out of scope for this PR).

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -881,6 +881,10 @@ VideoDecoder::AudioFramesOutput VideoDecoder::getFramesPlayedInRangeAudio(
       AVFrameStream avFrameStream = decodeAVFrame([startPts](AVFrame* avFrame) {
         return startPts < avFrame->pts + getDuration(avFrame);
       });
+      // TODO: it's not great that we are getting a FrameOutput, which is
+      // intended for videos. We should consider bypassing
+      // convertAVFrameToFrameOutput and directly call
+      // convertAudioAVFrameToFrameOutputOnCPU.
       auto frameOutput = convertAVFrameToFrameOutput(avFrameStream);
       firstFramePtsSeconds =
           std::min(firstFramePtsSeconds, frameOutput.ptsSeconds);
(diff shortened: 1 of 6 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -838,7 +838,7 @@ VideoDecoder::FrameBatchOutput VideoDecoder::getFramesPlayedInRange(
   return frameBatchOutput;
 }
 
-torch::Tensor VideoDecoder::getFramesPlayedInRangeAudio(
+VideoDecoder::AudioFramesOutput VideoDecoder::getFramesPlayedInRangeAudio(
```
NicolasHug, line 841: I decided to create a new `AudioFramesOutput` struct instead of relying on the existing `FrameOutput`, which contains unnecessary fields like `streamIndex` and `durationSeconds`. For now, those aren't needed for audio. No super strong opinion though.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -883,7 +884,9 @@ torch::Tensor VideoDecoder::getFramesPlayedInRangeAudio(
         return cursor_ < avFrame->pts + getDuration(avFrame);
       });
       auto frameOutput = convertAVFrameToFrameOutput(avFrameStream);
-      tensors.push_back(frameOutput.data);
+      firstFramePtsSeconds =
+          std::min(firstFramePtsSeconds, frameOutput.ptsSeconds);
+      frames.push_back(frameOutput.data);
```
NicolasHug, line 889: It's now a bit ugly that we are manipulating both a FrameOutput and an `AudioFramesOutput` here in this function.

This is mostly because `convertAVFrameToFrameOutput` returns a `FrameOutput`, but maybe we could bypass it and directly call `convertAudioAVFrameToFrameOutputOnCPU`. I'll write a TODO to investigate this.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: I think that's the right call, since the tensors are going to be very different.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Oh, yeah, that's cumbersome. If we're in an audio-only call, we shouldn't have to deal with video structures.

scotts (approved):

PR #555, opened 2025-03-13
Add `AudioDecoder.get_samples_played_in_range()` public method

NicolasHug:

### DIFF
--- a/src/torchcodec/_frame.py
+++ b/src/torchcodec/_frame.py
@@ -115,13 +115,16 @@ class FrameBatch(Iterable):
     def __repr__(self):
         return _frame_repr(self)
 
+
 @dataclass
 class AudioSamples(Iterable):
     """Audio samples with associated metadata."""
+
     # TODO-AUDIO: docs
     data: Tensor
     pts_seconds: float
     sample_rate: int
+
     def __post_init__(self):
         # This is called after __init__() when a Frame is created. We can run
         # input validation checks here.
--- a/src/torchcodec/_frame.py
+++ b/src/torchcodec/_frame.py
@@ -12,7 +12,7 @@ from torch import Tensor
 
 
 def _frame_repr(self):
-    # Utility to replace Frame and FrameBatch __repr__ method. This prints the
+    # Utility to replace __repr__ method of dataclasses below. This prints the
     # shape of the .data tensor rather than printing the (potentially very long)
     # data tensor itself.
     s = self.__class__.__name__ + ":\n"
@@ -111,6 +111,28 @@ class FrameBatch(Iterable):
 
     def __len__(self):
         return len(self.data)
+
+    def __repr__(self):
+        return _frame_repr(self)
+
+@dataclass
+class AudioSamples(Iterable):
+    """Audio samples with associated metadata."""
+    # TODO-AUDIO: docs
+    data: Tensor
+    pts_seconds: float
+    sample_rate: int
+    def __post_init__(self):
+        # This is called after __init__() when a Frame is created. We can run
+        # input validation checks here.
+        if not self.data.ndim == 2:
+            raise ValueError(f"data must be 2-dimensional, got {self.data.shape = }")
+        self.pts_seconds = float(self.pts_seconds)
+        self.sample_rate = int(self.sample_rate)
+
+    def __iter__(self) -> Iterator[Union[Tensor, float]]:
+        for field in dataclasses.fields(self):
+            yield getattr(self, field.name)
 
     def __repr__(self):
         return _frame_repr(self)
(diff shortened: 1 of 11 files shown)


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -37,3 +38,70 @@ def __init__(
         ) = get_and_validate_stream_metadata(
             decoder=self._decoder, stream_index=stream_index, media_type="audio"
         )
+        assert isinstance(self.metadata, core.AudioStreamMetadata)  # mypy
+
+    def get_samples_played_in_range(
+        self, start_seconds: float, stop_seconds: Optional[float] = None
+    ) -> AudioSamples:
```
NicolasHug, line 45: I feel like start_seconds default to the stream's beginning, but that can be done later.

src/torchcodec/_frame.py
```diff
@@ -114,3 +114,28 @@ def __len__(self):
 
     def __repr__(self):
         return _frame_repr(self)
+
+
```
NicolasHug, line 118: Below: the audio decoding API returns the new `AudioSamples` class rather than a pure Tensor. I think it has the following benefits:

- users can keep track of the sample_rate within that struct without having to handle it separately
- for some edge-cases, like with our mp3 test asset, the stream's beginning isn't 0. So we also returns `pts_seconds`, which may not always be equal to `start_seconds`.

test/decoders/test_ops.py
```diff
@@ -824,6 +824,32 @@ def get_reference_frames(start_seconds, stop_seconds):
             frames, get_reference_frames(start_seconds, stop_seconds)
         )
 
+    @pytest.mark.parametrize("asset", (NASA_AUDIO, NASA_AUDIO_MP3))
```
NicolasHug, line 827: This test is from https://github.com/pytorch/torchcodec/pull/554 (which we should merge first)

src/torchcodec/_frame.py
scotts: I think this makes sense, and it mirrors the video API. In the video API, `__getitem__()` returns just the tensor, but the named methods return `Frame` or `FrameBatch`. I think we should probably do that for audio as well.

src/torchcodec/_frame.py
```diff
@@ -114,3 +114,28 @@ def __len__(self):
...
+class AudioSamples(Iterable):
+    """Audio samples with associated metadata."""
+
+    # TODO-AUDIO: docs
+    data: Tensor
+    pts_seconds: float
+    sample_rate: int
+
+    def __post_init__(self):
+        # This is called after __init__() when a Frame is created. We can run
```
NicolasHug, line 129:
```suggestion
        # This is called after __init__() when an AudioSamples instance is created. We can run
```

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -854,7 +854,7 @@ VideoDecoder::AudioFramesOutput VideoDecoder::getFramesPlayedInRangeAudio(
 
   if (startSeconds == stopSeconds) {
     // For consistency with video
-    return AudioFramesOutput{torch::empty({0}), 0.0};
+    return AudioFramesOutput{torch::empty({0, 0}), 0.0};
```
NicolasHug, line 857: Drive-by, the video APIs return something of shape `(0, C, H, W)` (where C H W are from the metadata). Here, (0, 0) is the best we can do, at least it preserves the number of dimensions.

src/torchcodec/decoders/_audio_decoder.py
scotts: If we do that, we should also mirror that in the video API. We _could_ try to play the same trick that `range()` does, where the semantics of the first parameter changes based on the number of parameters, but I'd (softlly) rather not do that.

NicolasHug: Agreed, I'll update https://github.com/pytorch/torchcodec/issues/150 eventually with more of these defaults. That would be a good onboarding PR.

src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -37,3 +38,70 @@ def __init__(
...
+            output_pts_seconds = start_seconds
+        else:
+            # In normal cases we'll have first_pts <= start_pts, but in some
+            # edge cases it's possible to have first_pts > start_seconds,
+            # typically if the stream's first frame's pts isn't exactly 0.
+            offset_beginning = 0
+            output_pts_seconds = first_pts
+
+        num_samples = frames.shape[1]
+        last_pts = first_pts + num_samples / self.metadata.sample_rate
```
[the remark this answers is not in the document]
NicolasHug: Yes it's a float. I didn't add the `_seconds` everywhere, but they're all in seconds. I can do it if you think it improves clarity.

We do call `round()` just below, which I think is enough?

scotts (approved):

PR #556, opened 2025-03-13
Support all audio formats by converting to FLTP

NicolasHug:

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -23,6 +23,7 @@ extern "C" {
 #include <libavutil/imgutils.h>
 #include <libavutil/log.h>
 #include <libavutil/pixdesc.h>
+#include <libswresample/swresample.h>
 #include <libswscale/swscale.h>
 }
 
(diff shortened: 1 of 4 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1342,37 +1349,93 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
+  AVSampleFormat desiredSampleFormat = AV_SAMPLE_FMT_FLTP;
+
+  UniqueAVFrame convertedAVFrame;
+  if (sourceSampleFormat != desiredSampleFormat) {
+    convertedAVFrame = convertAudioAVFrameSampleFormat(
+        avFrameStream.avFrame, sourceSampleFormat, desiredSampleFormat);
+  }
+  const UniqueAVFrame& avFrame = (sourceSampleFormat != desiredSampleFormat)
+      ? convertedAVFrame
+      : avFrameStream.avFrame;
```
NicolasHug, line 1363: Above: I wasn't able to use the call to `convertAudioAVFrameSampleFormat` within the ternary expression.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1342,37 +1349,93 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
+  if (!streamInfo.swrContext) {
+    createSwrContext(
+        streamInfo, sampleRate, sourceSampleFormat, desiredSampleFormat);
+  }
+
+  UniqueAVFrame convertedAVFrame(av_frame_alloc());
+  TORCH_CHECK(
+      convertedAVFrame,
+      "Could not allocate frame for sample format conversion.");
+
```
NicolasHug, line 1409: Below: I realize we have previously avoided having any ffmpeg-related pre-proc directives in this file. I wonder however if it's really worth creating a separate util for this one line? I personally find the code easier to read this way, but no strong opinion.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1606,6 +1669,56 @@ void VideoDecoder::createSwsContext(
...
 }
 
+void VideoDecoder::createSwrContext(
+    StreamInfo& streamInfo,
+    int sampleRate,
+    AVSampleFormat sourceSampleFormat,
+    AVSampleFormat desiredSampleFormat) {
+  SwrContext* swrContext = nullptr;
+
+  int status = AVSUCCESS;
```
NicolasHug, line 1679: Below: similar comment, is it worth having a separate function in `FFmpegCommon.h` to do the initializaton?

test/utils.py
```diff
@@ -444,6 +444,9 @@ def sample_format(self) -> str:
     },
 )
 
+# Note that the file itself is s32 sample format, but the reference frames are
+# stored as fltp. We can add the s32 original reference frames once we support
+# decoding to non-fltp format, but for now we don't need to.
```
NicolasHug, line 449: I haven't yet explained how the audio references were created. That'll come soon, just know that I have checked for exact equality with the output of TorchAudio before checking anything in.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: I'd like to consolidate all FFmpeg pre-proc directives in the FFmpeg utility. Most of them are just one line, but we already have 5 or so of them, and I anticipate having more. If we don't consolidate, I think we'd end up having the directives all over `VideoDecoder.cpp` which I think harms readability.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Actually, I just realized that `convertAudioAVFrameSampleFormat()` already has that exact signature. I think we could simplify things a lot if `convertAudioAVFrameSampleFormat()` became `maybe ConvertAudioAVFrameSampleFormat()` (or a better name?) and only did the conversion if needed.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1350,37 +1357,93 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
+  TORCH_CHECK(
+      status == AVSUCCESS,
+      "Could not allocate frame buffers for sample format conversion: ",
+      getFFMPEGErrorStringFromErrorCode(status));
+
+  auto numSampleConverted = swr_convert(
+      streamInfo.swrContext.get(),
+      convertedAVFrame->data,
+      convertedAVFrame->nb_samples,
+      (const uint8_t**)avFrame->data,
```
scotts, line 1437: `static_cast<uint8_t**>(avFrame->data)` to avoid C-style casts which will succeed even if we wouldn't want them to: https://en.cppreference.com/w/cpp/language/explicit_cast

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: Yes, for the same reasons.

src/torchcodec/decoders/_core/VideoDecoder.cpp
NicolasHug: I wasn't able to make it work.

The function cannot take a `UniqueAVFrame avFrame` as input, it has to be a ref (or const ref).

Then it has to return either `avFrame` (which is a ref) or `convertedAVFrame` which is a `UniqueAVFrame` allocated within that function.

I tried a few things, but nothing worked. It's possible I'm doing something wrong.

scotts: I think there may be a way to do it, but it requires doing some `std::move()` calls. I've had to do that on both sides: going in and out of functions. Let's commit this code as-is, and I can try playing with it later.

src/torchcodec/decoders/_core/VideoDecoder.cpp
scotts: I agree it's not great, and I think the answer may be for us to implement them as methods on our unique wrappers. Something like:
```
convertedAVFrame->setChannelLayout(srcFrame);
```
But that will require us to define the wrapper logic ourselves, as opposed to just making them an alias to `std::unique_ptr`.

scotts (approved):

PR #557, opened 2025-03-13
Add `sample_format` to audio metadata

NicolasHug:

This PR adds the `sample_format` field to `AudioStreamMetadata`. In terms of values, it'll be what `av_get_sample_fmt_name()` returns, so it's a bit "ffmpeg-like": flt, fltp, s16, s32... Maybe we'll want to expose a more user-friendly string? I'm not sure, either way, we can always re-map as follow-ups.


test/resources/sine_mono_s32.wav
NicolasHug, line 1: Also adding this new asset, not strictly needed for this PR, but still useful to check a format that's not fltp.
It will be needed in https://github.com/pytorch/torchcodec/pull/556 anyway. It's from TorchAudio.

scotts (approved):

PR #558, opened 2025-03-13
Enable Intel GPU support in torchcodec on Linux (xpu device)

dvrogozh:
This commit enables support for Intel GPUs in torchcodec. It adds:
* ffmpeg-vaapi for decoding
* VAAPI based color space conversion (decoding output to RGBA)
* RGBA surface import as torch tensor (on torch xpu device)
* RGBA to RGB24 tensor slicing

To build torchcodec with Intel GPU support:
* Install pytorch with XPU backend support. For example, with:
```
pip3 install torch --index-url https://download.pytorch.org/whl/xpu
```
* Install oneAPI development environment following https://github.com/pytorch/pytorch?tab=readme-ov-file#intel-gpu-support
* Build and install FFmpeg with `--enable-vaapi`
* Install torcheval (for tests): `pip3 install torcheval`
* Build torchcodec with: `ENABLE_XPU=1 python3 setup.py devel`

Notes:
* RGB24 is not supported color format on current Intel GPUs (as it is considered to be suboptimal due to odd alignments)
* Intel media and compute APIs can't seamlessly work with the memory from each other. For example, Intel computes's Unified Shared Memory pointers are not recognized by media APIs. Thus, lower level sharing via dma fds is needed. This alos makes this part of the solution OS dependent.
* Color space conversion algoriths might be quite different as it happens for Intel. This requires to check PSNR values instead of per-pixel atol/rtol differences.
* Installing oneAPI environment is neded due to https://github.com/pytorch/pytorch/issues/149075

This commit was primary verfied on Intel Battlemage G21 (0xe20b) and Intel Data Center GPU Flex (0x56c0).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -112,6 +112,7 @@ class CMakeBuild(build_ext):
         torch_dir = Path(torch.utils.cmake_prefix_path) / "Torch"
         cmake_build_type = os.environ.get("CMAKE_BUILD_TYPE", "Release")
         enable_cuda = os.environ.get("ENABLE_CUDA", "")
+        enable_xpu = os.environ.get("ENABLE_XPU", "")
         python_version = sys.version_info
         cmake_args = [
             f"-DCMAKE_INSTALL_PREFIX={self._install_prefix}",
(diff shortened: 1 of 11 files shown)


dvrogozh:
Thanks to @eromomon, here is a rebase of the xpu enabling PR. As expected, the core part now just has xpu device interface files added without changes in a common parts of torchcodec. Tests have a number of changes though - we will try to suggest some patches in separate PRs.

dvrogozh:
Rebased on top of latest changes in main to take in:
* #817 
* #812

dvrogozh:
Rebased on top of latest changes in main to take in:
* #814

PR #560, opened 2025-03-14
Add libswresample to FFmpeg source build

NicolasHug:
Similar to https://github.com/pytorch/torchcodec/pull/219, but for libswresample which we'll use to resample audio and convert audio formats.

I took the liberty of pushing those builds to S3 already, and they work as expected as can be seen in  https://github.com/pytorch/torchcodec/pull/562 and https://github.com/pytorch/torchcodec/pull/556

### DIFF
--- a/packaging/build_ffmpeg.sh
+++ b/packaging/build_ffmpeg.sh
@@ -61,7 +61,8 @@ tar -xf ffmpeg.tar.gz --strip-components 1
     --enable-avfilter \
     --enable-avformat \
     --enable-avutil \
-    --enable-swscale
+    --enable-swscale \
+    --enable-swresample
 
 make -j install
 ls ${prefix}/*
@@ -78,6 +79,7 @@ if [[ "$(uname)" == Darwin ]]; then
         avdevice=libavdevice.58
         avfilter=libavfilter.7
         swscale=libswscale.5
+        swresample=libswresample.2
     elif [[ ${major_ver} == 5 ]]; then
         avutil=libavutil.57
         avcodec=libavcodec.59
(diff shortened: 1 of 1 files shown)


packaging/build_ffmpeg.sh
```diff
@@ -78,27 +79,31 @@ if [[ "$(uname)" == Darwin ]]; then
...
         swscale=libswscale.6
+        swresample=libswresample.4
     elif [[ ${major_ver} == 6 ]]; then
         avutil=libavutil.58
         avcodec=libavcodec.60
         avformat=libavformat.60
         avdevice=libavdevice.60
         avfilter=libavfilter.9
         swscale=libswscale.7
+        swresample=libswresample.4
```
NicolasHug, line 98: Yup, it's version 4 twice in a row, this isn't a typo (logs were previously failing)!

packaging/build_ffmpeg.sh
```diff
@@ -155,6 +160,13 @@ if [[ "$(uname)" == Darwin ]]; then
         ${prefix}/lib/${swscale}.dylib
     ${otool} -l ${prefix}/lib/${swscale}.dylib | grep -B2 ${prefix}
 
+    ${install_name_tool} \
+        -change ${prefix}/lib/${avutil}.dylib @rpath/${avutil}.dylib \
+        -delete_rpath ${prefix}/lib \
+        -id @rpath/${swresample}.dylib \
+        ${prefix}/lib/${swresample}.dylib
+    ${otool} -l ${prefix}/lib/${swresample}.dylib | grep -B2 ${prefix}
+
```
NicolasHug, line 169: There is only a dependency on avutil:



```
~/dev/torchcodec (add_swresample*) » ldd $CONDA_PREFIX/lib/libswresample.so | grep av                                                                              nicolashug@nicolashug-fedora-PF2MMKSN
	libavutil.so.58 => /home/nicolashug/.miniconda3/envs/codec/lib/./libavutil.so.58 (0x00007f2924800000)
```

I don't have access to a mac so I cannot provide the output of `otool` on the actual dylib as requested on https://github.com/pytorch/torchcodec/pull/219#issuecomment-2369161421, but that should be fine?

scotts (approved):

PR #564, opened 2025-03-14
Add core support for decoding from Python file-like objects

scotts:
The purpose of this PR is to provide functionality in the core API that allows users to provide a Python [file-like object](https://docs.python.org/3/glossary.html#term-file-like-object) as the video for us to decode. Specifically, we're exposing:
```py
def create_from_file_like(
    file_like: Union[io.RawIOBase, io.BytesIO], seek_mode: Optional[str] = None
) -> torch.Tensor:
```
Everything else in this PR is in direct support of what we need to do to support this new function.

On the kind of file-like object, note that we accept and test both [`io.RawIOBase`](https://docs.python.org/3/library/io.html#io.RawIOBase) and [`io.BytesIO`](https://docs.python.org/3/library/io.html#io.BytesIO). I'm confident we should support `RawIOBase`, as it's unbuffered, byte oriented reading. I'm less sure about `BytesIO`, because it is *buffered* byte oriented reading. The current tests pass, but we're not stressing it much.

On what we had to do to expose this new function:

1. Implement pybind11 functions that don't go through PyTorch C++ custom ops.
2. Split the shared libraries into three: `libtorchcodec_decoderN.so`, `libtorchcodec_custom_opsN.so`, `libtorchcodec_pybind_opsN.so`.
3. Generalize how we handle [AVIOContext](https://www.ffmpeg.org/doxygen/2.1/structAVIOContext.html) objects in the C++ VideoDecoder.

Going over each in turn:

**Why directly use pybind11?**

We're already using PyTorch C++ custom ops for our interface between Python and C++, and they already have some dependence on pybind11, so why do we need to create non-custom op pybind11 functions?

From what I can tell, the custom ops were not really intended for what we want to do, which is call C++ code which will itself make callbacks back up to user-provide Python functions. That is, when the user passes us a file-like object, we want FFmpeg to call the read and seek methods on the Python file-like object for all reading and seeking. I don't think custom ops are designed for that kind of dynamic callbacks to Python code. (Custom ops definitely *can* call Python custom ops, but they need to be registered as such ahead of time. We want to call arbitrary Python functions.) I may be wrong here, and we can explore that in the future.

What's I'm more confident of is that we need to store an actual reference to the Python file-like object in the C++ side. Using pybind11 directly, that's easy: we keep a pointer to a `py::object*`. PyTorch custom ops only accept tensors as arguments. We're already smuggling pointers through tensors in the rest of the core API, but we're going from C++ to Python and back to C++. When we store a pointer from C++ in a tensor to go back up to Python, we know it's the right thing. In this instance, we would want to smuggle a pointer from Python through a tensor to C++ - but Python doesn't have pointers. We may be able to get something that will work most of the time by just asking for `id(file_like)`, but even then, I'm not sure how to reliably turn that into a `py::object*` on the C++ side.

If we just use a pybind11 function, all of this difficulty goes away. The cost is that using a file-like object is definitely not going to be compatible with torch.export and torch.compile, but I'm not sure how to make that known to the PyTorch custom ops. This warrants further investigation.

**Why do we need multiple shared libraries?**

I *think* things are simpler if we have a shared library just for the custom ops and a separate shared library just for the pybind11 ops. That then means we need a third library which holds the actual decoder logic. I was not able to get anything working until I made this division, as I am currently using [`importlib.util.spec_from_file_location()`](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location) and [`importlib.util.module_from_spec()`](https://docs.python.org/3/library/importlib.html#importlib.util.spec_from_file_location) to load the pybind11 module. We just use `torch.ops.load_library()` for the custom ops; that function has machinery that then exposes available functions as fields of the module.

What I'm currently doing works on Linux, but is failing on Mac, so *something* is wrong. It may be possible we don't need to do the split.

**Generalization of handling AVIOContext**

Custom reading and seeking is done in FFmpeg by setting up an `AVIOContext` object. You call [`avio_alloc_context()`](https://ffmpeg.org/doxygen/2.1/avio_8h.html#a853f5149136a27ffba3207d8520172a5) where you provide a pointer to some external state, and then functions for read, seek and write. Then, during decoding, when FFmpeg needs to get more bytes from the file, it calls the callbacks, providing a pointer to the external state. You're responsible for managing that external state in your callbacks.

We already were using this for when users provided us the entire file as just plain bytes. I generalized this handling into three classes:

1. `AVIOContexHolder` which is a base class that knows how to allocate an `AVIOContext`. It cannot be instantiated directly. The `VideoDecoder` can be instantiated with an `AVIOContextHolder` and it uses it appropriately.
2. `AVIOBytesContext` which is the existing functionality we already had. It derives from `AVIOContextHolder`.
3. `AVIOFileLikeContext` which is the new functionality, and it also derives from `AVIOContextHolder`.


src/torchcodec/decoders/_core/ops.py
```diff
@@ -26,19 +31,37 @@ def load_torchcodec_extension():
...
+                pybind_ops_module_name,
+                pybind_ops_library_path,
+            )
+            if spec is None:
+                raise ImportError(
+                    f"Unable to load spec for module {pybind_ops_module_name} from path {pybind_ops_library_path}"
+                )
+
+            global _pybind_ops
+            _pybind_ops = importlib.util.module_from_spec(spec)
```
scotts, line 55: This does not yet work on Mac. See: https://github.com/pytorch/torchcodec/actions/runs/13924707065/job/38966320796?pr=564

We're unable to load the spec from the file on line 45. When digging around, it seems like we should be using `ctypes.CDLL` to load a shared library, and that's what `torch.ops.load_library()` does under the hood. Hence why I kept that in for the pybind ops.

TorchAudio does both: https://github.com/pytorch/audio/blob/c670ad81fda266b6598aeeef434583eb98197ae8/src/torio/_extension/utils.py#L105

The common case for pybind11 is that you have:

1. A shared library named `foo.so`.
2. A module declared as `foo` in the `PYBIND11_MODULE` macro in C++.
3. `foo.so` lives in a place Python knows to look.

Then `import foo` should just work.

src/torchcodec/decoders/_core/VideoDecoderOps.cpp
```diff
@@ -63,13 +65,14 @@ TORCH_LIBRARY(torchcodec_ns, m) {
...
 namespace {
+
 at::Tensor wrapDecoderPointerToTensor(
     std::unique_ptr<VideoDecoder> uniqueDecoder) {
   VideoDecoder* decoder = uniqueDecoder.release();
 
   auto deleter = [decoder](void*) { delete decoder; };
   at::Tensor tensor =
-      at::from_blob(decoder, {sizeof(VideoDecoder)}, deleter, {at::kLong});
+      at::from_blob(decoder, {sizeof(VideoDecoder*)}, deleter, {at::kLong});
```
scotts, line 75: Drive-by: we have been allocating enough space for an entire `VideoDecoder` _object_. But we're not storing the entire object, just a pointer.

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "src/torchcodec/decoders/_core/AVIOFileLikeContext.h"
+#include <torch/types.h>
+
+namespace facebook::torchcodec {
+
+AVIOFileLikeContext::AVIOFileLikeContext(py::object fileLike)
+    : fileLike_{UniquePyObject(new py::object(fileLike))} {
```
NicolasHug, line 13: I'm struggling to parse

```Cpp
new py::object(fileLike))
```

We're allocating a new `py::object` from an existing `py::object`, which is `fileLike`? What kind of copy is involved here (if any?)

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+}
+
+int AVIOFileLikeContext::read(void* opaque, uint8_t* buf, int buf_size) {
+  auto fileLike = static_cast<UniquePyObject*>(opaque);
+
+  int num_read = 0;
+  while (num_read < buf_size) {
+    int request = buf_size - num_read;
+    // TODO: It is maybe more efficient to grab the lock once in the
+    // surrounding scope?
```
NicolasHug, line 36: That's quite likely, IIRC when I was working with Cython stuff, acquiring / releasing the GIL wasn't trivially cheap

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+    if (chunk_len == 0) {
+      break;
+    }
+    TORCH_CHECK(
+        chunk_len <= request,
+        "Requested up to ",
+        request,
+        " bytes but, received ",
+        chunk_len,
+        " bytes. The given object does not confirm to read protocol of file object.");
```
NicolasHug, line 50:
```suggestion
        " bytes. The given object does not conform to read protocol of file object.");
```

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+  auto fileLike = static_cast<UniquePyObject*>(opaque);
+
+  int num_read = 0;
+  while (num_read < buf_size) {
+    int request = buf_size - num_read;
+    // TODO: It is maybe more efficient to grab the lock once in the
+    // surrounding scope?
+    py::gil_scoped_acquire gil;
+    auto chunk = static_cast<std::string>(
+        static_cast<py::bytes>((*fileLike)->attr("read")(request)));
```
NicolasHug, line 39: Can you help me understand why casting to `string`? Is it for convenience of keeping both the data and the size together?

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+        request,
+        " bytes but, received ",
+        chunk_len,
+        " bytes. The given object does not confirm to read protocol of file object.");
+    memcpy(buf, chunk.data(), chunk_len);
+    buf += chunk_len;
+    num_read += chunk_len;
+  }
+  return num_read == 0 ? AVERROR_EOF : num_read;
+}
```
NicolasHug, line 56: Above: this can come later, but we should definitely add a test on an object that can only return a small amount of bytes at once, so as to stress test the `while` logic above.

src/torchcodec/decoders/_core/AVIOBytesContext.cpp
```diff
@@ -0,0 +1,68 @@
...
+      ret = offset;
+      break;
+    default:
+      break;
+  }
+
+  return ret;
+}
+
+} // namespace facebook::torchcodec
```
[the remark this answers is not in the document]
scotts: I used to have no new files, but I ran into problems with testing. That is, I had `AVIOContextHolder` defined in `FFMPEGCommon.h`, `AVIOBytesContext` defined in `VideoDecoderOps.cpp` and `AVIOFileLikeContext` defined in `PyBindOps.cpp`. That ran into problems with testing in C++, as I needed access to the class declaration for `AVIOBytesContext` in `VideoDecoderTest.cpp`. Then I broke things down this way.

The reason I broke things out in this manner is that it's normal C++ convention to define separate files for classes, and to keep the implementation out of the header. In both instances, I'm trying to limit visibility; that is, the custom ops code doesn't need to know about `AVIOFileLikeContext`, and I think it's easier to reason about things when definitions are limited to only the scope they're needed.

I also pulled `AVIOContextHolder` out of `FFMPEGCommon.h` because it has significant logic unto itself. Right now, `FFMPEGCommon.h` is mostly about thin wrappers around FFmpeg stuff. 

In this breakdown, I think there's some clarify in the fact that `AVIOContextHolder` and the classes that derive it actually don't know about `VideoDecoder`.

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
scotts: Correct. We're invoking `py::object`'s [copy constructor](https://en.cppreference.com/w/cpp/language/copy_constructor) on `fileLike`. That resulting object is dynamically allocated on the heap, and its pointer is returned by `new`. The pointer to that object is then owned by `AVIOFileLikeContext::fileLike_`.

It may be instructive to contrast this to a stack allocated object, in statements that appear in normal code blocks:
```cpp
py::object allocatedOnStack(fileLike);
py::object* allocatedOnHeap = new py::object(fileLike);
```

src/torchcodec/decoders/_core/PyBindOps.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+namespace facebook::torchcodec {
+
+// In principle, this should be able to return a tensor. But when we try that,
+// we run into the bug reported here:
+//
+//   https://github.com/pytorch/pytorch/issues/136664
+//
+// So we instead launder the pointer through an int, and then use a conversion
+// function on the custom ops side to launder that int into a tensor.
```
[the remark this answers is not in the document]
scotts: Yes, exactly. I want this functionality to slot into our current APIs, not having a mirror set of APIs.

src/torchcodec/decoders/_core/ops.py
```diff
@@ -15,6 +18,8 @@
     _get_extension_path,
 )
 
+_pybind_ops: Optional[ModuleType] = None
+
 
 def load_torchcodec_extension():
```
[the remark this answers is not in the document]
scotts: I'd love to. :)

src/torchcodec/decoders/_core/ops.py
```diff
@@ -125,6 +158,11 @@ def create_from_tensor_abstract(
     return torch.empty([], dtype=torch.long)
 
 
+@register_fake("torchcodec_ns::_convert_to_tensor")
+def _convert_to_tensor_abstract(decoder_ptr: int) -> torch.Tensor:
+    return torch.empty([], dtype=torch.long)
```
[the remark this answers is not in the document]
scotts: It's needed because we can't handle that logic on the C++ side. That is, pybind11 should be able to just return a tensor. But when we do that, we run into a PyTorch core bug: https://github.com/pytorch/pytorch/issues/136664. In order to avoid that bug, we can't return a tensor from pybind11.

Our work-around, then, is to launder the pointer though an `int` on the pybind11 side, and then use a PyTorch custom op to launder that `int` into a tensor. We need to expose `_convert_to_tensor()` to Python because we put those calls together on the Python side. Notably, we never expose `_convert_to_tensor()` outside of `ops.py`. It's only used to implement `create_from_file_like()`.

test/decoders/test_ops.py
```diff
@@ -351,10 +355,18 @@ def test_create_decoder(self, create_from, device):
...
+        elif create_from == "bytes":
             with open(path, "rb") as f:
                 video_bytes = f.read()
             decoder = create_from_bytes(video_bytes)
+        elif create_from == "file_like_rawio":
+            decoder = create_from_file_like(open(path, mode="rb", buffering=0), "exact")
+        elif create_from == "file_like_bufferedio":
+            decoder = create_from_file_like(
+                open(path, mode="rb", buffering=-4096), "exact"
+            )
```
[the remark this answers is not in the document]
scotts: Whoops, I misread the docs. It should be positive.

src/torchcodec/decoders/_core/ops.py
```diff
@@ -26,19 +31,37 @@ def load_torchcodec_extension():
...
+        decoder_library_name = f"libtorchcodec_decoder{ffmpeg_major_version}"
+        custom_ops_library_name = f"libtorchcodec_custom_ops{ffmpeg_major_version}"
+        pybind_ops_library_name = f"libtorchcodec_pybind_ops{ffmpeg_major_version}"
         try:
-            torch.ops.load_library(_get_extension_path(library_name))
+            torch.ops.load_library(_get_extension_path(decoder_library_name))
+            torch.ops.load_library(_get_extension_path(custom_ops_library_name))
+
+            pybind_ops_library_path = _get_extension_path(pybind_ops_library_name)
+            torch.ops.load_library(pybind_ops_library_path)
```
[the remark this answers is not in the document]
scotts: Good question! I don't know. It's what TorchAudio does:

https://github.com/pytorch/audio/blob/c670ad81fda266b6598aeeef434583eb98197ae8/src/torio/_extension/utils.py#L101-L109

That is, it uses both `torch.ops.load_library()` and functions from `importlib`. I've gotten Linux working without the `torch.ops.load_library()` on the pybind library. What we have here was my last attempt at seeing what might work for Mac.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -7,58 +7,119 @@ find_package(Torch REQUIRED)
...
-    else()
-        list(APPEND sources CPUOnlyDevice.cpp)
-    endif()
-    add_library(${library_name} SHARED ${sources})
-    set_property(TARGET ${library_name} PROPERTY CXX_STANDARD 17)
+function(make_torchcodec_sublibrary
+    library_name
+    sources
+    dependent_libraries
+    ffmpeg_include_dirs)
```
NicolasHug, line 14: I could be wrong but it doesn't seem like we need to explicitly add `ffmpeg_include_dirs` anywhere. IIRC, the ffmpeg target we have are smart targets that will automatically add they INCLUDE dirs when calling `target_link_libraries`.

I tried removing those locally and things worked.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -7,58 +7,119 @@ find_package(Torch REQUIRED)
...
 
-    target_include_directories(
-        ${library_name}
+    add_library(${library_name} SHARED ${sources})
+    set_target_properties(${library_name} PROPERTIES CXX_STANDARD 17)
+    target_include_directories(${library_name}
         PRIVATE
         ./../../../../
         "${TORCH_INSTALL_PREFIX}/include"
         ${Python3_INCLUDE_DIRS}
```
NicolasHug, line 22: We probably don't need to worry about this (at least for now), but I suspect the `libtorchcodec_decoder?.so` libraries don't (or shouldn't) need the Python headers?

Similarly the pybind library probably doesn't need the torch headers.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -76,12 +137,10 @@ if(DEFINED ENV{BUILD_AGAINST_ALL_FFMPEG_FROM_S3})
...
-
-	make_torchcodec_library(libtorchcodec7 ffmpeg7)
-	make_torchcodec_library(libtorchcodec6 ffmpeg6)
-	make_torchcodec_library(libtorchcodec5 ffmpeg5)
-	make_torchcodec_library(libtorchcodec4 ffmpeg4)
-
+    make_torchcodec_libraries(7 ffmpeg7 $ffmpeg7_INCLUDE_DIRs)
+    make_torchcodec_libraries(6 ffmpeg6 $ffmpeg6_INCLUDE_DIRS)
+    make_torchcodec_libraries(4 ffmpeg4 $ffmpeg4_INCLUDE_DIRS)
+    make_torchcodec_libraries(5 ffmpeg5 $ffmpeg5_INCLUDE_DIRS)
```
NicolasHug, line 143: See comment about `ffmpeg_include_dirs`, I think this could just be:

```suggestion
    make_torchcodec_libraries(7 ffmpeg7 )
    make_torchcodec_libraries(6 ffmpeg6)
    make_torchcodec_libraries(4 ffmpeg4)
    make_torchcodec_libraries(5 ffmpeg5)
```

(same below with ${LIBAV_INCLUDE_DIRS}).

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -7,58 +7,119 @@ find_package(Torch REQUIRED)
...
-        list(APPEND sources CudaDevice.cpp)
-    else()
-        list(APPEND sources CPUOnlyDevice.cpp)
-    endif()
-    add_library(${library_name} SHARED ${sources})
-    set_property(TARGET ${library_name} PROPERTY CXX_STANDARD 17)
+function(make_torchcodec_sublibrary
+    library_name
+    sources
+    dependent_libraries
```
NicolasHug, line 13: Nit, I could be reading this the wrong way, but for me a "dependent library" is a **downstream** dependency, i.e. a library that depends on `library_name`. Here, we use "dependent library" to just mean "dependency", i.e. a library that `library_name` depends on?

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -7,58 +7,119 @@ find_package(Torch REQUIRED)
...
-    # We already set the library_name to be libtorchcodecN, so we don't want
-    # cmake to add another "lib" prefix. We do it this way because it makes it
-    # easier to find references to libtorchcodec in the code (e.g. via `git
-    # grep`)
-    set_target_properties(${library_name} PROPERTIES PREFIX "")
+function(make_torchcodec_libraries
+    ffmpeg_major_version
+    ffmpeg_target
+    ffmpeg_include_dirs)
+
```
NicolasHug, line 40: This can certainly be done in another PR but I think it would help to write a big comment to explain what we build, and why we have 3 different extensions. Maybe a link to your main comment on https://github.com/pytorch/torchcodec/pull/564/ is enough.

src/torchcodec/decoders/_core/AVIOFileLikeContext.h
```diff
@@ -0,0 +1,49 @@
...
+      if (obj) {
+        py::gil_scoped_acquire gil;
+        delete obj;
+      }
+    }
+  };
+
+  using UniquePyObject = std::unique_ptr<py::object, PyObjectDeleter>;
+  UniquePyObject fileLike_;
+};
```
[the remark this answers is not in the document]
scotts: Definitely, if it wasn't immediately obvious to you, we should document it.

src/torchcodec/decoders/_core/AVIOFileLikeContext.h
```diff
@@ -0,0 +1,49 @@
...
+  //
+  //   https://pybind11.readthedocs.io/en/stable/advanced/misc.html#common-sources-of-global-interpreter-lock-errors
+  struct PyObjectDeleter {
+    inline void operator()(py::object* obj) const {
+      if (obj) {
+        py::gil_scoped_acquire gil;
+        delete obj;
+      }
+    }
+  };
```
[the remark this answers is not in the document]
scotts: Actually _storing_ a `py::object` on the C++ side is not the common case. More common is that your C++ function receives a `py::object` through a pybind registered function, and you only do this inside of that pybind function. pybind guarantees that the GIL is held when it calls your C++ function.

Note that the PyTorch custom ops do _not_ hold the GIL, also on purpose. That's why we have to be GIL-aware in our code.

src/torchcodec/decoders/_core/AVIOContextHolder.h
```diff
@@ -0,0 +1,64 @@
...
+//        1. A read callback function.
+//        2. A seek callback function.
+//        3. A write callback function. (Not supported yet; it's for encoding.)
+//        4. A pointer to some context object that has the same lifetime as the
+//           AVIOContext itself. This context object holds the custom state that
+//           tracks the custom behavior of reading, seeking and writing. It is
+//           provided upon AVIOContext creation and to the read, seek and
+//           write callback functions.
+//      While it's not required, it is natural for the derived classes to make
+//      all of the above members. Base classes need to call
```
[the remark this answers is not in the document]
scotts: The functions could actually just be static functions defined in the .cpp file. They wouldn't have access to the private members of the derived class, although that also can be gotten around through the use of `friend`. As for the the state, technically it would also be defined statically in the .cpp file, but that's harder to get right as you want the state to have the same lifetime as the object itself. So, possible, just very tricky.

The point I'm making here is there is nothing in `AVIOContextHolder` that actually forces you to make the functions and state members of the deriving class. It's just really convenient to do so.

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
scotts: This is code copied from TorchAudio, and I thought it was necessary for that reason. But looking at the docs, I think we can just call a C++ version of `len()` on the returned `py::bytes`. I'll look into that.

scotts: Looking more: it's so that we have a pointer to the data that we can pass to `memcpy()`.

src/torchcodec/decoders/_core/CMakeLists.txt
scotts: If it's at all confusing, let's change the name. I think `library_dependencies` clears that up.

src/torchcodec/decoders/_core/CMakeLists.txt
scotts: Yup! Works for me too. I'm not sure what was going on when I added them. That will simplify the compilation as we don't need to pass as much info around.

src/torchcodec/decoders/_core/CMakeLists.txt
scotts: It's probably true that `libtorchcodec_decoder?.so` doesn't need the Python headers, but `libtorchcodec_pybind_ops?.so` does need the torch headers because it includes `VideoDecoder.h` which requires the torch headers.

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -3,62 +3,133 @@ project(TorchCodec)
...
 
-    target_include_directories(
-        ${library_name}
-        PRIVATE
-        ./../../../../
-        "${TORCH_INSTALL_PREFIX}/include"
-        ${Python3_INCLUDE_DIRS}
+    set(decoder_library_dependencies
+        ${ffmpeg_target}
+        ${TORCH_LIBRARIES}
```
scotts, line 59: Note that I took out `${Python3_LIBRARIES}` as a dependency. That caused problems with pybind11 importing on Mac: https://github.com/pybind/pybind11/issues/3907#issuecomment-1170412764

src/torchcodec/decoders/_core/ops.py
scotts: Works now! I'll make comments in the appropriate places.

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
NicolasHug: Shouldn't we be able to get both a pointer and the size from `py::bytes`? It might be simpler to keep `string`, but a small comment may help conveying the intent/reason.

src/torchcodec/decoders/_core/CMakeLists.txt
NicolasHug: IIUC the problem wasn't really with `libtorchcodec_decoder` (which shouldn't need Python anyway), it was with the `libtorchcodec_pybind_ops` library - which was pulling `libtorchcodec_decoder`?

src/torchcodec/decoders/_core/CMakeLists.txt
```diff
@@ -3,62 +3,157 @@ project(TorchCodec)
...
+    # 3. Create libtorchcodec_pybind_opsN.so.
+    set(pybind_ops_library_name "libtorchcodec_pybind_ops${ffmpeg_major_version}")
+    set(pybind_ops_sources
+        AVIOFileLikeContext.cpp
+        PyBindOps.cpp
+    )
+    set(pybind_ops_dependencies
+       ${decoder_library_name}
+       pybind11::module
+    )
```
NicolasHug, line 117: If my understanding is correct, maybe add a comment like:

> In theory we should also add Python3_LIBRARIES as a dependency, but this is causing issues on MacOS. See `undefined dynamic_lookup` below.

src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp
scotts: We actually can't get the pointer. We can get the size from all `py::object` instances, but there's nothing to access the underlying data pointer. (Other than going through the actual `PyObject` struct, but I'd rather stick to the pybind11 API.)

However! Other people have asked the same thing: https://github.com/pybind/pybind11/issues/2517, which lead to someone else implementing something just as good, which is getting a `std::string_view`: https://github.com/pybind/pybind11/issues/1807. That will avoid the copy and get what we need. Good catch!

src/torchcodec/decoders/_core/CMakeLists.txt
scotts: Yes, because `libtorchcodec_pybind_ops` depends on `libtorchcodec_decoder`, and `libtorchcodec_decoder` used to depend on `${Python3_LIBRARIES}`, we ran into problems at runtime.

Now we're good because `libtorchcodec_decoder` does not depend on `${Python3_LIBRARIES}`, `libtorchcodec_custom_ops` does, and `libtorchcodec_pybind_ops` does not (it gets what it needs through `pybind11::module`).

src/torchcodec/decoders/_core/AVIOBytesContext.cpp
```diff
@@ -0,0 +1,70 @@
...
+int AVIOBytesContext::read(void* opaque, uint8_t* buf, int buf_size) {
+  auto dataContext = static_cast<DataContext*>(opaque);
+  TORCH_CHECK(
+      dataContext->current <= dataContext->size,
+      "Tried to read outside of the buffer: current=",
+      dataContext->current,
+      ", size=",
+      dataContext->size);
+
+  int64_t numBytesRead = std::min(
```
scotts, line 29: Note that this avoids the negative buffer problem. Before, we were doing a narrowing conversion: we casted the `int64_t` value to an `int` in order to compare it to an `int`. That caused a large positive value to become a negative value. It's safer to cast the `int` to an `int64_t` and then compare both values as `int64_t`. This is fine because `std::memcpy()` takes a `size_t`, which the `int64_t` can safely convert to.

src/torchcodec/decoders/_core/ops.py
```diff
@@ -15,30 +18,61 @@
...
+    #
+    #   1. torch.ops.load_library(): For PyTorch custom ops. Loading libraries
+    #      through PyTorch registers the custom ops with PyTorch's runtime and
+    #      the ops can be accessed through torch.ops after loading.
+    #
+    #   2. importlib: For pybind11 modules. We load them dynamically, rather
+    #      than using a plain import statement. A plain import statement only
+    #      works when the module name and file name match exactly, and the
+    #      shared library file is in the import path. Our shared libraries do
+    #      not meet those conditions.
```
NicolasHug, line 43: Nit: I think we should remove this part from the sentence:

> and the shared library file is in the import path

Because the so/dylib files we produce are definitely in the "import path" (they're right next to other files we happily import). My point being that the above is *not* one of the conditions we fail to meet.

src/torchcodec/decoders/_core/ops.py
```diff
@@ -15,30 +18,61 @@
...
+    # libtorchcodec_*5.so, and libtorchcodec_*4.so. Each of these correspond to an
     # ffmpeg major version. This should cover all potential ffmpeg versions
     # installed on the user's machine.
     #
     # On fbcode, _get_extension_path() is overridden and directly points to the
     # correct .so file, so this for-loop succeeds on the first iteration.
+    #
+    # Note that we use two different methods for loading shared libraries:
+    #
+    #   1. torch.ops.load_library(): For PyTorch custom ops. Loading libraries
```
NicolasHug, line 35: Nit: not just custom ops, it's also for the "pure C++" decoder library

NicolasHug (approved):
Thanks a lot for the great work @scotts !

Since the PR is growing quite large, I'm happy to go ahead and merge this now if you'd like, and iterate on the tests as follow-ups.

NicolasHug (approved):
Latest changes LGTM still

PR #565, opened 2025-03-16
Fix first PTS bug

NicolasHug:
Fixes https://github.com/pytorch/torchcodec/issues/553.

The fix is just to set 

```C
streamInfo.codecContext->pkt_timebase = streamInfo.stream->time_base;
```

as is done in `ffprobe` (which, from https://github.com/pytorch/torchcodec/issues/553 , was correct): https://github.com/FFmpeg/FFmpeg/blob/1eafbf820312d45b31907e16877ae780022598c4/fftools/ffprobe.c#L3974

I looked and debugged with a source-build of `ffprobe` for a bit, trying to see if it was doing something clever with frames / packets pts. But I couldn't find anything. And the above is the only thing I could find that we weren't already doing with the AVCodecContext. So, while I don't necessarily understand *why* this is a fix, I'm fairly confident that this is indeed the right fix.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -467,6 +467,7 @@ void VideoDecoder::addStream(
   TORCH_CHECK_EQ(retVal, AVSUCCESS);
 
   streamInfo.codecContext->thread_count = ffmpegThreadCount.value_or(0);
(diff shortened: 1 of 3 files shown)


scotts (approved):

scotts:
Great fix! This must have also been a problem with video as well, but we never had any examples that exposed it.

PR #571, opened 2025-03-17
Simplify UniqueAVFrame handling in audio decoder

NicolasHug:
Follow-up to https://github.com/pytorch/torchcodec/pull/556#discussion_r1999237323.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -1361,14 +1361,8 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
       static_cast<AVSampleFormat>(avFrameStream.avFrame->format);
   AVSampleFormat desiredSampleFormat = AV_SAMPLE_FMT_FLTP;
 
-  UniqueAVFrame convertedAVFrame;
-  if (sourceSampleFormat != desiredSampleFormat) {
-    convertedAVFrame = convertAudioAVFrameSampleFormat(
-        avFrameStream.avFrame, sourceSampleFormat, desiredSampleFormat);
-  }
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1395,11 +1389,13 @@ void VideoDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
 
 UniqueAVFrame VideoDecoder::convertAudioAVFrameSampleFormat(
-    const UniqueAVFrame& avFrame,
+    UniqueAVFrame& avFrame,
     AVSampleFormat sourceSampleFormat,
-    AVSampleFormat desiredSampleFormat
+    AVSampleFormat desiredSampleFormat) {
+  if (sourceSampleFormat == desiredSampleFormat) {
+    return std::move(avFrame);
+  }
```
NicolasHug, line 1397: @scotts  WDYT - this is wrong, isn't it? This will end up destructing the `avFrameStream.avFrame` that was passed as a parameter to `convertAudioAVFrameToFrameOutputOnCPU`?

scotts: On this being wrong - I don't know, that depends on what we need for the calling side. What this will do is take ownership away from the input `avFrame` and give it to the output `UniqueAVFrame`. The question then is: is that okay for `avFrameStream` in the calling context? That is, is it okay for the `avFrameStream` in the calling context to be modified such that it no longer owns the `avFrame`? Because as a consequence of the move, `avFrameStream.avFrame` will become `nullptr`.

And if it's _not_ okay, then should `avFrameStream` in `convertAudioAVFrameToFrameOutputOnCPU()` actually be a const reference?

PR #572, opened 2025-03-17
Remove `AVFrameStream` struct

NicolasHug:
Continuing the work towards making  the decoder properly single-stream. This PR addresses https://github.com/pytorch/torchcodec/pull/483#discussion_r1930887776.

The `AVFrameStream` struct is obsolete, we don't need to keep track of the stream - we know it's `activeStreamIndex_`.

### DIFF
--- a/src/torchcodec/decoders/_core/VideoDecoder.cpp
+++ b/src/torchcodec/decoders/_core/VideoDecoder.cpp
@@ -583,9 +583,9 @@ VideoDecoder::FrameOutput VideoDecoder::getNextFrame() {
 VideoDecoder::FrameOutput VideoDecoder::getNextFrameInternal(
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
   validateActiveStream();
-  AVFrameStream avFrameStream = decodeAVFrame(
+  UniqueAVFrame avFrame = decodeAVFrame(
       [this](AVFrame* avFrame) { return avFrame->pts >= cursor_; });
-  return convertAVFrameToFrameOutput(avFrameStream, preAllocatedOutputTensor);
+  return convertAVFrameToFrameOutput(avFrame, preAllocatedOutputTensor);
 }
 
 VideoDecoder::FrameOutput VideoDecoder::getFrameAtIndex(int64_t frameIndex) {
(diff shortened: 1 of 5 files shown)


src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -1302,7 +1301,7 @@ void VideoDecoder::convertAVFrameToFrameOutputOnCPU(
 }
 
 int VideoDecoder::convertAVFrameToTensorUsingSwsScale(
-    const AVFrame* avFrame,
+    UniqueAVFrame& avFrame,
```
NicolasHug, line 1304: Unless I'm misunderstanding, pretending the `AVFrame*` was `const` was a big lie, because of the call to `sws_scale(..., avFrame->data, ...)` below. So I made the ref non-const.

scotts: I think it still makes sense for it to be const. We have a const reference to a unique pointer, which means that we can't change the contents of the unique pointer. That is, we can't call `release()` or `reset()` on the unique pointer. But we can make changes to the object the unique pointer points to.

This is similar to the difference between a "const pointer" and "pointer to const." A const pointer is a pointer that cannot change what it is pointing to. A pointer to const is a pointer to an object that cannot change - but you can change what the pointer points to.

Basically, anytime we don't want _ownership_ of the unique pointer to change, we should use const.

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -376,31 +359,33 @@ class VideoDecoder {
...
       FrameOutput& frameOutput,
       std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt);
 
-  torch::Tensor convertAVFrameToTensorUsingFilterGraph(const AVFrame* avFrame);
+  torch::Tensor convertAVFrameToTensorUsingFilterGraph(
+      const UniqueAVFrame& avFrame);
 
   int convertAVFrameToTensorUsingSwsScale(
-      const AVFrame* avFrame,
+      UniqueAVFrame& avFrame,
```
scotts, line 388: I think it's worth reasoning through if all of these should be const references - see my above.

scotts (approved):

PR #577, opened 2025-03-20
Remove `VideoStreamOptions` string constructor

NicolasHug:
The string constructor was only used in C++ tests. It's a lot of code for little benefit.

The ops were [already relying](https://github.com/pytorch/torchcodec/blob/ae19a7882752823e5cd9a8c580f01150dbc6e3ec/src/torchcodec/decoders/_core/VideoDecoderOps.cpp#L180-L218) on the (arguably cleaner) default constructor.

I checked and couldn't find any internal use of the string constructor either.


NicolasHug:
CI is red for unrelated reason. GPU and C++ tests are passing, so this PR is good to go.

scotts (approved):

PR #582, opened 2025-03-20
Add AudioDecoder docs and tutorial

NicolasHug:
This PR adds most docs for audio decoding, but more will come (like the metadata rendering stuff). This PR needs both https://github.com/pytorch/torchcodec/pull/584 and https://github.com/pytorch/torchcodec/pull/586 to be merged first.

Docs currently look like this:

Screenshot 2025-03-21 at 10-37-10 Decoding audio streams with AudioDecoder — TorchCodec 0 3 0a0 8ed45a7 Documentation


examples/audio_decoding.py
```diff
@@ -0,0 +1,111 @@
...
+# Here it's 0.025 seconds, even though we asked for samples starting from 0. Not
+# all streams start exactly at 0! This is not a bug in TorchCodec, this is a
+# property of the file that was defined when it was encoded.
+#
+# We only output the *start* of the samples, not the end or the duration. Those can
+# be easily derived from the number of samples and the sample rate:
+
+duration_seconds = samples.data.shape[1] / samples.sample_rate
+print(f"Duration = {int(duration_seconds // 60)}m{int(duration_seconds % 60)}s.")
+
```
NicolasHug, line 84: As I'm writing this, I feel like we should just set the `duration_seconds` field ourselves and make it part of `AudioSamples`? It's not like memory is a problem, and it would be consistent with video frames.

scotts: Yes, I think it would be useful. I was wondering the same thing when reading the tutorial.

scotts (approved):

PR #591, opened 2025-03-24
Move "scanned" metadata as video-only and add `begin_stream_from_header`

NicolasHug:
This PR does a bunch of related things:

- We add a new `begin_stream_seconds_from_header` field to audio and video metadata. It is derived from the AVStream's `start_time` field. This field is useful e.g. in our tutorial example because it is now clear that the stream doesn't exactly starts at 0.
- We move the `begin_stream_seconds_from_content` and `end_stream_seconds_from_content` fields as video-only: these fields are derived from a scan, so there's no point having them for audio.
- As a consequence of that, we move the `duration_seconds` metadata as video-only, because it is derived from `begin_stream_seconds_from_content` and `end_stream_seconds_from_content`. This means that audio metadata now only contains `duration_seconds_from_header`.
- We now allow negative `start_seconds` in the call to `get_samples_played_in_range`: the input check was based on `begin_stream_seconds_from_content` so that didn't make much sense. We still have tests that make sure we provide good error messages in the incorrect cases.

### DIFF
--- a/src/torchcodec/decoders/_audio_decoder.py
+++ b/src/torchcodec/decoders/_audio_decoder.py
@@ -11,7 +11,10 @@ from torch import Tensor
 
 from torchcodec import AudioSamples
 from torchcodec.decoders import _core as core
-from torchcodec.decoders._decoder_utils import create_decoder
+from torchcodec.decoders._decoder_utils import (
+    create_decoder,
+    ERROR_REPORTING_INSTRUCTIONS,
+)
 
 
 class AudioDecoder:
--- a/src/torchcodec/decoders/_audio_decoder.py
+++ b/src/torchcodec/decoders/_audio_decoder.py
@@ -11,10 +11,7 @@ from torch import Tensor
 
 from torchcodec import AudioSamples
 from torchcodec.decoders import _core as core
-from torchcodec.decoders._decoder_utils import (
-    create_decoder,
-    get_and_validate_stream_metadata,
-)
+from torchcodec.decoders._decoder_utils import create_decoder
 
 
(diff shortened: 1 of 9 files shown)


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -13,7 +13,7 @@
 from torchcodec.decoders import _core as core
 from torchcodec.decoders._decoder_utils import (
     create_decoder,
-    get_and_validate_stream_metadata,
+    ERROR_REPORTING_INSTRUCTIONS,
```
NicolasHug, line 16: Changes below: we previously had a common util for audio and video that extracted and validated the metadata: `get_and_validate_stream_metadata()`. Since we moved a bunch of fields as video-only, this util wasn't generic enough anymore to justify its existence, hence a few edits here and in the `video_decoder.py` file.

src/torchcodec/decoders/_core/VideoDecoderOps.cpp
```diff
@@ -473,6 +473,10 @@ std::string get_stream_json_metadata(
   if (streamMetadata.numFrames.has_value()) {
     map["numFrames"] = std::to_string(*streamMetadata.numFrames);
   }
+  if (streamMetadata.beginStreamFromHeader.has_value()) {
+    map["beginStreamFromHeader"] =
+        std::to_string(*streamMetadata.beginStreamFromHeader);
+  }
```
scotts, line 479: This reminded me to create Issue #593.

scotts (approved):

PR #594, opened 2025-03-24
Add` get_all_samples() `method to `AudioDecoder`

NicolasHug:
The most common decoding pattern for audio will be to decode an entire stream at once. Right now, the main way to do this is to call `decoder.get_samples_played_in_range()` which is a awkward:

- it's a long name
- we're not relying on the `range` feature

This PR introduces the `decoder.get_all_samples()` method which is a more natural alias, hopefully providing a better UX for the most common usage.

### DIFF
--- a/src/torchcodec/decoders/_audio_decoder.py
+++ b/src/torchcodec/decoders/_audio_decoder.py
@@ -78,6 +78,9 @@ class AudioDecoder:
     def get_all_samples(self) -> AudioSamples:
         """Returns all the audio samples from the source.
 
+        To decode samples in a specific range, use
+        :meth:`~torchcodec.decoders.AudioDecoder.get_samples_played_in_range`.
+
         Returns:
             AudioSamples: The samples within the file.
         """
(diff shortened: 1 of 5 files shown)


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -75,6 +75,14 @@ def __init__(
             sample_rate if sample_rate is not None else self.metadata.sample_rate
         )
 
+    def get_all_samples(self) -> AudioSamples:
+        """Returns all the audio samples from the source.
+
+        Returns:
+            AudioSamples: The samples within the file.
+        """
+        return self.get_samples_played_in_range()
```
scotts, line 84: Can we add a brief comment explaining that the defaults for `get_samples_played_in_range()` do what we want, and that this method is a way of advertising that to folks?

scotts (approved):

src/torchcodec/decoders/_audio_decoder.py
NicolasHug: Absolutely, do you want it as a comment or in the docstring?

PR #596, opened 2025-03-25
Do nullptr check on result from av_get_sample_fmt_name()

scotts:
We need to check the result of [`av_get_sample_fmt_name()`](https://ffmpeg.org/doxygen/6.1/group__lavu__sampfmts.html#ga31b9d149b2de9821a65f4f5612970838) because it can return `nullptr`.


jiangyurong609 (approved):

NicolasHug (approved):
Thanks for the fix!

Issue #605, opened 2025-03-28
Make C++ level device interface generic (device agnostic)


dvrogozh:
See: [src/torchcodec/decoders/_core/DeviceInterface.h](https://github.com/pytorch/torchcodec/blob/b83048001d5ae9f3aa7024a091acf78a7be12a0a/src/torchcodec/decoders/_core/DeviceInterface.h)

Currently defined device interface is CUDA specific by naming and usage. As discussed in https://github.com/pytorch/torchcodec/issues/559#issuecomment-2744913703 this complicates adding support for other device backends. It does seem a right step to refactor device interface, make it generic and allow easy registration of non-CUDA device specific backends. Creating this issue to track that.

CC: @scotts, @NicolasHug

PR #606, opened 2025-03-28
Make device interface generic

dvrogozh:
Fixes: #605

Changes:
* Device interface made device agnostic by intorducing `class DeviceInterface` from which specific backends should inherit their device specific implementations
* Implemented `CudaDevice` derived from `DeviceInterface`
* Created device interface registration mechanism (`registerDeviceInterface`)
* Created device interface creation mechanism (`createDeviceInterface`)

These changes allow to replace CUDA specific code in `VideoDecoder.cpp` and `VideoDecoderOps.cpp` by device agnostic code.


src/torchcodec/decoders/_core/DeviceInterface.h
```diff
@@ -23,25 +24,41 @@ namespace facebook::torchcodec {
...
-    std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt);
-
-void releaseContextOnCuda(
-    const torch::Device& device,
-    AVCodecContext* codecContext);
-
-std::optional<const AVCodec*> findCudaCodec(
-    const torch::Device& device,
-    const AVCodecID& codecId);
+struct DeviceInterface {
```
scotts, line 27: We should prefer `class` over `struct`. The structs we have are more data-holders rather than propers objects with a well defined interface. Since the whole point of this is the well-defined interface, `DeviceInterface` and everything that implements it should be a class.

src/torchcodec/decoders/_core/DeviceInterface.h
```diff
@@ -23,25 +24,41 @@ namespace facebook::torchcodec {
...
+ protected:
+  torch::Device device_;
+};
+
+using CreateDeviceInterfaceFn =
+    std::function<DeviceInterface*(const std::string& device)>;
+
+bool registerDeviceInterface(
+    const std::string device_type,
+    const CreateDeviceInterfaceFn create_interface);
```
scotts, line 59: Coding style: in this project, C++ code follows `camelCase` rather than `snake_case`. The few exceptions are when we have C++ code that exactly mirrors a Python function, where we keep the C++ names the same as the Python names.

src/torchcodec/decoders/_core/DeviceInterface.h
```diff
@@ -23,25 +24,41 @@ namespace facebook::torchcodec {
...
+  // support CUDA and others only support CPU.
+  virtual void initializeContext(AVCodecContext* codecContext) = 0;
+
+  virtual void convertAVFrameToFrameOutput(
+      const VideoDecoder::VideoStreamOptions& videoStreamOptions,
+      UniqueAVFrame& avFrame,
+      VideoDecoder::FrameOutput& frameOutput,
+      std::optional<torch::Tensor> preAllocatedOutputTensor = std::nullopt) = 0;
+
+  virtual void releaseContext(AVCodecContext* codecContext) = 0;
```
scotts, line 48: Do we need this to be a named member function? That is, isn't it appropriate for this to just happen in the destructor?

dvrogozh: Note that as currently defined `DeviceInterface` does not store `AVCodecContext`. It merely initializes it on a call to `initializeContext()`. i.e. `AVCodecContext* codecContext` argument in this method is both input and output one.

Note also that not storying of `AVCodecContext` is convenient if we potentially want same `DeviceInterface` capable to handle multiple contexts at once. Do we?

src/torchcodec/decoders/_core/VideoDecoder.h
```diff
@@ -138,7 +140,7 @@ class VideoDecoder {
     std::optional<int> height;
     std::optional<ColorConversionLibrary> colorConversionLibrary;
     // By default we use CPU for decoding for both C++ and python users.
-    torch::Device device = torch::kCPU;
+    std::shared_ptr<DeviceInterface> device;
```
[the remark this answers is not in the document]
dvrogozh: Some video clips might contain multiple video streams (multiple audio streams, subtitles, etc.). As of now `DeviceInterface` is associated with a single stream and allows to potentially process different streams on different accelerators. Did you consider to support such scenario in the future? /We still can handle that even we will move `DeviceInterface` to `VideoDecoder` storing it in a vector or map indexed by streams. That is design decision to make./

src/torchcodec/decoders/_core/DeviceInterface.cpp
```diff
@@ -0,0 +1,58 @@
...
+  std::scoped_lock lock(g_interface_mutex);
+  TORCH_CHECK(
+      g_interface_map.find(device_type) == g_interface_map.end(),
+      "Device interface already registered for ",
+      device_type);
+  g_interface_map.insert({device_type, create_interface});
+  return true;
+}
+
+std::shared_ptr<DeviceInterface> createDeviceInterface(
```
scotts, line 39: Here are in other places where we deal with this return value, we want to use `unique_ptr`. We want to enforce that only one entity owns the device at any one time. Unfortunately, this does mean we have to use ugly `std::move()` in a bunch of places.

src/torchcodec/decoders/_core/VideoDecoder.cpp
```diff
@@ -391,7 +388,7 @@ torch::Tensor VideoDecoder::getKeyFrameIndices() {
 void VideoDecoder::addStream(
     int streamIndex,
     AVMediaType mediaType,
-    const torch::Device& device,
+    DeviceInterface* device,
```
scotts, line 391: I think it's easier if `addStream()` still accepts a `const torch::Device&`. That means that `addStream()` would be responsible for calling `createDeviceInterface()`, which I think is fine. That would keep our concept of devices as an implementation detail for `VideoDecoder`. From the outside, we accept a `torch::Device` and internally we create the corresponding special logic for decoding on that device.

scotts:
@dvrogozh, this is great progress! I'd also like to see all of the `VideoDecoder::*OnCPU()` functions moved into a `CpuDevice` class. A bunch of the initialize and release member functions will be no-ops, and that's fine. That way it's clear what is device-agnostic code and what is device-specific code. I think that will also clean-up the dispatch logic inside of `VideoDecoder`.

scotts (changes requested):
@dvrogozh, this is great progress! I'd also like to see all of the `VideoDecoder::*OnCPU()` functions moved into a `CpuDevice` class. A bunch of the initialize and release member functions will be no-ops, and that's fine. That way it's clear what is device-agnostic code and what is device-specific code. I think that will also clean-up the dispatch logic inside of `VideoDecoder`.

dvrogozh:
> I'd also like to see all of the VideoDecoder::*OnCPU() functions moved into a CpuDevice class.

@scotts : I thought to do that in a follow up PR. Do you want me to do that in a single shot now?

src/torchcodec/decoders/_core/DeviceInterface.h
dvrogozh: done

src/torchcodec/decoders/_core/DeviceInterface.h
dvrogozh: done. Hopefully everywhere in new changes.

src/torchcodec/decoders/_core/DeviceInterface.h
dvrogozh: On a second thought, I removed `releaseContext`. Actually I overlooked that you have `StreamInfo` where you store stream specific stuff. So, couple changes laid quite nicely after I released that. Including no need for `releaseContext` and `std::unique_ptr` for `DeviceInterface`.

src/torchcodec/decoders/_core/VideoDecoder.cpp
dvrogozh: done. Note that this required few other changes for `DeviceInterface`. For example, I switch from `std::string` to `torch::DeviceType` in device interface registration. This also required to introduce a new `createTorchDevice` to be able to filter out unsupported devices against registered device interfaces early.

dvrogozh:
Rebased on top of merged #609 and #611.

dvrogozh:
@scotts : can you, please, review again?

src/torchcodec/_core/CudaDevice.cpp
```diff
@@ -158,39 +161,34 @@ AVBufferRef* getCudaContext(const torch::Device& device) {
...
+
   // It is important for pytorch itself to create the cuda context. If ffmpeg
   // creates the context it may not be compatible with pytorch.
   // This is a dummy tensor to initialize the cuda context.
   torch::Tensor dummyTensorForCudaInitialization = torch::empty(
-      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device));
-  codecContext->hw_device_ctx = getCudaContext(device);
+      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device_));
+  ctx_ = getCudaContext(device_);
+  codecContext->hw_device_ctx = av_buffer_ref(ctx_);
```
scotts, line 187: Interesting - we weren't previously calling `av_buffer_ref()` (https://www.ffmpeg.org/doxygen/2.5/group__lavu__buffer.html#gaa40ce7d3ede946a89d03323bbd7268c1) in this code. Did we perhaps have a bug previously, in that we were not properly tracking new references?

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+torch::Device createTorchDevice(const std::string device) {
+  // TODO: remove once DeviceInterface for CPU is implemented
+  if (device == "cpu") {
+    return torch::kCPU;
+  }
+
+  std::scoped_lock lock(g_interface_mutex);
+  std::string deviceType = getDeviceType(device);
+  TORCH_CHECK(
```
scotts, line 47: I think this code will be more clear if we don't embed the logic inside the `TORCH_CHECK()`. That is, do the search and only check the return value inside the `TORCH_CHECK()`.

src/torchcodec/_core/CudaDevice.cpp
dvrogozh: The reason that `av_buffer_ref()` is needed here is a change in `addToCacheIfCacheHasCapacity` signature: `AVCodecContext* codecContext` replaced by `AVBufferRef* hwContext` and a drop of `codecContext->hw_device_ctx = nullptr`.  There are at least 2 places which reference hw device context:
1. `CudaDevice`
2. FFmpeg's `AVCodecContext`

`AVCodecContext` uses ref/unref to work with the device context, but `CudaDevice` does not do that though formally it should have called `ref/unref` as well. However, previously that was balanced by calling `codecContext->hw_device_ctx = nullptr` right before destroying `AVCodecContext`, i.e. you have "stolen"  the reference from the `AVCodecContext` and avoid ffmpeg code to unreference the context.

Overall, there are 2 things to fix here:
1. Add ref for device context in `CudaDevice` and stop nulling the ctx pointer in `AVCodecContext` - that's what I fix in this PR
2. Another fix needed is adding `av_buffer_unref` on the cache overflow - I will submit a separate PR for that:
https://github.com/pytorch/torchcodec/blob/e611e2985ac9c0486438c7908724f4d269b92b09/src/torchcodec/_core/CudaDevice.cpp#L58-L61

dvrogozh: > Another fix needed is adding `av_buffer_unref` on the cache overflow - I will submit a separate PR for that

Actually, existing code will handle this correctly - `codecContext->hw_device_ctx = nullptr` won't be called and device context will be freed during release of `AVCodecContext`. But I need to add `av_buffer_unref` in my PR to balance +1 reference I've added.

src/torchcodec/decoders/_core/VideoDecoder.h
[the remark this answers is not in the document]
scotts: The current C++ decoder was originally implemented to support decoding from multiple streams at once, hence why we have methods for adding a stream and tracking info per-stream. But, that functionality never actually worked correctly. Getting it to work correctly would have required a lot more changes and testing, and we only expose single-stream decoding in our public API. So we instead decided to just make it explicit: the C++ decoder only works on one stream at a time. That's more explicit now in the new name.

We actually have a TODO to move some of the current fields of `StreamInfo` into `SingleStreamDecoder` itself for this reason. The device interface is another field that belongs in `SingleStreamDecoder` itself rather than in `StreamInfo`. In general, we're trying to keep the code as simple as possible, which means not keeping things around that we "might need" in the future. If we need it in the future, we'll implement it then. And if we want to implement mulistream decoding, we may end up implementing an entirely new C++ class.

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -357,6 +356,8 @@ class SingleStreamDecoder {
     // Used to know whether a new FilterGraphContext or UniqueSwsContext should
     // be created before decoding a new frame.
     DecodedFrameContext prevFrameContext;
+
+    std::unique_ptr<DeviceInterface> deviceInterface;
```
scotts, line 360: Let's move this to be a member of `SingleStreamDecoder` based on my previous comment about this class only supporting single stream decoding.

scotts:
@dvrogozh, this is looking great! Let's make that one change to where the device lives, and also figure out why the C++ tests are failing.

scotts:
Also: making a proper CPU device is fine for a follow-up PR, there's a bunch of changes already in this one, and it's self-consistent.

src/torchcodec/_core/CudaDevice.cpp
dvrogozh: Fixed (see `addToCacheIfCacheHasCapacity` and `CudaDevice` destructor).

src/torchcodec/_core/DeviceInterface.cpp
dvrogozh: done

src/torchcodec/_core/SingleStreamDecoder.h
dvrogozh: done

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -493,6 +492,7 @@ class SingleStreamDecoder {
   SeekMode seekMode_;
   ContainerMetadata containerMetadata_;
   UniqueDecodingAVFormatContext formatContext_;
+  std::unique_ptr<DeviceInterface> deviceInterface;
```
scotts, line 495: Class variable members end with an underscore (`foo_`).

scotts:
Agreed on moving `SingleStreamDecoder::VideoStreamOptions` and `SingleStreamDecoder::FrameOutput` in separate, new headers in a follow-up PR. Let's put `VideoStreamOptions` in its own header, and then the entire family of `FrameOutput`, `FrameBatchOutput` and `AudioFramesOutput` into one new header.

This looks great, thanks for making TorchCodec more general for devices!

scotts (approved):

scotts:
@dvrogozh, I'm merging as-is, you can address the class member name in the follow-up PRs.

src/torchcodec/_core/SingleStreamDecoder.h
dvrogozh: ah, I overlooked moving it from the struct. Will fix in a follow up PR.

dvrogozh:
> Agreed on moving `SingleStreamDecoder::VideoStreamOptions` <...>

Filed to track:
* https://github.com/pytorch/torchcodec/issues/618

> Making a proper CPU device is fine for a follow-up PR

Filed to track:
* https://github.com/pytorch/torchcodec/issues/619

src/torchcodec/_core/CudaDevice.h
```diff
@@ -0,0 +1,34 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include "src/torchcodec/_core/DeviceInterface.h"
+
+namespace facebook::torchcodec {
+
+class CudaDevice : public DeviceInterface {
```
NicolasHug, line 13: The term "device" should be reserved for the concept of pytorch device object. Here, the `CudaDevice` class name is misleading because it doesn't refer to a device, it refers to a device **interface**. @dvrogozh , do you mind submitting a PR to rename CudaDevice into `CudaDeviceInterface` (file names and file classes)?

dvrogozh: @NicolasHug : no problem, will do.

dvrogozh: @NicolasHug : filed #626

PR #607, opened 2025-03-29
Fix VideoDecoder device argument to accept torch.device

### LINKED ISSUE #602: `VideoDecoder` API accepts `device` argument as `str` or `torch.device`, but underlying C++ API `add_video_stream` only takes in a `str`.
### 🐛 Describe the bug

Repro:

```
device = torch.device('cuda:0')
VideoDecoder(video, device=device)
``` 

fails with complaints that 
```
trainer/0 [0]: Declaration: torchcodec_ns::add_video_stream(Tensor(a!) decoder, *, int? width=None, int? height=None, int? num_threads=None, str? dimension_order=None, int? stream_index=None, str? device=None) -> ()
trainer/0 [0]:Cast error details: Unable to cast Python instance of type <class 'torch.device'> to C++ type '?' (#define PYBIND11_DETAILED_ERROR_MESSAGES or compile in debug mode for details)
trainer/0 [0]:Skipping video tensor([ 0,  0,  0,  ..., 49, 48, 48], dtype=torch.uint8) due to decoding error: torchcodec_ns::add_video_stream() Expected a value of type 'Optional[str]' for argument 'device' but instead found type 'device'.
```

`VideoDecoder(video, device=str(self.device))` fixes the issue.

### Versions

N/A

soheil-star01:
### Summary

This PR fixes an issue where passing a `torch.device` to the `VideoDecoder` constructor causes a type error when passed to the underlying C++ `add_video_stream` function, which expects a `str`.

### Changes

- Convert `torch.device` to `str` inside `__init__` if necessary.

### Why

Fixes bug #602: `torch.device` instances were not compatible with the expected C++ type (`Optional[str]`), causing decoding to fail.

Closes #602.

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -8,7 +8,7 @@ import numbers
 from pathlib import Path
 from typing import Literal, Optional, Tuple, Union
 
-from torch import device, Tensor
+from torch import device as torch_device, Tensor
 
 from torchcodec import Frame, FrameBatch
 from torchcodec.decoders import _core as core
@@ -72,7 +72,7 @@ class VideoDecoder:
         stream_index: Optional[int] = None,
         dimension_order: Literal["NCHW", "NHWC"] = "NCHW",
         num_ffmpeg_threads: int = 1,
-        device: Optional[Union[str, device]] = "cpu",
+        device: Optional[Union[str, torch_device]] = "cpu",
         seek_mode: Literal["exact", "approximate"] = "exact",
     ):
         allowed_seek_modes = ("exact", "approximate")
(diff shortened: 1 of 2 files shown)


NicolasHug (commented):
Thank you for the PR @soheil-star01 . Changes look good, before we merge could you please write a small non-regression test in https://github.com/pytorch/torchcodec/blob/main/test/decoders/test_decoders.py, to make sure that passing a `torch.device()` now works? Something as simple as this should be enough:

```py
def test_device_instance(self):
	# Non-regression test for https://github.com/pytorch/torchcodec/issues/602	
	VideoDecoder(path, device=torch.device("cpu"))
```

soheil-star01:
> Thank you for the PR @soheil-star01 . Changes look good, before we merge could you please write a small non-regression test in https://github.com/pytorch/torchcodec/blob/main/test/decoders/test_decoders.py, to make sure that passing a `torch.device()` now works? Something as simple as this should be enough:
> 
> ```python
> def test_device_instance(self):
> 	# Non-regression test for https://github.com/pytorch/torchcodec/issues/602	
> 	VideoDecoder(path, device=torch.device("cpu"))
> ```

@NicolasHug Thank you. Done!

NicolasHug (approved):
Thank you for the PR @soheil-star01 !

PR #609, opened 2025-04-01
Move `torchcodec/decoders/_core` into `torchcodec/_core`

NicolasHug:
The `_core` folder currently contains all the C++ code, the metadata and the custom ops. Part of it will also be relevant for encoding (typically FFmpeg utils, and potentially sample format conversion). As a result, this PR moves the `_core` folder away from `decoders/`.

Public APIs and import remain unchanged

There will be 2 follow-ups:

- the `test/` structure should be updated
- the "VideoDecoder" C++  class and file names should be renamed into just "Decoder", since they are now relevant for both video and audio.

Heads up @dvrogozh that this PR will likely create merge conflicts with https://github.com/pytorch/torchcodec/pull/606, but they should be straightforward to resolve.


dvrogozh (approved):
Thank you for heads up, @NicolasHug. Rebase of #606 should not be a problem.

PR #610, opened 2025-04-01
Add ops test for file-like seeks

scotts:
Additional test for file-like behavior; see test comments for more.

### DIFF
--- a/test/decoders/test_ops.py
+++ b/test/decoders/test_ops.py
@@ -4,6 +4,7 @@
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
 
+import io
 import os
 from functools import partial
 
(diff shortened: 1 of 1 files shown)


test/decoders/test_ops.py
```diff
@@ -922,6 +923,76 @@ def get_all_frames(asset, sample_rate=None, stop_seconds=None):
...
+        # Test to ensure that seeks and reads are actually going through the
+        # methods on the IO object.
+        #
+        # Note that we do not check the number of reads in this test past the
+        # initialization step. That is because the number of reads that FFmpeg
+        # issues is dependent on the size of the internal buffer, the amount of
+        # data per frame and the size of the video file. We can't control
+        # the size of the buffer from the Python layer and we don't know the
+        # amount of data per frame. We also can't know the amount of data per
+        # frame from first principles, because it is data-depenent.
```
NicolasHug, line 938: Not sure if that would help, but we could store the number of samples of each frames in the .json files that we check in

NicolasHug (approved):

test/decoders/test_ops.py
```diff
@@ -922,6 +923,76 @@ def get_all_frames(asset, sample_rate=None, stop_seconds=None):
...
+                self.num_seeks = 0
+                self.num_reads = 0
+
+            def read(self, size: int) -> bytes:
+                self.num_reads += 1
+                return self._file.read(size)
+
+            def seek(self, offset: int, whence: int) -> bytes:
+                self.num_seeks += 1
+                return self._file.seek(offset, whence)
```
NicolasHug, line 952: Just noting this so we don't forget, we should test for proper errors if the user provides a malformed object that e.g. doesn't support read() or seek() correctly. E.g. if `read()` doesn't exist, right now we only error:

```py
    def create_from_file_like(
        file_like: Union[io.RawIOBase, io.BytesIO], seek_mode: Optional[str] = None
    ) -> torch.Tensor:
        assert _pybind_ops is not None
>       return _convert_to_tensor(_pybind_ops.create_from_file_like(file_like, seek_mode))
E       NotImplementedError
```

which isn't super informative. I guess the check can be done on the public Pythons side rather than on the ops?


Also, we should try to hit some of the error checks, e.g. catching in the tests that we properly raise https://github.com/pytorch/torchcodec/blob/f416dcf8cb709d679cd739d2a56a761d8658468f/src/torchcodec/decoders/_core/AVIOFileLikeContext.cpp#L53-L59 when doing something bad like

```py
return self._file.read(size + 10)
```

Issue #618, opened 2025-04-04
Move stream options and frame output structs to dedicated header(s)


dvrogozh:
As discussed in:
* https://github.com/pytorch/torchcodec/pull/606#issuecomment-2779316247
* https://github.com/pytorch/torchcodec/pull/606#issuecomment-2779571470

A follow up after #606 - need to move the following structures to separate header(s) and break circular dependency between `DeviceInterface.h` and `SingleStreamDecoder.h`:
* `SingleStreamDecoder::VideoStreamOptions`
* `SingleStreamDecoder::FrameOutput`
* `SingleStreamDecoder::FrameBatchOutput`
* `SingleStreamDecoder::AudioFramesOutput`

CC: @scotts, @NicolasHug

dvrogozh:
See #620 for the proposed fix.

PR #620, opened 2025-04-04
Move stream options and frame output structs to dedicated headers

dvrogozh:
Fixes: #618 

Commit moves stream options and frame output structs to dedicated headers:
* `Frame.h` now defines `FrameOutput`, `FrameBatchOutput` and `AudioFrameOutput`
* `Stream.h` now defines `VideoStreamOptions`, `AudioStreamOptions`, `StreamMetadata` and `ContainerMetadata`

Commit also resolves circular dependency between `DeviceInterface.h` and `SingleStreamDecoder.h`, forward declaration of `class DeviceInterface` no longer needed.

CC: @scotts, @NicolasHug


src/torchcodec/_core/Stream.h
```diff
@@ -0,0 +1,99 @@
...
+  // By default we use CPU for decoding for both C++ and python users.
+  torch::Device device = torch::kCPU;
+};
+
+struct AudioStreamOptions {
+  AudioStreamOptions() {}
+
+  std::optional<int> sampleRate;
+};
+
```
NicolasHug, line 48: Thanks for the PR @dvrogozh . Was it necessary to move the StreamMetadata and ContainerMetadata classes into this new header?
If yes, then I'd suggest moving them into their own `metadata.h` header
If not, then it might be best to leave them in the SingleStreamDecoder files

In both cases, I'd also suggest to rename Stream.h into just StreamOptions.h since this file should only contain option-related structs.


Otherwise, this PR LGTM

dvrogozh: In the last update I've pushed:
* `Stream.h` renamed to `StreamOptions.h`
* Created `Medata.h` to hold `StreamMetadata` and `ContainerMetadata`

@NicolasHug , @scotts : The reason why I moved metadata to header was that `StreamMetadata` is used by `FrameBatchOutput` which I also moved:

https://github.com/pytorch/torchcodec/blob/8b19f455046837f572490fe403918b57f7c1989a/src/torchcodec/_core/SingleStreamDecoder.h#L181-L184

That being said, note that I moved `ContainerMetadata` only for consistency - it's not used by any of the structures we really need to be in headers. Let me know if you want to avoid moving `ContainerMetadata`. Also, if we don't want to separately expose `FrameBatchOutput` we can avoid moving metadata to separate header entirely.

scotts (approved):

src/torchcodec/_core/Stream.h
scotts: @dvrogozh, looks good, thanks for cleaning this up!

PR #623, opened 2025-04-07
Add bit_rate parameter to encoder

NicolasHug:


test/test_ops.py
```diff
@@ -1127,28 +1134,26 @@ def test_round_trip(self, tmp_path):
         )
         encode_audio(encoder)
 
-        # TODO-ENCODING: tol should be stricter. We need to increase the encoded
-        # bitrate, and / or encode into a lossless format.
+        # TODO-ENCODING: tol should be stricter. We probably need to encode
+        # into a lossless format.
```
NicolasHug, line 1138: I tried to pass bit_rate=4000 and was able to decrease tolerance from 0.07 to 0.05, but that's clearly not strict enough still.

scotts (approved):

PR #626, opened 2025-04-07
Rename CudaDevice to CudaDeviceInterface

dvrogozh:
Renaming as requested in:
* https://github.com/pytorch/torchcodec/pull/606#discussion_r2030880769

CC: @NicolasHug, @scotts


NicolasHug (approved):
Thanks for following up @dvrogozh !

PR #629, opened 2025-04-09
Move sample rate and sample format conversion utils into `FFMPEGCommon.cpp`

NicolasHug:
This PR moves the sample rate and sample format conversions utils from SingleStreamDecoder into FFMPEGCommon. Sample format conversion is needed for encoding too, so we need to make them common.

Specifically:

- `SingleStreamDecoder::createSwrContext` is removed and its logic is not part of FFMPEGCommon's `allocateSwrContext`, which  was renamed into `createSwrContext`
- `SingleStreamDecoder::convertAudioAVFrameSampleFormatAndSampleRate` is moved too

Inputs are slightly modified to account for the fact that there's no `streamInfo_` anymore. Other than that, this is just copy/pasting code.


src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -155,9 +156,77 @@ SwrContext* allocateSwrContext(
...
+      getFFMPEGErrorStringFromErrorCode(status),
+      ". If the error says 'Invalid argument', it's likely that you are using "
+      "a buggy FFmpeg version. FFmpeg4 is known to fail here in some "
+      "valid scenarios. Try to upgrade FFmpeg?");
   return swrContext;
 }
 
+UniqueAVFrame convertAudioAVFrameSampleFormatAndSampleRate(
+    const UniqueSwrContext& swrContext,
+    const UniqueAVFrame& srcAVFrame,
```
scotts, line 172: Nit: we should be consistent about `src` and `source`. I have a preference for `src`, as it's a universal abbreviation, particularly when paired with `dst`. But if we say `source` in a lot of other places, we should stick with that.

scotts: Double nit, and I recognize this name already existed: `convertAudioAVFrameSampleFormatAndSampleRate()` is very long, and I feel like we're encoding parameter names that modify the operation into the name. I feel like it's clearer as just `convertAudioAVFrame()`.

scotts (approved):

src/torchcodec/_core/FFMPEGCommon.cpp
NicolasHug: Sounds good, I'll merge as-is and follow-up with a PR to address these

PR #630, opened 2025-04-09
Encoding: support wav, flac etc.

NicolasHug:
This PR adds support for encoding formats where the encoder doesn't natively support FLTP, which is the sample format of the input waveform. We use `swresample` to convert the input FLTP AVFrames into a sample format that the encoder support.


src/torchcodec/_core/Encoder.cpp
```diff
@@ -92,14 +92,13 @@ AudioEncoder::AudioEncoder(
   validateSampleRate(*avCodec, sampleRate);
   avCodecContext_->sample_rate = sampleRate;
 
-  // Note: This is the format of the **input** waveform. This doesn't determine
-  // the output.
```
NicolasHug, line 96: My original comment was wrong: that's not the format of the input waveform. It's the format of the input AVFrame that we pass to `avcodec_send_frame()`. And it needs to be a format that the codec supports.

src/torchcodec/_core/Encoder.h
```diff
@@ -26,12 +26,14 @@ class AudioEncoder {
  private:
   void encodeInnerLoop(
       AutoAVPacket& autoAVPacket,
-      const UniqueAVFrame& avFrame);
+      const UniqueAVFrame& srcAVFrame);
   void flushBuffers();
+  AVSampleFormat findOutputSampleFormat(const AVCodec& avCodec);
```
NicolasHug, line 31: maybe this shouldn't be a method and just a function in a private namespace?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -140,11 +132,50 @@ AudioEncoder::AudioEncoder(
...
+  // since this is the format of the input waveform. If FLTP isn't supported
+  // then we'll need to convert the AVFrame's format. Our heuristic is to encode
+  // into the format with the highest resolution.
+  if (avCodec.sample_fmts == nullptr) {
+    // Can't really validate anything in this case, best we can do is hope that
+    // FLTP is supported by the encoder. If not, FFmpeg will raise.
+    return AV_SAMPLE_FMT_FLTP;
+  }
+
+  std::vector<AVSampleFormat> preferredFormatsOrder = {
```
scotts, line 146: We should make this a static `const` so that we only initialize it once - I think this is on the critical path for encoding performance. I think it's cleanest to make it a `const` global in just this cpp file rather than making it a static class member. 

(Since we're in C++17, we can't make it a `constexpr`, but we could potentially use `std::array` instead: https://stackoverflow.com/questions/33241909/cannot-create-constexpr-stdvector. I don't know how much that matters, but I was curious.)

src/torchcodec/_core/Encoder.cpp
```diff
@@ -140,11 +132,50 @@ AudioEncoder::AudioEncoder(
...
+      AV_SAMPLE_FMT_S64,
+      AV_SAMPLE_FMT_S32P,
+      AV_SAMPLE_FMT_S32,
+      AV_SAMPLE_FMT_S16P,
+      AV_SAMPLE_FMT_S16,
+      AV_SAMPLE_FMT_U8P,
+      AV_SAMPLE_FMT_U8};
+
+  for (AVSampleFormat preferredFormat : preferredFormatsOrder) {
+    for (auto i = 0; avCodec.sample_fmts[i] != -1; ++i) {
```
scotts, line 161: In general, I think we should be explicit about our integers, so we should say `int` here. Particularly since we're also relying on auto conversion of `int` and an enum type. I think `auto` is appropriate when the induction variable is an actual iterator class, for example, I prefer:
```
for (auto it = container.begin(); it != container.end(); ++it) {}
```
Versus:
```
for (ContainerType<ContainedType>::iterator it = container.begin(); it != container.end(); ++it) {}
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -140,11 +132,50 @@ AudioEncoder::AudioEncoder(
...
+      if (avCodec.sample_fmts[i] == preferredFormat) {
+        return preferredFormat;
+      }
+    }
+  }
+  // Should never happen, but just in case
+  return avCodec.sample_fmts[0];
```
scotts, line 168: If it shouldn't happen, should we raise an error instead? If we think it shouldn't happen but it's also preferable to just pick our most preferred format, let's explain why.

NicolasHug: It may happen if our `preferredFormatsOrder` list is incomplete, which may (unlikely) happen in future FFmpeg versions. I'll make a comment

src/torchcodec/_core/Encoder.cpp
```diff
@@ -120,12 +119,6 @@ AudioEncoder::AudioEncoder(
       "avcodec_open2 failed: ",
       getFFMPEGErrorStringFromErrorCode(status));
 
-  TORCH_CHECK(
-      avCodecContext_->frame_size > 0,
-      "frame_size is ",
-      avCodecContext_->frame_size,
-      ". Cannot encode. This should probably never happen?");
```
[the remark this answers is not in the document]
scotts: Interesting - it might be worth noting why from the docs (https://ffmpeg.org/doxygen/6.0/structAVCodecContext.html#aec57f0d859a6df8b479cd93ca3a44a33), which I admit to not understanding when we turn 0 into our default.

src/torchcodec/_core/Encoder.h
scotts: If it doesn't use any class members, then, yeah, let's make it a function in anonymous namespace.

scotts (approved):

PR #635, opened 2025-04-10
Support encoding into a bytes tensor

NicolasHug:
This PR adds support for encoding audio samples into a bytes (uint8) tensor. We allocate the output tensor to 10MB, and double the allocated size up to 320MB if needed (after which we error out).

We rely on a new `AVIOToTensorContext` class, which inherits from `AVIOContextHolder`.

The C++ `Encoder` class now takes two **optional** and **mutually exclusive** parameters: `fileName` and `format`. `fileName` is used when encoding to a file, while `format` is used when encoding to a tensor. 

The core API has changed: we don't expose an Encoder object anymore. We only expose two stateless functions:

- `encode_audio_to_file(filename : str, ...)`
- `encoded_tensor = encode_audio_to_tensor(format : str, ...)`

I suspect that the public Python API will look similar to this, but final design is still WIP.


src/torchcodec/_core/AVIOBytesContext.cpp
```diff
@@ -13,7 +13,7 @@ AVIOBytesContext::AVIOBytesContext(const void* data, int64_t dataSize)
     : dataContext_{static_cast<const uint8_t*>(data), dataSize, 0} {
   TORCH_CHECK(data != nullptr, "Video data buffer cannot be nullptr!");
   TORCH_CHECK(dataSize > 0, "Video data size must be positive");
-  createAVIOContext(&read, &seek, &dataContext_);
+  createAVIOContext(&read, nullptr, &seek, &dataContext_);
```
NicolasHug, line 16: Below: created a new `AVIOToTensorContext` class instead of using `AVIOBytesContext`. The constructor is different, and so is the underlying `DataContext` struct.

I'm happy to consider alternative solutions.

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -59,8 +59,9 @@ function(make_torchcodec_libraries
     set(decoder_library_name "libtorchcodec_decoder${ffmpeg_major_version}")
     set(decoder_sources
         AVIOContextHolder.cpp
+        AVIOBytesContext.cpp
```
NicolasHug, line 62: `Encoder.[cpp, h]` rely on `AVIOToTensorContext`, and specifically on the `AVIOToTensorContext::getOutputTensor()` method. They don't rely on the `AVIOContextHolder` base class, like the decoder. For this reason we have to add `AVIOBytesContext.cpp` to the source dependency here.

Alternatively, I *think* we could make `getOutputTensor` a virtual method of the base class? But this method wouldn't make much sense for the existing child classes (like `AVIOBytesContext`), so this doesn't sounds like a great OOP design.

src/torchcodec/_core/AVIOContextHolder.cpp
```diff
@@ -22,13 +23,17 @@ void AVIOContextHolder::createAVIOContext(
       buffer != nullptr,
       "Failed to allocate buffer of size " + std::to_string(bufferSize));
 
-  avioContext_.reset(avio_alloc_context(
+  TORCH_CHECK(
+      (seek != nullptr) && ((write != nullptr) ^ (read != nullptr)),
+      "seek method must be defined, and either write or read must be defined. "
+      "But not both!")
```
NicolasHug, line 29: We may relax the mutual-exclusivity check above eventually, if we implement both `write` and `read` within the same class. For now, mutual-exclusivity is assumed and enforced, because we use the existence of `write` to set the `write_flag` below.

src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -177,4 +177,13 @@ bool canSwsScaleHandleUnalignedData();
...
 void setFFmpegLogLevel();
 
+AVIOContext* AVIOAllocContext(
+    uint8_t* buffer,
+    int buffer_size,
+    int write_flag,
+    void* opaque,
+    int (*read_packet)(void* opaque, uint8_t* buf, int buf_size),
+    int (*write_packet)(void* opaque, const uint8_t* buf, int buf_size),
+    int64_t (*seek)(void* opaque, int64_t offset, int whence));
```
scotts, line 187: Let's move the convenience function type aliases of `AVIO*Function` to here, and then use those in all of our interfaces. That should be easier to grep for, and also easier to tell at a glance that the signatures are the same in different places.

Also, since `AVIOAllocContext` is a function and not a type, I think we should name it `avioAllocContext()`. Otherwise I think readers will assume we're instantiating an object of type `AVIOAllocContext` until they look at the declaration.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -78,9 +79,13 @@ AudioEncoder::~AudioEncoder() {}
 AudioEncoder::AudioEncoder(
     const torch::Tensor wf,
     int sampleRate,
-    std::string_view fileName,
+    std::optional<std::string_view> fileName,
+    std::optional<std::string_view> formatName,
```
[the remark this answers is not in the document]
scotts: I think we should follow the same pattern set out in `SingleStreamDecoder`: one constructor that takes the file name, and another constructor that takes the `AVIOToTensorContext` as an argument. That means it's the callers responsibility to create the `AVIOToTensorContext` object; that's how it currently works in `SingleStreamDecoder`. That would also require pulling all common initialization into some `initializeEncoder()` function that both constructors would call.

The value, to me, is that it's more clear when creating the object that you're doing it correctly, and then when reading the initialization code, it's more clear what's required for, and correct for, each path.

src/torchcodec/_core/CMakeLists.txt
scotts: Agreed on not making `getOutputTensor()` virtual on the base class, since it's only applicable to one of the derived classes. We could potentially pull `AVIOToTensorContext` out of `AVIOBytesContext.[h|cpp]` to limit what gets put into `libtorchcodec_decoderN.so`.

I think the issue here is that the `Encoder` class needs to call `getOutputTensor()`, which means that it must actually hold a reference to an actual `AVIOToTensorContext` rather than just the base `AVIOContextHolder`. (Which is what `SingleStreamDecoder` does.)  A potential way around this is for `AVIOToTensorContext` to *accept* a tensor rather than creating its own. The caller, however, would still need to keep track of how many bytes it asked it to encode in order to do the final narrow on the tensor, so that's not necessarily any cleaner. Your call on what you think makes the most sense.

scotts (approved):

scotts:
Approving to unblock, but I do think we should follow the same two-constructor pattern we have in `SingleStreamDecoder`.

src/torchcodec/_core/FFMPEGCommon.h
scotts: Yes, correct. We can only use the type aliases when we declare some variable that has that type.

PR #638, opened 2025-04-12
File like tutorial

scotts:
Tutorial for streaming data through a file-like object.

### DIFF
--- a/examples/file_like.py
+++ b/examples/file_like.py
@@ -72,7 +72,6 @@ decoder = VideoDecoder(pre_downloaded_raw_video_bytes)
 
 print(f"Video size in MB: {len(pre_downloaded_raw_video_bytes) / 1024 / 1024}")
 print(decoder.metadata)
-print()
 
 # %%
 # We can see that the video is about 253 MB, has the resolution 1920x1080, is
@@ -132,7 +131,6 @@ print()
 
 print("Direct url to FFmpeg:")
 bench(direct_url_to_ffmpeg)
-print()
 
 # %%
 # Decoding the already downloaded video is clearly the fastest. Having to
@@ -161,7 +159,6 @@ def stream_while_decode():
 
 print("Stream while decode: ")
 bench(stream_while_decode)
-print()
 
 # %%
 # Streaming the data through a file-like object is about 4.3x faster than
@@ -217,7 +214,6 @@ first_frame = counter_decoder[0]
 print("Decoding the first frame required "
       f"{file_op_counter.num_reads - init_reads} additional reads and "
       f"{file_op_counter.num_seeks - init_seeks} additional seeks.")
-print()
 
 # %%
 # Performance: local file path versus local file-like object
@@ -242,7 +238,6 @@ print()
 
 print("Decode from existing open file object:")
 bench(decode_from_existing_open_file_object)
-print()
 
 # %%
 import shutil
--- a/examples/file_like.py
+++ b/examples/file_like.py
@@ -48,6 +48,7 @@ def bench(f, average_over=20, warmup=2):
     med = times.median().item()
     print(f"{med = :.2f}ms +- {std:.2f}")
 
+
 # %%
 # Performance: downloading first versus streaming
 # -----------------------------------------------
@@ -94,6 +95,7 @@ print()
 # this to avoid scanning the entire video during initialization, which would
 # require downloading the entire video even if we only want to decode the first
 # frame. See :ref:`sphx_glr_generated_examples_approximate_mode.py` for more.
+
 
 def decode_from_existing_download():
     decoder = VideoDecoder(
@@ -145,6 +147,7 @@ print()
 # `Filesystem interfaces for Python <https://filesystem-spec.readthedocs.io/en/latest/?badge=latest>`_.
 
 import fsspec
+
 
 def stream_while_decode():
     # The `client_kwargs` are passed down to the aiohttp module's client
--- a/examples/file_like.py
+++ b/examples/file_like.py
@@ -9,15 +9,21 @@
 Streaming data through file-like support
 ===================================================================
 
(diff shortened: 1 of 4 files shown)


examples/file_like.py
```diff
@@ -0,0 +1,245 @@
...
+"""
+===================================================================
+Streaming data through file-like support
+===================================================================
+
+In this example, we will show how to decode streaming data. That is, when files
+do not reside locally, we will show how to only download the data segments that
+are needed to decode the frames you care about. We accomplish this capability
+with Python
+`file-like objects <https://docs.python.org/3/glossary.html#term-file-like-object>`_."""
```
NicolasHug, line 16: Maybe indicate that we'll be using the `VideoDecoder` in this example, but everything also works for `AudioDecoder`?

Also: we should update this link in the `VideoDecoder` docstring, and create a similar one for `AudioDecoder`:

https://github.com/pytorch/torchcodec/blob/2c137e75661fd3c04b38901480defdf86a9183a9/src/torchcodec/decoders/_video_decoder.py#L32

examples/file_like.py
```diff
@@ -0,0 +1,245 @@
...
+# its size and content:
+
+from torchcodec.decoders import VideoDecoder
+
+nasa_url = "https://download.pytorch.org/torchaudio/tutorial-assets/stream-api/NASAs_Most_Scientifically_Complex_Space_Observatory_Requires_Precision-MP4.mp4"
+
+pre_downloaded_raw_video_bytes = get_url_content(nasa_url)
+decoder = VideoDecoder(pre_downloaded_raw_video_bytes)
+
+print(f"Video size in MB: {len(pre_downloaded_raw_video_bytes) / 1024 / 1024}")
```
NicolasHug, line 73: Nit: use `//` instead of `/` so this renders as `253` instead of `253.05233764648438`.

examples/file_like.py
```diff
@@ -0,0 +1,245 @@
...
+# implements its own read and seek methods that only download data from a URL as
+# needed. Rather than implementing our own, we can use such objects from the
+# `fsspec <https://github.com/fsspec/filesystem_spec>`_ module that provides
+# `Filesystem interfaces for Python <https://filesystem-spec.readthedocs.io/en/latest/?badge=latest>`_.
+
+import fsspec
+
+
+def stream_while_decode():
+    # The `client_kwargs` are passed down to the aiohttp module's client
```
NicolasHug, line 151: Q: it's not obvious what aiohttp is in this context, as it's not mentioned before. I see it was added as a dependency in `requirements.txt` - is it actually needed? If yes then we probably want to indicate in the tutorial that users will need to `pip install` it, and if not, we can remove it from `requirements.txt`?

scotts: Yeah, we do actually need to install it separately. Good call out, I'll mention in the text that it's an additional library to install.

examples/file_like.py
```diff
@@ -0,0 +1,302 @@
...
+# is much better, as its about 2.5x faster than downloading the video first.
+#
+# We can do better, and the way how is to use a file-like object which
+# implements its own read and seek methods that only download data from a URL as
+# needed. Rather than implementing our own, we can use such objects from the
+# `fsspec <https://github.com/fsspec/filesystem_spec>`_ module that provides
+# `Filesystem interfaces for Python <https://filesystem-spec.readthedocs.io/en/latest/?badge=latest>`_.
+# Note that using these capabilities from the fsspec` library also requires the
+# `aiohttp <https://docs.aiohttp.org/en/stable/>`_ module. You can install both with
+# ``pip install fsspec aiohttp``.
```
scotts, line 151: @NicolasHug, the code snippet for `pip install` renders kinda ugly. Rather than being one block of inline code, it renders each word separately. Do you know a better way to do it?

examples/file_like.py
scotts: Keeping unresolved until I actually replace the text in the source files.

examples/file_like.py
NicolasHug: Unfortunately no, this is a pytorch-sphinx-theme issue, and it has been a problem for years :(
https://github.com/pytorch/pytorch_sphinx_theme/issues/107

I hope it'll be fixed in the new theme (not sure what the status of that is).

To avoid the uglyness, maybe we can just write `"pip install ..."` in quotes, without using double backticks. Or use single backticks so it renders in italics.

scotts: Got it. Unfortunate. I'll play around with it a little.

examples/file_like.py
```diff
@@ -0,0 +1,303 @@
...
+    def read(self, size: int) -> bytes:
+        self.num_reads += 1
+        return self._file.read(size)
+
+    def seek(self, offset: int, whence: int) -> bytes:
+        self.num_seeks += 1
+        return self._file.seek(offset, whence)
+
+
+# Let's now get a file-like object from our class defined above, providing it a
```
NicolasHug, line 223: Just in case, if you wanted this to render as text instead of as a comment, we'll need `# %%` above. But keeping it as a comment is just fine, if that was the intended rendering.

NicolasHug (approved):
Looks fantastic, thank you!

The example takes a bit of time to build (~3 minutes), so it does increase the duration of our docs job. I suspect it means that the doc build job now becomes our longest CI job (as it also relies on the CUDA build, which is slow).

But that's an OK price to pay for now IMHO. If it becomes too slow, we can explore reducing `average_over`, or avoid benchmarking the `download_before_decode()` function which is the obvious offender.

examples/file_like.py
scotts: Yeah, this is intentionally a code comment. I felt there was enough going on it would help readers to have a little narration.

scotts:
Time spent is a good point. Let me reduce `average_over` to 10 instead of 20. That still should be enough samples to reduce variation.

PR #648, opened 2025-04-22
Remove torchvision test dependency

NicolasHug:
I think we shouldn't need torchvision since https://github.com/pytorch/torchcodec/pull/612

### DIFF
--- a/.github/workflows/linux_cuda_wheel.yaml
+++ b/.github/workflows/linux_cuda_wheel.yaml
@@ -105,7 +105,7 @@ jobs:
         run: ${CONDA_RUN} python -m pip install --upgrade pip
       - name: Install PyTorch
         run: |
-          ${CONDA_RUN} python -m pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu${{ env.cuda_version_without_periods }}
+          ${CONDA_RUN} python -m pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu${{ env.cuda_version_without_periods }}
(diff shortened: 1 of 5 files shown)


.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -105,7 +105,7 @@ jobs:
         run: ${CONDA_RUN} python -m pip install --upgrade pip
       - name: Install PyTorch
         run: |
-          ${CONDA_RUN} python -m pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu${{ env.cuda_version_without_periods }}
+          ${CONDA_RUN} python -m pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu${{ env.cuda_version_without_periods }}
```
scotts, line 108: Looks like we still need torchvision here because of the GPU benchmarks.

NicolasHug: aaaaah I forgot we run those. We probably... shouldn't, since we never check :shrug:

scotts: I think there's some value in doing it, as it makes sure that the GPU benchmarks don't become stale. That is, it's not about the actual performance result from the benchmark, it's more a check that it's still usable. There may also be value in doing that on the CPU side as well.

scotts:
Current CI failures look like the pre-existing problem, so this should improve things.

scotts (approved):

PR #664, opened 2025-04-29
Audio: Fix output shape when start==stop

NicolasHug:
Fixes https://github.com/pytorch/torchcodec/issues/661

Output shape is now (num_channels, 0) instead of (0, 0). This is more consistent with the video decoder as well, so I think we can consider this a bug fix.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -848,12 +848,13 @@ AudioFramesOutput SingleStreamDecoder::getFramesPlayedInRangeAudio(
             std::to_string(*stopSecondsOptional) + ").");
   }
 
+  StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
+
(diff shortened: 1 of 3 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -848,13 +848,14 @@ AudioFramesOutput SingleStreamDecoder::getFramesPlayedInRangeAudio(
             std::to_string(*stopSecondsOptional) + ").");
   }
 
+  StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
+
   if (stopSecondsOptional.has_value() && startSeconds == *stopSecondsOptional) {
     // For consistency with video
-    return AudioFramesOutput{torch::empty({0, 0}), 0.0};
+    auto numChannels = getNumChannels(streamInfo.codecContext);
```
scotts, line 855: Nit: I prefer we're explicit about our ints. It's just as easy to read.

scotts (approved):

PR #666, opened 2025-04-29
Ensure correct encoding for non-contiguous WF

NicolasHug:
TSIA!

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -13,10 +13,8 @@ torch::Tensor validateWf(torch::Tensor wf) {
       wf.dtype() == torch::kFloat32,
       "waveform must have float32 dtype, got ",
       wf.dtype());
-  // TODO-ENCODING check contiguity of the input wf to ensure that it is indeed
-  // planar (fltp).
   TORCH_CHECK(wf.dim() == 2, "waveform must have 2 dimensions, got ", wf.dim());
-  return wf;
+  return wf.contiguous();
 }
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -13,10 +13,8 @@ torch::Tensor validateWf(torch::Tensor wf) {
       wf.dtype() == torch::kFloat32,
       "waveform must have float32 dtype, got ",
       wf.dtype());
-  // TODO-ENCODING check contiguity of the input wf to ensure that it is indeed
-  // planar (fltp).
   TORCH_CHECK(wf.dim() == 2, "waveform must have 2 dimensions, got ", wf.dim());
-  return wf;
+  return wf.contiguous();
```
NicolasHug, line 17: Confirming that this is load-bearing: newly added test fails if we just return `wf` without calling `contiguous()`.

test/test_ops.py
```diff
@@ -1284,6 +1284,35 @@ def test_encode_to_tensor_long_output(self):
 
         torch.testing.assert_close(self.decode(encoded_tensor), samples)
 
+    def test_contiguity(self):
```
scotts, line 1287: Can you provide some narration in the test about what exactly we're testing? I'm afraid I'm not familiar enough with tensor manipulation to fully understand everything. For example, I'm surprised `assert_close(contiguous_samples, non_contiguous_samples)` passes when they have different strides.

scotts (approved):

PR #667, opened 2025-04-29
Encoding: address some TODOs

NicolasHug:
This PR addresses a few outstanding TODO-ENCODING entries. See diff for (hopefully obvious!) changes.

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -394,8 +394,6 @@ void encode_audio_to_file(
       .encode();
 }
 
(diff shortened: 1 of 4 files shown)


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -394,8 +394,6 @@ void encode_audio_to_file(
       .encode();
 }
 
-// TODO-ENCODING is "format" a good parameter name?? It kinda conflicts with
-// "sample_format" which we may eventually want to expose.
```
NicolasHug, line 398: torchaudio.load() uses `format` as well, so I think that's fine. We haven't discussed the public API yet anyway.

PR #674, opened 2025-05-09
add opencv benchmark

jinhohwang-meta:
This PR added the benchmark for opencv. The benchmark shows that there is gap. Hope the torchcodec team can find out where the gap is from and improve the decoding performance. 

Local benchmark 

```
1 threads: --------------------------------------------------------------------------------------------------------------------------------
      OpenCV            |            22.4            |            22.6           |       6.4        |        9.3        |        18.1
      TorchCodecPublic  |            39.4            |            28.1           |       9.5        |        9.8        |        26.0
```

Image

### DIFF
--- a/benchmarks/decoders/benchmark_decoders_library.py
+++ b/benchmarks/decoders/benchmark_decoders_library.py
@@ -145,6 +145,74 @@ class TorchVision(AbstractDecoder):
         ]
         return frames
 
+class OpenCVDecoder(AbstractDecoder):
+    def __init__(self):
+        import cv2.videoio_registry as vr
+
+        self._print_each_iteration_time = False
+        api_pref = None
+        for backend in vr.getStreamBufferedBackends():
+            if not vr.hasBackend(backend):
+                continue
+            if not vr.isBackendBuiltIn(backend):
+                _, abi, api = vr.getStreamBufferedBackendPluginVersion(backend)
+                if (abi < 1 or (abi == 1 and api < 2)):
+                    continue
+            api_pref = backend
+            break
+        self._backend = api_pref
+
+    def decode_frames(self, video_file, pts_list):
+        import cv2
+
+        cap = cv2.VideoCapture(video_file, self._backend, [])
+        if not cap.isOpened():
+            raise ValueError("Could not open video stream")
+
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+        approx_frame_numbers = [int(pts * fps) for pts in pts_list]
+
+        current_frame = 0
+        frames = []
+        while True:
+            ok = cap.grab()
+            if not ok:
+                break
+            if current_frame in approx_frame_numbers:  # only decompress needed
+                ret, frame = cap.retrieve()
+                if ret:
+                    frames.append(frame)
+
+            if len(frames) == len(approx_frame_numbers):
+                break
+            current_frame += 1
+        cap.release()
+        return frames
+
+    def decode_first_n_frames(self, video_file, n):
+        import cv2
+
+        cap = cv2.VideoCapture(video_file, self._backend, [])
+        if not cap.isOpened():
+            raise ValueError("Could not open video stream")
+
+        frames = []
+        for i in range(n):
+            ok = cap.grab()
+            if not ok:
+                break
+            ret, frame = cap.retrieve()
+            if ret:
+                frames.append(frame)
+        cap.release()
+        return frames
+
+    def decode_and_resize(self, video_file, pts_list, height, width, device):
+        import cv2
+        frames = [cv2.resize(frame, (width, height)) for frame in self.decode_frames(video_file, pts_list)]
+        return frames
+
 
 class TorchCodecCore(AbstractDecoder):
     def __init__(self, num_threads=None, color_conversion_library=None, device="cpu"):
--- a/benchmarks/decoders/benchmark_decoders_library.py
+++ b/benchmarks/decoders/benchmark_decoders_library.py
@@ -151,6 +151,7 @@ class OpenCVDecoder(AbstractDecoder):
 
         self._print_each_iteration_time = False
         api_pref = None
+        # Check backend abi/api for compatibility
         for backend in vr.getStreamBufferedBackends():
             if not vr.hasBackend(backend):
                 continue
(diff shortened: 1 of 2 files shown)


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+
+    def decode_frames(self, video_file, pts_list):
+        import cv2
+
+        cap = cv2.VideoCapture(video_file, self._backend, [])
+        if not cap.isOpened():
+            raise ValueError("Could not open video stream")
+
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
```
NicolasHug, line 173: `frames` seems unused and gets redefined below

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+
+        self._print_each_iteration_time = False
+        api_pref = None
+        for backend in vr.getStreamBufferedBackends():
+            if not vr.hasBackend(backend):
+                continue
+            if not vr.isBackendBuiltIn(backend):
+                _, abi, api = vr.getStreamBufferedBackendPluginVersion(backend)
+                if (abi < 1 or (abi == 1 and api < 2)):
+                    continue
```
NicolasHug, line 160: Can you add a comment explaining why this is needed?

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+        approx_frame_numbers = [int(pts * fps) for pts in pts_list]
+
+        current_frame = 0
+        frames = []
+        while True:
+            ok = cap.grab()
+            if not ok:
+                break
```
NicolasHug, line 181: Let's raise here instead of a `break`. This will ensure that we are not reporting results in case `cap.grab()` fails.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+    def decode_frames(self, video_file, pts_list):
+        import cv2
+
+        cap = cv2.VideoCapture(video_file, self._backend, [])
+        if not cap.isOpened():
+            raise ValueError("Could not open video stream")
+
+        fps = cap.get(cv2.CAP_PROP_FPS)
+        frames = int(cap.get(cv2.CAP_PROP_FRAME_COUNT))
+        approx_frame_numbers = [int(pts * fps) for pts in pts_list]
```
NicolasHug, line 174: Nit: `s/numbers/indices`. We use the term `indices` for this.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+
+        cap = cv2.VideoCapture(video_file, self._backend, [])
+        if not cap.isOpened():
+            raise ValueError("Could not open video stream")
+
+        frames = []
+        for i in range(n):
+            ok = cap.grab()
+            if not ok:
+                break
```
NicolasHug, line 204: Same here, we should raise instead of `break`

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -145,6 +145,74 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+            ret, frame = cap.retrieve()
+            if ret:
+                frames.append(frame)
+        cap.release()
```
NicolasHug, line 208: Let's assert that `len(frames) == n` to ensure no error went undetected

NicolasHug (commented):
Thanks a lot for the benchmark @jinhohwang-meta , that's very useful! I made a few comments below.

Based on how the code is converting pts to indices (using the average fps), I think a fairer comparison against torchcodec would be to benchmark torchcodec with the approximate mode (instead of exact mode). I think it corresponds to `torchcodec_core_nonbatch`, although I'm not 100% sure (CC @scotts )

NicolasHug (commented):
Thanks a lot for the benchmark @jinhohwang-meta , that's very useful! I made a few comments below.

Based on how the code is converting pts to indices (using the average fps), I think a fairer comparison against torchcodec would be to benchmark torchcodec with the approximate mode (instead of exact mode). I think it corresponds to `torchcodec_core_nonbatch`, although I'm not 100% sure (CC @scotts )

scotts:
The comparison we'll probably be most interested in is this one:
```
python benchmarks/decoders/benchmark_decoders.py --decoders opencv,torchcodec_public:seek_mode=exact,torchcodec_public:seek_mode=approximate,torchaudio --min-run-seconds 40
```
That is, it compares:

1. OpenCV
2. TorchCodec public API, exact seek mode
3. TorchCodec public API, approximate seek mode
4. TorchAudio

We'll also want to use some more videos - the ones that are part of the README benchmark are good to use. We generate those on the fly here: https://github.com/pytorch/torchcodec/blob/87b98e870faa0ee3f66404dad6450030a1c86827/benchmarks/decoders/generate_readme_data.py#L37-L55

jinhohwang-meta:
> Thanks a lot for the benchmark @jinhohwang-meta , that's very useful! I made a few comments below.
> 
> Based on how the code is converting pts to indices (using the average fps), I think a fairer comparison against torchcodec would be to benchmark torchcodec with the approximate mode (instead of exact mode). I think it corresponds to `torchcodec_core_nonbatch`, although I'm not 100% sure (CC @scotts )

@NicolasHug I actually did compare with approximate as well, but it showed similar results. @scotts asked me to push this for the team to follow on to find out the gap and I am very much interested in to see why as well.

benchmarks/decoders/benchmark_decoders_library.py
scotts: @jinhohwang-meta, can we be more descriptive of what the meaning of these backends are? Why are we checking for specific ranges of the ABI and API?

I think we'll also want `assert api_pref is not None` after the loop.

Dan-Flores:
This change was iterated on and merged in #711.

PR #678, opened 2025-05-13
AudioDecoder: specify desired num_channels

NicolasHug:
Closes https://github.com/pytorch/torchcodec/issues/675

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -293,18 +293,20 @@ void AudioEncoder::encodeInnerLoop(
   if (mustConvert) {
     if (!swrContext_) {
       swrContext_.reset(createSwrContext(
-          avCodecContext_,
           AV_SAMPLE_FMT_FLTP,
           avCodecContext_->sample_fmt,
           srcAVFrame->sample_rate, // No sample rate conversion
-          srcAVFrame->sample_rate));
+          srcAVFrame->sample_rate,
+          srcAVFrame,
+          getNumChannels(srcAVFrame) // No num_channel conversion
+          ));
     }
-    convertedAVFrame = convertAudioAVFrameSampleFormatAndSampleRate(
+    convertedAVFrame = convertAudioAVFrameSamples(
         swrContext_,
         srcAVFrame,
         avCodecContext_->sample_fmt,
         srcAVFrame->sample_rate, // No sample rate conversion
-        srcAVFrame->sample_rate);
+        getNumChannels(srcAVFrame)); // No num_channel conversion
     TORCH_CHECK(
         convertedAVFrame->nb_samples == srcAVFrame->nb_samples,
         "convertedAVFrame->nb_samples=",
--- a/src/torchcodec/_core/FFMPEGCommon.cpp
+++ b/src/torchcodec/_core/FFMPEGCommon.cpp
@@ -107,6 +107,10 @@ void setChannelLayout(
 
 namespace {
 #if LIBAVFILTER_VERSION_MAJOR > 7 // FFmpeg > 4
+
+// Returns:
+// - the srcAVFrame's channel layout if srcAVFrame has desiredNumChannels
+// - the default channel layout with desiredNumChannels otherwise.
 AVChannelLayout getDesiredChannelLayout(
     int desiredNumChannels,
     const UniqueAVFrame& srcAVFrame) {
@@ -118,8 +122,10 @@ AVChannelLayout getDesiredChannelLayout(
   }
   return desiredLayout;
 }
+
 #else
 
+// Same as above
 int64_t getDesiredChannelLayout(
     int desiredNumChannels,
     const UniqueAVFrame& srcAVFrame) {
@@ -134,6 +140,7 @@ int64_t getDesiredChannelLayout(
 #endif
 } // namespace
 
+// Sets dstAVFrame' channel layout to getDesiredChannelLayout(): see doc above
 void setChannelLayout(
     UniqueAVFrame& dstAVFrame,
     const UniqueAVFrame& srcAVFrame,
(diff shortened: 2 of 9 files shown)


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -280,9 +280,11 @@ void add_video_stream(
 void add_audio_stream(
     at::Tensor& decoder,
     std::optional<int64_t> stream_index = std::nullopt,
-    std::optional<int64_t> sample_rate = std::nullopt) {
+    std::optional<int64_t> sample_rate = std::nullopt,
+    std::optional<int64_t> num_channels = std::nullopt) {
   AudioStreamOptions audioStreamOptions;
   audioStreamOptions.sampleRate = sample_rate;
+  audioStreamOptions.numChannels = num_channels;
```
NicolasHug, line 287: This implicitly converts an optional int64 to an optional int. The same implicit conversion already happens for sample_rate, and for the VideoStreamOptions' fields like height, width, and ffmpegThreadCount.

I'll open an issue to fix / check these all at once.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -293,18 +293,20 @@ void AudioEncoder::encodeInnerLoop(
...
           avCodecContext_->sample_fmt,
           srcAVFrame->sample_rate, // No sample rate conversion
-          srcAVFrame->sample_rate));
+          srcAVFrame->sample_rate,
+          srcAVFrame,
+          getNumChannels(srcAVFrame) // No num_channel conversion
+          ));
     }
-    convertedAVFrame = convertAudioAVFrameSampleFormatAndSampleRate(
+    convertedAVFrame = convertAudioAVFrameSamples(
```
NicolasHug, line 304: I renamed the function because it converts to many things to list now. @scotts  I also remember that you wanted to align all the `sourceStuff` to `srcStuff` - which I'll do as a quick follow-up!

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -106,32 +105,80 @@ void setChannelLayout(
...
 
 SwrContext* createSwrContext(
-    UniqueAVCodecContext& avCodecContext,
     AVSampleFormat sourceSampleFormat,
     AVSampleFormat desiredSampleFormat,
     int sourceSampleRate,
-    int desiredSampleRate) {
+    int desiredSampleRate,
+    const UniqueAVFrame& srcAVFrame,
+    int desiredNumChannels) {
```
NicolasHug, line 169: slightly changed the signature to take an AVFrame instead of the AVCodecContext. This is where we get the source num_channel and source channel layout. It's more accurate to get it from the AVFrame, although they should both always be the same.

src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -40,6 +40,8 @@ class AudioDecoder:
             the :term:`best stream` is used.
         sample_rate (int, optional): The desired output sample rate of the decoded samples.
             By default, the samples are returned in their original sample rate.
+        num_channels (int, optional): The desired number of channels of the decoded samples.
+            By default, the original number of channels is used.
```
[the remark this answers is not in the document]
NicolasHug: Yes, it's the number of channels that the audio source was encoded with. We use similar phrasing for `sample_rate` above. I feel like it should be fairly clear what this means for audio users, but happy to consider alternatives

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1171,27 +1178,33 @@ void SingleStreamDecoder::convertAudioAVFrameToFrameOutputOnCPU(
   int desiredSampleRate =
       streamInfo.audioStreamOptions.sampleRate.value_or(sourceSampleRate);
 
+  int sourceNumChannels = getNumChannels(srcAVFrame);
```
scotts, line 1181: So the "original" channels are actually a property of the frame, which means it's *likely* the same in a stream. I don't think we need to worry about it now, but: any chance there are a variable number of channels in a stream?

NicolasHug: Conceptually, it's a property of the samples rather than a property of the frame (which is what FFmpeg exposes).

I cannot guarantee that it doesn't exist, but I think it's very unlikely: samples are continuous, unlike video frames which are more discrete. So having samples with different properties in a single stream would be the result of a major bug in the encoder (which I guess we can't always rule out).

I'll add a `TORCH_CHECK(getNumChannels(srcAVFrame) == getNumChannels(avCodecContext));` to make sure we don't get any surprise

src/torchcodec/_core/FFMPEGCommon.cpp
scotts: Nit: I prefer for the primary FFmpeg data structures to appear first in a function. I think of these functions as if they are methods on those structures, and that gets reinforced by making them the first parameter. I think I understand the current order, which `srcFrame` is placed next to `desiredNumChannels` because you're getting the source number of channels from `srcFrame`. But given what we're doing, I think of this function as "Create a `swrContext` from this `AVFrame` with these desire parameters."

scotts (approved):

src/torchcodec/_core/FFMPEGCommon.cpp
NicolasHug: I agree that for functions that look like methods, we want the FFmpeg struct to be the first. The reason I removed the avCodecContext from the top and now use an AVFrame is *precisely* because this function shouldn't read as a method.

> I think I understand the current order, which srcFrame is placed next to desiredNumChannels because you're getting the source number of channels from srcFrame

Yes, you're correct. Ideally, we shouldn't need to pass the AVFrame at all, we should just pass its `AVChannelLayout` as `sourceChannelLayout`. But unfortunately, `AVChannelLayout` only exists from FFmpeg 5. That's the main reason we're passing an `AVFrame`: it's so that we can call `createSwrContext()` uniformly across ffmpeg versions, and contain the preproc `#ifdef`  directives into FFmpegCommon.cpp.

Does that make sense?

Basically: we're passing an AVFrame because of some FFmpeg version constraint, but really this just be "source channel layout", and thus this isn't a method on an AVFrame at all.

src/torchcodec/decoders/_audio_decoder.py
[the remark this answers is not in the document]
NicolasHug: Edited to

> By default, the number of channels of the source is used

Did the same for sample_rate

PR #682, opened 2025-05-15
Fix pts <-> seconds conversions

NicolasHug:
:cold_face: 

We weren't multiplying by `timeBase.num`. All the videos we've been testing on so far had a `timebase.num == 1`, so the bug went unnoticed.


scotts (approved):
Good catch!

PR #683, opened 2025-05-15
Fallback to DTS if PTS info doesn't exist

NicolasHug:
Some videos, like https://huggingface.co/datasets/raushan-testing-hf/videos-test/blob/main/sample_video_2.avi mentioned in https://github.com/pytorch/torchcodec/issues/673, do not have any PTS value for their packets.

Up until now, this caused various weird issues like https://github.com/pytorch/torchcodec/issues/677 and https://github.com/pytorch/torchcodec/issues/676, and some silently-wrong error (e.g. index-based API with exact mode were always returning the first frame).

This PR allows for proper decoding of such poorly-encoded videos, by falling back to the DTS info if the PTS do not exist.

I added a test, which is skipped unconditionally. We should enable it back once we can check in such a poorly-encoded video. I made sure the test properly pass on https://huggingface.co/datasets/raushan-testing-hf/videos-test/blob/main/sample_video_2.avi with FFmpeg 4 and 6.

(needs https://github.com/pytorch/torchcodec/pull/682 first. Note that the `sample_video_2.avi` video is one of those videos were timebase.den != 1, which exposed the bug that  is fixed in https://github.com/pytorch/torchcodec/pull/682).

### DIFF
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -142,6 +142,9 @@ class VideoStreamMetadata(StreamMetadata):
             self.end_stream_seconds_from_content is None
             or self.begin_stream_seconds_from_content is None
             or self.num_frames is None
+            # Should never happen, but prevents ZeroDivisionError:
+            or self.end_stream_seconds_from_content
+            == self.begin_stream_seconds_from_content
         ):
             return self.average_fps_from_header
         return self.num_frames / (
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -986,6 +986,47 @@ class TestVideoDecoder:
             assert_frames_equal(ref_frame3, frames[1].data)
             assert_frames_equal(ref_frame5, frames[2].data)
 
+    @pytest.mark.parametrize("seek_mode", ("exact", "approximate"))
+    def test_pts_to_dts_fallback(self, seek_mode):
+        # Non-regression test for
+        # https://github.com/pytorch/torchcodec/issues/677 and
+        # https://github.com/pytorch/torchcodec/issues/676.
+        # More accurately, this is a non-regression test for videos which do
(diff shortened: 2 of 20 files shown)


src/torchcodec/_core/_metadata.py
```diff
@@ -142,6 +142,9 @@ def average_fps(self) -> Optional[float]:
             self.end_stream_seconds_from_content is None
             or self.begin_stream_seconds_from_content is None
             or self.num_frames is None
+            # Should never happen, but prevents ZeroDivisionError:
+            or self.end_stream_seconds_from_content
+            == self.begin_stream_seconds_from_content
```
NicolasHug, line 147: This isn't actually needed anymore, i.e. the dts fallback already fixed the metadata issue. But it's probably safer to have this regardless?

scotts: Yeah, let's keep it. I think a single frame video is allowed, and we would encounter this problem with such a video.

test/test_decoders.py
```diff
@@ -986,6 +986,47 @@ def get_some_frames(decoder):
...
+        # https://github.com/pytorch/torchcodec/issues/676.
+        # More accurately, this is a non-regression test for videos which do
+        # *not* specify pts values (all pts values are N/A and set to
+        # INT64_MIN), but specify *dts* value - which we fallback to.
+        #
+        # The test video we have is from
+        # https://huggingface.co/datasets/raushan-testing-hf/videos-test/blob/main/sample_video_2.avi
+        # We can't check it into the repo due to potential licensing issues, so
+        # we have to unconditionally skip this test.#
+        # TODO: encode a video with no pts values to unskip this test. Couldn't
```
scotts, line 1002: Create an issue?

scotts (approved):
Another great catch!

PR #685, opened 2025-05-15
Remove 'unstable API' notice from README

NicolasHug:
The APIs we expose are stable by now, I feel like we can safely remove these notices?


scotts (approved):
_Enthusiastically_ approved!

PR #692, opened 2025-05-21
Add `AudioEncoder` public Python API

NicolasHug:
This PR exposes a simple `AudioEncoder` object, following "Option 2" from https://github.com/pytorch/torchcodec/pull/690.

More tests and doc will come in follow-ups. There are already plenty of tests in `test_ops.py`


scotts (approved):
We should add some basic tests that succeed in follow-up PRs.

PR #693, opened 2025-05-22
Audio encoding: support custom `num_channels`

NicolasHug:
This PR adds a `num_channels` parameter to the audio encoder, which allows users to specify the number of channels of the **encoded** data (e.g. encode a stereo tensor into a mono file).

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -55,6 +55,20 @@ void validateSampleRate(const AVCodec& avCodec, int sampleRate) {
       supportedRates.str());
 }
 
+void print_supported_channel_layouts(const AVCodec *codec) {
+    if (!codec->ch_layouts) {
+        printf("No specific channel layouts supported by this encoder.\n");
+        return;
+    }
+    const AVChannelLayout *layout = codec->ch_layouts;
+    while (layout->order != AV_CHANNEL_ORDER_UNSPEC) {
+        char layout_name[256];
+        av_channel_layout_describe(layout, layout_name, sizeof(layout_name));
+        printf("Supported channel layout: %s\n", layout_name);
+        layout++;
+    }
+}
+
 static const std::vector<AVSampleFormat> preferredFormatsOrder = {
     AV_SAMPLE_FMT_FLTP,
     AV_SAMPLE_FMT_FLT,
@@ -101,7 +115,8 @@ AudioEncoder::AudioEncoder(
     const torch::Tensor wf,
     int sampleRate,
     std::string_view fileName,
-    std::optional<int64_t> bitRate)
+    std::optional<int64_t> bitRate,
+    std::optional<int64_t> numChannels)
     : wf_(validateWf(wf)) {
   setFFmpegLogLevel();
   AVFormatContext* avFormatContext = nullptr;
@@ -121,7 +136,7 @@ AudioEncoder::AudioEncoder(
       "avio_open failed: ",
       getFFMPEGErrorStringFromErrorCode(status));
 
-  initializeEncoder(sampleRate, bitRate);
+  initializeEncoder(sampleRate, bitRate, numChannels);
 }
 
 AudioEncoder::AudioEncoder(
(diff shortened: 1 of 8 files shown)


src/torchcodec/encoders/_audio_encoder.py
```diff
@@ -31,23 +31,27 @@ def to_file(
...
         *,
         bit_rate: Optional[int] = None,
+        num_channels: Optional[int] = None,
     ) -> Tensor:
         return _core.encode_audio_to_tensor(
             wf=self._samples,
             sample_rate=self._sample_rate,
             format=format,
             bit_rate=bit_rate,
+            num_channels=num_channels,
```
NicolasHug, line 56: There are no tests for the public API right now. I will soon migrate most of the existing encoder ops tests into testing the public Python APIs.

src/torchcodec/_core/Encoder.h
```diff
@@ -44,6 +50,9 @@ class AudioEncoder {
   UniqueAVCodecContext avCodecContext_;
   int streamIndex_;
   UniqueSwrContext swrContext_;
+  // TODO-ENCODING: desiredNumChannels should just be part of an options struct,
+  // see other TODO above.
+  int desiredNumChannels_ = -1;
 
   const torch::Tensor wf_;
```
scotts, line 57: I think a comment here that `wf` stands for "wave form", and it's the original audio data passed to us by the user would be helpful. I know this is not directly related to the changes in this PR, but I keep having to remind myself of this fact.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -228,7 +233,7 @@ void AudioEncoder::encode() {
   avFrame->format = AV_SAMPLE_FMT_FLTP;
   avFrame->sample_rate = avCodecContext_->sample_rate;
   avFrame->pts = 0;
-  setChannelLayout(avFrame, avCodecContext_);
+  setDefaultChannelLayout(avFrame, static_cast<int>(wf_.sizes()[0]));
```
scotts, line 236: I had to think about this for a few moments to convince myself it's correct, so it may be worth putting in a comment: the _default_ channel layout should be the channel layout of the provided waveform.

src/torchcodec/_core/Encoder.h
NicolasHug: I have a TODO somewhere to rename `wf` to `samples`, like in our Python API. That should make it more obvious

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -88,21 +88,67 @@ void setDefaultChannelLayout(
...
+    // If we can't validate, we must assume it'll be fine. If not, FFmpeg will
+    // eventually raise.
+    return;
+  }
+  for (auto i = 0; avCodec.ch_layouts[i].order != AV_CHANNEL_ORDER_UNSPEC;
+       ++i) {
+    if (numChannels == avCodec.ch_layouts[i].nb_channels) {
+      return;
+    }
+  }
```
scotts, line 115: A comment here saying that we've now entered the error path might be helpful - I think this is less obvious because we're in an `#if` block.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -88,21 +88,67 @@ void setDefaultChannelLayout(
...
+}
 
+void validateNumChannels(const AVCodec& avCodec, int numChannels) {
+#if LIBAVFILTER_VERSION_MAJOR > 7 // FFmpeg > 4
+  if (avCodec.ch_layouts == nullptr) {
+    // If we can't validate, we must assume it'll be fine. If not, FFmpeg will
+    // eventually raise.
+    return;
+  }
+  for (auto i = 0; avCodec.ch_layouts[i].order != AV_CHANNEL_ORDER_UNSPEC;
```
scotts, line 110: I'm not sure that this is correct. I think it will often work, but [the docs say](https://ffmpeg.org/doxygen/6.1/structAVCodec.html#a710e3bd3081124ef3364b0c520379dd8) that `ch_layouts` is:
> Array of supported channel layouts, terminated with a zeroed layout.

That does imply that the terminating struct will have `AV_CHANNEL_ORDER_UNSPEC` for its order because that corresponds to 0 in the `AVChannelOrder` enum, but I think that's also a valid order for a real layout.

I think it may be safer to say `avCodec.ch_layouts[i] != AVChannelLayout{0}`. That's also not obvious, so we could define `const auto emptyAVChannelLayout = AVChannelLayout{0}` and use that in the comparison.

PR #700, opened 2025-05-27
Encoding: allow user-defined encoded sample rate

NicolasHug:
This PR allows the user to specify a desired sample rate for the encoded output.

Unlike other conversions (format, num_channels), this one is far from trivial. It involves, in some specific scenarios, the use of an intermediate FIFO to store samples before they get encoded. I tried to document all of this in the `Note: [Encoding loop, sample rate conversion and FIFO]`, so I recommend starting the review from there.

### DIFF
--- a/src/torchcodec/_core/Encoder.h
+++ b/src/torchcodec/_core/Encoder.h
@@ -95,11 +95,8 @@ class AudioEncoder {
 // │                         - num channels                                                   │
 // │                         - format                                                         │
 // │                         - sample rate                                                    │
-// │                        If sample rate,                                                   │
-// │                        stores data in                                                    │
-// │                        swr buffers                                                       │
-// │                        which will need                                                   │
-// │                        to be flushed                                                     │
+// │                        If sample rate is converted, stores data in swr buffers,          │
+// │                        which will need to be flushed by maybeFlushSwrBuffers()           │
 // │                                                                                          │
 // │                               ▲                                                          │
 // │                               │                 ┌─EncodeFrameThroughFifo()──────────────┐│
@@ -114,7 +111,10 @@ class AudioEncoder {
 // └──────────────────────────────────────────────┼──────────────────────┼────────────────────┘
 //                                                │                      │
 //  AVFrame from  maybeFlushSwrBuffers()       ───┘                      │
-//  Only if sample rate conversion was needed
-//                                                                 nullptr, to flush
-//                                                                 FFmpeg buffers
+//  Only if sample rate conversion was needed                       nullptr, to flush
+//  The call to maybeFlushSwrBuffers() will                         FFmpeg buffers
+//  also instruct to flush the FIFO, if it exists.
+//
+//
+//
 /* clang-format on */


src/torchcodec/encoders/_audio_encoder.py
```diff
@@ -15,7 +15,9 @@ class AudioEncoder:
             tensor of shape ``(num_channels, num_samples)``, or a 1D tensor in
             which case ``num_channels = 1`` is assumed. Values must be float
             values in ``[-1, 1]``.
-        sample_rate (int): The sample rate of the **input** ``samples``.
+        sample_rate (int): The sample rate of the **input** ``samples``. The
+            sample rate of the necoded output can be specified using the
```
scotts: `necoded` -> `encoded`

src/torchcodec/_core/Encoder.h
```diff
@@ -45,13 +40,72 @@ class AudioEncoder {
...
+// │    input                                     │  │ │              │  │                   ││
+// │    samples                                   │  │ │              │  │                   ││
+// │                                              │  │ │              │  │                   ││
+// │                                              │  │ └────► FIFO ───┘  │                   ││
+// │                                              │  └───────────────────┼───────────────────┘│
+// └──────────────────────────────────────────────┼──────────────────────┼────────────────────┘
+//                                                │                      │
+//  AVFrame from  maybeFlushSwrBuffers()       ───┘                      │
+//  Only if sample rate conversion was needed
+//                                                                 nullptr, to flush
```
scotts: This comment is art.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -318,31 +319,84 @@ UniqueAVFrame AudioEncoder::maybeConvertAVFrame(const UniqueAVFrame& avFrame) {
...
+        avFrame->nb_samples,
+        "This is unexpected, please report on the TorchCodec bug tracker.");
+  }
+  return convertedAVFrame;
+}
+
+void AudioEncoder::encodeFrameThroughFifo(
+    AutoAVPacket& autoAVPacket,
+    const UniqueAVFrame& avFrame,
+    bool andFlushFifo) {
```
scotts: Nit: `andFlushFifo` -> `flushFifo`

src/torchcodec/_core/Encoder.cpp
```diff
@@ -381,11 +435,39 @@ void AudioEncoder::encodeInnerLoop(
...
+  UniqueAVFrame avFrame = allocateAVFrame(
+      numRemainingSamples,
+      outSampleRate_,
+      outNumChannels_,
+      avCodecContext_->sample_fmt);
+  int actualNumRemainingSamples = swr_convert(
+      swrContext_.get(), avFrame->data, avFrame->nb_samples, NULL, 0);
+  avFrame->nb_samples = actualNumRemainingSamples;
+
+  encodeFrameThroughFifo(autoAVPacket, avFrame, /*andFlushFifo=*/true);
```
scotts: I had to think about what's going on here for a while, let me know if I'm correct: we sometimes need to encode a frame through a FIFO because of various conversion reasons. We also need to make sure we flush our buffers, sometimes, in order to get the final frames. And those frames _also_ need to sometimes go through the FIFO for the same reasons as above. But when we do the final flushing encoding through a FIFO, we need to do some special logic to make sure we're not leaving a frame behind in the FIFO. Correct?

We have the excellent comment in the header, but I think saying something here and at the top of the while loop in `encodeFrameThroughFifo()` could help readers put all this together.

NicolasHug: Yes that's all correct. Something I may not have been made clear yet, and I will add comments to that end:

The use of the FIFO applies to **all frames**, or to none of them. i.e. a given `Encoder` instance will either:

- create and use the FIFO for every single frame
- not use the FIFO.

I hope I was able to clarify all this a bit in [`fa0856a` (#700)](https://github.com/pytorch/torchcodec/pull/700/commits/fa0856aa71b343476e75dfea6fb5c12e4695426e)

PR #704, opened 2025-05-29
Refactor audio sample conversion in encoder

NicolasHug:
At a high level this PR just moves our AVFrame conversion logic away from `encodeInnerLoop()` into a new `maybeConvertAVFrame()` method, which is now called in `encode()`.

It makes more sense to call this conversion logic *before* entering `encodeInnerLoop()`, and this becomes more evident as I'm working on sample rate conversion in another branch.

This is mainly copy/pasted but the diff isn't trivial and has a few subtle changes that I'll describe in the comments below.

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -282,10 +282,13 @@ void AudioEncoder::encode() {
     // encoded frame would contain more samples than necessary and our results
     // wouldn't match the ffmpeg CLI.
     avFrame->nb_samples = numSamplesToEncode;
-    encodeInnerLoop(autoAVPacket, avFrame);
 
-    avFrame->pts += static_cast<int64_t>(numSamplesToEncode);
+    UniqueAVFrame convertedAVFrame = maybeConvertAVFrame(avFrame);
+    encodeInnerLoop(autoAVPacket, convertedAVFrame);
+
     numEncodedSamples += numSamplesToEncode;
+    // TODO-ENCODING set frame pts correctly, and test against it.
+    // avFrame->pts += static_cast<int64_t>(numSamplesToEncode);
   }
   TORCH_CHECK(numEncodedSamples == numSamples, "Hmmmmmm something went wrong.");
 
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -282,10 +282,13 @@ void AudioEncoder::encode() {
...
     avFrame->nb_samples = numSamplesToEncode;
-    encodeInnerLoop(autoAVPacket, avFrame);
 
-    avFrame->pts += static_cast<int64_t>(numSamplesToEncode);
+    UniqueAVFrame convertedAVFrame = maybeConvertAVFrame(avFrame);
+    encodeInnerLoop(autoAVPacket, convertedAVFrame);
+
     numEncodedSamples += numSamplesToEncode;
+    // TODO-ENCODING set frame pts correctly, and test against it.
+    // avFrame->pts += static_cast<int64_t>(numSamplesToEncode);
```
NicolasHug, line 291: This is a drive-by. I realized that commenting this out was letting our tests pass, so this isn't load bearing. We weren't setting the pts on the convertedAVFrame anyway, so our logic was already wrong. I'm adding this TODO to investigate that later.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -298,42 +301,43 @@ void AudioEncoder::encode() {
...
-    convertedAVFrame = convertAudioAVFrameSamples(
-        swrContext_,
-        srcAVFrame,
+UniqueAVFrame AudioEncoder::maybeConvertAVFrame(const UniqueAVFrame& avFrame) {
+  if (static_cast<AVSampleFormat>(avFrame->format) ==
+          avCodecContext_->sample_fmt &&
+      getNumChannels(avFrame) == outNumChannels_) {
+    // Note: the clone references the same underlying data, it's a cheap copy.
+    return UniqueAVFrame(av_frame_clone(avFrame.get()));
+  }
```
NicolasHug, line 310: The main logic change is this one above: the call to [`av_frame_clone`](https://ffmpeg.org/doxygen/trunk/group__lavu__frame.html#ga46d6d32f6482a3e9c19203db5877105b) when no conversion needs to be done. We want to return the original `avFrame` but we can't use `std::move`, because that would destruct the original `avFrame` prematurely (the one created in `encode()`).

NicolasHug: Thanks for the review! As discussed offline I'll merge now but I want to dig a little bit more about what we could do differently

PR #706, opened 2025-05-30
Add pytest tmp_path

Dan-Flores:
This PR adds a [tmp_path](https://docs.pytest.org/en/stable/how-to/tmp_path.html) fixture to `test_bad_input_parametrized`. By adding this argument, pytest will automatically create a temporary directory to store the decoder in when `to_file` is tested.

As a result, 'output.mp3' is not leftover after running the following test:
`pytest test -vvv -k "test_bad_input_parametrized[to_file]"`


NicolasHug (approved):
Good catch, thank you @Dan-Flores !

Issue #695, opened 2025-05-22
Live stream decoding


scotts:
### 🚀 The feature

Currently, both audio and video decoding depend on knowing both the total number of frames and the duration of the stream. As a consequence, we can't decode live streaming video or audio because we cannot know the total number of frames or the duration.

This feature would be to somehow enable decoding of live media. Because the APIs for `VideoDecoder` and `AudioDecoder` assume the ability to seek to arbitrary places in the stream, I suspect we might need a new top-level public API specifically for live streams.

### Motivation, pitch

On the _training_ side, I suspect this feature will not be used much. That is, training tends to happen on large corpuses of pre-existing videos. I doubt there's much demand to _train_ on a live stream. But doing _inference_ on a live video seems like it would be a common use-case, and we can't currently support it.

FredrikNoren:
We're interested in this. We're doing research on agents in video games, and for that we'd like to have a stream of frames from the game which we send to our policy.

@scotts As a workaround meanwhile; do you think it would work if we set the duration to some arbitrary large number in the header of the video?

scotts:
@FredrikNoren, I think that may work for now. But you'll also need to do the same for the number of frames. Please let know if it succeeds or fails!

I'd like to learn some more about your use-case. Are you basically just grabbing the next N frames that you can on the live stream? Are you ever seeking backward, or seeking forward more than 1 frame?

phillies:
Please put this higher up on the priority list 🥺 🙏 . This is an essential feature for a decoder in production settings, as inference is very often a live stream like you said. Since the torchaudio StreamReader is removed in torch 2.9.0, this missing feature is blocking us from using torch 2.9+.
Our use case is reading live camera feeds (rtsp, to be precise) and decode the frames directly into a GPU tensor.  StreamReader with h264_cuvid decoder was a great way to achieve this in <5ms per frame incl. color conversion from yuv to rgb. Would be great if TorchCodec could offer this in the future, too.

PR #708, opened 2025-05-31
Allow num_frames and duration to be absent in C++ decoder

scotts:
This is necessary to decode live streaming data (#695), but not sufficient. It's necessary because we need the C++ layer to be robust to the number of frames and duration being missing from the metadata, as that will not be available when decoding a live stream. But it's also not sufficient because the Python layer for `VideoDecoder` also asserts that metadata is present.

In fact, because `VideoDecoder` uses the number frames from the metadata in its `__len__()` method, we will need to expose a different public Python API to support live streaming. We should, however, be able to reuse the C++ layer.

I don't particularly love the way I'm currently achieving this, so I'm open to alternatives. I'm also concerned that when the number of frames or duration are missing, we might actually be open to segfaults. I _think_ we should just get the exceptions from the inner decoding loop about not being able to decode frames, but I'm not sure.

Discussion points:

1. Are we comfortable relaxing these requirements at the C++ level? We can (and do) still enforce some of this at the Python level if the Python APIs require this.
2. If yes to 1, is this the best way to do it?
3. We don't have any explicit tests for what happens when this data is not present. We don't have any files which have it missing. We could write some C++ tests that go in an just remove the values from the C++ `VideoDecoder`. That's ugly - is it worth it?

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -602,15 +602,21 @@ FrameBatchOutput SingleStreamDecoder::getFramesInRange(
   const auto& streamMetadata =
       containerMetadata_.allStreamMetadata[activeStreamIndex_];
   const auto& streamInfo = streamInfos_[activeStreamIndex_];
-  int64_t numFrames = getNumFrames(streamMetadata);
   TORCH_CHECK(
       start >= 0, "Range start, " + std::to_string(start) + " is less than 0.");
   TORCH_CHECK(
-      stop <= numFrames,
-      "Range stop, " + std::to_string(stop) +
-          ", is more than the number of frames, " + std::to_string(numFrames));
-  TORCH_CHECK(
       step > 0, "Step must be greater than 0; is " + std::to_string(step));
+
+  // Note that if we do not have the number of frames available in our metadata,
+  // then we assume that the upper part of the range is valid.
+  std::optional<int64_t> numFrames = getNumFrames(streamMetadata);
+  if (numFrames.has_value()) {
+    TORCH_CHECK(
+        stop <= numFrames.value(),
+        "Range stop, " + std::to_string(stop) +
+            ", is more than the number of frames, " +
+            std::to_string(numFrames.value()));
+  }
 
   int64_t numOutputFrames = std::ceil((stop - start) / double(step));
   const auto& videoStreamOptions = streamInfo.videoStreamOptions;
@@ -676,7 +682,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedAt(
       containerMetadata_.allStreamMetadata[activeStreamIndex_];
 
   double minSeconds = getMinSeconds(streamMetadata);
-  double maxSeconds = getMaxSeconds(streamMetadata);
+  std::optional<double> maxSeconds = getMaxSeconds(streamMetadata);
 
   // The frame played at timestamp t and the one played at timestamp `t +
   // eps` are probably the same frame, with the same index. The easiest way to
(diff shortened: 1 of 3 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1456,15 +1481,12 @@ int64_t SingleStreamDecoder::getPts(int64_t frameIndex) {
...
+std::optional<int64_t> SingleStreamDecoder::getNumFrames(
     const StreamMetadata& streamMetadata) {
   switch (seekMode_) {
     case SeekMode::exact:
       return streamMetadata.numFramesFromScan.value();
     case SeekMode::approximate: {
-      TORCH_CHECK(
-          streamMetadata.numFrames.has_value(),
-          "Cannot use approximate mode since we couldn't find the number of frames from the metadata.");
       return streamMetadata.numFrames.value();
```
NicolasHug: I think we mean to return `streamMetadata.numFrames` instead of `streamMetadata.numFrames.value()`?

NicolasHug: Maybe I'm missing something but I'm surprised this compiles. This function seems to always return `int64_t`, never a `std::optional<int64_t>` ?

NicolasHug:
I think it's reasonable to relax the existing requirements when the video doesn't have enough metadata to enforce them. The good news is we haven't had any user report indicating that our hard `TORCH_CHECK` were limiting, so such videos hopefully aren't common in the wild (except for live streams).

On point 2. I'm a bit puzzled about what the PR is doing (see my question above) but overall I don't see a much better alternative.

And re tests / potential segfaults: we definitely want to test this, but I agree that we want to avoid writing some C++ tests for that, it's hard to maintain. We could create an issue to remind us to generate such bad video once we have our own video encoder implemented, it should be easier at this point. Until then, to make sure we don't have segfaults, maybe we could just modify the code locally to e.g.  force numFrames to be `nullops` and see how our tests behave?

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: It's overly broad, but still true. That is, the following should compile just fine:

```
std::optional<int> alwaysOne() {
  return 1;
}
```
It's just that the return value is always an optional with the value 1.

NicolasHug: Ah, it reminds me of this great random number generator:

```py
def get_random_int():
    return 4
```

scotts:
@NicolasHug, I hope the purpose of PR is more clear now that I've corrected my code. :)

And on how to quickly check we don't segfault: good thinking, I just tried that. Specifically, I commented out the lines: https://github.com/pytorch/torchcodec/blob/ae50558cc95877c47d38c88f2ddf93da70f5aa2a/src/torchcodec/_core/SingleStreamDecoder.cpp#L126-L134
That means that even if the metadata is present in the file, we just ignore it. As expected, we get tons of test failures - in particular, constructing a Python `VideoDecoder` object asserts those values are not `None` in the metadata. But no segfaults!

NicolasHug (approved):
Nice, thanks for checking. I also tried the following, to bypass the instantiation failures related to missing metadata:

```diff
diff --git a/src/torchcodec/_core/SingleStreamDecoder.cpp b/src/torchcodec/_core/SingleStreamDecoder.cpp
index f4a285e..2947c64 100644
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -1487,7 +1487,8 @@ std::optional<int64_t> SingleStreamDecoder::getNumFrames(
     case SeekMode::exact:
       return streamMetadata.numFramesFromScan.value();
     case SeekMode::approximate: {
-      return streamMetadata.numFrames;
+      return std::nullopt;
     }
     default:
       throw std::runtime_error("Unknown SeekMode");
@@ -1512,7 +1513,8 @@ std::optional<double> SingleStreamDecoder::getMaxSeconds(
     case SeekMode::exact:
       return streamMetadata.maxPtsSecondsFromScan.value();
     case SeekMode::approximate: {
-      return streamMetadata.durationSeconds;
+      return std::nullopt;
     }
     default:
       throw std::runtime_error("Unknown SeekMode");
```

And all the tests are passing except for 2 tests with the following error, which doesn't seem to be problematic:

```
E       AssertionError: Regex pattern did not match.
E        Regex: 'must be less than'
E        Input: 'Requested next frame while there are no more frames left to decode.'
```

PR #711, opened 2025-06-04
add opencv benchmark

Dan-Flores:
This PR updates the changes in #674, which added a benchmark for decoding with the OpenCV library.

#### Changes in this PR:
- Removed iteration through available backends
- Removed API/ABI compatibility checks
- Added FFMPEG as default backend for OpenCV
- Added option to select `stream_index` in TorchCodecPublic, TorchAudioDecoder

#### Benchmark 1: nasa_13013.mp4 (320x180)
* The file "nasa_13013.mp4" has multiple video streams. This can be seen by calling `ffprobe test/resources/nasa_13013.mp4`.
* For accurate benchmarking, we should ensure the same video streams are being decoded. This was not true for the original benchmark in #674, hence the unexpected results.
  * [OpenCV decoder with FFMPEG backend does not support](https://github.com/opencv/opencv/blob/7ca36d9c508180e6db2a95f2c6cfaaea3750c10a/modules/videoio/src/cap_ffmpeg_impl.hpp#L2093-L2133) setting the video stream, height, or width parameters.
  * The updated benchmark below specifies a `stream_index` to TorchCodecPublic and TorchAudioDecoder. 

```
python benchmarks/decoders/benchmark_decoders.py --decoders torchcodec_public:seek_mode=approximate+stream_index=0,opencv,torchaudio:stream_index=0 --min-run-seconds 40
```
```
[--------------- video=/home/danielflores3/github/Dan-Flores/benchmarks/decoders/../../test/resources/nasa_13013.mp4 h264 480x270, 13.013s 29.97002997002997fps ---------------]
1 threads: ---------------------------------------------------------------------------------------------------------------------------------------------------------------------
      OpenCV[backend=FFMPEG]                                 |            32.9            |            32.0           |        9.9       |        10.9       |        26.6      
      TorchAudio:stream_index=0                              |            92.9            |            89.5           |       10.6       |        11.9       |        31.2      
      TorchCodecPublic:seek_mode=approximate+stream_index=0  |            35.8            |            31.6           |       10.3       |        11.1       |        24.9      

```
image

#### Benchmark 2: mandelbrot_1920x1080_120s.mp4
* To generate a video for this benchmark, the command below was used.
  * The generated video file has one video stream to ensure the decoders select the same video stream. 
```
ffmpeg -y -f lavfi -i mandelbrot=s=1920x1080 -t 120 -c:v h264 -r 60 -g 600 -pix_fmt yuv420p mandelbrot_1920x1080_120s.mp4
```

```
python benchmarks/decoders/benchmark_decoders.py --decoders torchcodec_public:seek_mode=approximate,opencv,torchaudio --min-run-seconds 40  --video-paths mandelbrot_1920x1080_120s.mp4
```
```
[---------------------------------------------- video=mandelbrot_1920x1080_120s.mp4 h264 1920x1080, 120.0s 60.0fps ---------------------------------------------]
1 threads: ------------------------------------------------------------------------------------------------------------------------------------------------------
      OpenCV[backend=FFMPEG]                  |          15314.1           |          15102.9          |       50.9       |        76.6       |       336.6      
      TorchAudio                              |           6798.7           |           8445.0          |       55.0       |        95.9       |       554.7      
      TorchCodecPublic:seek_mode=approximate  |           6824.3           |           5454.0          |       50.7       |        79.2       |       446.6      

```
Screenshot 2025-06-11 at 1 57 23 PM

#### Interpreting the Results:
* The updated results for benchmarking `nasa_13013.mp4` on the lower resolution stream shows similar performance by TorchCodec and OpenCV.
* The results of benchmarking the high resolution video show varied results.
   * OpenCV's decoder sequentially decodes frames. This approach results in lower performance when decoding random and uniform frames, but may be more optimized for decoding frames in order, due to the [grab()](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html#ae38c2a053d39d6b20c9c649e08ff0146) and [retrieve()](https://docs.opencv.org/3.4/d8/dfe/classcv_1_1VideoCapture.html#a9ac7f4b1cdfe624663478568486e6712) functions.
  * TorchCodec decoders perform significantly better than OpenCV at decoding random and uniform frames, but are slower than OpenCV when decoding a high number of sequential frames.


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -146,6 +146,72 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+    def __init__(self, backend):
+        import cv2
+
+        self._available_backends = {"FFMPEG": cv2.CAP_FFMPEG}
+        self._backend = self._available_backends.get(backend)
+
+        self._print_each_iteration_time = False
+
+    def decode_frames(self, video_file, pts_list):
+        import cv2
```
NicolasHug: It's best to only import in `__init__` and then store the module as a `self.cv2` attibute. Otherwise, we'd be benchmarking the `import` statement when calling the decoding method, and it may add noise to the results.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -146,6 +146,72 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+                ret, frame = cap.retrieve()
+                if ret:
+                    frames.append(frame)
+
+            if len(frames) == len(approx_frame_indices):
+                break
+            current_frame += 1
+        cap.release()
+        assert len(frames) == len(approx_frame_indices)
+        return frames
```
NicolasHug, line 185: I suspect what we're getting as output from opencv are numpy arrays. For a fair comparison with the other decoders, we should convert them to pytorch. I think `torch.from_numpy` is what we'd want to use for a fair comparison, as it returns a view and it's cheap.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -146,6 +146,72 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+                frames.append(frame)
+        cap.release()
+        assert len(frames) == n
+        return frames
+
+    def decode_and_resize(self, video_file, pts_list, height, width, device):
+        import cv2
+
+        frames = [
+            cv2.resize(frame, (width, height))
```
NicolasHug: We should note that openCV doesn't apply antialias, while the rest of the `decode_and_resize()` methods apply antialias by default. That makes these methods less comparable.

I would suggest *not* to expose `decode_and_resize` for openCV if we can, so as to prevent any confusion, but if we need to expose it for technical reasons then let's at least add a comment pointing out this discrepancy about antialiasing.


CC @scotts as it's relevant to the whole "resize inconsistency chaos"

NicolasHug (commented):
Thanks for the PR @Dan-Flores ! It looks good, I shared a few comments below. As we just discussed offline, it might be worth checking the output frames for validity, to make sure that the all the decoders we're benchmarking are returning similar frames.

benchmarks/decoders/benchmark_decoders_library.py
Dan-Flores: I confirmed opencv returned numpy arrays, so I added a call to `torch.from_numpy`. While validating that the frames were correct, I had to make a few more adjustments to the color and array order to get the same result as other decoders. I can remove them if we decide they are not needed for the benchmark.

benchmarks/decoders/benchmark_decoders_library.py
Dan-Flores: `decode_and_reize()` is needed to generate data for the README, so I've left a comment here.

Dan-Flores:
> Thanks for the PR @Dan-Flores ! It looks good, I shared a few comments below. As we just discussed offline, it might be worth checking the output frames for validity, to make sure that the all the decoders we're benchmarking are returning similar frames.

cc @NicolasHug  - It seems OpenCV only supports certain frame resolutions, so the benchmark on `nasa13013.mp4` was using a smaller resolution, and was not a fair comparison (see image below). I was unfortunately unable to modify the VideoCapture's resolution - [OpenCV's documentation](https://docs.opencv.org/3.4/d4/d15/group__videoio__flags__base.html#gaeb8dd9c89c10a5c63c139bf7c4f5704d) on these properties suggests that modifying them can fail unexpectedly. 

Screenshot 2025-06-06 at 1 35 19 PM

I ran the benchmark using a generated mandelbrot video at 1920x1080, which OpenCV is able to support.

NicolasHug:
Thanks for checking @Dan-Flores !

I think what's happening with `nasa_13013.mp4 ` is simply that opencv is decoding a different stream! We can see the streams with `ffprobe`:

```
~/dev/torchcodec  » ffprobe test/resources/nasa_13013.mp4                                        
...
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from 'test/resources/nasa_13013.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2avc1mp41
    encoder         : Lavf58.76.100
  Duration: 00:00:13.06, start: 0.000000, bitrate: 412 kb/s
  Stream #0:0[0x1](eng): Video: h264 (High) (avc1 / 0x31637661), yuv420p(tv, bt709, progressive), 320x180 [SAR 1:1 DAR 16:9], 71 kb/s, 25 fps, 25 tbr, 12800 tbn (default)
    Metadata:
      handler_name    : ?Mainconcept Video Media Handler
      vendor_id       : [0][0][0][0]
  Stream #0:1[0x2](eng): Audio: aac (LC) (mp4a / 0x6134706D), 8000 Hz, stereo, fltp, 72 kb/s (default)
    Metadata:
      handler_name    : #Mainconcept MP4 Sound Media Handler
      vendor_id       : [0][0][0][0]
  Stream #0:2[0x3](eng): Subtitle: mov_text (tx3g / 0x67337874), 0 kb/s (default)
    Metadata:
      handler_name    : SubtitleHandler
  Stream #0:3[0x4](eng): Video: h264 (High) (avc1 / 0x31637661), yuv420p(tv, bt709, progressive), 480x270 [SAR 1:1 DAR 16:9], 128 kb/s, 29.97 fps, 29.97 tbr, 30k tbn (default)
    Metadata:
      handler_name    : ?Mainconcept Video Media Handler
      vendor_id       : [0][0][0][0]
  Stream #0:4[0x5](eng): Audio: aac (LC) (mp4a / 0x6134706D), 16000 Hz, stereo, fltp, 128 kb/s (default)
    Metadata:
      handler_name    : #Mainconcept MP4 Sound Media Handler
      vendor_id       : [0][0][0][0]
  Stream #0:5[0x6](eng): Subtitle: mov_text (tx3g / 0x67337874), 0 kb/s (default)
    Metadata:
      handler_name    : SubtitleHandler
```

the mp4 containers has 2 video streams: one with 320x180 resolution and one with 480x270 resolution. OpenCV decodes the first one while we decode the second one, because it's what ffmpeg considers to be the ["best" stream](https://ffmpeg.org/doxygen/2.0/group__lavf__decoding.html#gaa6fa468c922ff5c60a6021dcac09aff9).
So to get a fair comparison we just need to tell opencv to decode the same stream.

And BTW, we probably want to double check we were decoding the same streams with our other benchmark backends (torchaudio etc.), but that can be done separately.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -828,7 +829,7 @@ def run_benchmarks(
         # are using different random pts values across videos.
         random_pts_list = (torch.rand(num_samples) * duration).tolist()
 
-        for decoder_name, decoder in decoder_dict.items():
+        for decoder_name, decoder in sorted(decoder_dict.items(), key=lambda x: x[0]):
```
Dan-Flores, line 857: This change was added to make it easier to compare benchmarks, I'm open to alternative approaches to sorting or removing it entirely.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -146,6 +146,80 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+            if not ok:
+                raise ValueError("Could not grab video frame")
+            if current_frame in approx_frame_indices:  # only decompress needed
+                ret, frame = cap.retrieve()
+                if ret:
+                    # OpenCV uses BGR, change to RGB
+                    frame = self.cv2.cvtColor(frame, self.cv2.COLOR_BGR2RGB)
+                    # Update to C, H, W
+                    frame = np.transpose(frame, (2, 0, 1))
+                    frame = torch.from_numpy(frame)
```
NicolasHug: Nit: move this "BGR array --> RGB tensor" logic in a separate helper that we can re-use across methods

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -293,10 +367,17 @@ def decode_first_n_frames(self, video_file, n):
...
+        self,
+        num_ffmpeg_threads=None,
+        device="cpu",
+        seek_mode="exact",
+        stream_index=None,
+    ):
         self._num_ffmpeg_threads = num_ffmpeg_threads
         self._device = device
         self._seek_mode = seek_mode
+        self._stream_index = int(stream_index) if stream_index else None
```
NicolasHug, line 392: I was a bit surprised to see this written like this, because if `stream_index` is `0` then it becomes `None`, which isn't what we want. Then I realized `stream_index` is a `str` (!), not an `int`, so the logic is sound.

I rarely advocate for type annotations but in this case, I think it could help to annotate the `stream_index` parameter of `__init__` as `Optional[str]`. We might as well do the same for `num_ffmpeg_threads`. (same below)

benchmarks/decoders/benchmark_decoders_library.py
scotts: Let's keep this, as it ensures that we always get the results printed in the same order. But a comment stating that is good, since it may not be obvious why. (We're actually _doing_ the experiments in the same order every time, which means we populate the results in the same way, which makes them displayed in the same way.)

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -146,6 +146,80 @@ def decode_and_resize(self, video_file, pts_list, height, width, device):
...
+        assert len(frames) == n
+        return frames
+
+    def decode_and_resize(self, video_file, pts_list, height, width, device):
+
+        # OpenCV doesn't apply antialias, while other `decode_and_resize()` implementations apply antialias by default.
+        frames = [
+            self.cv2.resize(frame, (width, height))
+            for frame in self.decode_frames(video_file, pts_list)
+        ]
```
scotts: @Dan-Flores, @NicolasHug, I actually agree that we should not expose this until we apply antialias. It's fine that we won't be able to generate a README graph before we do that. I'd rather that we fail when trying to do a bogus comparison than succeed and not realize it's a bogus comparison. :)

NicolasHug: SGTM, instead of *not* implementing the method, I suggest to define it and just raise, so that we remember why we didn't implement it in the first place:

```py
def decode_and_resize(self, *args, **kwags*):
    raise ValueError("OpenCV doesn't apply antialias while pytorch does by default, this is potentially an unfair comparison")
```

PR #717, opened 2025-06-09
Add Docs for AudioEncoder

NicolasHug:
This PR adds docstrings and a tutorial for the `AudioEncoder`, plus some minor changes listed in the comments below


docs/source/conf.py
NicolasHug, line 1: Changed in this file, as well as the files renaming, are meant to separate our ["tutorials" page](https://docs.pytorch.org/torchcodec/stable/generated_examples/index.html) into 2 separate sections: one for decoding, one for encoding.

src/torchcodec/encoders/_audio_encoder.py
```diff
@@ -16,8 +26,11 @@ def __init__(self, samples: Tensor, *, sample_rate: int):
             raise ValueError(
                 f"Expected samples to be a Tensor, got {type(samples) = }."
             )
+        if samples.ndim == 1:
+            # make it 2D and assume 1 channel
+            samples = samples[None, :]
```
NicolasHug: Drive-by, I think this makes sense, i.e. if the input tensor is 1D we assume it's 1 channel instead of raising an error

README.md
```diff
@@ -3,17 +3,18 @@
...
+videos and audio, TorchCodec is how you turn these into data.
 
 We achieve these capabilities through:
 
 * Pythonic APIs that mirror Python and PyTorch conventions.
-* Relying on [FFmpeg](https://www.ffmpeg.org/) to do the decoding. TorchCodec
-  uses the version of FFmpeg you already have installed. FFmpeg is a mature
-  library with broad coverage available on most systems. It is, however, not
-  easy to use. TorchCodec abstracts FFmpeg's complexity to ensure it is used
+* Relying on [FFmpeg](https://www.ffmpeg.org/) to do the decoding / encoding.
```
scotts: Nit: I prefer "x and y" as opposed to "x / y" in prose.

docs/source/conf.py
```diff
@@ -68,18 +68,24 @@ class CustomGalleryExampleSortKey:
...
+                "file_like.py",
+                "approximate_mode.py",
+                "sampling.py",
+            ]
+        else:
+            assert "examples/encoding" in self.src_dir
+            order = [
+                "audio_encoding.py",
+            ]
+
```
scotts, line 89: Can we add a comment explaining that we have two top-level galleries, and for that reason, we need to figure out which gallery we're using (decoding versus encoding)? I was real confused until I concluded that must be what's going on.

scotts:
Moving the examples into decoding and encoding subdirectories makes sense, but I'm curious: will that change the resulting tutorial URL? We already have some pointers to these tutorials floating around blogs and social media.

src/torchcodec/encoders/_audio_encoder.py
scotts: Is that the most idomatic way to do that? There's also unsqueeze, view or reshape. Minor, but I'm surprised at how this works.

scotts (approved):

examples/encoding/audio_encoding.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+samples, sample_rate = make_sinewave()
+
+print(f"Encoding samples with {samples.shape = } and {sample_rate = }")
+play_audio(samples, rate=sample_rate)
+
+# %%
+# We first instantiate an :class:`~torchcodec.encoders.AudioEncoder`. We pass it
+# the samples to be encoded. The samples must a 2D tensors of shape
```
Dan-Flores: Nit: "The samples must **be** a 2D tensors of shape"

src/torchcodec/encoders/_audio_encoder.py
NicolasHug: It's an idiom that comes from `numpy`, indexing with `None` just adds "empty" dimensions, it's an alias for `np.newaxis` which is slighty more explicit: https://numpy.org/devdocs//user/basics.indexing.html#dimensional-indexing-tools

I personally find it idiomatic because... I'm used to it.

I always have to look-up what `unsqueeze` does because I forget, but I suspect it's more commonly used. I'll switch to that.

PR #721, opened 2025-06-11
Turn BytesContext into FromTensorContext

### LINKED ISSUE #720: Creating a decoder from bytes does not adopt the memory
### 🐛 Describe the bug

The following can run into problems:
```python
import torch
import torchcodec

def create_decoder():
  file = open("video.mp4", "rb")
  return VideoDecoder(file.read())

decoder = create_decoder()
print(decoder[0])  # we may hit a runtime error
```
We're creating a `VideoDecoder` by providing the actual bytes of the video. The problem is that the `VideoDecoder` does _not_ take ownership of those bytes. So in the example above, the bytes object returned by `read()` goes out of scope outside of the `create_decoder()` call. Because of that, it's entirely possible (likely?) that the bytes are garbage collected. Then when we try to do actual decoding, we can hit a runtime error.

The solution is that we _should_ take ownership of the bytes. The pattern above is quite reasonable and common. We need to fix this by making sure that the C++ `VideoDecoder` takes proper ownership of the memory. We already pass the bytes in as a tensor:
https://github.com/pytorch/torchcodec/blob/539c9475d00795b4b81371bd0314d22433c301ae/src/torchcodec/_core/custom_ops.cpp#L199-L207
We can probably fix this by making the `AVIOBytesContext` actually adopt the tensor itself, rather than just a pointer to the bytes. Although this then re-raises the question if we should just merge that with `AVIOToTensorContext`.

### Versions

This affects version 0.4 and all prior.

scotts:
Fixes #720.

Reading from a raw `void*` is problematic because we have no way of claiming ownership of it. This PR makes sure reading from bytes only happens through a tensor. The tensor is effectively a smart pointer, so it ensures that we actually keep the data around. We now have both a tensor-reading context and a tensor-writing context. This PR also refactors both to share code - which we thought we may eventually need to do.

The test added by this PR fails without these changes in the C++ level.

### DIFF
--- a/src/torchcodec/_core/AVIOBytesContext.h
+++ b/src/torchcodec/_core/AVIOBytesContext.h
@@ -11,23 +11,19 @@
 
 namespace facebook::torchcodec {
 
+struct TensorContext {
+  torch::Tensor data;
+  int64_t current;
+};
+
 // For Decoding: enables users to pass in the entire video or audio as bytes.
 // Our read and seek functions then traverse the bytes in memory.
-class AVIOBytesContext : public AVIOContextHolder {
+class AVIOFromTensorContext : public AVIOContextHolder {
  public:
-  explicit AVIOBytesContext(const void* data, int64_t dataSize);
+  explicit AVIOFromTensorContext(torch::Tensor data);
 
  private:
-  struct DataContext {
-    const uint8_t* data;
-    int64_t size;
-    int64_t current;
-  };
-
-  static int read(void* opaque, uint8_t* buf, int buf_size);
-  static int64_t seek(void* opaque, int64_t offset, int whence);
-
-  DataContext dataContext_;
+  TensorContext tensorContext_;
 };
 
 // For Encoding: used to encode into an output uint8 (bytes) tensor.
@@ -37,18 +33,7 @@ class AVIOToTensorContext : public AVIOContextHolder {
   torch::Tensor getOutputTensor();
 
  private:
-  struct DataContext {
-    torch::Tensor outputTensor;
-    int64_t current;
-  };
-
-  static constexpr int64_t INITIAL_TENSOR_SIZE = 10'000'000; // 10MB
-  static constexpr int64_t MAX_TENSOR_SIZE = 320'000'000; // 320 MB
-  static int write(void* opaque, const uint8_t* buf, int buf_size);
-  // We need to expose seek() for some formats like mp3.
-  static int64_t seek(void* opaque, int64_t offset, int whence);
-
-  DataContext dataContext_;
+  TensorContext tensorContext_;
 };
 
 } // namespace facebook::torchcodec
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -196,15 +196,13 @@ at::Tensor create_from_tensor(
   TORCH_CHECK(
       video_tensor.scalar_type() == torch::kUInt8,
       "video_tensor must be kUInt8");
-  void* data = video_tensor.mutable_data_ptr();
-  size_t length = video_tensor.numel();
 
   SingleStreamDecoder::SeekMode realSeek = SingleStreamDecoder::SeekMode::exact;
   if (seek_mode.has_value()) {
     realSeek = seekModeFromString(seek_mode.value());
   }
 
-  auto contextHolder = std::make_unique<AVIOBytesContext>(data, length);
+  auto contextHolder = std::make_unique<AVIOFromTensorContext>(video_tensor);
 
   std::unique_ptr<SingleStreamDecoder> uniqueDecoder =
       std::make_unique<SingleStreamDecoder>(std::move(contextHolder), realSeek);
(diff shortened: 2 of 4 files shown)


src/torchcodec/_core/AVIOBytesContext.h
```diff
@@ -11,23 +11,19 @@
 
 namespace facebook::torchcodec {
 
+struct TensorContext {
+  torch::Tensor data;
+  int64_t current;
+};
+
```
scotts: I don't love that this implementation detail is now leaked.

src/torchcodec/_core/AVIOTensorContext.h
```diff
@@ -0,0 +1,39 @@
...
+#include <torch/types.h>
+#include "src/torchcodec/_core/AVIOContextHolder.h"
+
+namespace facebook::torchcodec {
+
+struct TensorContext {
+  torch::Tensor data;
+  int64_t current;
+};
+
```
scotts, line 20: I don't love that this is exposed outside of this file.

test/test_decoders.py
```diff
@@ -126,6 +126,27 @@ def test_metadata(self, seek_mode):
         assert decoder.metadata.height == 270
         assert decoder.metadata.width == 480
 
+    def test_create_bytes_ownership(self):
+        # Note that the bytes object we use to instantiate the decoder does not
```
NicolasHug:
```suggestion
    def test_create_bytes_ownership(self):
        # Non-regression test for https://github.com/pytorch/torchcodec/issues/720
        # Note that the bytes object we use to instantiate the decoder does not
```

test/test_decoders.py
```diff
@@ -126,6 +126,27 @@ def test_metadata(self, seek_mode):
...
+        # side has gone away, and if we don't have ownership on the C++ side, we
+        # will hit runtime errors or segfaults.
+        #
+        # Also note that if this test fails, OTHER tests will likely
+        # mysteriously fail. That's because a failure in this tests likely
+        # indicates memory corruption, and the memory we corrupt could easily
+        # cause problems in other tests. So if this test fails, fix this test
+        # first.
+        with open(NASA_VIDEO.path, "rb") as f:
+            decoder = VideoDecoder(f.read())
```
NicolasHug, line 147: I think we should call `gc.collect()` here, to force the GC to run and free the resources? It won't change anything in this branch, but in `main`, I think this would deterministically ensure that the resource gets freed. I may be wrong but I suspect there might be cases where this test *doesn't* segfault on `main` simply because the GC didn't collect things yet.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -196,15 +196,13 @@ at::Tensor create_from_tensor(
...
-  void* data = video_tensor.mutable_data_ptr();
-  size_t length = video_tensor.numel();
 
   SingleStreamDecoder::SeekMode realSeek = SingleStreamDecoder::SeekMode::exact;
   if (seek_mode.has_value()) {
     realSeek = seekModeFromString(seek_mode.value());
   }
 
-  auto contextHolder = std::make_unique<AVIOBytesContext>(data, length);
+  auto contextHolder = std::make_unique<AVIOFromTensorContext>(video_tensor);
```
NicolasHug, line 205: Can you confirm my understanding that the crux of the fix is this line, where we now pass-in the tensor which is ref-counted, instead of passing the raw underlying data which wasn't ref-counted and thus freed?

src/torchcodec/_core/AVIOTensorContext.h
NicolasHug: Curious how we could prevent it from leaking out? A base class with a definition as `private`? That sounds overkill?

src/torchcodec/_core/AVIOTensorContext.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+} // namespace
+
+AVIOFromTensorContext::AVIOFromTensorContext(torch::Tensor data)
+    : tensorContext_{data, 0} {
+  TORCH_CHECK(data.numel() > 0, "data must not be empty");
+  TORCH_CHECK(data.is_contiguous(), "data must be contiguous");
+  TORCH_CHECK(data.scalar_type() == torch::kUInt8, "data must be kUInt8");
+  createAVIOContext(&read, nullptr, &seek, &tensorContext_);
```
NicolasHug, line 108: Nit, since `read` etc. aren't members anymore, we may want to update this comment:

https://github.com/pytorch/torchcodec/blob/539c9475d00795b4b81371bd0314d22433c301ae/src/torchcodec/_core/AVIOContextHolder.h#L30-L31

NicolasHug (approved):
Thanks for the fix!!

NicolasHug (approved):
Thanks for the fix!!

src/torchcodec/_core/custom_ops.cpp
scotts: Yes, that is correct!

src/torchcodec/_core/AVIOTensorContext.h
scotts: The struct itself would still be visible to those that included the .h file. I may be able to put it in a `detail` namespace. Let me try something like that.

PR #725, opened 2025-06-12
Replacing throw with TORCH_CHECK

Krishn1412:
Replaced all the occurrences of throw in the cpp code with TORCH_CHECK.

TORCH_CHECK throws a runtime error, so the cpp test suite had to be updated as well.


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -446,13 +441,12 @@ void SingleStreamDecoder::addVideoStream(
...
-      !streamMetadata.averageFpsFromHeader.has_value()) {
-    throw std::runtime_error(
-        "Seek mode is approximate, but stream " +
-        std::to_string(activeStreamIndex_) +
-        " does not have an average fps in its metadata.");
-  }
+  TORCH_CHECK(
+      !(seekMode_ == SeekMode::approximate &&
+        !streamMetadata.averageFps.has_value()),
+      "Seek mode is approximate, but stream ",
```
scotts: I find this boolean expression much harder to reason about than the previous one. It's wordier, but let's instead do:
```c++
if (seekMode_ == SeekMode::approximate) {
  TORCH_CHECK(
    streamMetadata.averageFpsFromHeader.has_value(),
    "Same message as before");
}
```
I prefer that, as it reads closer to what we would say out loud: if the seek mode is approximate, assert that it has a value for fps.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1147,23 +1142,22 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
...
-      throw SingleStreamDecoder::EndOfFileException(
-          "Requested next frame while there are no more frames left to "
-          "decode.");
-    }
-    throw std::runtime_error(
-        "Could not receive frame from decoder: " +
+    TORCH_CHECK(
+        !(reachedEOF || status == AVERROR_EOF),
+        "Requested next frame while there are no more frames left to decode.");
+
```
scotts: We have to revert this check back to what it was previously. We turn the `EndOfFileException` into a Python `IndexError` in the Python-to-C++ layer:
https://github.com/pytorch/torchcodec/blob/f4a351c9da37eb976ea583b0f10785b86fe6f1a8/src/torchcodec/_core/custom_ops.cpp#L303-L307

scotts:
@Krishn1412, thanks for working on this! You're getting test failures because you'll need to pull changes from main - we changed a bunch of variable names in the past few days. There's also likely going to be some test failures related to not throwing the `EndOfFileException` that will need to be fixed.

src/torchcodec/_core/SingleStreamDecoder.cpp
Krishn1412: Should I change the custom_ops.cpp file instead to handle it using TORCH_CHECK runtime errors?

scotts: No, we should keep what's going on in custom_ops.cpp as-is. It has the behavior we want, and we're already using the most relevant PyTorch APIs to achieve it.

Krishn1412: got it!

scotts:
@Krishn1412, it looks like we're still getting some errors related to the exceptions we're now throwing versus the errors we expect to see. If you follow the [contributing guide](https://github.com/pytorch/torchcodec/blob/main/CONTRIBUTING.md), you can actually run all of these tests locally, too.

Krishn1412:
@scotts, thanks for the opportunity. Looking forward to more contributions!

PR #726, opened 2025-06-13
Encoder: validate pts of encoded frames

NicolasHug:


test/test_encoders.py
```diff
@@ -162,12 +261,22 @@ def test_against_cli(self, asset, bit_rate, num_channels, format, method, tmp_pa
...
             atol=atol,
         )
+        assert samples_by_us.pts_seconds == samples_by_ffmpeg.pts_seconds
+        assert samples_by_us.duration_seconds == samples_by_ffmpeg.duration_seconds
+        assert samples_by_us.sample_rate == samples_by_ffmpeg.sample_rate
+
+        if method == "to_file":
+            validate_frames_properties(actual=encoded_by_us, expected=encoded_by_ffmpeg)
+        else:
+            assert method == "to_tensor", "wrong test parametrization!"
```
NicolasHug, line 281: Interestingly enough, the call to `validate_frames_properties()` was passing on `main`. That's because the underlying encoder was able to infer `pts` valid values. But it was emitting warnings that were visible in DEBUG mode.

The other tests we have above (`assert ... not in captured.err`) make sure those warnings aren't emitted anymore, confirming we are now correctly setting the pts values explicitly.

PR #728, opened 2025-06-16
Verify decoder outputs

Dan-Flores:
#### Overview
While adding the OpenCV decoder in #711, we realized that OpenCV was decoding a different stream than TorchCodec, resulting in unexpected benchmark results. This PR implements a function to compare the outputs of decoders. 

#### Example usage
By adding `--verify-outputs` to the benchmark command, the script outputs a warning indicating that the output frames are different for TorchCodec in approximate mode compared to OpenCV on `nasa_13013.mp4`:
```
python benchmarks/decoders/benchmark_decoders.py --decoders torchaudio,torchcodec_public:seek_mode=approximate,opencv   --verify-outputs 
```
Screenshot 2025-06-13 at 4 19 47 PM

```
Error while comparing TorchCodecPublic:seek_mode=approximate and OpenCV[backend=FFMPEG]: The values for attribute 'shape' do not match: torch.Size([3, 270, 480]) != torch.Size([3, 180, 320]).
```

The dimensions match when `stream_index` is utilized, but the tensors are unequal:
```
python benchmarks/decoders/benchmark_decoders.py --decoders torchaudio:stream_index=0,torchcodec_public:seek_mode=approximate+stream_index=0,opencv   --verify-outputs --video-paths output.mp4
```
```

AssertionError: Tensor-likes are not equal!

Mismatched elements: 38759 / 172800 (22.4%)
Greatest absolute difference: 16 at index (0, 98, 96)
Greatest relative difference: inf at index (0, 44, 210)
```

For further testing, I generated a video using ffmpeg, where all results matched:
```
ffmpeg -f lavfi -i testsrc=duration=10:size=1280x720:rate=30 -vf "drawtext=fontfile=/path/to/font.ttf: text='%{pts\:hms}': x=(w-text_w)/2: y=(h-text_h)/2: fontsize=48: fontcolor=white: box=1: boxcolor=black@0.5: boxborderw=5, hue=H=2*t" -pix_fmt yuv420p output.mp4
```
```
python benchmarks/decoders/benchmark_decoders.py --decoders torchaudio:stream_index=0,torchcodec_public:seek_mode=approximate+stream_index=0,opencv   --verify-outputs --video-paths output.mp4
```
```
video=output.mp4, decoder=OpenCV[backend=FFMPEG]
Results of baseline TorchCodecPublic and OpenCV[backend=FFMPEG] match!
video=output.mp4, decoder=TorchAudio:stream_index=0
Results of baseline TorchCodecPublic and TorchAudio:stream_index=0 match!
video=output.mp4, decoder=TorchCodecPublic:seek_mode=approximate+stream_index=0
Results of baseline TorchCodecPublic and TorchCodecPublic:seek_mode=approximate+stream_index=0 match!
```

### DIFF
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -8,65 +8,16 @@ import argparse
 import importlib.resources
 import os
 import platform
-import typing
-from dataclasses import dataclass, field
 from pathlib import Path
 
 import torch
 
 from benchmark_decoders_library import (
-    AbstractDecoder,
-    DecordAccurate,
-    DecordAccurateBatch,
-    OpenCVDecoder,
+    decoder_registry,
     plot_data,
     run_benchmarks,
-    TorchAudioDecoder,
-    TorchCodecCore,
-    TorchCodecCoreBatch,
-    TorchCodecCoreCompiled,
-    TorchCodecCoreNonBatch,
-    TorchCodecPublic,
-    TorchCodecPublicNonBatch,
-    TorchVision,
+    verify_outputs,
 )
-
-
-@dataclass
-class DecoderKind:
-    display_name: str
-    kind: typing.Type[AbstractDecoder]
-    default_options: dict[str, str] = field(default_factory=dict)
-
-
-decoder_registry = {
-    "decord": DecoderKind("DecordAccurate", DecordAccurate),
-    "decord_batch": DecoderKind("DecordAccurateBatch", DecordAccurateBatch),
-    "torchcodec_core": DecoderKind("TorchCodecCore", TorchCodecCore),
-    "torchcodec_core_batch": DecoderKind("TorchCodecCoreBatch", TorchCodecCoreBatch),
-    "torchcodec_core_nonbatch": DecoderKind(
-        "TorchCodecCoreNonBatch", TorchCodecCoreNonBatch
-    ),
-    "torchcodec_core_compiled": DecoderKind(
-        "TorchCodecCoreCompiled", TorchCodecCoreCompiled
-    ),
-    "torchcodec_public": DecoderKind("TorchCodecPublic", TorchCodecPublic),
-    "torchcodec_public_nonbatch": DecoderKind(
-        "TorchCodecPublicNonBatch", TorchCodecPublicNonBatch
-    ),
-    "torchvision": DecoderKind(
-        # We don't compare against TorchVision's "pyav" backend because it doesn't support
-        # accurate seeks.
-        "TorchVision[backend=video_reader]",
-        TorchVision,
-        {"backend": "video_reader"},
-    ),
-    "torchaudio": DecoderKind("TorchAudio", TorchAudioDecoder),
-    "opencv": DecoderKind(
-        "OpenCV[backend=FFMPEG]", OpenCVDecoder, {"backend": "FFMPEG"}
-    ),
-}
-
 
 def in_fbcode() -> bool:
     return "FB_PAR_RUNTIME_FILES" in os.environ
@@ -148,6 +99,12 @@ def main() -> None:
         type=str,
         default="benchmarks.png",
     )
+    parser.add_argument(
+        "--verify-outputs",
+        help="Verify that the outputs of the decoders are the same",
+        default=False,
+        action=argparse.BooleanOptionalAction,
+    )
 
     args = parser.parse_args()
     specified_decoders = set(args.decoders.split(","))
@@ -176,6 +133,10 @@ def main() -> None:
         for entry in os.scandir(args.video_dir):
             if entry.is_file() and entry.name.endswith(".mp4"):
                 video_paths.append(entry.path)
+
+    if args.verify_outputs:
+        verify_outputs(decoders_to_run, video_paths, num_uniform_samples)
+        return
 
     results = run_benchmarks(
         decoders_to_run,
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -136,7 +136,6 @@ def main() -> None:
 
     if args.verify_outputs:
         verify_outputs(decoders_to_run, video_paths, num_uniform_samples)
-        return
 
     results = run_benchmarks(
         decoders_to_run,
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -136,7 +136,7 @@ def main() -> None:
                 video_paths.append(entry.path)
 
     if args.verify_outputs:
-        verify_outputs(decoders_to_run, video_paths, 1)
+        verify_outputs(decoders_to_run, video_paths, num_uniform_samples)
     else:
         results = run_benchmarks(
             decoders_to_run,
--- a/benchmarks/decoders/benchmark_decoders.py
+++ b/benchmarks/decoders/benchmark_decoders.py
@@ -19,6 +19,7 @@ from benchmark_decoders_library import (
     verify_outputs,
 )
 
+
 def in_fbcode() -> bool:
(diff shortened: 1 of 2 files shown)


benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1025,97 @@ def run_benchmarks(
...
+        # Compare the frames from all decoders to the frames from TorchCodecPublic
+        for curr_decoder_name, curr_decoder_frames in decoders_and_frames:
+            assert len(torchcodec_public_results) == len(curr_decoder_frames)
+            all_match = True
+            for f1, f2 in zip(torchcodec_public_results, curr_decoder_frames):
+                # Validate that the frames are the same with a tolerance
+                try:
+                    torch.testing.assert_close(f1, f2)
+                except Exception as e:
+                    tensorcat(f1)
```
Dan-Flores: This library is useful for visually comparing frames, but I am open to removing it since it is not necessary.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1025,97 @@ def run_benchmarks(
     compare = benchmark.Compare(results)
     compare.print()
     return df_data
+
+
+def verify_outputs(decoders_to_run, video_paths, num_samples):
+    # Import library to show frames that don't match
+    from tensorcat import tensorcat
+
+    # Reuse TorchCodecPublic decoder with options, if provided, as the baseline
```
Dan-Flores: Is `TorchCodecPublic` the correct decoder to use as the baseline?

scotts: Good question - yes, I think so. We have a large battery of unit tests testing for bit-for-bit correctness on TorchCodec, and it's easier to test on the public API.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1024,97 @@ def run_benchmarks(
...
+    if torchcodec_display_name := next(
+        (name for name in decoders_to_run if "TorchCodecPublic" in name),
+        None,
+    ):
+        torchcodec_public_decoder = decoders_to_run[torchcodec_display_name]
+    # Create default TorchCodecPublic decoder to use as a baseline
```
scotts, line 1045: This means that the reference decoder will be subject to the options that the user provides, such as `seek_mode`. I think we shouldn't try to use the options the user provided, but instead decide what the reference decoder is, and always use that. That means that we probably shouldn't use the default options, but decide what options we want to use. Regarding `seek_mode`, I think we should probably use `exact` as the reference.

Dan-Flores: One catch is that given the new 'stream_index' option, we rely on the user to indicate which stream the benchmarks should compare. Without this option, the benchmark for OpenCV vs TorchCodecPublic would not match.

I agree that `exact` should be used as the reference. I've updated the code to only reuse the `stream_index` argument.

benchmarks/decoders/benchmark_decoders_library.py
scotts: Yeah, I think we should remove it. Even though most users won't run the benchmarks, we still want to keep dependencies down.

I also think we can simplify this part, and just do the plain assertion. For this kind of validation, it's generally better to get a failure than it is to just print the result to stdout. Our CI, and a lot of other monitoring infra, will use the process exit code to determine if there was a problem. Doing what we're currently doing will mean that we'd need to parse stdout to figure out if there was a status if we wanted to hook this up to CI.

benchmarks/decoders/benchmark_decoders.py
```diff
@@ -177,6 +135,9 @@ def main() -> None:
             if entry.is_file() and entry.name.endswith(".mp4"):
                 video_paths.append(entry.path)
 
+    if args.verify_outputs:
+        verify_outputs(decoders_to_run, video_paths, num_uniform_samples)
+
```
scotts: I think this option is most useful if it's mutually exclusive with running the actual benchmarks. That way someone can specify it to quickly test the benchmark's correctness. So here I think we should make running the benchmarks and printing the benchmark results the else branch here.

benchmarks/decoders/benchmark_decoders_library.py
scotts: Oh! But we want to use the stream_index that the user provided. Good call. :)

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1024,99 @@ def run_benchmarks(
...
+    # Get frames using each decoder
+    for video_file_path in video_paths:
+        metadata = get_metadata(video_file_path)
+        metadata_label = f"{metadata.codec} {metadata.width}x{metadata.height}, {metadata.duration_seconds}s {metadata.average_fps}fps"
+        print(f"{metadata_label=}")
+
+        # Generate uniformly random PTS
+        duration = metadata.duration_seconds
+        pts_list = [i * duration / num_samples for i in range(num_samples)]
+
```
scotts, line 1057: Technically this is uniformly-spaced PTS values, or just evenly-spaced PTS values. Uniformly random would look like `pts_list = (torch.rand(num_samples) * duration).tolist()`.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1024,99 @@ def run_benchmarks(
...
+
+def decode_and_adjust_frames(
+    decoder, video_file_path, *, num_samples: int, pts_list: list[float]
+):
+    frames = []
+    # Decode non-sequential frames using decode_frames function
+    random_frames = decoder.decode_frames(video_file_path, pts_list)
+    # Extract the frames from the FrameBatch if necessary
+    if isinstance(random_frames, FrameBatch):
```
scotts: On comments: often the _what_ is something a reader can easily deduce. It's the _why_ that usually needs a comment. In this case, the why is that TorchCodec's batch APIs return a `FrameBatch`. We sometimes use these APIs in our experiments, and we just return that directly. But for all other decoders, we just return a list of frames.

benchmarks/decoders/benchmark_decoders_library.py
```diff
@@ -986,3 +1024,99 @@ def run_benchmarks(
...
+            print(f"video={video_file_path}, decoder={decoder_name}")
+
+            frames = decode_and_adjust_frames(
+                decoder,
+                video_file_path,
+                num_samples=num_samples,
+                pts_list=pts_list,
+            )
+            decoders_and_frames.append((decoder_name, frames))
+
```
scotts: Since we just want to assert that the frames are close (see comment below), I think we can simplify even further and just do that assertion here. Then we don't need to do a separate loop over `decoders_and_frames`. In fact, I think we don't even need to record `decoders_and_frames`, as we don't need to remember the frames after the assertion.

scotts (changes requested):
Thanks for this! I think addressing the remaining comments will simplify the verification and make it easier for us to hook this up to CI (eventually, if we want to).

benchmarks/decoders/benchmark_decoders_library.py
Dan-Flores: Thanks for the feedback, I'll remove the library and keep that in mind going forward.

benchmarks/decoders/benchmark_decoders_library.py
Dan-Flores: Good point, I was able to significantly simplify this portion.

scotts (approved):

scotts:
This is great! We should have been doing this all along. :) Let's go ahead and merge - I'm confident that the testing itself is correct, and I suspect we have a bug in how we're decoding with OpenCV. Let's create a new issue to investigate the problem with how we're using OpenCV, and label it as a "bug."

Issue #727, opened 2025-06-13
Approximate mode video decoding should tolerate one of duration or num_frames missing


scotts:
The Python `VideoDecoder` needs to know how many frames are in a stream. It asserts this at startup:
https://github.com/pytorch/torchcodec/blob/dd44f57180070572eee9e7d35d38f9c03569f6f9/src/torchcodec/decoders/_video_decoder.py#L387-L390
It needs to know this because the number of frames are the length of the decoder as a sequence:
https://github.com/pytorch/torchcodec/blob/dd44f57180070572eee9e7d35d38f9c03569f6f9/src/torchcodec/decoders/_video_decoder.py#L122-L123
Some videos are missing the number of frames in their metadata. For exact mode, this is not a problem because we compute the number of frames when we scan the stream. For approximate mode, we currently cannot instantiate a decoder for such videos. Approximate mode also requires the FPS - that's how we figure out indices. So for a video that is missing its number of frames, if it has its FPS and time duration, we should be able to compute the number of frames.

In terms of implementation, we should hide all of this inside of the metadata class rather than changing any logic in `VideoDecoder`. We already make `num_frames` a property and figure out which is the best value to return:
https://github.com/pytorch/torchcodec/blob/dd44f57180070572eee9e7d35d38f9c03569f6f9/src/torchcodec/_core/_metadata.py#L125-L133
We should also figure out if `num_frames_from_header` is `None`, and if it is, return the calculation. Doing this change is very straight-forward, but we'll need testing to make sure we don't run into segfaults.

scotts:
@mollyxu, does your recent work close this issue?

mollyxu:
Yep I think so! From the merged PR #1021 

https://github.com/meta-pytorch/torchcodec/blob/b35005da9c8ea9e6d0d7b173442b8f2e9267b27d/src/torchcodec/_core/Metadata.cpp#L82-L92

Issue #724, opened 2025-06-11
Encoder tests sometimes fail on reading input


scotts:
I've noticed that some of the encoder tests sometimes fail on reading input:
```
FAILED test/test_encoders.py::TestAudioEncoder::test_num_channels[to_file-2-2] - RuntimeError: Could not open input file: /tmp/pytest-of-root/pytest-0/test_num_channels_to_file_2_2_0/output.mp3 Invalid data found when processing input
FAILED test/test_encoders.py::TestAudioEncoder::test_num_channels[to_file-None-2] - RuntimeError: Could not open input file: /tmp/pytest-of-root/pytest-0/test_num_channels_to_file_None1/output.mp3 Invalid data found when processing input
FAILED test/test_encoders.py::TestAudioEncoder::test_num_channels[to_tensor-2-2] - RuntimeError: Failed to open input buffer: Invalid data found when processing input
FAILED test/test_encoders.py::TestAudioEncoder::test_num_channels[to_tensor-None-2] - RuntimeError: Failed to open input buffer: Invalid data found when processing input
```
When this happens locally, these tests usually pass after re-running. I also just saw an instance in CI: https://github.com/pytorch/torchcodec/actions/runs/15593694391/job/43919750761#step:15:730

NicolasHug:
Reopening as this is still a problem.

We tried to address it in https://github.com/pytorch/torchcodec/pull/755 and then in https://github.com/pytorch/torchcodec/pull/773, which were probably needed regardless, but the original issue persists.

Issue #710, opened 2025-06-03
Add test video with missing number of frames and duration


scotts:
PR #708 relaxed the requirement at the C++ level that the number of frames and the duration of the video or audio file must be present. We are doing this in eventual support of live streaming (see #695). The tests pass, and we checked that when we hack the C++ code to remove obtaining the values, we don't run into any segfaults. But, we don't have any tests that ensure the correct behavior with the values missing.

We should:

1. Check in a small test video that has missing metadata for number of frames and duration.
2. Use it for some of our ops tests - in particular, range queries in both indices and time. Tests that use the Python`VideoDeocder` will always fail because we assert the presence of these values at the Python level.

PR #732, opened 2025-06-18
Calculate num_frames if missing from metadata

Dan-Flores:
This PR enables the instantiation of a VideoDecoder without `num_frames_from_content` or `num_frames_from_header` in the metadata, to resolve issue #727. Without these variables, the VideoDecoder will calculate the number of frames using `average_fps_from_header` and `duration_seconds_from_header`.

### Tests in test_decoders.py
* The return value of `_get_stream_json_metadata` is mocked to ensure VideoDecoder is instantiated without `num_frames_from_content` or `num_frames_from_header`. 
* Various `get_frame...` methods are tested on this decoder.

### Tests in test_metadata.py
* `test_calculate_num_frames_using_fps_and_duration` instantiates `VideoStreamMetadata` without `num_frames_from_content` or `num_frames_from_header`, and  checks that various values of `average_fps` and `duration_seconds` result in the correct number of frames.

* In `test_num_frames_fallback`, the test where both `num_frames_from_content` and `num_frames_from_header` are `None` is no longer valid. A similar test case is added in `test_calculate_num_frames_using_fps_and_duration`.

### DIFF
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -129,8 +129,15 @@ class VideoStreamMetadata(StreamMetadata):
         """
         if self.num_frames_from_content is not None:
             return self.num_frames_from_content
-        else:
+        elif self.num_frames_from_header is not None:
             return self.num_frames_from_header
+        elif (
+            self.average_fps_from_header is not None
+            and self.duration_seconds_from_header is not None
+        ):
+            return int(self.average_fps_from_header * self.duration_seconds_from_header)
+        else:
+            return None
 
     @property
     def average_fps(self) -> Optional[float]:
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -6,6 +6,8 @@
 
 import contextlib
 import gc
+import json
+from unittest.mock import patch
 
 import numpy
 import pytest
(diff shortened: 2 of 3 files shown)


src/torchcodec/_core/_metadata.py
```diff
@@ -129,8 +129,15 @@ def num_frames(self) -> Optional[int]:
         """
```
scotts, line 140: In the doc string, we should also say what we do if num_frames is missing from all of the metadata.

src/torchcodec/_core/_metadata.py
```diff
@@ -129,8 +129,15 @@ def num_frames(self) -> Optional[int]:
...
         if self.num_frames_from_content is not None:
             return self.num_frames_from_content
-        else:
+        elif self.num_frames_from_header is not None:
             return self.num_frames_from_header
+        elif (
+            self.average_fps_from_header is not None
+            and self.duration_seconds_from_header is not None
+        ):
```
scotts, line 148: I just went down an interesting rabbit-hole of asking "Should we instead use the `average_fps` property?" And the answer is no! :) If `self.average_fps_from_header is None`, then `self.average_fps` actually calls this property, `self.num_frames` and we'd (I think) hit infinite recursion. I do wonder if there is a way for us to be more systematic about this, where we could more centrally define the logic of under what circumstances we use which metadata.

test/test_decoders.py
```diff
@@ -738,6 +740,76 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
...
+        frames = decoder.get_frames_at(indices)
+        for ref, frame in zip(ref_frames, frames.data):
+            torch.testing.assert_close(ref, frame)
+
+        # Test get_frames_played_in_range to get all frames
+        assert decoder.metadata.end_stream_seconds is not None
+        all_frames = decoder.get_frames_played_in_range(
+            decoder.metadata.begin_stream_seconds, decoder.metadata.end_stream_seconds
+        )
+        assert_frames_equal(all_frames.data, decoder[:])
```
scotts: Can we also add a simple `assert len(decoder) == CORRECT_VALUE`?

scotts (changes requested):
I had initially approved, but it looks like the tests are failing on Mac - let's make sure we're using the right tolerances there.

scotts:
Thank you! Also, this is only half of #727. We also want to be able to handle if `num_frames` is present but `duration` is not. This half, though, is definitely the more valuable side, as core operations in the Python `VideoDecoder` depend on `num_frames`.

test/test_decoders.py
```diff
@@ -738,6 +740,77 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
...
+        assert decoder.metadata.num_frames_from_header is None
+        assert decoder.metadata.num_frames_from_content is None
+        assert decoder.metadata.duration_seconds is not None
+        assert decoder.metadata.average_fps is not None
+        assert decoder.metadata.num_frames == int(
+            decoder.metadata.duration_seconds * decoder.metadata.average_fps
+        )
+        assert len(decoder) == 390
+
+        # Test get_frames_in_range
```
scotts: Let's add a comment here that we're really just testing the logic in the Python `VideoDecoder` class when we do the decoding. We do checks that involve `num_frames` at the Python layer, and that logic will use the mocked metadata. But the C++ layer does not use that mocked metadata, and will instead use what it read from the video file itself.

We should also create an issue about adding a video file that has the number of frames missing in its metadata so we can actually test the C++ logic as well. This test would then use that video file, and not need any mocking.

scotts:
Thanks, @Dan-Flores! I looked into the failures, and they're all known (#740, #734, #724).

test/test_decoders.py
Dan-Flores: Good point! I'll remove some unnecessary python calls here, and add a comment indicating that the python logic is the only aspect being tested.

It looks like issue #710 covers this subject.

Dan-Flores:
I've updated this PR to complete the other half of #727, and updated the description as well.

test/test_metadata.py
```diff
@@ -143,6 +143,93 @@ def test_num_frames_fallback(
...
+        average_fps_from_header=average_fps_from_header,
+        stream_index=0,
+    )
+
+    assert metadata.num_frames == expected_num_frames
+
+
+@pytest.mark.parametrize(
+    "duration_seconds_from_header, begin_stream_seconds_from_content, end_stream_seconds_from_content, expected_duration_seconds",
+    [(60, 5, 20, 15), (60, 1, None, 60), (None, 0, 10, 10)],
```
NicolasHug: Nit for completeness, I think we can also test the case when only `begin_stream_seconds_from_content` is None?

```suggestion
    [(60, 5, 20, 15), (60, 1, None, 60), (60, None, 1, 60), (None, 0, 10, 10)],
```

test/test_metadata.py
```diff
@@ -143,6 +143,93 @@ def test_num_frames_fallback(
...
+        average_fps_from_header=5,
+        stream_index=0,
+    )
+
+    assert metadata.duration_seconds == expected_duration_seconds
+
+
+@pytest.mark.parametrize(
+    "num_frames_from_header, average_fps_from_header, expected_duration_seconds",
+    [(100, 10, 10), (100, None, None), (None, None, None)],
```
NicolasHug: Similarly here we may want to test when only `num_frames_from_header` is missing:

```suggestion
    [(100, 10, 10), (100, None, None),  (None, 10, None), (None, None, None)],
```

src/torchcodec/_core/_metadata.py
```diff
@@ -88,14 +88,22 @@ def duration_seconds(self) -> Optional[float]:
         fall back to ``duration_seconds_from_header``.
```
NicolasHug: Similarly to what was done for the `num_frames` properties, we can slightly update the docstring above to describe the new fall-back

NicolasHug (approved):
Thanks for the PR, and great work on the tests/mocking!

src/torchcodec/_core/_metadata.py
```diff
@@ -85,17 +85,27 @@ class VideoStreamMetadata(StreamMetadata):
     def duration_seconds(self) -> Optional[float]:
         """Duration of the stream in seconds. We try to calculate the duration
         from the actual frames if a :term:`scan` was performed. Otherwise we
-        fall back to ``duration_seconds_from_header``.
+        fall back to ``duration_seconds_from_header``. If that value is None,
```
scotts: Nit:
```suggestion
        fall back to ``duration_seconds_from_header``. If that value is also None,
```

src/torchcodec/_core/_metadata.py
```diff
@@ -123,14 +133,22 @@ def end_stream_seconds(self) -> Optional[float]:
 
     @property
     def num_frames(self) -> Optional[int]:
-        """Number of frames in the stream. This corresponds to
-        ``num_frames_from_content`` if a :term:`scan` was made, otherwise it
-        corresponds to ``num_frames_from_header``.
+        """Number of frames in the stream (int or None).
+        This corresponds to ``num_frames_from_content`` if a :term:`scan` was made,
+        otherwise it corresponds to ``num_frames_from_header``. If it is None,
```
scotts: Nit:
```suggestion
        otherwise it corresponds to ``num_frames_from_header``. If that value is also None,
```

scotts (approved):
All the new stuff looks good too, good to merge!

PR #735, opened 2025-06-21
Only enforce the check for video and audio streams

entron:


NicolasHug:
Hi @entron , can you please share more context about the reasons for this change?

entron:
hi @NicolasHug, I have one video with chapters in the meta data and torchcodec throws this error. If I remove the captions in the meta data it works but I think maybe it is better to simply ignore none video/audio stream all together?

NicolasHug:
Thanks for the reply @entron . Any chance you could share such video, or help me generate one? Sorry for the extra work, I'm a bit short on time right now and having a reproducing example would greatly help.

entron:
@NicolasHug, I don't know how to generate such a video it is just one bad example probably. The Chapters are like:

```
Chapters:
    Chapter #0:0: start -10.000000, end 0.000000
      Metadata:
        title           : Chapter 1
    Chapter #0:1: start 0.000000, end 10.000000
      Metadata:
        title           : Chapter 2
        ...
        
```

NicolasHug:
I'll close this, I think we ended up removing the corresponding lines due to a similar issue (possibly the same!)

Issue #733, opened 2025-06-19
query video  pixel aspect ratio (PAR) or sample aspect ratio (SAR)


carandraug:
Pixels in video file formats are not necessarily "square". Such non-square rectangular pixels are defined by their pixel aspect ratio (PAR) or sample aspect ratio (SAR) (see also [wikipedia's page on Pixel Aspect Ratio](https://en.wikipedia.org/wiki/Pixel_aspect_ratio))

ffmpeg supports themand ffprobe will report them but torchcodec returns the exact number of pixels without any information about the pixel aspect ratio. Here's the output of `ffprobe` on a file with non-square pixels (note the SAR (Sample Aspect Ratio) which is not 1:1):

```
$ ffprobe -version
ffprobe version 6.1.1 Copyright (c) 2007-2023 the FFmpeg developers
[...]
$ ffprobe -hide_banner mp4/512/5090101893225793677.mp4
Input #0, mov,mp4,m4a,3gp,3g2,mj2, from 'mp4/512/5090101893225793677.mp4':
  Metadata:
    major_brand     : isom
    minor_version   : 512
    compatible_brands: isomiso2avc1mp41
    encoder         : Lavf58.76.100
  Duration: 00:59:34.20, start: 0.000000, bitrate: 865 kb/s
  Stream #0:0[0x1](und): Video: h264 (High) (avc1 / 0x31637661), yuv420p(progressive), 640x512 [SAR 64:45 DAR 16:9], 731 kb/s, 25 fps, 25 tbr, 12800 tbn (default)
    Metadata:
      handler_name    : VideoHandler
      vendor_id       : [0][0][0][0]
  Stream #0:1[0x2](eng): Audio: aac (LC) (mp4a / 0x6134706D), 48000 Hz, stereo, fltp, 128 kb/s (default)
    Metadata:
      handler_name    : SoundHandler
      vendor_id       : [0][0][0][0]
```

Would be nice if torchcodec would at least provide information about the pixel aspect ratio (and since most downstream tasks will likely expect square pixels, maybe also provide frames scaled accordingly).

scotts:
[...]

PR #737, opened 2025-06-22
VideoStreamMetadata.sample_aspect_ratio: new metadata field (#733)

carandraug:
New field to `VideoStreamMetadata` to at least get information about the stream sample/pixel aspect ratio.  Getting this information is the minimum required to support non-square pixels.

---

This addresses issue #733 (which was deemed "reasonable" which is why I'm implementing it, thank you for considering this).

There is one thing that annoys me about this PR. The names for properties in C++ `StreamMetadata` and json returned by `get_stream_json_metadata` match. I could have separate num/den properties in `StreamMetadata` but I don't think that's quite right because they're actually two parts of the same value. On the json side of things I made it two separate properties because it's `std::map<std::string, std::string>` (although it would be trivial to make it a colon separated string and parse it on the python side of things).

### DIFF
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -83,7 +83,7 @@ class VideoStreamMetadata(StreamMetadata):
     We recommend using the ``average_fps`` attribute instead."""
     pixel_aspect_ratio: Optional[Fraction]
     """Pixel Aspect Ratio (PAR), also known as Sample Aspect Ratio
-    (SAR --- not to be confused with Sample Aspect Ratio, also SAR),
+    (SAR --- not to be confused with Storage Aspect Ratio, also SAR),
     is the ratio between the width and height of each pixel
     (``fractions.Fraction`` or None)."""
 


scotts:
@carandraug, thanks for tackling this! This PR actually raises an interesting design question: so far, we've only exposed metadata as strings, floats or ints. The value that's closest to the SAR that we currently have is the FPS, which we return as a float. In FFmpeg, it's an `AVRational`. We have not yet returned a rational as-is, without turning it into a float.

For your use-case, is it easier to have a rational, or a float? If we keep a rational, I think we should actually make it a [`numbers.Rational`](https://docs.python.org/3/library/numbers.html#numbers.Rational) rather than just a tuple. @NicolasHug, thoughts?

NicolasHug:
I think it would have to be a `fractions.Fraction` because `Rational` is an ABC, but agreed this is better than a plain tuple as it supports natural operations like `+`, `float(.)`, etc.

Now that I think about it, we could have exposed fps as a fraction as well, but it's probably too late to do so without breaking BC?

carandraug:
@scotts writes:
> For your use-case, is it easier to have a rational, or a float? If we keep a rational, I think we should actually make it a [`numbers.Rational`](https://docs.python.org/3/library/numbers.html#numbers.Rational) rather than just a tuple. @NicolasHug, thoughts?

From afar, I guess using a rational would be slightly simpler but I see no reason why I couldn't work a float though. However, it is trivial to get a float from a rational but not the inverse. If someone ever really needs the fraction, it will be too late to change so I'd suggest to return a rational now.

Shall I add a commit to this PR to use a `fractions.Fraction` instead of a `tuple`? Is the rest of the PR ok?

src/torchcodec/_core/Metadata.h
```diff
@@ -45,6 +46,7 @@ struct StreamMetadata {
   // Video-only fields derived from the AVCodecContext.
   std::optional<int64_t> width;
   std::optional<int64_t> height;
+  std::optional<std::pair<int, int>> sampleAspectRatio;
```
NicolasHug: We're using FFmpeg types here already, so I think we should define this as [AVRational](https://ffmpeg.org/doxygen/6.0/structAVRational.html) instead of a tuple

src/torchcodec/_core/_metadata.py
```diff
@@ -80,6 +80,12 @@ class VideoStreamMetadata(StreamMetadata):
     average_fps_from_header: Optional[float]
     """Averate fps of the stream, obtained from the header (float or None).
     We recommend using the ``average_fps`` attribute instead."""
+    sample_aspect_ratio: Optional[tuple[int, int]]
```
NicolasHug: Because this is a public name, I think we should name it `pixel_aspect_ratio`, since it is only relevant for pixels. We use the term `sample` to denote audio data in the rest of our public APIs. It's fine to keep the name `sampleAspectRatio` to be close to FFmpeg in the private parts of the code (like in C++), but here, "pixel" will be more descriptive.

And as mentioned in previous discussion, let's make this a Fraction just to be on the safe side.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -601,6 +601,12 @@ std::string get_stream_json_metadata(
   if (streamMetadata.height.has_value()) {
     map["height"] = std::to_string(*streamMetadata.height);
   }
+  if (streamMetadata.sampleAspectRatio.has_value()) {
+    map["sampleAspectRatioNum"] =
+        std::to_string((*streamMetadata.sampleAspectRatio).first);
+    map["sampleAspectRatioDen"] =
+        std::to_string((*streamMetadata.sampleAspectRatio).second);
+  }
```
NicolasHug, line 609: I think what you're doing here is OK. I don't see a much cleaner way around the json intermediate layer. @scotts  any concern?

scotts: Agreed. We're turning what is fundamentally two values into one object. This is a reasonable way to do that.

scotts:
@NicolasHug, you bring up an interesting point about our existing rates: should we have also made them `numbers.Fractions`? Maybe, but something that's implied in our API is that the time unit we both report and accept is always seconds. Our rationale was that's more ergonomic for users.

If we find that folks start needing the underlying rational, we can support that in a backwards-compatible way. We can always change our implementation in C++ and the Python layer to have the actual fraction. The existing APIs would become properties that return the float, and we could add `*_rational` accessors. But this is a layer of complexity I'd only want us to take on if users see a big need for it.

carandraug:
I have applied all the requested changes:
* `std::pair` -> `AVRational`
* `sample_aspect_ratio` -> `pixel_aspect_ratio`
* `tuple[int, int]` -> `Fraction`

There is one thing that is annoying me with the use of `Fraction`. For the common case of `Fraction(1, 1), this prints as `1`. I think it would be nice if the fraction format was always used:

```
VideoStreamMetadata:
[...]
  pixel_aspect_ratio: 1  # would be nice if this was 1/1
[...]
```

I could force it on `StreamMetadata.__repr__` maybe, not sure how you feel about it.

carandraug:
> I could force it on `StreamMetadata.__repr__` maybe, not sure how you feel about it.

To be clear, I mean something along the lines of:

```
diff --git a/src/torchcodec/_core/_metadata.py b/src/torchcodec/_core/_metadata.py
index 054cfc3..39830e9 100644
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -41,7 +41,12 @@ class StreamMetadata:
     def __repr__(self):
         s = self.__class__.__name__ + ":\n"
         for field in dataclasses.fields(self):
-            s += f"{SPACES}{field.name}: {getattr(self, field.name)}\n"
+            attr = getattr(self, field.name)
+            if isinstance(attr, Optional[Fraction]):
+                # Once we require Python 3.13 we can use format(.., "#")
+                s += f"{SPACES}{field.name}: {attr.numerator}/{attr.denominator}\n"
+            else:
+                s += f"{SPACES}{field.name}: {attr}\n"
         return s
```

src/torchcodec/_core/_metadata.py
```diff
@@ -80,6 +81,11 @@ class VideoStreamMetadata(StreamMetadata):
     average_fps_from_header: Optional[float]
     """Averate fps of the stream, obtained from the header (float or None).
     We recommend using the ``average_fps`` attribute instead."""
+    pixel_aspect_ratio: Optional[Fraction]
+    """Pixel Aspect Ratio (PAR), also known as Sample Aspect Ratio
+    (SAR --- not to be confused with Sample Aspect Ratio, also SAR),
```
NicolasHug: > known as **Sample Aspect Ratio** (SAR --- not to be confused with **Sample Aspect Ratio**

Emphasis mine above, is there a typo? Both are the same.

But maybe we don't even need to mention Sample Aspect Ratio at all :thinking: ?

NicolasHug (approved):
Thanks for the PR @carandraug , LGTM with a comment below, let's see if tests are happy.

On the `__repr__`: I personally I don't think it's worth the hassle, but let's consider that in a separate PR so we can get this one merged first.

src/torchcodec/_core/_metadata.py
carandraug: Yes, it is a typo. It should be Storage Aspect Ratio. It is fixed now on a separate commit.

I think it's worth to, at least, mention sample aspect ratio since it's very a common term for this (e.g., that's what ffmpeg uses). If a user is familiar with sample aspect ratio only, then this hint helps them in understanding what this is about (otherwise they will have to got read on PAR before knowing it's exactly the same thing). I could drop the note about Storage Aspect Ration being also SAR but because there's so much confusion with this terms, I think it's better to be explicit that we're talking about SAR).

To be really explicit, I could add to the docstring "if this value is not 1/1, then the pixels are not square".

NicolasHug:
Merging, thank you for the PR @carandraug !

Issue #738, opened 2025-06-23
Error in Decoding audio streams example


dhunstack:
### 📚 The doc issue

The section on [`Custom sample rate`](https://docs.pytorch.org/torchcodec/stable/generated_examples/audio_decoding.html#custom-sample-rate) in the `Decoding audio streams with AudioDecoder` has an error.

The text says
```
The ouput will sound the same, but note that the number of samples greatly increased:
```

This is incorrect, as the sample rate has been reduced from `44100` to `16000` and the number of samples in output has actually reduced from `4297722` to `1559264` for the original audio. Also the example file used is music audio instead which loses high frequency content above `8000Hz` on sampling rate reduction.

An improved version of the text could be
```
The number of samples has reduced from `4297722` to `1559264`, in the ratio `44100` to `16000`.
The downsampled output loses some high frequency content, which may or may not sound the same depending on the playback device.
```

NicolasHug:
Thanks for the catch and for the PR @dhunstack

dhunstack:
Thanks, I'll close this issue since the fix has been merged

PR #739, opened 2025-06-23
[fix] Fix error in Audio Decoder example

dhunstack:
This commit fixes the incorrect text in `Custom sample rate` section of the `Decoding audio streams with AudioDecoder` example, as described in issue #738

### DIFF
--- a/examples/audio_decoding.py
+++ b/examples/audio_decoding.py
@@ -93,10 +93,9 @@ play_audio(samples)
 # ------------------
 #
 # We can also decode the samples into a desired sample rate using the
-# ``sample_rate`` parameter of :class:`~torchcodec.decoders.AudioDecoder`.
-# The number of samples has reduced from `4297722` to `1559264`, in the 
-# ratio `44100` to `16000`. The downsampled output loses high frequency 
-# content, which may or may not sound the same depending on the playback device.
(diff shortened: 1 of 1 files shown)


examples/audio_decoding.py
```diff
@@ -93,9 +93,10 @@ def play_audio(samples):
 # ------------------
 #
 # We can also decode the samples into a desired sample rate using the
-# ``sample_rate`` parameter of :class:`~torchcodec.decoders.AudioDecoder`. The
-# ouput will sound the same, but note that the number of samples greatly
-# increased:
+# ``sample_rate`` parameter of :class:`~torchcodec.decoders.AudioDecoder`.
+# The number of samples has reduced from `4297722` to `1559264`, in the 
+# ratio `44100` to `16000`. The downsampled output loses high frequency 
+# content, which may or may not sound the same depending on the playback device.
```
NicolasHug:
```suggestion
# ``sample_rate`` parameter of :class:`~torchcodec.decoders.AudioDecoder`. The
# ouput will sound similar, but note that the number of samples greatly
# decreased:```

NicolasHug (approved):
Thanks for the catch and for the PR @dhunstack . I slightly modified your changes to not include absolute number of samples, as those may change if we ever modify the input file.
I retained your point noting that the audio may not sound exactly the same as well.

PR #741, opened 2025-06-25
test/: fix failing tests due to missing argument (fixup d6ce5703)

carandraug:
The new tests introduced in ffac96c8 (#732) are failing since d6ce5703 (#737) because d6ce5703 was merged by rebase (testes passed on the PR but it was branched off before #732 was merged).


NicolasHug (approved):
I should have merged with `main` before landing the PR. Thanks for catching this @carandraug !

PR #742, opened 2025-06-25
StreamMetadata: always display Fractions in num/den form

carandraug:
During the review of PR #733 I suggested that fractions should always be displayed in num/den form:

> On the `__repr__`: I personally I don't think it's worth the hassle, but let's consider that in a separate PR so we can get this one merged first.

_Originally posted by @NicolasHug in https://github.com/pytorch/torchcodec/pull/737#pullrequestreview-2954198659_

So here's that separate PR.

### DIFF
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -41,7 +41,14 @@ class StreamMetadata:
     def __repr__(self):
         s = self.__class__.__name__ + ":\n"
         for field in dataclasses.fields(self):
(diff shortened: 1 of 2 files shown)


NicolasHug:
Hey @carandraug thanks for the PR and sorry for the (very!) late review. I don't think we'll merge this, we'd like to keep the  `__repr__` reasonably simple, and I think the existing printed results are fine. IIRC, this was mainly for `pixel_aspect_ratio: 1 ` to render as `1/1`, which I think is a matter of personal preference.

Issue #150, opened 2024-08-05
Extend `SimpleVideoDecoder` index-based APIs


NicolasHug:
For consistency with tensor indexing semantics and with builtin Python indexing of sequences, we should extend the index-based APIs of `SimpleVideoDecoder`, namely `__getitem__()`, `get_frame_at()` and `get_frames_at()` to support:

- negative indexing. `-n` just means `len(decoder) - n`.
- upper bound in slices to be greated than `len(decoder)`.

In short we should allow for that kind of stuff:

```
>>> torch.arange(10)[-3:1000]
tensor([7, 8, 9])
```

In terms of implementation, we can implement this within the `SimpleVideoDecoder` class in Python. No need to extend the core or C++ APIs.

NicolasHug:
I think this was largely addressed now.

PR #743, opened 2025-06-25
Support negative index in SimpleVideoDecoder

Dan-Flores:
This PR adds the missing functionality described in #150, specifically adding support for negative indexing in `get_frame_at()` and `get_frames_at()`.

In both functions, a negative index `n` is converted to `len(decoder) - n`. The tests are updated to reflect this support as well.

The other item in that issue, allowing "upper bound in slices to be greated than `len(decoder)`", is already supported by `getitem_slice`. Specifically, this section of `_video_decoder.py` handles negative and out-of-range indices:
```
    def _getitem_slice(self, key: slice) -> Tensor:
        assert isinstance(key, slice)

        start, stop, step = key.indices(len(self))
```

I added a new test case to `test_getitem_slice` to make this more explicit.

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -328,6 +328,19 @@ class TestVideoDecoder:
         )
         assert_frames_equal(ref386_389, slice386_389)
 
+        # slices with upper bound greater than len(decoder) are supported
+        slice387_389 = decoder[-3:10000].to(device)
+        assert slice387_389.shape == torch.Size(
+            [
+                3,
+                NASA_VIDEO.num_color_channels,
+                NASA_VIDEO.height,
+                NASA_VIDEO.width,
+            ]
+        )
+        ref387_389 = NASA_VIDEO.get_frame_data_by_range(387, 390).to(device)
+        assert_frames_equal(ref387_389, slice387_389)
+
         # an empty range is valid!
         empty_frame = decoder[5:5]
         assert_frames_equal(empty_frame, NASA_VIDEO.empty_chw_tensor.to(device))
@@ -437,6 +450,11 @@ class TestVideoDecoder:
             expected_frame_info.duration_seconds, rel=1e-3
         )
 
+        # test negative frame index
+        frame_minus1 = decoder.get_frame_at(-1)
+        ref_frame_minus1 = NASA_VIDEO.get_frame_data_by_index(389).to(device)
+        assert_frames_equal(ref_frame_minus1, frame_minus1.data)
+
         # test numpy.int64
         frame9 = decoder.get_frame_at(numpy.int64(9))
         assert_frames_equal(ref_frame9, frame9.data)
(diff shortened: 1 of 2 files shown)


test/test_decoders.py
```diff
@@ -520,8 +549,11 @@ def test_get_frames_at(self, device, seek_mode):
     def test_get_frames_at_fails(self, device, seek_mode):
         decoder = VideoDecoder(NASA_VIDEO.path, device=device, seek_mode=seek_mode)
 
-        with pytest.raises(RuntimeError, match="Invalid frame index=-1"):
-            decoder.get_frames_at([-1])
+        expected_converted_index = -10000 + len(decoder)
+        with pytest.raises(
+            RuntimeError, match=f"Invalid frame index={expected_converted_index}"
```
Dan-Flores, line 554: This error is from `SingleStreamDecoder.cpp`. Should we also be checking valid indices at the Python level, in `_video_decoder.py`?

test/test_decoders.py
```diff
@@ -520,8 +549,11 @@ def test_get_frames_at(self, device, seek_mode):
     def test_get_frames_at_fails(self, device, seek_mode):
         decoder = VideoDecoder(NASA_VIDEO.path, device=device, seek_mode=seek_mode)
 
-        with pytest.raises(RuntimeError, match="Invalid frame index=-1"):
-            decoder.get_frames_at([-1])
+        expected_converted_index = -10000 + len(decoder)
```
Dan-Flores, line 552: It is not necessary to get the expected index here, but I added it to match the other tests. Alternatively, the test can simply match the text: "Invalid frame index".

scotts: It's fine to do, but I usually default to just matching the text that doesn't change.

test/test_decoders.py
scotts: Good observation - at the Python level, we currently don't check the input when we get a list of indices (`get_frames_at()`) or a list of timestamps (`get_frames_played_at()`). The weak rationale was that we'd have to walk the list to do the check, and that seemed wasteful. Now that we're walking the list in `get_frames_at()` to deal with negative indices, we might as well do error checking at the same time.

The question, then, is what to do in `get_frames_played_at()`? For completeness, I think it does make sense to do error checking at the Python level there, but let's do that in a separate PR. We can also do some refactoring there, and make `_getitem_slice()` call `VideoDecoder.get_frames_in_range()` in order to get that error checking at the Python level as well.

scotts (approved):
Let's do the error checking in `get_frames_at()` now too, but approved to unblock.

NicolasHug (approved):
Thanks @Dan-Flores , LGTM!

I forgot to write it down in https://github.com/pytorch/torchcodec/issues/150, but eventually it'd be good to accept the same semantics for `get_frames_in_range()`. We we can merge this PR as-is though.

PR #746, opened 2025-07-01
Convert negative frame indices in C++, convert slices in Python

Dan-Flores:
This PR adds the follow ups from #743:
* Converting negative indices is done in C++: `getFrameAtIndexInternal`:
  * This function is called in `getFrameAtIndex`, `getFramesAtIndices`, `getFramesInRange`, `getFramesPlayedInRange`.
* We rely on python's slice `.indices()` function to convert negative or None indices in `get_frames_in_range`.
* Errors in `validateFrameIndex` are changed to `std::out_of_range`, to better reflect the IndexError.

The relevant tests are updated as well:
* Updated `test_get_frame_at_index` to test returning a negative index.
* Added `test_get_frames_at_indices_negative_indices` to test returning a negative index.
  * Added `test_get_frames_at_indices_fail_on_invalid_negative_indices` to ensure negative indices fail when an invalid negative index is passed in.
* Added `test_get_frames_in_range_slice_indices_syntax` to test negative indices, capping upper bound indices, and none as stop.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -1627,7 +1627,8 @@ void SingleStreamDecoder::validateFrameIndex(
     throw std::out_of_range(
         "Invalid frame index=" + std::to_string(frameIndex) +
         " for streamIndex=" + std::to_string(streamMetadata.streamIndex) +
-        "; must be greater than or equal to 0, or be a valid negative index");
+        "; negative indices must have an absolute value less than the number of frames, "
+        "and the number of frames must be known.");
   }
 
   // Note that if we do not have the number of frames available in our metadata,
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -492,7 +492,7 @@ class TestVideoDecoder:
 
         with pytest.raises(
             IndexError,
-            match="must be greater than or equal to 0, or be a valid negative index",
+            match="negative indices must have an absolute value less than the number of frames",
         ):
             frame = decoder.get_frame_at(-10000)  # noqa
 
@@ -557,7 +557,7 @@ class TestVideoDecoder:
 
         with pytest.raises(
             IndexError,
-            match="must be greater than or equal to 0, or be a valid negative index",
+            match="negative indices must have an absolute value less than the number of frames",
         ):
             decoder.get_frames_at([-10000])
 
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -202,7 +202,7 @@ class TestVideoDecoderOps:
         add_video_stream(decoder, device=device)
         with pytest.raises(
             IndexError,
-            match="must be greater than or equal to 0, or be a valid negative index",
+            match="negative indices must have an absolute value less than the number of frames",
         ):
             invalid_frames, *_ = get_frames_at_indices(
                 decoder, frame_indices=[-10000, -3000]


test/test_decoders.py
```diff
@@ -772,6 +769,66 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
...
+            decoder.get_frames_in_range(start=1000, stop=10)
+
+        with pytest.raises(IndexError, match="Start index -\\d+ is out of bounds"):
+            decoder.get_frames_in_range(start=-1000, stop=10)
+
+        with pytest.raises(
+            IndexError,
+            match="Stop index \\(-\\d+\\) must not be less than the start index",
+        ):
+            decoder.get_frames_in_range(start=0, stop=-1000)
```
Dan-Flores: I considered possible edge cases for `stop`, I believe these tests should cover them. 
We assert that `start` is within `0 <= start < self._num_frames`, and that `stop >= start`. 

As a result, we do not need to explicitly check that `stop` is within the valid range, since any higher value is reduced to `self._num_frames`, and any lower value cannot be below the lowest `start` value of 0.

test/test_decoders.py
```diff
@@ -549,13 +549,10 @@ def test_get_frames_at(self, device, seek_mode):
     def test_get_frames_at_fails(self, device, seek_mode):
         decoder = VideoDecoder(NASA_VIDEO.path, device=device, seek_mode=seek_mode)
 
-        expected_converted_index = -10000 + len(decoder)
-        with pytest.raises(
-            RuntimeError, match=f"Invalid frame index={expected_converted_index}"
-        ):
+        with pytest.raises(IndexError, match="Index -\\d+ is out of bounds"):
```
Dan-Flores: Since [pytest.raises](https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises) accepts regex patterns, it is not necessary to calculate the index that will appear in the error.

test/test_decoders.py
```diff
@@ -772,6 +769,66 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
             empty_frames.duration_seconds, NASA_VIDEO.empty_duration_seconds
         )
 
+    @pytest.mark.parametrize("device", cpu_and_cuda())
+    @pytest.mark.parametrize("stream_index", [3, None])
```
NicolasHug: Nit: I think we can remove the `stream_index` parametrization, because what we're testing here is (or should be) completely orthogonal to the stream concept.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -220,9 +220,14 @@ def get_frames_at(self, indices: list[int]) -> FrameBatch:
...
-            index if index >= 0 else index + self._num_frames for index in indices
-        ]
+        for i, index in enumerate(indices):
+            index = index if index >= 0 else index + self._num_frames
+            if not 0 <= index < self._num_frames:
+                raise IndexError(
+                    f"Index {index} is out of bounds; must be in the range [0, {self._num_frames})."
+                )
+            else:
+                indices[i] = index
```
NicolasHug: In general, we should avoid modifying the user-provided input inplace, as this can be surprising and have unintended effects. Let's keep the list comprehension logic?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -247,6 +252,8 @@ def get_frames_in_range(self, start: int, stop: int, step: int = 1) -> FrameBatc
         Returns:
             FrameBatch: The frames within the specified range.
         """
+        start = start if start >= 0 else start + self._num_frames
+        stop = min(stop if stop >= 0 else stop + self._num_frames, self._num_frames)
```
NicolasHug: Just to note that eventually, we'll want to support `stop=None` and make it the default. This is what we have in the audio decoder and that's useful to decode all the frames at once. We should also add it to the time-based API `get_frames_played_in_range()`.

But we should do that separately.

test/test_decoders.py
```diff
@@ -772,6 +769,66 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
...
+            stream_index=stream_index,
+            device=device,
+            seek_mode=seek_mode,
+        )
+        # slices with upper bound greater than len(decoder) are supported
+        ref_frames387_389 = NASA_VIDEO.get_frame_data_by_range(
+            start=387, stop=390, stream_index=stream_index
+        ).to(device)
+        frames387_389 = decoder.get_frames_in_range(start=387, stop=1000)
+        print(f"{frames387_389.data.shape=}")
```
NicolasHug: Seems like a debugging left-over:

```suggestion
```

test/test_decoders.py
```diff
@@ -772,6 +769,66 @@ def test_get_frames_in_range(self, stream_index, device, seek_mode):
...
+                NASA_VIDEO.get_width(stream_index=stream_index),
+            ]
+        )
+        assert_frames_equal(ref_frames387_389, frames387_389.data)
+
+        # test that negative values in the range are supported
+        ref_frames386_389 = NASA_VIDEO.get_frame_data_by_range(
+            start=386, stop=390, stream_index=stream_index
+        ).to(device)
+        frames386_389 = decoder.get_frames_in_range(start=-4, stop=1000)
```
NicolasHug: Let's check for a negative `stop` value as well. This should exclude the last frame so you'll have to exclude it from the ref above.

```suggestion
        frames386_389 = decoder.get_frames_in_range(start=-4, stop=-1)
```

test/test_decoders.py
NicolasHug: Out of precaution and to be pedantic, maybe also test when both start and stop are positive:

```
decoder.get_frames_in_range(start=10, stop=0)
```

so that `stop` is fully independent of our negative index conversion logic

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -220,9 +220,14 @@ def get_frames_at(self, indices: list[int]) -> FrameBatch:
         Returns:
             FrameBatch: The frames at the given indices.
         """
-        indices = [
-            index if index >= 0 else index + self._num_frames for index in indices
-        ]
+        for i, index in enumerate(indices):
```
NicolasHug: I realize I'm sort of going back on the discussion from https://github.com/pytorch/torchcodec/pull/743/files#r2167398761, so I apologize for not bringing that up earlier. But I wonder if we actually want to do the validation and the conversion in Python.  For performance reasons, it might be better to keep everything in C++, as the `indices` list can potentially be quite long, especially when sampling a large number of frames for long videos. I suspect we can easily have `len(indices) > Nx1000`, typically when using our samplers on long videos:

https://github.com/pytorch/torchcodec/blob/103f714f81dc8c6095caa7bb1d30441583f54750/src/torchcodec/samplers/_index_based.py#L180

As a somewhat related concern I think we'll eventually want to support `indices` not just as a list, but also as a tensor, e.g. for users who create their own sampling strategy using `torch` utilities. And it'd be great if we could avoid copying that input tensor before passing it to the underlying decoding ops, which we can only do if we run the validation/conversion in C++.

In any case, we can and should merge this PR as-is regardless of our conclusion, because we are already doing such validation in `main`, and this PR doesn't add much on top of the existing one.

NicolasHug (commented):
Thanks for the PR @Dan-Flores . Left a few comments below, this mostly looks good. There's a discussion point I'd like to dig into with you and @scotts , but this isn't blocking for this PR.

src/torchcodec/decoders/_video_decoder.py
scotts: Agreed on both points - when I wrote the comment on #743, I was also wondering about performance, and figured let's go for consistency first, then deal with performance later. We can make the checking basically zero-cost on the C++ side if we do it as we iterate through the indices - but we'll want to _not_ do a `TORCH_CHECK()`, but throw a `std::out_of_range` so it becomes the right thing on the Python side. We can do all of that on a follow-up PR.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -568,8 +574,6 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
     auto indexInOutput = indicesAreSorted ? f : argsort[f];
     auto indexInVideo = frameIndices[indexInOutput];
 
-    validateFrameIndex(streamMetadata, indexInVideo);
-
```
Dan-Flores, line 626: I believe this call to `validateFrameIndex` is not needed, since `getFrameAtIndexInternal` will also call `validateFrameIndex`.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -247,6 +236,8 @@ def get_frames_in_range(self, start: int, stop: int, step: int = 1) -> FrameBatc
         Returns:
             FrameBatch: The frames within the specified range.
         """
+        # Adjust start / stop indices to enable indexing semantics, ex. [-10, 1000] returns the last 10 frames
+        start, stop, step = slice(start, stop, step).indices(self._num_frames)
```
Dan-Flores, line 234: This is how slices are handled in `VideoDecoder._getitem_slice`, using it here ensures consistent behavior.

test/test_decoders.py
```diff
@@ -487,10 +487,13 @@ def test_get_frame_at_tuple_unpacking(self, device):
     def test_get_frame_at_fails(self, device, seek_mode):
         decoder = VideoDecoder(NASA_VIDEO.path, device=device, seek_mode=seek_mode)
 
-        with pytest.raises(IndexError, match="out of bounds"):
+        with pytest.raises(
+            IndexError,
+            match="must be greater than or equal to 0, or be a valid negative index",
+        ):
```
Dan-Flores, line 496: The previous error check in `get_frame_at` was removed from Python. Now, this error comes from `validateFrameIndex`. 

I removed the single index error check in Python for consistency between the `get_frame(s)...` functions, but I am open to added them back if erroring earlier is preferred, please let me know if you have any preference @scotts @NicolasHug

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -526,6 +526,12 @@ FrameOutput SingleStreamDecoder::getFrameAtIndexInternal(
   const auto& streamInfo = streamInfos_[activeStreamIndex_];
   const auto& streamMetadata =
       containerMetadata_.allStreamMetadata[activeStreamIndex_];
+
+  std::optional<int64_t> numFrames = getNumFrames(streamMetadata);
+  if (numFrames.has_value()) {
+    // If the frameIndex is negative, we convert it to a positive index
+    frameIndex = frameIndex >= 0 ? frameIndex : frameIndex + numFrames.value();
+  }
```
scotts, line 588: ~If `numFrames` does _not_ have a value, then I think we need to assert that `frameIndex` is positive, and throw a `std::range_error` if it's not. Otherwise, we'll end up using a negative `frameIndex` further down.~

Ah, now I see that we actually do that in `validateFrameIndex()`, which we call right after. 👍

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: Hmm. Yes, although I had to think about it. What gave me pause is the fact that we also use it in the `if` branch, but if `indexInVideo == previousIndexInVideo` then we must have checked it during the previous iteration so we're safe.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1559,21 +1563,23 @@ void SingleStreamDecoder::validateScannedAllStreams(const std::string& msg) {
...
-      frameIndex >= 0,
-      "Invalid frame index=" + std::to_string(frameIndex) +
-          " for streamIndex=" + std::to_string(streamMetadata.streamIndex) +
-          "; must be greater than or equal to 0");
+  if (frameIndex < 0) {
+    throw std::out_of_range(
+        "Invalid frame index=" + std::to_string(frameIndex) +
+        " for streamIndex=" + std::to_string(streamMetadata.streamIndex) +
+        "; must be greater than or equal to 0, or be a valid negative index");
+  }
```
scotts, line 1632: It could end up being wordy, but we might want to mention that we can't receive any negative index if we don't know the number of frames. Maybe something like:
```c++
throw std::out_of_range(
        "Invalid frame index=" + std::to_string(frameIndex) +
        " for streamIndex=" + std::to_string(streamMetadata.streamIndex) +
        "; negative indices must have an absolute value that is less than the number of frames, "
        "and the number of frames must be known.");
  }
```

test/test_decoders.py
scotts: I think it makes sense to do the error checking once, on the C++ side. But we still have the error checking on the Python side for `get_frames_in_range()` - is that now also redundant?

scotts (approved):

test/test_decoders.py
Dan-Flores: Yes, I'll remove those checks since `.indices()` will prevent most of them. It does not alter a negative `step`, but we check and error on that case in `SingleStreamDecoder::getFramesInRange`.

PR #752, opened 2025-07-04
Windows: Add FFmpeg build job

NicolasHug:
This PR adds Windows builds for our non-GPL FFmpeg binaries that we will then upload to S3.

This is passing, see e.g. https://github.com/pytorch/torchcodec/actions/runs/16081032115/job/45385732283?pr=752

### DIFF
--- a/.github/workflows/build_ffmpeg.yaml
+++ b/.github/workflows/build_ffmpeg.yaml
@@ -76,34 +76,19 @@ jobs:
       fail-fast: false
       matrix:
         ffmpeg-version: ["4.4.4", "5.1.4", "6.1.1", "7.0.1"]
-    runs-on: windows-latest
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


scotts (approved):
I can see we're actually invoking the same build scrip that Linux and Mac use - which is excellent - but we're doing a bunch of setup before that. Any insight into what that's is doing?

NicolasHug:
My understanding of `vc_env_helper.bat` is that it's a generic helper to sett up the right configuration for the windows compiler. The script takes a command as input, e.g. our ffmpeg build script or whatever else like `pip install -e .`, and executes that command with the right environment. I don't have much more insight into it, there's a very similar helper in torchvision, so I was comfortable copy/pasting it blindly: https://github.com/pytorch/vision/blob/main/packaging/windows/internal/vc_env_helper.bat

An I think `packaging/build_ffmpeg.bat` does more of the same thing, i.e. setting up the env correctly, specifically for the FFmpeg build this time.

PR #754, opened 2025-07-05
Support encoding to file-like object

NicolasHug:
This PR adds the `to_file_like()` method to the `AudioEncoder`. This allows users to encode samples into a custom file-like object that supports `seek` and `write` methods.

~Marking this as draft because there are unresolved points (see below), but this is still ready for a solid first review round.~


src/torchcodec/_core/ops.py
```diff
@@ -153,6 +153,59 @@ def create_from_file_like(
...
+        sample_rate,
+        format,
+        file_like,
+        bit_rate,
+        num_channels,
+    )
+
+    # This check is useless but it's critical to keep it to ensures that samples
+    # is still alive during the call to encode_audio_to_file_like.
+    assert samples.is_contiguous()
```
NicolasHug, line 207: I hate that we have to do this but I do not see any other obvious way to keep the input `samples` alive for the duration of the call.
Claude is saying that we could just pass `samples` as a py::object. We won't be able to turn it back to a tensor (as mentioned in the code comment above), but claude claims that passing it as a parameter will ensure that pybind will keep it alive. I cannot verify this.

@scotts, any thoughts?

src/torchcodec/_core/AVIOFileLikeContext.h
```diff
@@ -17,13 +17,16 @@ namespace facebook::torchcodec {
 
 // Enables uers to pass in a Python file-like object. We then forward all read
 // and seek calls back up to the methods on the Python object.
-class AVIOFileLikeContext : public AVIOContextHolder {
+// TODO: explain this. We probably don't want it.
+class __attribute__((visibility("default"))) AVIOFileLikeContext
+    : public AVIOContextHolder {
```
NicolasHug: So, this has been causing some headaches for me.

Forget about the `__attribute__((visibility("default"))) ` for a  second.
In this PR, the `AVIOFileLikeContext` class is now a direct dependency of the `Encoder` [1]. It means that this file must be compiled along with the `libtorchcodec_decoderN` library. Before (i.e. in `main`), this file was only compiled with `libtorchcodec_pybind_opsN`.

And we know from https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes that this class *should* be compiled with "hidden" visibility: it holds a `py::object fileLike_` attribute, which itself is set to "hidden" by pybind.

But if we were to set `AVIOFileLikeContext` to "hidden", then we'd have another problem: we'd get the same kind of warning, but for the `Encoder`, because the `Encoder` would not be "hidden":

```
Encoder declared with greater visibility than the type of its field Encoder::avioFileLikeContext_
```

We can't set the Encoder as hidden, that would make it invisible to custom ops and other things. It just wouldn't work.


[1] See comment below in this PR about the need for 2 different constructors: one for `AVIOFileLikeContext`, one for `AVIOToTensorContext`.

------

So here with `__attribute__((visibility("default")))` we are explicitly setting the visibility of `AVIOFileLikeContext` to "default".
"default" actually means **public**, i.e. **NOT hidden**. This override pybind's desire of having the `py::object fileLike_` to be hidden, and avoid the related warnings/errors. I *do not* think this is a great solution though, because IIUC, we're not actually solving the issue, we're mainly silencing a warning by forcing something to be "default"/public when it shouldn't. I think we are thus exposing ourselves to symbol conflicts:

> As to why -fvisibility=hidden is necessary, because pybind modules could have been compiled under different versions of pybind itself, it is also important that the symbols defined in one module do not clash with the potentially-incompatible symbols defined in another. While Python extension modules are usually loaded with localized symbols (under POSIX systems typically using dlopen with the RTLD_LOCAL flag), this Python default can be changed, but even if it isn’t it is not always enough to guarantee complete independence of the symbols involved when not using -fvisibility=hidden.

----

If we don't want to do this (by this I mean setting `AVIOFileLikeContext` to "default"), then I think we'll need to avoid relying on `AVIOFileLikeContext` within the `Encoder`. That means we will only want to interact with the base `AVIOContextHolder` class, and I think that means we'll need to expose the `getOutputTensor` as virtual, even know it will only make sense for `AVIOToTensorContext`. 

I hope this makes somewhat sense. I'm still wrapping my head around all this.

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -142,15 +143,6 @@ function(make_torchcodec_libraries
...
     )
-    # pybind11 limits the visibility of symbols in the shared library to prevent
-    # stray initialization of py::objects. The rest of the object code must
-    # match. See:
-    #   https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes
-    target_compile_options(
-        ${pybind_ops_library_name}
-        PUBLIC
-      "-fvisibility=hidden"
-    )
```
NicolasHug: This can be removed now because the new `__attribute__((visibility("default")))` has priority over any compilation flag. See other big comment above.

src/torchcodec/_core/pybind_ops.cpp
```diff
@@ -31,15 +33,58 @@ int64_t create_from_file_like(
...
+  m.def(
+      "encode_audio_to_file_like",
+      &encode_audio_to_file_like,
+      "data_ptr",
+      "shape",
+      "sample_rate",
+      "format",
+      "file_like",
+      "bit_rate",
+      "num_channels");
```
scotts: Oh, wow, we're not listing the parameters for `create_from_file_like()`. Are there benefits to doing so?

src/torchcodec/_core/pybind_ops.cpp
```diff
@@ -31,15 +33,58 @@ int64_t create_from_file_like(
...
+void encode_audio_to_file_like(
+    int64_t data_ptr,
+    const std::vector<int64_t>& shape,
+    int64_t sample_rate,
+    std::string_view format,
+    py::object file_like,
+    std::optional<int64_t> bit_rate = std::nullopt,
+    std::optional<int64_t> num_channels = std::nullopt) {
+  // We assume float32 *and* contiguity, this must be enforced by the caller.
+  auto tensor_options = torch::TensorOptions().dtype(torch::kFloat32);
```
scotts, line 54: If we keep this technique, we can probably allow all dtypes by passing in the dtype from the Python side as ints. I assume the Python and C++ enums agree on values, but even if they don't, we can figure out the mapping. Ugly, but possible.

src/torchcodec/_core/ops.py
scotts: On the keep-alive part, I believe Claude is right. If we pass something as a `py::object`, that gets properly reference-counted which will keep the object alive. When we launder a pointer as an int, there's no reference counting.

Of course, we would ideally just pass the tensor - but we run into problems passing tensors as tensors into the pybind11 code. The next simplest thing that we probably can't do for performance reasons is to copy the tensor into either bytes or a list, and then pass those as `py::object`. But since `samples` will be large, I don't think we want to do that.

Most workarounds I can think of are worse. One that might be just as bad, but could potentially apply to both this situation and decoder creation:

1. On the pybind11 side, we only create the `AVIOFileLikeContext`. We don't create the encoder or decoder. We do still accept the file-like objects, and they are still stored in the `AVIOFileLIkeContext`.
2. We return an int from the C++ side to the Python side where that int is a pointer to the `AVIOFileLikeContext`.
3. On the PyTorch custom ops side, we have functions for create-from-file-like and encode-to-file-like that accept the int value and do a `reinterpret_cast<AVIOFileLikeContext*>` in the C++. Those are then passed to the decoder or encode.

As it is right now, we're doing a lot of ugly pointer casting with tensors. The above may actually be better, as then the pybind11 code is only really concerned with creating AVIOFIleLikeContext objects. It doesn't even need to know about encoders and decoders.

src/torchcodec/_core/Encoder.h
```diff
@@ -20,13 +21,27 @@ class AudioEncoder {
...
+  // We need one constructor for each type of AVIOContextHolder. We can't have a
+  // single constructor that accepts the base AVIOContextHolder class and hold
+  // that as attribute, because we are calling the getOutputTensor() method on
+  // the AVIOToTensorContext, which is not available in the base class.
   AudioEncoder(
       const torch::Tensor& samples,
       int sampleRate,
       std::string_view formatName,
-      std::unique_ptr<AVIOToTensorContext> avioContextHolder,
+      std::unique_ptr<AVIOToTensorContext> AVIOToTensorContext,
```
scotts: Nit: variable name should start with lower case, maybe `avioToTensorContext`.

src/torchcodec/_core/Encoder.h
```diff
@@ -49,8 +64,8 @@ class AudioEncoder {
 
   const torch::Tensor samples_;
 
-  // Stores the AVIOContext for the output tensor buffer.
-  std::unique_ptr<AVIOToTensorContext> avioContextHolder_;
+  std::unique_ptr<AVIOToTensorContext> avioToTensorContext_;
+  std::unique_ptr<AVIOFileLikeContext> avioFileLikeContext_;
```
scotts, line 56: I think a simpler approach is to only store and accept a `AVIOContextHolder`, but when you need to access `AVIOToTensorContext` specific methods, do a [`std::dynamic_cast<AVIOToTensorContext*>`](https://www.en.cppreference.com/w/cpp/language/dynamic_cast.html). Dynamic casts are also ugly, but it localizes the ugliness to the one place we need it, and avoids having parallel variables and parallel constructors.

A dynamic cast, in C++, can be used to say "I only have a pointer to `Base`, but I know that this object is actually type `Derived`: return me a pointer to `Derived`." If it fails, you get back `nullptr`.

This _might_ solve the linking-visibility problem above, I'm not sure. The reason I think it might is that we remove `AVIOFileLikeContext` from the declaration of `Encoder`. It will only appear inside the implementation.

src/torchcodec/_core/AVIOFileLikeContext.h
scotts: @NicolasHug, we might be able to get around this, see my comment below about storing and accepting `AVIOContextHolder` and doing a dynamic cast to `AVIOToTensorContext` as needed.

NicolasHug: Yes, the dynamic cast allows us to only store the base `AVIOContextHolder` and avoid this problem. Thank you!

src/torchcodec/_core/Encoder.h
NicolasHug: Yes, it worked! Thank you!

src/torchcodec/_core/pybind_ops.cpp
NicolasHug: Nope, that's not needed, thanks for catching!

src/torchcodec/_core/pybind_ops.cpp
NicolasHug: Oh, I don't think we need to support more dtypes than just float32: the input samples that the user gives us must be float32 already. This comment is just here to explicitly state the assumptions that are made within `encode_audio_to_file_like`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -248,9 +248,10 @@ void AudioEncoder::initializeEncoder(
 torch::Tensor AudioEncoder::encodeToTensor() {
   TORCH_CHECK(
       avioContextHolder_ != nullptr,
-      "Cannot encode to tensor, avio context doesn't exist.");
+      "Cannot encode to tensor, avio tensor context doesn't exist.");
   encode();
-  return avioContextHolder_->getOutputTensor();
+  return dynamic_cast<AVIOToTensorContext*>(avioContextHolder_.get())
+      ->getOutputTensor();
```
scotts: Since `dynamic_cast` will return a `nullptr` if it can't do the cast, this a potential segfault risk. We should do a `TORCH_CHECK` on the result to make sure it's not a `nullptr` before accessing it.

scotts (approved):

PR #757, opened 2025-07-07
Adapt NPP calls for CUDA >= 12.9

Kh4L:
Recent CUDA versions don't support non-context NPP calls, so use the ctx-based API calls.
Also `CUDA 12.9+` deprecates `nppGetStreamContext`, so we need to build the NPP context manually.

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -224,41 +224,67 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
   // Use the user-requested GPU for running the NPP kernel.
   c10::cuda::CUDAGuard deviceGuard(device_);
 
-  NppiSize oSizeROI = {width, height};
-  Npp8u* input[2] = {avFrame->data[0], avFrame->data[1]};
+  cudaStream_t rawStream = at::cuda::getCurrentCUDAStream().stream();
+
+  // Build an NppStreamContext, either via the old helper or by hand on CUDA 12.9+
+  NppStreamContext nppCtx{};
+  #if CUDA_VERSION < 12090
+    NppStatus ctxStat = nppGetStreamContext(&nppCtx);
+    TORCH_CHECK(ctxStat == NPP_SUCCESS, "nppGetStreamContext failed");
+    // override if you want to force a particular stream
+    nppCtx.hStream = rawStream;
+  #else
+    // CUDA 12.9+: helper was removed, we need to build it manually
+    int dev = 0;
+    cudaError_t err = cudaGetDevice(&dev);
+    TORCH_CHECK(err == cudaSuccess, "cudaGetDevice failed");
+    cudaDeviceProp prop{};
+    err = cudaGetDeviceProperties(&prop, dev);
+    TORCH_CHECK(err == cudaSuccess, "cudaGetDeviceProperties failed");
+
+    nppCtx.nCudaDeviceId                       = dev;
+    nppCtx.nMultiProcessorCount                = prop.multiProcessorCount;
+    nppCtx.nMaxThreadsPerMultiProcessor        = prop.maxThreadsPerMultiProcessor;
+    nppCtx.nMaxThreadsPerBlock                 = prop.maxThreadsPerBlock;
+    nppCtx.nSharedMemPerBlock                  = prop.sharedMemPerBlock;
+    nppCtx.nCudaDevAttrComputeCapabilityMajor  = prop.major;
+    nppCtx.nCudaDevAttrComputeCapabilityMinor  = prop.minor;
+    nppCtx.nStreamFlags                        = 0;
+    nppCtx.hStream                             = rawStream;
+  #endif
+
+  // Prepare ROI + pointers
+  NppiSize oSizeROI = { width, height };
+  Npp8u* input[2]  = { avFrame->data[0], avFrame->data[1] };
 
   auto start = std::chrono::high_resolution_clock::now();
   NppStatus status;
+
   if (avFrame->colorspace == AVColorSpace::AVCOL_SPC_BT709) {
-    status = nppiNV12ToRGB_709CSC_8u_P2C3R(
+    status = nppiNV12ToRGB_709CSC_8u_P2C3R_Ctx(
         input,
         avFrame->linesize[0],
         static_cast<Npp8u*>(dst.data_ptr()),
         dst.stride(0),
-        oSizeROI);
+        oSizeROI,
+        nppCtx);
   } else {
-    status = nppiNV12ToRGB_8u_P2C3R(
+    status = nppiNV12ToRGB_8u_P2C3R_Ctx(
         input,
         avFrame->linesize[0],
         static_cast<Npp8u*>(dst.data_ptr()),
         dst.stride(0),
-        oSizeROI);
+        oSizeROI,
+        nppCtx);
   }
   TORCH_CHECK(status == NPP_SUCCESS, "Failed to convert NV12 frame.");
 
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -224,41 +224,66 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
   }
   TORCH_CHECK(status == NPP_SUCCESS, "Failed to convert NV12 frame.");
 
-  // Make the pytorch stream wait for the npp kernel to finish before using the
-  // output.
-  at::cuda::CUDAEvent nppDoneEvent;
-  at::cuda::CUDAStream nppStreamWrapper =
-      c10::cuda::getStreamFromExternal(nppGetStream(), device_.index());
-  nppDoneEvent.record(nppStreamWrapper);
-  nppDoneEvent.block(at::cuda::getCurrentCUDAStream());
```
NicolasHug, line 298: @Kh4L Can you confirm my understanding that the `nppiNV12ToRGB_8u_P2C3R_Ctx` call well properly wait on the stream before returning?

Kh4L: Thank you!

> Do I understand correctly that in 12.9 we have to use the context-based API, while at the same time the context creation helper was removed?! This sounds error prone, is there any way we could avoid manually building and setting the context attributes?

Unfortunately, there is currently no alternative way to build this.
Since I’m not part of the npp team, I can’t comment on their design choices

> @Kh4L Can you confirm my understanding that the nppiNV12ToRGB_8u_P2C3R_Ctx call well properly wait on the stream before returning?

That’s correct. We bind the NPP context to the active CUDA stream so we can leverage CUDA stream management rather than performing a blocking sync

https://github.com/pytorch/torchcodec/pull/757/files#diff-37d8a09669d3f009b6850f6e66888b6875d805064933148fce3a637cc7694712R254

Kh4L:
@NicolasHug LMK if you need anything else! 

the assertion error doesn't seem related to my change

NicolasHug:
Nothing else to do on your side @Kh4L , thank you. I'll merge this soon, I'll just try to extract all the `#ifdef` stuff into their own single function before merging.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -224,41 +224,66 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
+  NppStreamContext nppCtx{};
+#if CUDA_VERSION < 12090
+  NppStatus ctxStat = nppGetStreamContext(&nppCtx);
+  TORCH_CHECK(ctxStat == NPP_SUCCESS, "nppGetStreamContext failed");
+  // override if you want to force a particular stream
+  nppCtx.hStream = rawStream;
+#else
+  // CUDA 12.9+: helper was removed, we need to build it manually
+  int dev = 0;
+  cudaError_t err = cudaGetDevice(&dev);
```
NicolasHug: Hi @Kh4L , I actually have some questions before moving forward:

1. Should we just rely on the existing `device_`  attribute instead of calling `cudaGetDevice(&dev)`, or are they actually equivalent?
2. Would it make sense to cache the `nppCtx` across calls? In this PR it looks like we're creating the context over and over for every single frame that needs to be decoded. I wonder if it might be beneficial to cache it in the class and re-use it?

Thanks for your help so far, I'm still trying to build familiarity with that part of the code base.

Kh4L: Hi @NicolasHug 

Just saw your comment.

Yes we should be able to rely on the existing `device_`, thanks for pushing a commit with the proposed changes!

NicolasHug:
Let me try to summarize the changes and add context for other reviewers, and for future reference.

From their [release notes](https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#npp-release-12-9), CUDA 12.9 deprecates:

- calls to non-context APIs. We now need to use `nppiNV12ToRGB_8u_P2C3R_Ctx` instead of `nppiNV12ToRGB_8u_P2C3R`
- the `nppGetStreamContext()` helper, which we were previously using to create an `NppContext`. They explicitly recommend to manually create such context, as done here: https://github.com/NVIDIA/CUDALibrarySamples/tree/master/NPP#example-using-application-managed-context.

Removing  the non-context APIs makes sense to me. I don't understand the logic behind removing `nppGetStreamContext()`. [And I'm not the only one](https://forums.developer.nvidia.com/t/nppgetstreamcontext-disappeared-from-cuda-12-9/332269/10). But OK. 

We now have to manually create the `NppContext`, which is the bullk of this PR. Note that `opencv` is doing something very similar as well: https://github.com/opencv/opencv/pull/27288/files.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -265,37 +303,37 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
 
-  // Use the user-requested GPU for running the NPP kernel.
-  c10::cuda::CUDAGuard deviceGuard(device_);
+  // TODO cache the NppStreamContext! It currently gets re-recated for every
+  // single frame. The cache should be per-device, similar to the existing
+  // hw_device_ctx cache. When implementing the cache logic, the
+  // NppStreamContext hStream and nStreamFlags should not be part of the cache
+  // because they may change across calls.
+  NppStreamContext nppCtx = createNppStreamContext(
+      static_cast<int>(getFFMPEGCompatibleDeviceIndex(device_)));
```
NicolasHug, line 312: A note on the cache: I originally implemented the "cache" as a  simple `nppCtx_` attribute on the `CudaDeviceInterface` class: [`565896e` (#757)](https://github.com/pytorch/torchcodec/pull/757/commits/565896e4935113eab2a50dc66124548b0acb0025)


But I don't think that would be correct: the `CudaDeviceInterface` instance is global, and we only have one single instance for *all* CUDA devices. And we can't use one single `NppContext` for all CUDA devices - we need one `NppContext` **per** device. 

So, we need a per-device cache for the `NppContext`, similar to our existing `hw_device_ctx` cache. I'm leaving that for an immediate follow-up.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -265,37 +303,37 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
     dst = allocateEmptyHWCTensor(height, width, device_);
   }
 
-  // Use the user-requested GPU for running the NPP kernel.
-  c10::cuda::CUDAGuard deviceGuard(device_);
```
NicolasHug, line 269: This guard isn't needed anymore as we now explicitly pass the current device to the NppContext creation.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -265,37 +303,37 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
   }
   TORCH_CHECK(status == NPP_SUCCESS, "Failed to convert NV12 frame.");
-
-  // Make the pytorch stream wait for the npp kernel to finish before using the
-  // output.
-  at::cuda::CUDAEvent nppDoneEvent;
-  at::cuda::CUDAStream nppStreamWrapper =
-      c10::cuda::getStreamFromExternal(nppGetStream(), device_.index());
-  nppDoneEvent.record(nppStreamWrapper);
-  nppDoneEvent.block(at::cuda::getCurrentCUDAStream());
```
NicolasHug, line 298: These syncs aren't needed anymore because we now explicitly ask `Npp` to rely on pytorch's current stream.

NicolasHug (approved):

PR #759, opened 2025-07-08
Generalize pybind_ops module name

scotts:
Enables the name used for the pybind11 C++ module to be configured at compile time. Note that making this work requires making sure that the compile time name also matches the name we return dynamically at runtime; see the comments.

Right now, we're not taking advantage of this capability. We plan to in the near future.

### DIFF
--- a/src/torchcodec/_internally_replaced_utils.py
+++ b/src/torchcodec/_internally_replaced_utils.py
@@ -58,4 +58,7 @@ def _get_pybind_ops_module_name(ffmpeg_major_version: str) -> str:
     # when we compile _core/pybind_ops.cpp. If the values do not match, we will
     # not be able to import the C++ shared library as a Python module at
     # runtime.
+    #
+    # The parameter ffmpeg_major_version is unused externally, but used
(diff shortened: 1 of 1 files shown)


src/torchcodec/_internally_replaced_utils.py
```diff
@@ -51,3 +51,11 @@ def _load_pybind11_module(module_name: str, library_path: str) -> ModuleType:
         )
 
     return importlib.util.module_from_spec(spec)
+
+
+def _get_pybind_ops_module_name(ffmpeg_major_version: str) -> str:
+    # Note that this value must match the value used as PYBIND_OPS_MODULE_NAME
+    # when we compile _core/pybind_ops.cpp. If the values do not match, we will
+    # not be able to import the C++ shared library as a Python module at
+    # runtime.
```
NicolasHug, line 60: Maybe just add a comment clarifying that the `ffmpeg_major_version` is only required for the internal version of this util

NicolasHug (approved):

PR #669, opened 2025-05-04
test(video): add test cases for variable framerate videos

XueSongTap:
Add comprehensive unit tests for variable framerate video decoding functionality:
- Test basic decoding of variable framerate content
- Verify exact seeking mode works correctly with variable frame rates
- Compare exact vs approximate seeking behavior
- Validate frame timing patterns match expected variable rate
- Test sequential decoding with monotonically increasing timestamps
- Ensure batch frame retrieval works with variable durations

These tests ensure the decoder properly handles videos where the frame rate
changes throughout the content, particularly during seek operations.

Related to #653

selt test:
```
(torchcodec-env) yxc@yxc-MS-7B89:~/code/2504/torchcodec_dev$ pytest test/test_variable_fps_video.py -v
=============================================== test session starts ===============================================
platform linux -- Python 3.10.17, pytest-8.3.5, pluggy-1.5.0 -- /home/yxc/.local/share/mamba/envs/torchcodec-env/bin/python3.10
cachedir: .pytest_cache
rootdir: /home/yxc/code/2504/torchcodec_dev
configfile: pyproject.toml
collected 12 items

test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_basic_decoding[cpu] PASSED               [  8%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_basic_decoding[cuda] SKIPPED (CUDA n...) [ 16%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_exact_seeking_mode[cpu] PASSED           [ 25%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_exact_seeking_mode[cuda] SKIPPED (CU...) [ 33%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_approximate_seeking_mode_behavior[cpu] PASSED [ 41%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_approximate_seeking_mode_behavior[cuda] SKIPPED [ 50%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_frame_timing_pattern[cpu] PASSED         [ 58%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_frame_timing_pattern[cuda] SKIPPED (...) [ 66%]
test/test_variable_fps_video.py::TestVariableFPSVideoDecoder::test_sequential_decoding[cpu] PASSED          [ 75%]
[...]

PR #448, opened 2025-01-09
fix: Solve CUDA AV1 decoding

hugo-ijw:
fixes #443
Hello there, love the library!
It looks like there is an FFmpeg bug when selecting an AV1 decoder while specifying an HW decoder; a SW decoder is always chosen (cf. https://github.com/FFmpeg/FFmpeg/commit/ad67ea9eeed2cc8edab8662b7503ce4a49247b1d).

I also added small details to the contributing doc and an AV1 sample video for tests (I do not know if the GPU used in your CI/CD is compatible with CUDA AV1 decoding, let's see).

PR #761, opened 2025-07-09
Variable FPS tests

scotts:
Creating a new PR based off of #669. I can't push to that PR, so we need a new one.

This PR adds a new video which has variable FPS, commits some reference frames from it, and then uses it in two tests. Notably we only test with exact seeking mode. These tests would fail in approximate mode because we would get the wrong frames.

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1139,7 +1139,6 @@ class TestVideoDecoder:
         with pytest.raises(AssertionError, match="not equal"):
             torch.testing.assert_close(decoder[0], decoder[10])
 
-
     @pytest.mark.parametrize("device", cpu_and_cuda())
     def test_var_fps_single_frame(self, device):
         # We are explicitly only testing exact mode because we know approximate
@@ -1159,7 +1158,6 @@ class TestVideoDecoder:
         assert_frames_equal(frame_150, ref_frame_150)
         assert_frames_equal(frame_299, ref_frame_299)
 
-
     @pytest.mark.parametrize("device", cpu_and_cuda())
     def test_var_fps_range(self, device):
         # We are explicitly only testing exact mode because we know approximate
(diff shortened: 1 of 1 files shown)


test/generate_reference_resources.sh
```diff
@@ -56,12 +56,6 @@ for frame in "${FRAMES[@]}"; do
   ffmpeg -y -i "$VIDEO_PATH" -vf select="eq(n\,$frame)" -vsync vfr -q:v 2 "$VIDEO_PATH.stream0.frame$frame_name.bmp"
 done
 
-for bmp in "$RESOURCES_DIR"/*.bmp
-do
-  python3 "$TORCHCODEC_PATH/test/convert_image_to_tensor.py" "$bmp"
-  rm -f "$bmp"
-done
-
```
scotts, line 64: Drive by. I realized we had this logic twice: here and again at the end of the script. We only want it once. The dupe was introduced back in #448.

scotts:
Status: the current video in the PR is failing on CUDA with errors similar to what we fixed with #448. I want to re-create the video, but I am unable to. The comment in the script uses the command:
```
ffmpeg -f lavfi -i testsrc=duration=10:size=128x128:rate=30 -vf "setpts=PTS*if(gt(N\,30)\,1.5\,1)" -c:v libx264 -bf 0 var_fps_video.mp4
```
But I don't have libx264 in my environment. I do have libopenh264, but when I use that to encode the video, I see the warnings:
```
[libopenh264 @ 0x47a3c0] [OpenH264] this = 0x0x644690, Warning:layerId(0) doesn't support profile(578), change to UNSPECIFIC profile
[libopenh264 @ 0x47a3c0] [OpenH264] this = 0x0x644690, Warning:bEnableFrameSkip = 0,bitrate can't be controlled for RC_QUALITY_MODE,RC_BITRATE_MODE and RC_TIMESTAMP_MODE without enabling skip frame.
[libopenh264 @ 0x47a3c0] [OpenH264] this = 0x0x644690, Warning:Change QP Range from(0,51) to (12,42)
```
And the resulting video does _not_ have a variable frame rate. Shelving this for now until we can reliably recreate a variable framer rate video.

Issue #466, opened 2025-01-22
Allow receiving an external frame index


scotts:
### 🚀 The feature

We have had several request for the ability for `VideoDecoder` to accept an *external frame index* upon startup. That is, in exact mode, we build up a frame index which tells us the mapping between frame indices and pts. If users could provide that index to us, we could be in exact mode but not have to scan the entire file at startup. (We would, however, be the the mercy of the accuracy of the external index.)

### Motivation, pitch

The main motivation is performance. There is potentially less need for this feature now that we have approximate mode (#427), which also avoids scanning the entire file. But the combination of receiving an external index and exact mode could perhaps yield the best of both worlds: exact seeks and fast startup.

The reason this comes up is that in many environments, videos are pre-processed before being used for training and inference. Creating and storing the external index for a particular video would be relatively low-cost, and it would be paid once.

NicolasHug:
From a UX perspective, we may want to think of the user-provided index a third `seek_mode`? The reason I say this is because when the index is provided by the user, we end up in a hybrid mode that doesn't really fit neither approximate nor exact. It's more of a combination of both:

- approximate: we don't scan, we don't rely on an index
- exact: we scan and rely on the index we built ourselves
- user-provided index: we don't scan (like in approx), but still use an index (like in exact).

scotts:
That's a good point. I was thinking of the two concepts as being orthogonal, but you're right, they're related. That is, it doesn't make much sense to provide an external index if you're also using approximate mode - why provide us with something that you then ask us to not use? Ideally, our API would prevent such logical inconsistencies.

Dan-Flores:
Additionally, we should consider which format(s) we want to support for an external frame index. One easy way to get a frame index is ffprobe's `-show_frames` command, which supports several formats including CSV and JSON.
```
ffprobe -i input.mp4 -select_streams v:0 -show_frames -print_format json
```
Ideally, I think we should support the most common format, or multiple common formats.

scotts:
@Dan-Flores, good question! I actually see this as two questions:

1. What is the _schema_ we should accept? That is, what data do we need, what are the types of that data, and how is it structured?
2. Given a schema, what _format_ do we accept? This is the question of, say, CSV or JSON. The same schema can exist in multiple formats.

I do, however, have a single answer to both questions: I think we should do the schema returned by `ffprobe`, and we should use JSON. My reasoning:

[...]

PR #764, opened 2025-07-10
Add stream_index seek mode, read frame index and update metadata

Dan-Flores:
This PR works towards enabling decoders to receive an external frame index, as described in #466. It is only enabled in the private C++ code, and not exposed in the Python yet. 

### The changes:
* The stream index is expected to be passed into `add_video_stream` in the format of 3 tensors in a tuple. 
  * The reason it is not passed into the SingleStreamDecoder constructor (as discussed in #466)  is that a `stream_index` needs to be specified to update the decoder's metadata from the ffprobe output, and that `stream_index` is explicitly available during the `add_video_stream`  call. 
* The metadata updates happen in the added function `readFrameIndexUpdateMetadataAndIndex`, which parallels the function used by exact mode, `scanFileAndUpdateMetadataAndIndex`. 
  * It assumes PTS ints are passed in, since that is the expected type used in `FrameInfo`.

### Testing:
* To create a frame index in the format needed, a new`frame_index` property is added to `TestContainerFile`, which runs, parses, and sorts the ffprobe output.
* I updated `test_get_metadata` to allow the new seek mode to work.
* I added `test_seek_mode_frame_index`, which tests a variety of frame grabbing functions in the new seek mode.

### DIFF
--- a/src/torchcodec/_core/Frame.h
+++ b/src/torchcodec/_core/Frame.h
@@ -45,6 +45,20 @@ struct AudioFramesOutput {
   double ptsSeconds;
 };
 
+// FrameMappings is used for the custom_frame_mappings seek mode to store
+// metadata of frames in a stream. The size of all tensors in this struct must
+// match.
+struct FrameMappings {
+  // 1D tensor of int64, each value is the PTS of a frame in timebase units.
+  torch::Tensor all_frames;
+  // 1D tensor of bool, each value indicates if the corresponding frame in
+  // all_frames is a key frame.
+  torch::Tensor is_key_frame;
+  // 1D tensor of int64, each value is the duration of the corresponding frame
+  // in all_frames in timebase units.
+  torch::Tensor duration;
+};
+
 // --------------------------------------------------------------------------
 // FRAME TENSOR ALLOCATION APIs
 // --------------------------------------------------------------------------
--- a/src/torchcodec/_core/Frame.h
+++ b/src/torchcodec/_core/Frame.h
@@ -45,20 +45,6 @@ struct AudioFramesOutput {
   double ptsSeconds;
 };
 
-// FrameMappings is used for the custom_frame_mappings seek mode to store
-// metadata of frames in a stream. The size of all tensors in this struct must
-// match.
-struct FrameMappings {
-  // 1D tensor of int64, each value is the PTS of a frame in timebase units.
-  torch::Tensor all_frames;
-  // 1D tensor of bool, each value indicates if the corresponding frame in
-  // all_frames is a key frame.
-  torch::Tensor is_key_frame;
-  // 1D tensor of int64, each value is the duration of the corresponding frame
-  // in all_frames in timebase units.
-  torch::Tensor duration;
-};
-
 // --------------------------------------------------------------------------
 // FRAME TENSOR ALLOCATION APIs
 // --------------------------------------------------------------------------
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -348,8 +348,8 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
 
   streamMetadata.numFramesFromContent = all_frames.size(0);
   for (int64_t i = 0; i < all_frames.size(0); ++i) {
-    // FrameInfo struct utilizes PTS
-    FrameInfo frameInfo = {.pts = all_frames[i].item<int64_t>()};
+    FrameInfo frameInfo;
+    frameInfo.pts = all_frames[i].item<int64_t>();
     frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
     streamInfos_[streamIndex].allFrames.push_back(frameInfo);
     if (frameInfo.isKeyFrame) {
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -346,7 +346,7 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   streamMetadata.numFramesFromContent = all_frames.size(0);
   for (int64_t i = 0; i < all_frames.size(0); ++i) {
     // FrameInfo struct utilizes PTS
-    FrameInfo frameInfo = {.pts=all_frames[i].item<int64_t>()};
+    FrameInfo frameInfo = {.pts = all_frames[i].item<int64_t>()};
     frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
     frameInfo.nextPts = (i + 1 < all_frames.size(0))
         ? all_frames[i + 1].item<int64_t>()
@@ -354,7 +354,7 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
     streamInfos_[streamIndex].allFrames.push_back(frameInfo);
     if (frameInfo.isKeyFrame) {
       streamInfos_[streamIndex].keyFrames.push_back(frameInfo);
-    } 
+    }
   }
 }
 
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -346,12 +346,15 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   streamMetadata.numFramesFromContent = all_frames.size(0);
   for (int64_t i = 0; i < all_frames.size(0); ++i) {
     // FrameInfo struct utilizes PTS
-    FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
+    FrameInfo frameInfo = {.pts=all_frames[i].item<int64_t>()};
     frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
     frameInfo.nextPts = (i + 1 < all_frames.size(0))
         ? all_frames[i + 1].item<int64_t>()
         : INT64_MAX;
     streamInfos_[streamIndex].allFrames.push_back(frameInfo);
+    if (frameInfo.isKeyFrame) {
+      streamInfos_[streamIndex].keyFrames.push_back(frameInfo);
+    } 
   }
 }
 
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -198,6 +198,41 @@ int SingleStreamDecoder::getBestStreamIndex(AVMediaType mediaType) {
 // VIDEO METADATA QUERY API
 // --------------------------------------------------------------------------
 
+void SingleStreamDecoder::sortAllFrames() {
+  for (auto& [streamIndex, streamInfo] : streamInfos_) {
+    std::sort(
+        streamInfo.keyFrames.begin(),
+        streamInfo.keyFrames.end(),
+        [](const FrameInfo& frameInfo1, const FrameInfo& frameInfo2) {
+          return frameInfo1.pts < frameInfo2.pts;
+        });
+    std::sort(
+        streamInfo.allFrames.begin(),
+        streamInfo.allFrames.end(),
+        [](const FrameInfo& frameInfo1, const FrameInfo& frameInfo2) {
+          return frameInfo1.pts < frameInfo2.pts;
+        });
+
+    size_t keyFrameIndex = 0;
+    for (size_t i = 0; i < streamInfo.allFrames.size(); ++i) {
+      streamInfo.allFrames[i].frameIndex = i;
+      if (streamInfo.allFrames[i].isKeyFrame) {
+        TORCH_CHECK(
+            keyFrameIndex < streamInfo.keyFrames.size(),
+            "The allFrames vec claims it has MORE keyFrames than the keyFrames vec. There's a bug in torchcodec.");
+        streamInfo.keyFrames[keyFrameIndex].frameIndex = i;
+        ++keyFrameIndex;
+      }
+      if (i + 1 < streamInfo.allFrames.size()) {
+        streamInfo.allFrames[i].nextPts = streamInfo.allFrames[i + 1].pts;
+      }
+    }
+    TORCH_CHECK(
+        keyFrameIndex == streamInfo.keyFrames.size(),
+        "The allFrames vec claims it has LESS keyFrames than the keyFrames vec. There's a bug in torchcodec.");
+  }
+}
+
 void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
   if (scannedAllStreams_) {
     return;
@@ -283,39 +318,7 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
       getFFMPEGErrorStringFromErrorCode(status));
 
   // Sort all frames by their pts.
-  for (auto& [streamIndex, streamInfo] : streamInfos_) {
-    std::sort(
-        streamInfo.keyFrames.begin(),
-        streamInfo.keyFrames.end(),
-        [](const FrameInfo& frameInfo1, const FrameInfo& frameInfo2) {
-          return frameInfo1.pts < frameInfo2.pts;
-        });
-    std::sort(
-        streamInfo.allFrames.begin(),
-        streamInfo.allFrames.end(),
-        [](const FrameInfo& frameInfo1, const FrameInfo& frameInfo2) {
-          return frameInfo1.pts < frameInfo2.pts;
-        });
-
-    size_t keyFrameIndex = 0;
-    for (size_t i = 0; i < streamInfo.allFrames.size(); ++i) {
-      streamInfo.allFrames[i].frameIndex = i;
-      if (streamInfo.allFrames[i].isKeyFrame) {
-        TORCH_CHECK(
-            keyFrameIndex < streamInfo.keyFrames.size(),
-            "The allFrames vec claims it has MORE keyFrames than the keyFrames vec. There's a bug in torchcodec.");
-        streamInfo.keyFrames[keyFrameIndex].frameIndex = i;
-        ++keyFrameIndex;
-      }
-      if (i + 1 < streamInfo.allFrames.size()) {
-        streamInfo.allFrames[i].nextPts = streamInfo.allFrames[i + 1].pts;
-      }
-    }
-    TORCH_CHECK(
-        keyFrameIndex == streamInfo.keyFrames.size(),
-        "The allFrames vec claims it has LESS keyFrames than the keyFrames vec. There's a bug in torchcodec.");
-  }
-
+  sortAllFrames();
   scannedAllStreams_ = true;
 }
 
@@ -356,6 +359,8 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
       streamInfos_[streamIndex].keyFrames.push_back(frameInfo);
     }
   }
+  // Sort all frames by their pts
+  sortAllFrames();
 }
 
 ContainerMetadata SingleStreamDecoder::getContainerMetadata() const {
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -199,6 +199,10 @@ int SingleStreamDecoder::getBestStreamIndex(AVMediaType mediaType) {
 // --------------------------------------------------------------------------
 
 void SingleStreamDecoder::sortAllFrames() {
+  // Sort the allFrames and keyFrames vecs in each stream, and also sets
+  // additional fields of the FrameInfo entries like nextPts and frameIndex
+  // This is called at the end of a scan, or when setting a user-defined frame
+  // mapping.
   for (auto& [streamIndex, streamInfo] : streamInfos_) {
     std::sort(
         streamInfo.keyFrames.begin(),
@@ -350,7 +354,7 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   for (int64_t i = 0; i < all_frames.size(0); ++i) {
     FrameInfo frameInfo;
     frameInfo.pts = all_frames[i].item<int64_t>();
-    frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
+    frameInfo.isKeyFrame = is_key_frame[i].item<bool>();
     streamInfos_[streamIndex].allFrames.push_back(frameInfo);
     if (frameInfo.isKeyFrame) {
       streamInfos_[streamIndex].keyFrames.push_back(frameInfo);
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -321,10 +321,10 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
 
 void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
     int streamIndex,
-    CustomFrameMappings customFrameMappings) {
-  auto& all_frames = customFrameMappings.all_frames;
-  auto& is_key_frame = customFrameMappings.is_key_frame;
-  auto& duration = customFrameMappings.duration;
+    std::tuple<at::Tensor, at::Tensor, at::Tensor> customFrameMappings) {
+  auto& all_frames = std::get<0>(customFrameMappings);
+  auto& is_key_frame = std::get<1>(customFrameMappings);
+  auto& duration = std::get<2>(customFrameMappings);
   TORCH_CHECK(
       all_frames.size(0) == is_key_frame.size(0) &&
           is_key_frame.size(0) == duration.size(0),
@@ -468,7 +468,8 @@ void SingleStreamDecoder::addStream(
 void SingleStreamDecoder::addVideoStream(
     int streamIndex,
     const VideoStreamOptions& videoStreamOptions,
-    std::optional<CustomFrameMappings> customFrameMappings) {
+    std::optional<std::tuple<at::Tensor, at::Tensor, at::Tensor>>
+        customFrameMappings) {
   addStream(
       streamIndex,
       AVMEDIA_TYPE_VIDEO,
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -326,7 +326,8 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   auto& is_key_frame = std::get<1>(customFrameMappings);
   auto& duration = std::get<2>(customFrameMappings);
   TORCH_CHECK(
-      all_frames.size(0) == is_key_frame.size(0) && is_key_frame.size(0) == duration.size(0),
+      all_frames.size(0) == is_key_frame.size(0) &&
+          is_key_frame.size(0) == duration.size(0),
       "all_frames, is_key_frame, and duration from custom_frame_mappings were not same size.");
 
   auto& streamMetadata = containerMetadata_.allStreamMetadata[streamIndex];
@@ -346,8 +347,7 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   for (int64_t i = 0; i < all_frames.size(0); ++i) {
     // FrameInfo struct utilizes PTS
     FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
-    frameInfo.isKeyFrame =
-        (is_key_frame[i].item<int64_t>() == 1);
+    frameInfo.isKeyFrame = (is_key_frame[i].item<int64_t>() == 1);
     frameInfo.nextPts = (i + 1 < all_frames.size(0))
         ? all_frames[i + 1].item<int64_t>()
         : INT64_MAX;
@@ -468,7 +468,8 @@ void SingleStreamDecoder::addStream(
 void SingleStreamDecoder::addVideoStream(
     int streamIndex,
     const VideoStreamOptions& videoStreamOptions,
-    std::optional<std::tuple<at::Tensor, at::Tensor, at::Tensor>> customFrameMappings) {
+    std::optional<std::tuple<at::Tensor, at::Tensor, at::Tensor>>
+        customFrameMappings) {
   addStream(
       streamIndex,
       AVMEDIA_TYPE_VIDEO,
@@ -498,7 +499,8 @@ void SingleStreamDecoder::addVideoStream(
     TORCH_CHECK(
         customFrameMappings.has_value(),
         "Please provide frame mappings when using custom_frame_mappings seek mode.");
-    readCustomFrameMappingsUpdateMetadataAndIndex(streamIndex, customFrameMappings.value());
+    readCustomFrameMappingsUpdateMetadataAndIndex(
+        streamIndex, customFrameMappings.value());
   }
 }
 
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -319,12 +319,12 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
   scannedAllStreams_ = true;
 }
 
-void SingleStreamDecoder::readFrameIndexUpdateMetadataAndIndex(
+void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
     int streamIndex,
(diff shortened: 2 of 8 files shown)


test/test_metadata.py
```diff
@@ -82,7 +79,7 @@ def test_get_metadata(metadata_getter):
     assert best_video_stream_metadata.begin_stream_seconds_from_header == 0
     assert best_video_stream_metadata.bit_rate == 128783
     assert best_video_stream_metadata.average_fps == pytest.approx(29.97, abs=0.001)
-    assert best_video_stream_metadata.pixel_aspect_ratio is None
+    assert best_video_stream_metadata.pixel_aspect_ratio == Fraction(1, 1)
```
Dan-Flores: Since `add_video_stream` is always being called, this value is set instead of `None`.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -266,15 +269,20 @@ void add_video_stream(
     std::optional<int64_t> num_threads = std::nullopt,
     std::optional<std::string_view> dimension_order = std::nullopt,
     std::optional<int64_t> stream_index = std::nullopt,
-    std::optional<std::string_view> device = std::nullopt) {
+    std::optional<std::string_view> device = std::nullopt,
+    std::optional<std::string_view> color_conversion_library = std::nullopt,
```
Dan-Flores: This is not related to the main change in this PR, but I added the `color_conversion_library` argument, since it is available in `_add_video_stream`, but not `add_video_stream`.

Currently, [tests utilize _add_video_stream](https://github.com/pytorch/torchcodec/blob/eb51e0e97f456cf92c5c6987f6ebeae824ebab5a/test/test_ops.py#L454) whenever `color_conversion_library` is needed. Please let me know if this was intentional, and I can remove this change.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
+    frameInfo.isKeyFrame =
+        (i < key_frames.size(0) && key_frames[i].item<int64_t>() == 1);
+    frameInfo.nextPts = (i + 1 < all_frames.size(0))
+        ? all_frames[i + 1].item<int64_t>()
+        : INT64_MAX;
+    streamInfos_[streamIndex].allFrames.push_back(frameInfo);
+  }
+  readFrameIndex_ = true;
```
Dan-Flores: I'm not sure `readFrameIndex_` is necessary, as there should be an error if `add_video_stream` is called more than once, but I added this check for consistency.

test/utils.py
```diff
@@ -121,6 +122,7 @@ class TestContainerFile:
     default_stream_index: int
     stream_infos: Dict[int, Union[TestVideoStreamInfo, TestAudioStreamInfo]]
     frames: Dict[int, Dict[int, TestFrameInfo]]
+    frame_index_data: tuple[torch.Tensor, torch.Tensor, torch.Tensor] | None = None
```
NicolasHug: This line is causing [errors on 3.9](https://github.com/pytorch/torchcodec/actions/runs/16186249349/job/45693405808?pr=764):

```
test/utils.py:127: in TestContainerFile
    frame_index_data: tuple[torch.Tensor, torch.Tensor, torch.Tensor] | None = None
E   TypeError: unsupported operand type(s) for |: 'types.GenericAlias' and 'NoneType'
```

I think we'd have to do `from __future__ import annotations` for `|` to be a valid 3.9 typing operator. But in general we've just been using `Optional` when something can be None.

Personally, I prefer the way you've done it here. `Optional` is IMO a bad name, because it conflicts with the notion of "optional parameter", i.e. a parameter which can have a default. That's very different from a parameter that can be ``None``, which is what `Optional` defines. In any case, we should probably just keep using `Optional` here, and move towards `Union` or `|` everywhere in the code base at some point.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
   scannedAllStreams_ = true;
 }
 
+void SingleStreamDecoder::readFrameIndexUpdateMetadataAndIndex(
+    int streamIndex,
+    std::tuple<at::Tensor, at::Tensor, at::Tensor> frameIndex) {
```
NicolasHug: Nit: we're currently using both `at::Tensor` and `torch::Tensor` throughout the codebase, we should probably align on one single use. `torch::Tensor` seems to be used more often:

```
~/dev/torchcodec (seek_mode_frame_index*) » git grep torch::Tensor | wc -l       
82
(codec) ---------------------------------------------------------------------------------------------------------------
~/dev/torchcodec (seek_mode_frame_index*) » git grep at::Tensor | wc -l      
38
```

Maybe just open an issue to follow-up on this?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  if (readFrameIndex_) {
+    return;
+  }
```
NicolasHug: Calling `readFrameIndexUpdateMetadataAndIndex` twice with different `frameIndex` values might lead to unexpected results, because the frame index won't be updated on the second call.

As you mentioned in your comment, our code never actually calls `readFrameIndexUpdateMetadataAndIndex` twice, so this sequence of calls shouldn't occur in practice. However, for someone new to the codebase, it might be surprising this early return. To address this, I suggest either:

- Removing the `readFrameIndex_` logic and allowing `readFrameIndexUpdateMetadataAndIndex` to be called multiple times, updating the index each time.
- Raising a clear error if it's called more than once, instead of exiting early

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: I left a comment on that just below

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  auto& key_frames = std::get<1>(frameIndex);
+  auto& duration = std::get<2>(frameIndex);
+
+  auto& streamMetadata = containerMetadata_.allStreamMetadata[streamIndex];
+
+  // Get the last index for key_frames and duration
+  auto last_idx = all_frames.size(0) - 1;
+  streamMetadata.beginStreamPtsFromContent = all_frames[0].item<int64_t>();
+  streamMetadata.endStreamPtsFromContent =
+      all_frames[last_idx].item<int64_t>() + duration[last_idx].item<int64_t>();
```
NicolasHug: I would assume `-1` would work like it does in Python, but I'm not 100% sure:

```suggestion
      all_frames[-1].item<int64_t>() + duration[-1].item<int64_t>();
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  if (readFrameIndex_) {
+    return;
+  }
+  auto& all_frames = std::get<0>(frameIndex);
+  auto& key_frames = std::get<1>(frameIndex);
+  auto& duration = std::get<2>(frameIndex);
+
+  auto& streamMetadata = containerMetadata_.allStreamMetadata[streamIndex];
+
+  // Get the last index for key_frames and duration
```
NicolasHug: Nit: the comment mentions `key_frames` but we're actually using `all_frames`.

Also, we should `TORCH_CHECK` here that all 3 tensors are of the same length!

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,45 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+
+  streamMetadata.endStreamPtsSecondsFromContent =
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
+    frameInfo.isKeyFrame =
+        (i < key_frames.size(0) && key_frames[i].item<int64_t>() == 1);
```
NicolasHug: The `i < key_frames.size(0)` check shouldn't be needed if we do the `TORCH_CHECK` validation mentioned above.
Also, I think `key_frames` should be a `bool` tensor, and I'd suggest to rename it to `is_key_frame`

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -53,6 +53,13 @@ class SingleStreamDecoder {
   // the allFrames and keyFrames vectors.
   void scanFileAndUpdateMetadataAndIndex();
 
+  // Reads the user provided frame index and updates each StreamInfo's index,
+  // i.e. the allFrames and keyFrames vectors, and the
+  // endStreamPtsSecondsFromContent
+  void readFrameIndexUpdateMetadataAndIndex(
+      int streamIndex,
+      std::tuple<at::Tensor, at::Tensor, at::Tensor> frameIndex);
```
NicolasHug: Here or somewhere else, we should document what are the expected lenfth, dtype, and associated semantic of each tensor.

We could also consider making `FrameIndex` a struct, instead of relying on a `tuple`.

src/torchcodec/_core/custom_ops.cpp
NicolasHug: This was intentional :)

We wanted to make the library selection something very, very private, which is why we are only exposing it through `_core._add_video_stream()`, with two (!) leading underscores.

 IIRC, we still had to expose it to Python for testing reasons, but we shouldn't add `color_conversion_library` to the "still private but slightly more public" `_core.add_video_stream()` API.

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -29,7 +29,7 @@ class SingleStreamDecoder {
   // CONSTRUCTION API
   // --------------------------------------------------------------------------
 
-  enum class SeekMode { exact, approximate };
+  enum class SeekMode { exact, approximate, custom_frame_mappings };
```
NicolasHug, line 32: I don't know if we'll want to *publicly* expose a new `seek_mode` in Python, but I think for the C++ side this a reasonable approach. @scotts  curious  if you have any opinion on this?

test/test_ops.py
```diff
@@ -448,6 +448,61 @@ def test_frame_pts_equality(self):
             )
             assert pts_is_equal
 
+    def test_seek_mode_custom_frame_mappings_fails(self):
+        decoder = create_from_file(str(NASA_VIDEO.path), "custom_frame_mappings")
```
NicolasHug: Nit: here and in other places, use keyword arguments for extra clarity. We typically don't use kw for the path though (1st param), as it's very obvious what it is.

```suggestion
        decoder = create_from_file(str(NASA_VIDEO.path), seek_mode"custom_frame_mappings")
```

test/utils.py
```diff
@@ -223,6 +227,56 @@ def get_frame_info(
...
+    # This property is used to get the frame mappings for the custom_frame_mappings seek mode.
+    @property
+    def custom_frame_mappings(
+        self, stream_index: Optional[int] = None
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        if stream_index is None:
+            stream_index = self.default_stream_index
+        if self.custom_frame_mappings_data is None:
+            self.get_custom_frame_mappings(stream_index)
+        return self.custom_frame_mappings_data
```
NicolasHug: I could be wrong but I think there's a potential bug here: if `custom_frame_mappings()` is called twice e.g.

```
custom_frame_mappings(stream_index=0)
custom_frame_mappings(stream_index=1)
```

then the second call will still return the mapping of `stream_index=0`, because that's the one that was cached?

Perhaps the easiest way to fix this is would be to make `custom_frame_mappings_data` a dict, where the keys are the stream indices, and the values are the individual mappings. This is also how the other fields (`frames`, `stream_infos`, etc.) are implemented.

-----

In any case, good job on the implementation overall, I like that we compute the mappings on the fly and cache them. This avoids us having to check those long files into the repo.

test/utils.py
```diff
@@ -223,6 +227,56 @@ def get_frame_info(
...
+            ).stdout
+        )
+        custom_frame_mappings_data = ([], [], [])
+        frames = show_frames_result["frames"]
+        for frame in frames:
+            custom_frame_mappings_data[0].append(float(frame["pts"]))
+            custom_frame_mappings_data[1].append(frame["key_frame"])
+            custom_frame_mappings_data[2].append(float(frame["duration"]))
+
+        (pts_list, key_frame_list, duration_list) = custom_frame_mappings_data
```
NicolasHug: Instead of defining `custom_frame_mappings_data`, let's just define the 3 individual lists? i.e.

```py
pts_list = []
key_frame_list = []
duration_list = []
for frame in frames: ...
```

or even just create all 3 by list comprehension. It's 3X slower but I don't think we care.

test/utils.py
Dan-Flores: Good catch, this would cause a bug if repeatedly called. I'll update `custom_frame_mappings_data` to be a dict, and remove the `@property` decorator so that providing a `stream_index` is possible.

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: I opened #767 to follow up on this change seperately.

test/utils.py
```diff
@@ -121,6 +122,9 @@ class TestContainerFile:
     default_stream_index: int
     stream_infos: Dict[int, Union[TestVideoStreamInfo, TestAudioStreamInfo]]
     frames: Dict[int, Dict[int, TestFrameInfo]]
+    custom_frame_mappings_data: Dict[
```
NicolasHug: Nit, since we should only access the mappins through the `get_custom_frame_mappings()` function, I'd suggest to "mark" that field as private:

```suggestion
    _custom_frame_mappings_data: Dict[
```

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -53,6 +53,13 @@ class SingleStreamDecoder {
   // the allFrames and keyFrames vectors.
   void scanFileAndUpdateMetadataAndIndex();
 
+  // Reads the user provided frame index and updates each StreamInfo's index,
+  // i.e. the allFrames and keyFrames vectors, and
+  // endStreamPtsSecondsFromContent
+  void readCustomFrameMappingsUpdateMetadataAndIndex(
+      int streamIndex,
+      std::tuple<at::Tensor, at::Tensor, at::Tensor> customFrameMappings);
+
```
NicolasHug, line 58: Just bumping https://github.com/pytorch/torchcodec/pull/764#discussion_r2197160670 again, which may have been missed:

- we should document what are the expected length, dtype, and associated semantic of each tensor.
- I'd also recommend relying on a new `FrameMappings` struct instead of a 3-tuple.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  streamMetadata.beginStreamPtsSecondsFromContent =
+      *streamMetadata.beginStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.endStreamPtsSecondsFromContent =
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
```
NicolasHug: I'm not very familiar with that kind of initialization, but it's not immediately obvious that it's setting the `frameInfo.pts` field. Let's set the field explicitly instead, for clarity.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+    frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
+    frameInfo.nextPts = (i + 1 < all_frames.size(0))
+        ? all_frames[i + 1].item<int64_t>()
+        : INT64_MAX;
+    streamInfos_[streamIndex].allFrames.push_back(frameInfo);
```
NicolasHug, line 358: We should also make sure to update the `streamInfos_[streamIndex].keyFrames` index, if isKeyFrame is true!

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {all_frames[i].item<int64_t>()};
+    frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
+    frameInfo.nextPts = (i + 1 < all_frames.size(0))
+        ? all_frames[i + 1].item<int64_t>()
+        : INT64_MAX;
```
NicolasHug: Let's also set the `frameInfo`'s `frameIndex` field. I think it should be `i`?
EDIT: it should actually be set after we are sure the sequence is sorted, see other comment below.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+    streamInfos_[streamIndex].allFrames.push_back(frameInfo);
+  }
+}
+
```
NicolasHug, line 366: I realize reading this that there is an additional design decision we'll have to make: whether we expect the index to be already sorted, or not.

In the existing `scanFileAndUpdateMetadataAndIndex()` function we are reading packets in order, and yet we are sorting them afterwards:

https://github.com/pytorch/torchcodec/blob/86e952f26ca7d27da2e8362a91019a9b3980ed96/src/torchcodec/_core/SingleStreamDecoder.cpp#L285-L298


I suspect that frame mappings coming from `ffprobe` won't be ordered in general. I think we have the following options:

- expect the input mapping to be sorted - that may not be a great UX
- sort the mapping in Python - this is kinda what this PR is doing, by sorting the mappings in the tests - but we should remove that and have that logic within the code, not the tests
- sort the mappings in C++

I think the simpler, safer choice is to sort in C++ and rely on the same sorting logic that we have in `scanFileAndUpdateMetadataAndIndex()`. Curious what your thoughts are @Dan-Flores @scotts ?

test/test_metadata.py
```diff
@@ -31,23 +31,26 @@ def _get_container_metadata(path, seek_mode):
...
-        functools.partial(_get_container_metadata, seek_mode="approximate"),
-        functools.partial(_get_container_metadata, seek_mode="exact"),
-    ),
-)
-def test_get_metadata(metadata_getter):
-    with_scan = (
-        metadata_getter.keywords["seek_mode"] == "exact"
-        if isinstance(metadata_getter, functools.partial)
-        else False
+@pytest.mark.parametrize("seek_mode", ["approximate", "exact", "custom_frame_mappings"])
```
NicolasHug: I think you may have removed `get_container_metadata_from_header` from the previous parametrization

src/torchcodec/_core/SingleStreamDecoder.h
scotts: To elaborate on defining a new `FrameMappings` struct: I think of the code in `custom_ops.cpp` as the _bridge_ layer between the C++ logic and the Python logic. So that's the layer where we would turn a tuple of tensors into a struct. That way, in the C++ logic, we're (as much as possible) operating on proper types with meaningful field names.

As a point of comparison, we do something similar with batched output. `SingleStreamDecoder` returns a `FrameBatchOutput` struct, and the code in `custom_ops.cpp` turns that into a tuple of tensors.

src/torchcodec/_core/SingleStreamDecoder.h
scotts: Agreed. Right now, the `SingleStreamDecoder`'s `seekMode_` is what we use to determine if we should do a file scan or not: https://github.com/pytorch/torchcodec/blob/d444567ad281dc8f6b184621beb2e032c0f7e549/src/torchcodec/_core/SingleStreamDecoder.cpp#L183-L185

Since the point of this feature is to avoid a file scan while still maintaining seek accuracy, I think it makes sense for it to be a new seek mode on the C++ side. We can figure out what to do with the public API later.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  TORCH_CHECK(
+      all_frames.size(0) == is_key_frame.size(0) &&
+          is_key_frame.size(0) == duration.size(0),
+      "all_frames, is_key_frame, and duration from custom_frame_mappings were not same size.");
+
+  auto& streamMetadata = containerMetadata_.allStreamMetadata[streamIndex];
+
+  streamMetadata.beginStreamPtsFromContent = all_frames[0].item<int64_t>();
+  streamMetadata.endStreamPtsFromContent =
+      all_frames[-1].item<int64_t>() + duration[-1].item<int64_t>();
```
scotts, line 344: TIL that tensors on the C++ side also support negative indices! I'm so used to that being not allowed in C++ arrays and standard containers that I initially thought this was undefined behavior!

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -319,6 +319,42 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+
+  auto avStream = formatContext_->streams[streamIndex];
+  streamMetadata.beginStreamPtsSecondsFromContent =
+      *streamMetadata.beginStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.endStreamPtsSecondsFromContent =
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
```
scotts, line 352: We should probably use the `ptsToSeconds()` function here, but it looks like we're also not using in the scanning function. And we can probably better define `ptsToSeconds()` to use `av_q2d()`. Let's address that elsewhere; created #770 to track.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: Agreed on sorting on the C++ side; we can do it quickly on the C++ side, and it's much nicer to users. We should extract the existing logic out to a utility function called only in this cpp file, and call it in both places.

I _think_ we sort in the scan function because I _think_ it's possible for the actual packets to be _not_ in PTS order.

Dan-Flores: Thanks for the feedback! Sorting on the C++ side makes sense, I've updated the PR to reuse the sorting code from the scan function.

src/torchcodec/_core/SingleStreamDecoder.h
Dan-Flores: Thanks for the reference. I added `FrameMappings` struct to `Frame.h` and utilized it on the C++ side, please let me know if the design pattern needs any adjustments.

test/test_metadata.py
Dan-Flores: Good catch! I've added it back, and made some adjustments to the checked value of `pixel_aspect_ratio` and to `_get_container_metadata` to work around the need to call `add_video_stream` for this new seek mode.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -283,40 +318,49 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {.pts = all_frames[i].item<int64_t>()};
+    frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
+    frameInfo.nextPts = (i + 1 < all_frames.size(0))
+        ? all_frames[i + 1].item<int64_t>()
+        : INT64_MAX;
```
NicolasHug: I think this is set in `sortAllFrames()` now?

test/utils.py
```diff
@@ -223,6 +227,50 @@ def get_frame_info(
...
+                check=True,
+                capture_output=True,
+                text=True,
+            ).stdout
+        )
+        pts_list = torch.tensor([float(frame["pts"]) for frame in result["frames"]])
+        is_key_frame_list = torch.tensor(
+            [frame["key_frame"] for frame in result["frames"]]
+        )
+        duration_list = torch.tensor(
```
NicolasHug: Nit: these aren't list anymore, they're tensors now. For consistency, we could just give them the same names that they have in C++, i.e. `all_frames`, `is_key_frame` and `duration`.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -283,40 +318,49 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+  streamMetadata.beginStreamPtsSecondsFromContent =
+      *streamMetadata.beginStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.endStreamPtsSecondsFromContent =
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    // FrameInfo struct utilizes PTS
+    FrameInfo frameInfo = {.pts = all_frames[i].item<int64_t>()};
```
NicolasHug: This  seems to be causing compilation errors

src/torchcodec/_core/Frame.h
```diff
@@ -45,6 +45,20 @@ struct AudioFramesOutput {
...
+  // 1D tensor of int64, each value is the PTS of a frame in timebase units.
+  torch::Tensor all_frames;
+  // 1D tensor of bool, each value indicates if the corresponding frame in
+  // all_frames is a key frame.
+  torch::Tensor is_key_frame;
+  // 1D tensor of int64, each value is the duration of the corresponding frame
+  // in all_frames in timebase units.
+  torch::Tensor duration;
+};
+
```
NicolasHug: Thanks for creating the separate and for the comments!

If we can, let's put this into `src/torchcodec/_core/SingleStreamDecoder.h` instead. The existing structs in this file are mostly related to input/output structs (i.e. video frames, or audio samples).

test/utils.py
```diff
@@ -223,6 +227,53 @@ def get_frame_info(
...
         return self.frames[stream_index][idx]
 
+    # This function is used to get the frame mappings for the custom_frame_mappings seek mode.
+    def get_custom_frame_mappings(
+        self, stream_index: Optional[int] = None
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+        # Ensure all tests using this function are skipped if the FFmpeg version is 4 or 5
+        # FFprobe on FFmpeg 4 and 5 does not return complete metadata
+        if get_ffmpeg_major_version() == 4 or get_ffmpeg_major_version() == 5:
+            pytest.skip("FFprobe on FFmpeg 4 and 5 does not return complete metadata")
```
NicolasHug: That's a smart way to do it, but the main problem with having the `pytest.skip()` call here is that it's not immediately obvious whether a test will be skipped or not just by reading it. We try to add those `skip` statements either as a `pytest.mark.skipif()` decorator where the test is defined, or within the test body. This way, we know what to expect when looking at the test, and that can help debugging too.

In this case, I think we should just add the decorator directly on the affected tests, e.g.:

```
@pytest.mark.skipif(get_ffmpeg_major_version() in (4, 5), reason="ffprobe isn't accurate on ffmpeg 4 and 5")
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -198,6 +198,41 @@ int SingleStreamDecoder::getBestStreamIndex(AVMediaType mediaType) {
 // VIDEO METADATA QUERY API
 // --------------------------------------------------------------------------
 
+void SingleStreamDecoder::sortAllFrames() {
```
NicolasHug, line 201: Let's add a small comment to explain what this does

```suggestion
void SingleStreamDecoder::sortAllFrames() {
  // Sort the allFrames and keyFrames vecs in each stream, and also sets
  // additional fields of the FrameInfo entries like nextPts and frameIndex
  // This is called at the end of a scan, or when setting a user-defined frame
  // mapping.
 
```

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: Just bumping this again as it may have been missed

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -283,40 +318,46 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
+      *streamMetadata.beginStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.endStreamPtsSecondsFromContent =
+      *streamMetadata.endStreamPtsFromContent * av_q2d(avStream->time_base);
+
+  streamMetadata.numFramesFromContent = all_frames.size(0);
+  for (int64_t i = 0; i < all_frames.size(0); ++i) {
+    FrameInfo frameInfo;
+    frameInfo.pts = all_frames[i].item<int64_t>();
+    frameInfo.isKeyFrame = (is_key_frame[i].item<bool>() == true);
```
NicolasHug: This should work?

```suggestion
    frameInfo.isKeyFrame = is_key_frame[i].item<bool>();
```

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -53,20 +53,45 @@ class SingleStreamDecoder {
...
+    // in all_frames in timebase units.
+    torch::Tensor duration;
+  };
+
+  // Reads the user provided frame index and updates each StreamInfo's index,
+  // i.e. the allFrames and keyFrames vectors, and
+  // endStreamPtsSecondsFromContent
+  void readCustomFrameMappingsUpdateMetadataAndIndex(
+      int streamIndex,
+      FrameMappings customFrameMappings);
```
NicolasHug: Let's make this private, probably in the `INITIALIZERS` section

test/test_metadata.py
```diff
@@ -37,18 +49,24 @@ def _get_container_metadata(path, seek_mode):
         get_container_metadata_from_header,
         functools.partial(_get_container_metadata, seek_mode="approximate"),
         functools.partial(_get_container_metadata, seek_mode="exact"),
+        functools.partial(_get_container_metadata, seek_mode="custom_frame_mappings"),
```
NicolasHug, line 52: Nicely done, this test is quite tricky to handle but you made it work with what I think was the simplest change.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: @NicolasHug, @Dan-Flores: Unfortunately, that style of struct initialization is not available in C++17, which is what we're using. What's confusing is that it's been in _C_ since [C99](https://en.cppreference.com/w/c/language/struct_initialization.html), but it's not in C++ until [C++23](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html).

test/utils.py
Dan-Flores: Thank you for the suggestion, this approach is very readable. I implemented it in `test_seek_mode_custom_frame_mappings`, but for `test_get_metadata` I used pytest.skip to ensure we only skipped when necessary.

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: I will address #770 after completing this change, so we can update + test all locations in one PR.

PR #777, opened 2025-07-13
Avoid silently wrong results on 10bit CUDA

NicolasHug:
Addresses https://github.com/pytorch/torchcodec/issues/736 and https://github.com/pytorch/torchcodec/issues/598.

Some 10bit videos are properly decoded by NVDEC, but are passed as-is to the NPP color conversion functions, which expect 8bits videos. This results in garbage output. This PR adds a check to prevent such incorrect conversion. It does *not* add support for 10 bit videos on CUDA. It just avoids silently wrong results.


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -196,10 +196,48 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
     UniqueAVFrame& avFrame,
     FrameOutput& frameOutput,
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
+  // We check that avFrame->format == AV_PIX_FMT_CUDA. This only ensures the
+  // AVFrame is on GPU memory. It can be on CPU memory if the video isn't
+  // supported by NVDEC for whatever reason: NVDEC falls back to CPU decoding in
+  // this case, and our check fails.
+  // TODO: we could send the frame back into the CPU path, and rely on
```
scotts, line 203: Let's create an issue to track.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -196,10 +196,48 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
+  // TODO: we should investigate how to can perform color conversion for
+  // non-8bit videos. This is supported on CPU.
+  TORCH_CHECK(
+      avFrame->hw_frames_ctx != nullptr,
+      "The AVFrame does not have a hw_frames_ctx. "
+      "That's unexpected, please report this to the TorchCodec repo.");
+
+  AVPixelFormat actualFormat =
+      reinterpret_cast<AVHWFramesContext*>(avFrame->hw_frames_ctx->data)
+          ->sw_format;
```
scotts: I got tripped up reading this because I'm not used to seeing a `reinterpret_cast` inside of an expression. I think it's easier to read as:
```cpp
hwFramesCtx = reinterpret_cast<AVHWFramesContext*>(avFrame->hw_frames_ctx->data);
AVPixelFormat actualFormat = hwFramesCtx->sw_format;
```
Also: TIL that `AVBufferRef*` is effectively `Object*` for FFmpeg, and they use it to store lots of different generic stuff.

test/test_decoders.py
```diff
@@ -1138,6 +1141,31 @@ def test_pts_to_dts_fallback(self, seek_mode):
...
+            match = "The AVFrame is p010le, but we expected AV_PIX_FMT_NV12."
+        else:
+            match = "Expected format to be AV_PIX_FMT_CUDA, got yuv420p10le."
+        with pytest.raises(RuntimeError, match=match):
+            decoder.get_frame_at(0)
+
+    @pytest.mark.parametrize("asset", (H264_10BITS, H265_10BITS))
+    def test_10bit_videos_cpu(self, asset):
+        # This just validates that we can decode 10-bit videos on CPU.
+        # TODO validate against the ref that the decoded frames are correct
```
Dan-Flores, line 1164: What is required to test that the decoded frames are correct? It seems we would need to check in additional png files from each 10-bit video to the `test/resources/`, are there other changes needed?

NicolasHug: Correct, we'd have to check in a few reference frames to compare against. We usually store those as raw `.pt` files.

src/torchcodec/_core/CudaDeviceInterface.cpp
NicolasHug: Following up in https://github.com/pytorch/torchcodec/issues/776

PR #778, opened 2025-07-13
Add tutorial for parallel decoding

NicolasHug:


examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+    "-stream_loop", "49",  # repeat video 50 times
+    "-i", str(short_video_path),
+    "-c", "copy",
+    str(long_video_path)
+]
+subprocess.run(ffmpeg_command, check=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+decoder = VideoDecoder(long_video_path, seek_mode="approximate")
+metadata = decoder.metadata
+
```
scotts, line 122: Can we put most of the logic above that appears at module scope into a function, and then call that? It's a big chunk of code, and it's not obvious what from it lives on. I think having the stuff above in functions helps readers understand what's going on. So I think something like this:
```python
def generate_long_video():
    # most of the module scope logic

short_duration_path, long_duration_path = generate_long_video()
```

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+print(f"Total frames: {metadata.num_frames}")
+
+
+# %%
+# .. _start_parallel_decoding:
+#
+# Frame sampling strategy
+# -----------------------
+#
+# For this tutorial, we'll sample frames at a target rate of 2 FPS from our long
```
scotts: This might be a me-thing, but I was tripped up by the "2 FPS from our long video" part. That is, I initially thought that meant we were targeting the decoding itself to happen at the observable speed of 2 FPS. This might be because I know our benchmarks report FPS as their metric. Rather, we mean that we want to sample 2 FPS in the reference frame of the video's time. Phrasing that I think would have helped me understand quicker:

> For this tutorial, we'll sample a frame every 2 seconds from our long video.

Also, just realized that in the existing text, "inference" is mispelled.

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+print(f"Speedup vs sequential: {speedup:.2f}x with {NUM_CPUS} FFmpeg threads.")
+
+
+# %%
+# Method 3: multiprocessing
+# -------------------------
+#
+# Process-based parallelism distributes work across multiple Python processes.
+
+def decode_with_multiprocessing(indices: List[int], num_processes: int, video_path=long_video_path):
```
scotts: Let's put each of these parameters on a separate line - on my system, the rendering for this line wraps.

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+    """Decode frames using multiple processes with joblib."""
+    chunks = split_indices(indices, num_chunks=num_processes)
+
+    results = Parallel(n_jobs=num_processes, backend="loky", verbose=0)(
```
scotts, line 209: What's the "loky" backend?

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+# %%
+# Method 4: Joblib multithreading
+# -------------------------------
+#
+# Thread-based parallelism uses multiple threads within a single process.
+# TorchCodec releases the GIL, so this can be very effective.
+
+def decode_with_multithreading(indices: List[int], num_threads: int, video_path=long_video_path):
```
scotts: Same as above - let's use line breaks for parameters.

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+    """Decode frames using multiple threads with joblib."""
+    chunks = split_indices(indices, num_chunks=num_threads)
+
+    results = Parallel(n_jobs=num_threads, prefer="threads", verbose=0)(
```
scotts, line 237: Does "prefer" mean that in some situations it might end up using processes? Does it default to processes, since we didn't say this above?

scotts (approved):
Excellent tutorial! I think this will be extremely helpful for our users!

examples/decoding/parallel_decoding.py
```diff
@@ -0,0 +1,245 @@
...
+    decoder = VideoDecoder(video_path, num_ffmpeg_threads=num_threads, seek_mode="approximate")
+    return decoder.get_frames_at(indices)
+
+
+NUM_CPUS = cpu_count()
+
+times, result_ffmpeg = bench(decode_with_ffmpeg_parallelism, all_indices, num_threads=NUM_CPUS)
+ffmpeg_time = report_stats(times, unit="s")
+speedup = sequential_time / ffmpeg_time
+print(f"Speedup vs sequential: {speedup:.2f}x with {NUM_CPUS} FFmpeg threads.")
```
Dan-Flores: nit: Perhaps for additional clarity, we could write out the comparison instead of using "vs":

```
print(f"Speedup compared to sequential: {speedup:.2f}x using {NUM_CPUS} FFmpeg threads.")
```

examples/decoding/parallel_decoding.py
NicolasHug: Fair, fps is overloaded and that confuses me too, sometimes. I'll add comments to clarify the context

examples/decoding/parallel_decoding.py
NicolasHug: It's a multi-processing backend for joblib: https://github.com/joblib/loky

I added a comment. In general, I don't want to go too much over the details of `joblib` in this tutorial, because as I mentioned at the top, the concepts covered here are joblib-agnostic. The reader should just trust that the `decode_with*` functions are doing the right thing

examples/decoding/parallel_decoding.py
NicolasHug: It basically mean "use threads, unless this gets overridden by something with higher priority, like a context manager". There is more about this in the docstring for "backend" https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html

PR #784, opened 2025-07-23
Consistently use ptsToSeconds

Dan-Flores:
As discussed in #764, this PR updates various methods in SingleStreamDecoder.cpp to use `ptsToSeconds` when performing calculations with the `time_base`. 

Additionally, `ptsToSeconds` is updated to use `av_q2d` to ensure the  `AVRational`s are correctly handled.

Tests are updated as well to reflect the added precision:
* In VideoDecoderTest.cpp, comparisons of `ptsSeconds` are updated from `EXPECT_EQ` to `EXPECT_DOUBLE_EQ`.
* In test_ops.py, `pytest.approx` is used to compare `pts_seconds`.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -751,7 +751,7 @@ class TestAudioDecoderOps:
                 frame, asset.get_frame_data_by_index(frame_index)
             )
             frame_info = asset.get_frame_info(frame_index)
-            assert pts_seconds == frame_info.pts_seconds
+            assert pts_seconds == pytest.approx(frame_info.pts_seconds)
             assert duration_seconds == frame_info.duration_seconds
             frame_index += 1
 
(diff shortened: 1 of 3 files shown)


test/test_ops.py
```diff
@@ -955,7 +955,7 @@ def test_pts(self, asset):
                 frames, asset.get_frame_data_by_index(frame_index)
             )
 
-            assert pts_seconds == start_seconds
+            assert pts_seconds == pytest.approx(start_seconds)
```
NicolasHug, line 958: I'm surprised we need to use `approx` so just to be on the safe side I removed it and we're getting:

```
FAILED test/test_ops.py::TestAudioDecoderOps::test_next[asset0] - assert 0.5760000000000001 == 0.576
FAILED test/test_ops.py::TestAudioDecoderOps::test_next[asset1] - assert 0.9359999999999999 == 0.936
FAILED test/test_ops.py::TestAudioDecoderOps::test_pts[asset0] - assert 0.5760000000000001 == 0.576
FAILED test/test_ops.py::TestAudioDecoderOps::test_pts[asset1] - assert 0.9359999999999999 == 0.936
```

Which is probably OK. But I still find it surprising considering the implementation of  [av_q2d](https://ffmpeg.org/doxygen/7.0/rational_8h_source.html#l00104) is just

```
static inline double av_q2d(AVRational a){
     return a.num / (double) a.den;
 }

```

PR #790, opened 2025-07-27
Support 10 bits videos on CUDA

NicolasHug:
This is part 1. We still need to handle those that are decoded on the CPU instead.


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -196,54 +196,171 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
+            avFrame->data[1]) // UV plane (16-bit interleaved)
+    };
+
+    // Choose color matrix based on colorspace
+    const Npp32f (*aTwist)[4];
+    
+    // TODO use even more accurage values from
+    // https://ffmpeg.org/doxygen/trunk/yuv2rgb_8c_source.html#l00047
+    // Need to devide by 65536 to get the floats
+    // BT.709 matrix (HDTV)
```
NicolasHug: Or just adapt matrix from https://github.com/pytorch/torchcodec/pull/791/files

PR #794, opened 2025-07-30
Improve `pytest` default options

NicolasHug:
We have a "slow" test that takes >1 minute to complete. I got a bit tired of having to always manually specify `-k "not slow"` in my test options, so I made it the default: the tests that are marked as 'slow' are now skipped by default. We always run them in the CI though.

Similarly, I always find myself having to add `-v`, so it's now the default. I hope this isn't controversial.

### DIFF
--- a/.github/workflows/macos_wheel.yaml
+++ b/.github/workflows/macos_wheel.yaml
@@ -122,4 +122,4 @@ jobs:
 
       - name: Run Python tests
         run: |
-          pytest test --ignore-addopts -v
+          pytest --override-ini="addopts=-v" test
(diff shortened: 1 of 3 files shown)


.github/workflows/macos_wheel.yaml
```diff
@@ -122,4 +122,4 @@ jobs:
 
       - name: Run Python tests
         run: |
-          pytest test -vvv
+          pytest --override-ini="addopts=-v" test
```
Dan-Flores, line 125: Should the CI continue to use maximum verbosity?

NicolasHug: We've been using `-v` only for the CUDA tests already. Personally I find `-vvv` too versbose and thus difficult to debug, because the traceback is really long. Thank you for the review - I'll merge as-is, but I'm happy to put back `-vvv` if you prefer.

PR #798, opened 2025-07-31
[WIP] native nvcuvid backend

NicolasHug:


src/torchcodec/_core/CustomNvdecDeviceInterface.cpp
```diff
@@ -0,0 +1,553 @@
...
+           cached.OutputFormat == requested.OutputFormat &&
+           cached.bitDepthMinus8 == requested.bitDepthMinus8;
+  }
+  
+  static std::mutex mutex_;
+  static std::vector<CachedDecoder> cached_;
+};
+
+std::mutex DecoderCache::mutex_;
+std::vector<CachedDecoder> DecoderCache::cached_;
```
scotts, line 158: I think it's more clear to make `mutex_` and `cached_` normal member variables of `DecoderCache`, and then declare an instance of `DecoderCache` that is itself static.

src/torchcodec/_core/CustomNvdecDeviceInterface.cpp
```diff
@@ -0,0 +1,553 @@
...
+
+// Global cleanup helper
+struct CacheCleanup {
+  ~CacheCleanup() {
+    DecoderCache::clearCache();
+  }
+};
+static CacheCleanup g_cache_cleanup;
+
```
scotts, line 167: Doing the above would, I think, enable you to just make this cleanup logic the destructor for `DecoderCache`.

PR #799, opened 2025-08-01
Add custom_frame_mappings to VideoDecoder init

Dan-Flores:
This PR enables `custom_frame_mappings` to be used in the Python VideoDecoder.

* Implements `read_custom_frame_mappings()`. This parses a JSON str or JSON file to extract `all_frames`, `is_key_frame`, `duration`.
* Updates the `seek_mode` if `custom_frame_mappings` is passed in to avoid the exact mode scan during initialization.
* Tests are added to check that the frames are being decoded correctly.

## Benchmarking
I wrote a short benchmarking script to test the initialization times of `custom_frame_mappings` versus `exact`.

`custom_frame_mapping` mode was quicker than `exact` mode if the `ffprobe` command used to generate the frame mapping json used the `-show_entries` option to reduce the JSON size.

With this optimization the performance improvement increased with longer videos.
Without this optimization, using the full `ffprobe` output, `exact` mode was faster.

The results on `nasa_13013.mp4`:
```
exact:
med = 3.38ms +- 2.54
custom_frame_mappings:
med = 3.02ms +- 0.86
```
The results on a generated video, `mandelbrot_1920x1080_120s.mp4`:
```
exact:
med = 29.51ms +- 7.56
custom_frame_mappings:
med = 16.69ms +- 9.38
```

The benchmarking code:
<details>    

    import subprocess
    import torch

    from torchcodec.decoders._video_decoder import VideoDecoder

        for _ in range(warmup):
            f(*args, **kwargs)

        times = []
        for _ in range(num_exp):
            start = perf_counter_ns()
            f(*args, **kwargs)
            end = perf_counter_ns()
            times.append(end - start)

        mul = {
            "ns": 1,
            "µs": 1e-3,
            "ms": 1e-6,
            "s": 1e-9,
        }[unit]
        times = times * mul
        std = times.std().item()
        med = times.median().item()
        return med

    def main() -> None:
        """Benchmarks the init of VideoDecoder with different seek_modes"""
        resources_dir = "/Users/danielflores3/torchcodec/test/resources/"

        # video=resources_dir+"nasa_13013.mp4"
        video=resources_dir+"mandelbrot_1920x1080_120s.mp4"

        mappings_json = subprocess.run(
                [
                    "ffprobe",
                    "-i",
                    f"{video}",
                    "-select_streams",
                    "0",
                    "-show_frames",
                    "-show_entries", "frame=pts,key_frame,duration",
                    "-of",
                    "json",
                ],
                check=True,
                capture_output=True,
                text=True,
            ).stdout
        
        print("exact:")
        report_stats(bench(VideoDecoder, source=video, seek_mode="exact", stream_index=0))

        print("custom_frame_mappings:")
        report_stats(bench(VideoDecoder, source=video, custom_frame_mappings=mappings_json, stream_index=0))

    if __name__ == "__main__":
        main()

```
</details>

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -65,8 +65,9 @@ class VideoDecoder:
             :ref:`sphx_glr_generated_examples_decoding_approximate_mode.py`
         custom_frame_mappings (str, bytes, or file-like object, optional):
             Mapping of frames to their metadata, typically generated via ffprobe.
-            This enables accurate frame seeking without requiring a full video scan
-            as in `exact` mode. Expected JSON format:
+            This enables accurate frame seeking without requiring a full video scan.
+            Do not set seek_mode when custom_frame_mappings is provided.
+            Expected JSON format:
 
             .. code-block:: json
 
@@ -79,6 +80,8 @@ class VideoDecoder:
                         }
                     ]
                 }
+
+            Alternative field names "pkt_pts" and "pkt_duration" are also supported.
 
     Attributes:
         metadata (VideoStreamMetadata): Metadata of the video stream.

### CODE REVIEW

**Comment by NicolasHug** [comment]
Thanks for the benchmarks! That's quite interesting. It'd be interesting to use the `core` API instead of the `VideoDecoder` in the benchmarks, so that we can measure the time for `add_video_stream()` specifically when passing the frame mapping as tensors.

If that's faster than `add_video_stream(seek_mode="exact")`, as we would expect, then it may be that what's slowing down the frame-mappings is the `io` or the `json` parsing parts?

**Comment by Dan-Flores** [comment]
>... then it may be that what's slowing down the frame-mappings is the `io` or the `json` parsing parts?

I believe this is it - I updated the PR description with my updated benchmarking script and results. By reducing the JSON size to contain only the necessary information, the performance of `custom_frame_mapping` improves significantly.

Edit: Here's my alternative code for benchmarking the `core` API by passing the frame mapping as tensors. It shows a similar performance improvement to reducing the JSON, so this could be another viable design approach. 

<details>

    import subprocess
    import torch

    from torchcodec import _core
    from torchcodec.decoders._video_decoder import VideoDecoder, read_custom_frame_mappings


        for _ in range(warmup):
            f(*args, **kwargs)

        times = []
        for _ in range(num_exp):
            start = perf_counter_ns()
            f(*args, **kwargs)
            end = perf_counter_ns()
            times.append(end - start)

        mul = {
            "ns": 1,
            "µs": 1e-3,
            "ms": 1e-6,
            "s": 1e-9,
        }[unit]
        times = times * mul
        std = times.std().item()
        med = times.median().item()
        return med

    def main() -> None:
        """Benchmarks the init of VideoDecoder with different seek_modes"""
        resources_dir = "/Users/danielflores3/torchcodec/test/resources/"

        # video=resources_dir+"nasa_13013.mp4"
        video=resources_dir+"mandelbrot_1920x1080_120s.mp4"

        mappings_json = subprocess.run(
                [
                    "ffprobe",
                    "-i",
                    f"{video}",
                    "-select_streams",
                    "0",
                    "-show_frames",
                    "-show_entries", "frame=pts,key_frame,duration",
                    "-of",
                    "json",
                ],
                check=True,
                capture_output=True,
                text=True,
            ).stdout
        
        # Get tensors of each, Pass into core function directly
        custom_frame_mappings_data = read_custom_frame_mappings(mappings_json)
        args = [("exact", None), ("custom_frame_mappings", custom_frame_mappings_data)]
        for seek_mode, frame_mappings_data in args:
            # benchmark speed up in add_video_stream
            print(f"{seek_mode=}")
            report_stats(bench(init_add_stream, 
                video,
                seek_mode,
                frame_mappings_data,
            ))

    def init_add_stream(video, seek_mode, frame_mappings_data):
        decoder = _core.create_from_file(str(video), seek_mode)
        _core.add_video_stream( 
                decoder,
                stream_index=0,
                dimension_order="NCHW",
                num_threads=1,
                device="cpu",
                custom_frame_mappings=frame_mappings_data,
            )

    if __name__ == "__main__":
        main()

</details>

File: test/test_ops.py [hunk 1]
```diff
@@ -505,10 +504,6 @@ def test_seek_mode_custom_frame_mappings_fails(self):
                 decoder, stream_index=0, custom_frame_mappings=different_lengths
             )
 
-    @pytest.mark.skipif(
-        get_ffmpeg_major_version() in (4, 5),
-        reason="ffprobe isn't accurate on ffmpeg 4 and 5",
-    )
```
Dan-Flores, line 511: ffprobe does include the fields `pts` and `duration` [in older versions](https://github.com/FFmpeg/FFmpeg/blob/d151d3aecbe2a478c44be7faca56caae1848ca87/doc/APIchanges#L1799), using the name `pkt_pts` and `pkt_duration`, so we can enable these tests by falling back to the `pts...` keys.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -88,6 +90,18 @@ def __init__(
...
+            if seek_mode not in ("exact", "custom_frame_mappings"):
+                raise ValueError(
+                    "While setting custom frame mappings, do not set `seek_mode`."
+                )
+            # Set seek mode to avoid exact mode scan
+            seek_mode = "custom_frame_mappings"
+        custom_frame_mappings_data = (
+            read_custom_frame_mappings(custom_frame_mappings)
+            if custom_frame_mappings is not None
+            else None
```
scotts: Is it ever valid for `seek_mode == "custom_frame_mappings"` and `custom_frame_mappings_data is None` to both be true? If no, then we should probably raise an error here.

Actually, shouldn't line 93 mean that `custom_frame_mappings` *cannot* be `None`?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +394,40 @@ def _get_and_validate_stream_metadata(
...
+) -> tuple[Tensor, Tensor, Tensor]:
+    try:
+        if hasattr(custom_frame_mappings, "read"):
+            input_data = json.load(custom_frame_mappings)
+        else:
+            input_data = json.loads(custom_frame_mappings)
+    except json.JSONDecodeError:
+        raise ValueError(
+            "Invalid custom frame mappings. "
+            "It should be a valid JSON string or a JSON file object."
```
scotts: In our type annotation, we only say we accept string-like objects, not file-like objects. In other parts of the code, I've indicated a file-like object with `io.RawIOBase` and `io.BufferedReader`: https://github.com/pytorch/torchcodec/blob/c3eea9f9f42d3d51f9c53ba19be6c11cc88c21dd/src/torchcodec/decoders/_video_decoder.py#L74-L77

For our doc string, however, we just say "file-like object" since those actual types are probably meaningless to most users: https://github.com/pytorch/torchcodec/blob/c3eea9f9f42d3d51f9c53ba19be6c11cc88c21dd/src/torchcodec/decoders/_video_decoder.py#L26

Which introduces a good point: since this a top-level API that we expect users to call, we need a docstring. :)

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -79,7 +80,8 @@ def __init__(
         dimension_order: Literal["NCHW", "NHWC"] = "NCHW",
         num_ffmpeg_threads: int = 1,
         device: Optional[Union[str, torch_device]] = "cpu",
-        seek_mode: Literal["exact", "approximate"] = "exact",
+        seek_mode: Literal["exact", "approximate", "custom_frame_mappings"] = "exact",
+        custom_frame_mappings: Optional[Union[bytes, bytearray, str]] = None,
```
scotts: These two values should also be explained in the docstring.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +394,40 @@ def _get_and_validate_stream_metadata(
...
+            input_data = json.load(custom_frame_mappings)
+        else:
+            input_data = json.loads(custom_frame_mappings)
+    except json.JSONDecodeError:
+        raise ValueError(
+            "Invalid custom frame mappings. "
+            "It should be a valid JSON string or a JSON file object."
+        )
+    # These keys are prefixed with "pkt_" in ffmpeg 4 and ffmpeg 5
+    pts_key = "pkt_pts" if "pts" not in input_data["frames"][0] else "pts"
```
scotts: Nit, but it would be helpful to users to do explicit error checks on our assumptions about the data:

1. Assert that `input_data` is non-empty.
2. Assert that `"frames"` is a valid key.
3. Assert that one of the valid pts and duration keys exist. (Rather than assuming it must be the other one.)

If any of the above isn't true, we'll still fail, but it will just be with `KeyError` exceptions that the cause of may not be obvious to users.

test/test_decoders.py
```diff
@@ -1279,6 +1280,109 @@ def test_10bit_videos_cpu(self, asset):
...
+            ValueError,
+            match="Invalid custom frame mappings. It should be a valid JSON string or a JSON file object.",
+        ):
+            with open(invalid_json_path, "r") as f:
+                VideoDecoder(
+                    NASA_VIDEO.path,
+                    stream_index=0,
+                    device=device,
+                    custom_frame_mappings=f.read(),
+                )
```
scotts: Similar to the comments I left in `read_custom_frame_mappings()`, I think we should also test for valid JSON that does not follow the schema that FFmpeg produces.

test/utils.py
```diff
@@ -267,31 +267,38 @@ def get_custom_frame_mappings(
...
+            check=True,
+            capture_output=True,
+            text=True,
+        ).stdout
+        return result
+
+    def create_custom_frame_mappings(self, stream_index: int) -> None:
+        result = json.loads(self.generate_custom_frame_mappings(stream_index))
+        # These keys are prefixed with "pkt_" in ffmpeg 4 and ffmpeg 5
+        pts_key = "pkt_pts" if "pts" not in result["frames"][0] else "pts"
```
scotts: I think it's simpler to just call `read_custom_frame_mappings()` here. I believe it's accomplishing the same thing, and as that function evolves, so to will this utils function need to evolve.

src/torchcodec/decoders/_video_decoder.py
Dan-Flores: Good point, that case would not be valid. 
Currently, creating a VideoDecoder with `seek_mode = custom_frame_mappings` and `custom_frame_mappings=None` will throw an error in `add_video_stream`, which is called on line 121. 

Do you think it would be helpful to error sooner?

src/torchcodec/decoders/_video_decoder.py
Dan-Flores: Thanks for the references, I'll add a docstring and use the term `file-like object` for uniformity.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -62,7 +63,24 @@ class VideoDecoder:
...
+                    "frames": [
+                        {
+                            "pts": 0,
+                            "duration": 1001,
+                            "key_frame": 1
+                            ...  # Other metadata fields can be included
+                        },
+                        ...
+                    ]
+                }
```
scotts, line 82: We should also say something about the relationship with `seek_mode`.

test/test_decoders.py
```diff
@@ -1279,6 +1280,120 @@ def test_10bit_videos_cpu(self, asset):
...
+                    if expected_match == "seek_mode"
+                    else "custom_frame_mappings"
+                ),
+            )
+
+    @pytest.mark.parametrize("device", all_supported_devices())
+    def test_custom_frame_mappings_init_fails_invalid_json(self, tmp_path, device):
+        invalid_json_path = tmp_path / "invalid_json"
+        with open(invalid_json_path, "w+") as f:
+            f.write("""'{"invalid": "json"'""")
```
scotts: Nit: this is not obviously invalid - it's invalid because it's missing the closing `}`, but I had to read really closely to catch that! Part of the reason I had difficulty is that there's so much quoting going on. I think it might be better to do something like:
```python
  f.write("garbage input")
```
That is, something that doesn't even look like JSON at a glance.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +416,58 @@ def _get_and_validate_stream_metadata(
         end_stream_seconds,
         num_frames,
     )
+
+
+def read_custom_frame_mappings(
```
NicolasHug: Nit

```suggestion
def _read_custom_frame_mappings(
```

This makes it more obvious that it is private. It is already private even without the leading underscore, because it's only exposed within `_video_decoder.py` which itself has the underscore. But explicitly having an underscore makes it even more obvious, and consistent with the other private functions defined in this file

src/torchcodec/decoders/_video_decoder.py
NicolasHug: We can also add that in order to accomodate for different FFmpeg versions we also allow `pkt_pts` and `pkt_duration` ?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -79,16 +97,34 @@ def __init__(
         dimension_order: Literal["NCHW", "NHWC"] = "NCHW",
         num_ffmpeg_threads: int = 1,
         device: Optional[Union[str, torch_device]] = "cpu",
-        seek_mode: Literal["exact", "approximate"] = "exact",
+        seek_mode: Literal["exact", "approximate", "custom_frame_mappings"] = "exact",
```
NicolasHug: IIRC we decided not to expose `custom_frame_mappings` at the public level, and just force users to rely on the default seek_mode (approximate) if they pass `custom_frames_mapping`? That's consistent with the existing (correct) check below:

```py
        if custom_frame_mappings is not None and seek_mode == "approximate":
            raise ValueError(
```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +416,58 @@ def _get_and_validate_stream_metadata(
...
+    first_frame = input_data["frames"][0]
+    pts_key = next((key for key in ("pts", "pkt_pts") if key in first_frame), None)
+    duration_key = next(
+        (key for key in ("duration", "pkt_duration") if key in first_frame), None
+    )
+    key_frame_present = "key_frame" in first_frame
+
+    if not pts_key or not duration_key or not key_frame_present:
+        raise ValueError(
+            "Invalid custom frame mappings. The 'pts', 'duration', and 'key_frame' keys are required in the frame metadata."
```
NicolasHug: we probably want to indicate that `pkt_pts` and `pkt_duration` are also OK?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +416,58 @@ def _get_and_validate_stream_metadata(
...
+        )
+
+    frame_data = [
+        (float(frame[pts_key]), frame["key_frame"], float(frame[duration_key]))
+        for frame in input_data["frames"]
+    ]
+    all_frames, is_key_frame, duration = map(Tensor, zip(*frame_data))
```
NicolasHug: I think this should be `torch.tensor`, not `Tensor` which is a type

Also in general, list comprehensions are preferred to `map` (but we can keep it as-is)

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -379,3 +416,58 @@ def _get_and_validate_stream_metadata(
...
+    assert (
+        len(all_frames) == len(is_key_frame) == len(duration)
+    ), "Mismatched lengths in frame index data"
```
NicolasHug: For user-facing assertions we prefer `if + ValueError` rather than `assert` which will lead to an `AssertionError`. The ValueError is more explicitly about bad user-input.

test/test_decoders.py
```diff
@@ -1279,6 +1280,120 @@ def test_10bit_videos_cpu(self, asset):
         decoder = VideoDecoder(asset.path)
         decoder.get_frame_at(10)
 
+    def setup_frame_mappings(tmp_path: str, file: bool, stream_index: int):
```
NicolasHug: Nit tmp_path isn't a str, it's Path, otherwise we wouldn't be able to apply `/` to it

test/test_decoders.py
```diff
@@ -1279,6 +1280,120 @@ def test_10bit_videos_cpu(self, asset):
...
+            open(custom_frame_mappings, "r")
+            if hasattr(custom_frame_mappings, "read")
+            else contextlib.nullcontext()
+        ) as custom_frame_mappings:
+            decoder = VideoDecoder(
+                NASA_VIDEO.path,
+                stream_index=stream_index,
+                device=device,
+                custom_frame_mappings=custom_frame_mappings,
+            )
```
NicolasHug, line 1320: Nit: everything below can probably be out of the context manager. It's preferable to end the CM's scope as soon as it's not needed.

src/torchcodec/decoders/_video_decoder.py
scotts: Yes, I believe @NicolasHug is right here. I completely forgot about that.

test/test_decoders.py
Dan-Flores: Good point - since type annotations are not used elsewhere in this file I will remove them from this function.

PR #800, opened 2025-08-01
[wip] Add npp_stream_ctx cache

Dan-Flores:


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -187,16 +191,44 @@ NppStreamContext createNppStreamContext(int deviceIndex) {
...
 
-  // TODO when implementing the cache logic, move these out. See other TODO
-  // below.
-  nppCtx.hStream = at::cuda::getCurrentCUDAStream(deviceIndex).stream();
-  err = cudaStreamGetFlags(nppCtx.hStream, &nppCtx.nStreamFlags);
-  TORCH_CHECK(
-      err == cudaSuccess,
-      "cudaStreamGetFlags failed: ",
-      cudaGetErrorString(err));
+  return nppCtxPtr;
```
NicolasHug, line 194: Oh, I think this is the problem: we're returning a pointer to `nppCtx`, which was allocated on the stack, not on the heap. So the pointer's value (and what it points to) is only valid within the `createNppStreamContext` function, not after it returns.

PR #807, opened 2025-08-05
Use MSVC toolchain when building FFmpeg on Windows

NicolasHug:
As done in torchaudio: https://github.com/pytorch/audio/blob/6fbc710b617f79b992ef2ebc7f95e818aa390293/.github/scripts/ffmpeg/build.sh#L20-L22

### DIFF
--- a/packaging/build_ffmpeg.sh
+++ b/packaging/build_ffmpeg.sh
@@ -20,6 +20,10 @@
 set -eux
 
 prefix="${FFMPEG_ROOT}"
+args=""
+if [[ "$OSTYPE" == "msys" ]]; then
+   args="--toolchain=msvc"
+fi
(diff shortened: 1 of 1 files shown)


packaging/build_ffmpeg.sh
```diff
@@ -20,6 +20,10 @@
 set -eux
 
 prefix="${FFMPEG_ROOT}"
+args=""
+if [[ "$OSTYPE" == "msys" ]]; then
```
scotts, line 24: Further down, we use `$(uname)` to determine we're running on Mac (Darwin). Can we use the same mechanism here, or if we can't, can we change the mechanism further down? I think using one mechanism to determine platform will make this script easier to understand and maintain.

NicolasHug:
I'm not sure we need this after all, closing for now

PR #814, opened 2025-08-13
Implement abstract per-GPU cache helper class

dvrogozh:
This commit introduces per-GPU cache helper class based on original design of the cache in CUDA device interface. New helper allows to store and reuse HW contexts which are expensive to create. Class is templated and allows storing different object types. At the moment it's applied to cache CUDA ffmpeg device contexts (`AVBufferRef`) and NPP contexts (`NppStreamContext`). Class can be further reused in non-CUDA device implementations.

CC: @scotts, @NicolasHug, @eromomon


src/torchcodec/_core/Cache.h
```diff
@@ -0,0 +1,123 @@
...
+template <typename T>
+bool PerGpuCache<T>::addIfCacheHasCapacity(
+    const torch::Device& device,
+    T&& obj) {
+  torch::DeviceIndex deviceIndex = getNonNegativeDeviceIndex(device);
+  if (static_cast<size_t>(deviceIndex) >= cache_.size()) {
+    return false;
+  }
+  cache_[deviceIndex]->addIfCacheHasCapacity(std::move(obj));
+  return true;
```
NicolasHug: Shouldn't we return the result from `cache_[deviceIndex]->addIfCacheHasCapacity()` here, instead of always returning `true`? IIUC the call to `cache_[deviceIndex]->addIfCacheHasCapacity()` may still fail if the capacity is exceeded?

```suggestion
  return cache_[deviceIndex]->addIfCacheHasCapacity(std::move(obj));
```

src/torchcodec/_core/Cache.h
```diff
@@ -0,0 +1,123 @@
...
+ private:
+  int capacity_;
+  std::mutex mutex_;
+  std::vector<T> cache_;
+};
+
+template <typename T>
+bool Cache<T>::addIfCacheHasCapacity(T&& obj) {
+  std::scoped_lock lock(mutex_);
+  if (capacity_ >= 0 && cache_.size() >= (size_t)capacity_) {
```
NicolasHug:
```suggestion
  if (capacity_ >= 0 && cache_.size() >= static_cast<size_t>(capacity_)) {
```

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -171,34 +134,37 @@ NppStreamContext createNppStreamContext(int deviceIndex) {
...
+  nppCtx->nMaxThreadsPerMultiProcessor = prop.maxThreadsPerMultiProcessor;
+  nppCtx->nMaxThreadsPerBlock = prop.maxThreadsPerBlock;
+  nppCtx->nSharedMemPerBlock = prop.sharedMemPerBlock;
+  nppCtx->nCudaDevAttrComputeCapabilityMajor = prop.major;
+  nppCtx->nCudaDevAttrComputeCapabilityMinor = prop.minor;
 
   return nppCtx;
 }
 
+std::unique_ptr<NppStreamContext> getNppStreamContext(
```
NicolasHug: Nit: instead of creating this new `getNppStreamContext` function, maybe we can merge it with `createNppStreamContext` and rename `createNppStreamContext` to `getNppStreamContext` ?

NicolasHug (commented):
Thanks a lot for the PR @dvrogozh , and for addressing our NppCtx cache TODOs! I made minor comments below, but this looks great overall. I'd also let @scotts take a look.

src/torchcodec/_core/Cache.h
dvrogozh: Definitely we should. Thank you for the catch.

src/torchcodec/_core/CudaDeviceInterface.cpp
dvrogozh: I am fine with that. Merged function will be short enough vs. what we have in `getCudaContext` and 2 create functions. I will update soon.

dvrogozh: done

src/torchcodec/_core/Cache.h
```diff
@@ -0,0 +1,122 @@
...
+    return false;
+  }
+  return cache_[deviceIndex]->addIfCacheHasCapacity(std::move(obj));
+}
+
+template <typename T>
+T PerGpuCache<T>::get(const torch::Device& device) {
+  torch::DeviceIndex deviceIndex = getNonNegativeDeviceIndex(device);
+  if (static_cast<size_t>(deviceIndex) >= cache_.size()) {
+    return {};
```
scotts: Is it ever valid to receive a `deviceIndex` that is outside of what the cache has been initialized to handle? I would think no, and that this should probably be a `TORCH_CHECK`?

src/torchcodec/_core/Cache.h
```diff
@@ -0,0 +1,122 @@
...
+  cache_.push_back(std::move(obj));
+  return true;
+}
+
+template <typename T>
+T Cache<T>::get() {
+  std::scoped_lock lock(mutex_);
+  if (!cache_.size())
+    return {};
+
```
scotts, line 61: We're depending on default object construction to indicate "empty cache." That works okay when we have pointers, but we have implemented this completely generically. I see several alternatives:

1. Explicitly make the objects in cache either `T*` or `std::unique_ptr<T>`. That way we can return `nullptr` when the cache is empty. I think that is more clear about expectations and usage.
2. Keep it as `T` and return an `optional<T>`. (Seems overkill if we know it's always going to contain pointers.)
3. Keep it as `T` and add a `Cache::empty()` member function that we expect callers to check, and make empty calls to `get()` throw.

src/torchcodec/_core/Cache.h
```diff
@@ -0,0 +1,122 @@
...
+    return false;
+  }
+  cache_.push_back(std::move(obj));
+  return true;
+}
+
+template <typename T>
+T Cache<T>::get() {
+  std::scoped_lock lock(mutex_);
+  if (!cache_.size())
```
scotts: Nit: I prefer calls to `cache_.empty()` instead of using an integer in a boolean context.

src/torchcodec/_core/Cache.h
dvrogozh: To clarify the 3d option, you mean the following?
```
if (!cache.empty()) item = cache.get();
...
T Cache<T>::get() {
  TORCH_CHECK(cache_.size(), ...);
}
```
This way there will be a race condition on fetching from cache in parallel.

I think 1st option makes more sense to me right now. I will implement it on Monday if there will be no objections.

scotts: Yes, I meant the third option, and yes, I agree there would be race conditions in a multithreaded environment. I also prefer the first option.

scotts:
@dvrogozh, thank you for working on this! These abstractions are very much needed. The small tweaks to the implementation we talked about should be enough.

src/torchcodec/_core/Cache.h
dvrogozh: Done

src/torchcodec/_core/Cache.h
dvrogozh: done

src/torchcodec/_core/Cache.h
dvrogozh: Implemented cache which explicitly works with `std::unique_ptr`. Please, take another look.

scotts (approved):
@dvrogozh, this looks great, thank you!

dvrogozh:
> `FAILED test/test_encoders.py::TestAudioEncoder::test_num_channels`

These ci failures seem unrelated.

scotts:
@dvrogozh, yup, agreed, it's a problem we know about and should start tracking. I'm going to merge.

PR #817, opened 2025-08-14
test: rename cpu_and_cuda to all_supported_devices

dvrogozh:
This commit renames a `cpu_and_cuda()` function used for pytest device parameterization to `all_supported_devices()`. This is helpful for those who work on multi-gpu support in torchcodec as `cpu_and_cuda()` was one of the function required to be patched to allow running torchcodec tests on other devices. Rename steps out from binding function to cuda and allows to reduce the change scope to enable non-cuda devices.

CC: @scotts, @NicolasHug, @eromomon


NicolasHug (approved):

NicolasHug (approved):
This sounds reasonable, thanks @dvrogozh

PR #818, opened 2025-08-18
Add internal usage logging

scotts:
Add *internal* torch-based logging for all of our top-level APIs.

Usage of this call is similar to using the logging module in the Python stdlib. It can be used for debugging purpose to log which methods are used and by default it is inactive, unless the user manually subscribes a logger via the [`SetAPIUsageLogger`](https://github.com/pytorch/pytorch/blob/eb3b9fe719b21fae13c7a7cf3253f970290a573e/c10/util/Logging.cpp#L114) function. Please note it is triggered only once for the same API call within a process. It does not collect any data from open-source users since it is no-op by default. For more information, please refer to the [PyTorch note on API usage logging](https://pytorch.org/docs/stable/notes/large_scale_deployments.html#api-usage-logging).

### DIFF
--- a/src/torchcodec/decoders/_audio_decoder.py
+++ b/src/torchcodec/decoders/_audio_decoder.py
@@ -8,8 +8,8 @@ import io
 from pathlib import Path
 from typing import Optional, Union
 
+import torch
 from torch import Tensor
-from torch._C import _log_api_usage_once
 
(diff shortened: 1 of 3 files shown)


scotts:
@NicolasHug, yes, definitely, that was an oversight on my part.

Issue #820, opened 2025-08-19
Refactor our resource generation script


scotts:
We have a script, [generate_reference_resources.sh](https://github.com/pytorch/torchcodec/blob/4af0bfe5f294415e14754f3ef9c5c3fc22b0d858/test/generate_reference_resources.sh), that we use to generate all of our reference frames for testing. *However*, we also have command in there that generates an actual audio mp3 from one of our reference videos: https://github.com/pytorch/torchcodec/blob/4af0bfe5f294415e14754f3ef9c5c3fc22b0d858/test/generate_reference_resources.sh#L45

On a clean repo, running this script should be an expensive no-op. That is, it should do a lot of work generating all of the test references, but what it generates should match exactly, bit-for-bit, what's already in the repo. Lately, however, when I run this script, the generated mp3 file from this line, `test/resources/nasa_13013.mp4.audio.mp3`, is different from what we have checked in.

Questions we should resolve:

1. Why is the newly generated mp3 different? This may be due to me having a different FFmpeg installed; I believe I used FFmpeg 4 back when first generating that mp3 and now I use FFmpeg 6 in my development environment.
2. Should we just side-step the above question entirely and say that this script should exclusively be about generating references *frames* from checked-in media files?
3. Since we're here, should we refactor the script into Python? My personal rule I like to follow is that once a bash script has a single branch, I would rather do it in Python. We crossed that threshold a while ago and now have nested for-loops. We also call a Python script, [`convert_image_to_tensor.py`](https://github.com/pytorch/torchcodec/blob/4af0bfe5f294415e14754f3ef9c5c3fc22b0d858/test/convert_image_to_tensor.py) from within this bash script, so we consolidate it all in one.
4. Even if we stick with bash, we should remove the redundant bmp generation. (Trying to do this bit is why I discovered this problem.)

NicolasHug:
1. I think it's very likely due to a different FFmpeg version as you noted, and generally different codec versions.
2. Yes, I think it'd be easier to consider this script to only be about generating the `.pt` expected frames. Precisely because of the problem we have with point 1, i.e. FFmpeg-version variability. So far, for a lot of videos, our workflow has been to generate the video locally with `ffmpeg` on an arbitrary version, and check it in into the repo. As long as our tests pass on all FFmpeg versions, this is just as robust, as probably a lot simpler.
3. Agreed that it'd be easier to maintain in Python
4. Agreed

PR #828, opened 2025-08-20
Update resource generation script to python

Dan-Flores:
This PR updates the resource generation script to be in python and integrates the `convert_image_to_tensor.py` script, as discussed in #820.
* Additionally, the MP3 audio file generation command is moved to `test/utils.py`, similar to the other TestVideo/Audio definitions.

Note:
In testing, I observed that the generated reference resources differed slightly from the currently checked-in files, but the resulting tensors were the same between the shell and python script, so its likely a difference in FFmpeg.


test/generate_reference_resources.py
```diff
@@ -0,0 +1,144 @@
...
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+# Run this script to update the resources used in unit tests. The resources are all derived
+# from source media already checked into the repo.
+
+
+def convert_image_to_tensor(image_path):
+    if not os.path.exists(image_path):
```
NicolasHug: Here and everywhere else in this file, let's try to modernize the codebase and rely on `pathlib` and `pathlib.Path` functionalities, instead of `os.*`.

test/generate_reference_resources.py
```diff
@@ -0,0 +1,144 @@
...
+        return
+    # Get base filename without extension
+    base_filename = os.path.splitext(image_path)[0]
+    pil_image = Image.open(image_path)
+    img_tensor = torch.from_numpy(np.asarray(pil_image))
+    print(img_tensor.shape)
+    print(img_tensor.dtype)
```
NicolasHug: I noticed they were present in the previous util, but they were probably debug leftovers that we can now remove?

test/generate_reference_resources.py
NicolasHug: The comment above about using `pathlib` still stands, but I wonder whether the "if file exists" check makes sense? Shouldn't we expect the file to exist, and error if not?

test/generate_reference_resources.py
```diff
@@ -0,0 +1,144 @@
...
+            "-i",
+            VIDEO_PATH,
+            "-vf",
+            f"select=eq(n\\,{frame})",
+            "-vsync",
+            "vfr",
+            "-q:v",
+            "2",
+            output_bmp,
+        ]
```
NicolasHug: I haven't looked at the `cmd` definition in details, but maybe there are opportunities to factorize some of the calls?

test/generate_reference_resources.py
```diff
@@ -0,0 +1,144 @@
...
+import numpy as np
+
+import torch
+from PIL import Image
+
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
```
scotts: I think this copyright notice should be at the very top - although that might conflict with the `#!` line, so we should make sure it can still be invoked the intended way.

Dan-Flores:
> Can you clarify how you observed that the generated references resources were different? I'm a little confused because I would assume that in order to compare the reference resources, we'd be looking at the resulting tensors (which, as you noted, appear to be equal)

I apologize for the confusing language @NicolasHug, when I mentioned generated references resources, I was referring to the tensors. 
The tensors output by the existing shell script and the new python script on my machine are identical with each other, but are slightly different from the tensors currently checked in. To confirm the difference was small, I ran a python script with `torch.load` and `torch.allclose` and a tolerance of 3 was enough to pass.

test/generate_reference_resources.py
Dan-Flores: There was one issue I encountered in the for loop over these arrays:
```
STREAMS = [0, 3]
FRAMES = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 15, 20, 25, 30, 35, 386, 387, 388, 389]
```

One stream index does not have as many frames, hence why stream 3's checked in frames go up to `nasa_13013.mp4.stream3.frame000389.pt`, but stream 0's checked in frames end at `nasa_13013.mp4.stream0.frame000035.pt`.
In these cases, FFmpeg errors quietly so the script does not create or delete the frames:

```
[vost#0:0/bmp @ 0x145f23cb0] No filtered frames for output stream, trying to initialize anyway.
...
[out#0/image2 @ 0x145f23150] Output file is empty, nothing was encoded(check -ss / -t / -frames parameters if used)
```

PR #831, opened 2025-08-22
Move filter graph to stand alone class

dvrogozh:
Changes:
* Defined stand alone class for the filter graph (based on definition in `CpuDeviceInterface`)
* Generalized filter graph to:
  * Cover HW backends (by accepting HW frames context)
  * Support configurable filter pipelines (by accepting string pipeline description). For example, it's possible to pass VAAPI or CUDA specific filters:
```
FiltersContext.filters = "scale=480:270:sws_flags=bilinear"; # for CPU (current usage)
FiltersContext.filters = "scale_vaapi=480:270:format=rgba";  # for VAAPI
```

These changes will allow:
* Easily define ffmpeg driven decoding + video processing pipeline for any backend supported by ffmpeg (for multi-gpu support)
* Add ffmpeg media filters support to torchcodec (any filter, not only scale or color conversion)

### DIFF
--- a/src/torchcodec/_core/CpuDeviceInterface.cpp
+++ b/src/torchcodec/_core/CpuDeviceInterface.cpp
@@ -6,11 +6,6 @@
 
 #include "src/torchcodec/_core/CpuDeviceInterface.h"
 
-extern "C" {
-#include <libavfilter/buffersink.h>
-#include <libavfilter/buffersrc.h>
-}
-
 namespace facebook::torchcodec {
 namespace {
 
@@ -19,20 +14,6 @@ static bool g_cpu = registerDeviceInterface(
     [](const torch::Device& device) { return new CpuDeviceInterface(device); });
 
 } // namespace
-
-bool CpuDeviceInterface::DecodedFrameContext::operator==(
-    const CpuDeviceInterface::DecodedFrameContext& other) {
-  return decodedWidth == other.decodedWidth &&
-      decodedHeight == other.decodedHeight &&
-      decodedFormat == other.decodedFormat &&
-      expectedWidth == other.expectedWidth &&
-      expectedHeight == other.expectedHeight;
-}
-
-bool CpuDeviceInterface::DecodedFrameContext::operator!=(
-    const CpuDeviceInterface::DecodedFrameContext& other) {
-  return !(*this == other);
-}
 
 CpuDeviceInterface::CpuDeviceInterface(const torch::Device& device)
     : DeviceInterface(device) {
@@ -132,8 +113,9 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
 
     frameOutput.data = outputTensor;
   } else if (colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH) {
-    if (!filterGraphContext_.filterGraph || prevFrameContext_ != frameContext) {
-      createFilterGraph(frameContext, videoStreamOptions, timeBase);
+    if (!filterGraphContext_ || prevFrameContext_ != frameContext) {
+      filterGraphContext_ = std::make_unique<FilterGraph>(
+          frameContext, videoStreamOptions, timeBase);
       prevFrameContext_ = frameContext;
     }
     outputTensor = convertAVFrameToTensorUsingFilterGraph(avFrame);
@@ -187,14 +169,8 @@ int CpuDeviceInterface::convertAVFrameToTensorUsingSwsScale(
 
 torch::Tensor CpuDeviceInterface::convertAVFrameToTensorUsingFilterGraph(
     const UniqueAVFrame& avFrame) {
-  int status = av_buffersrc_write_frame(
-      filterGraphContext_.sourceContext, avFrame.get());
-  TORCH_CHECK(
-      status >= AVSUCCESS, "Failed to add frame to buffer source context");
+  UniqueAVFrame filteredAVFrame = filterGraphContext_->convert(avFrame);
 
-  UniqueAVFrame filteredAVFrame(av_frame_alloc());
-  status = av_buffersink_get_frame(
-      filterGraphContext_.sinkContext, filteredAVFrame.get());
   TORCH_CHECK_EQ(filteredAVFrame->format, AV_PIX_FMT_RGB24);
 
   auto frameDims = getHeightAndWidthFromResizedAVFrame(*filteredAVFrame.get());
@@ -208,108 +184,6 @@ torch::Tensor CpuDeviceInterface::convertAVFrameToTensorUsingFilterGraph(
   };
   return torch::from_blob(
       filteredAVFramePtr->data[0], shape, strides, deleter, {torch::kUInt8});
-}
-
-void CpuDeviceInterface::createFilterGraph(
-    const DecodedFrameContext& frameContext,
-    const VideoStreamOptions& videoStreamOptions,
-    const AVRational& timeBase) {
-  filterGraphContext_.filterGraph.reset(avfilter_graph_alloc());
-  TORCH_CHECK(filterGraphContext_.filterGraph.get() != nullptr);
-
-  if (videoStreamOptions.ffmpegThreadCount.has_value()) {
-    filterGraphContext_.filterGraph->nb_threads =
-        videoStreamOptions.ffmpegThreadCount.value();
-  }
-
-  const AVFilter* buffersrc = avfilter_get_by_name("buffer");
-  const AVFilter* buffersink = avfilter_get_by_name("buffersink");
-
-  std::stringstream filterArgs;
-  filterArgs << "video_size=" << frameContext.decodedWidth << "x"
-             << frameContext.decodedHeight;
-  filterArgs << ":pix_fmt=" << frameContext.decodedFormat;
-  filterArgs << ":time_base=" << timeBase.num << "/" << timeBase.den;
-  filterArgs << ":pixel_aspect=" << frameContext.decodedAspectRatio.num << "/"
-             << frameContext.decodedAspectRatio.den;
-
-  int status = avfilter_graph_create_filter(
-      &filterGraphContext_.sourceContext,
-      buffersrc,
-      "in",
-      filterArgs.str().c_str(),
-      nullptr,
-      filterGraphContext_.filterGraph.get());
-  TORCH_CHECK(
-      status >= 0,
-      "Failed to create filter graph: ",
-      filterArgs.str(),
-      ": ",
-      getFFMPEGErrorStringFromErrorCode(status));
-
-  status = avfilter_graph_create_filter(
-      &filterGraphContext_.sinkContext,
-      buffersink,
-      "out",
-      nullptr,
-      nullptr,
-      filterGraphContext_.filterGraph.get());
-  TORCH_CHECK(
-      status >= 0,
-      "Failed to create filter graph: ",
-      getFFMPEGErrorStringFromErrorCode(status));
-
-  enum AVPixelFormat pix_fmts[] = {AV_PIX_FMT_RGB24, AV_PIX_FMT_NONE};
-
-  status = av_opt_set_int_list(
-      filterGraphContext_.sinkContext,
-      "pix_fmts",
-      pix_fmts,
-      AV_PIX_FMT_NONE,
-      AV_OPT_SEARCH_CHILDREN);
-  TORCH_CHECK(
-      status >= 0,
-      "Failed to set output pixel formats: ",
-      getFFMPEGErrorStringFromErrorCode(status));
-
-  UniqueAVFilterInOut outputs(avfilter_inout_alloc());
-  UniqueAVFilterInOut inputs(avfilter_inout_alloc());
-
-  outputs->name = av_strdup("in");
-  outputs->filter_ctx = filterGraphContext_.sourceContext;
-  outputs->pad_idx = 0;
-  outputs->next = nullptr;
-  inputs->name = av_strdup("out");
-  inputs->filter_ctx = filterGraphContext_.sinkContext;
-  inputs->pad_idx = 0;
-  inputs->next = nullptr;
-
-  std::stringstream description;
-  description << "scale=" << frameContext.expectedWidth << ":"
-              << frameContext.expectedHeight;
-  description << ":sws_flags=bilinear";
-
-  AVFilterInOut* outputsTmp = outputs.release();
-  AVFilterInOut* inputsTmp = inputs.release();
-  status = avfilter_graph_parse_ptr(
-      filterGraphContext_.filterGraph.get(),
-      description.str().c_str(),
-      &inputsTmp,
-      &outputsTmp,
-      nullptr);
-  outputs.reset(outputsTmp);
-  inputs.reset(inputsTmp);
-  TORCH_CHECK(
-      status >= 0,
-      "Failed to parse filter description: ",
-      getFFMPEGErrorStringFromErrorCode(status));
-
-  status =
-      avfilter_graph_config(filterGraphContext_.filterGraph.get(), nullptr);
-  TORCH_CHECK(
-      status >= 0,
-      "Failed to configure filter graph: ",
-      getFFMPEGErrorStringFromErrorCode(status));
 }
 
 void CpuDeviceInterface::createSwsContext(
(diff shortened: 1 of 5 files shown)


dvrogozh:
Pushed a fix for the extra `;` typo which was failing some ci checks.

dvrogozh:
Latest CI errors don't seem to be related with the PR. 2 errors:
```
Error: Unable to download artifact(s): Artifact not found for name: pytorch_torchcodec__3.9_cu126_x86_64
```
and
```
/Users/ec2-user/runner/_work/torchcodec/torchcodec/pytorch/torchcodec/src/torchcodec/_core/custom_ops.cpp:87:47: error: must specify at least one argument for '...' parameter of variadic macro [-Werror,-Wgnu-zero-variadic-macro-arguments]
  TORCH_INTERNAL_ASSERT(tensor.is_contiguous());
                                              ^
/Users/ec2-user/runner/_work/_temp/conda_environment_17164967222/lib/python3.10/site-packages/torch/include/c10/util/Exception.h:425:9: note: macro 'TORCH_INTERNAL_ASSERT' defined here
#define TORCH_INTERNAL_ASSERT(cond, ...)                                         \
* Getting build dependencies for wheel...
```

I believe the second error is addressed already in #847, so I will need to rebase. Not sure if the first one is addressed. @scotts : should I rebase now or you need time to fix 1st issue?

dvrogozh:
Rebased on top of latest master. I hope this will help ci to pass as some fixes for ci has recently landed.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -84,13 +65,22 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
...
+  filtersContext.outputWidth = expectedOutputWidth;
+  filtersContext.outputHeight = expectedOutputHeight;
+  filtersContext.outputFormat = AV_PIX_FMT_RGB24;
+  filtersContext.timeBase = timeBase;
+
+  std::stringstream filters;
+  filters << "scale=" << expectedOutputWidth << ":" << expectedOutputHeight;
+  filters << ":sws_flags=bilinear";
+
+  filtersContext.filters = filters.str();
```
NicolasHug: I think the `filters` field isn't used by `swscale`, yet it gets created unconditionally. Maybe this should be moved to the branch where `colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH`?
And we may want to rename the `filters` field to something more explicit about filtergraph - e.g. `filtergraphStr` or something like that?

src/torchcodec/_core/FilterGraph.cpp
```diff
@@ -0,0 +1,150 @@
...
+  const AVFilter* buffersrc = avfilter_get_by_name("buffer");
+  const AVFilter* buffersink = avfilter_get_by_name("buffersink");
+
+  auto deleter = [](AVBufferSrcParameters* p) {
+    if (p) {
+      av_freep(&p);
+    }
+  };
+  std::unique_ptr<AVBufferSrcParameters, decltype(deleter)> srcParams(
+      nullptr, deleter);
```
NicolasHug: Could / should we move that to the other smart pointer logic we have in `FFMPEGCommon.h`?

src/torchcodec/_core/FilterGraph.h
```diff
@@ -0,0 +1,45 @@
...
+  FilterGraph(
+      const FiltersContext& filtersContext,
+      const VideoStreamOptions& videoStreamOptions);
+
+  UniqueAVFrame convert(const UniqueAVFrame& avFrame);
+
+ private:
+  UniqueAVFilterGraph filterGraph_;
+  AVFilterContext* sourceContext_ = nullptr;
+  AVFilterContext* sinkContext_ = nullptr;
```
NicolasHug, line 42: I guess this was pre-existing, but I'm curious if those are properly freed on destruction? Maybe they are freed  along with `filterGraph_`  by `avfilter_graph_free`?

src/torchcodec/_core/FilterGraph.h
```diff
@@ -0,0 +1,45 @@
...
+namespace facebook::torchcodec {
+
+struct FiltersContext {
+  int inputWidth = 0;
+  int inputHeight = 0;
+  AVPixelFormat inputFormat = AV_PIX_FMT_NONE;
+  AVRational inputAspectRatio = {0, 0};
+  int outputWidth = 0;
+  int outputHeight = 0;
+  AVPixelFormat outputFormat = AV_PIX_FMT_NONE;
```
NicolasHug, line 21: At the moment, in `main`, we are hard-cording the output format to `AV_PIX_FMT_RGB24`. Do we expect the `outputFormat` to be used differently in the future (maybe in the CUDA 10bit PR?) ?
I'm asking because if not, maybe we should still hard-code the value.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -114,9 +104,9 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
     outputTensor = preAllocatedOutputTensor.value_or(allocateEmptyHWCTensor(
         expectedOutputHeight, expectedOutputWidth, torch::kCPU));
 
-    if (!swsContext_ || prevFrameContext_ != frameContext) {
-      createSwsContext(frameContext, avFrame->colorspace);
-      prevFrameContext_ = frameContext;
+    if (!swsContext_ || prevFiltersContext_ != filtersContext) {
+      createSwsContext(filtersContext, avFrame->colorspace);
+      prevFiltersContext_ = std::move(filtersContext);
```
NicolasHug: Because the new `FiltersContext` has more field than the old `DecodedFrameContext`, we are now comparing more things to decide whether we should re-create a new swscontext.

I wonder if we should have a sws-specific comparision function that only compares the fields that are specific to swscale? My goal here is that we avoid re-creating the swscontext unnecessarily.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -132,9 +122,10 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
 
     frameOutput.data = outputTensor;
   } else if (colorConversionLibrary == ColorConversionLibrary::FILTERGRAPH) {
-    if (!filterGraphContext_.filterGraph || prevFrameContext_ != frameContext) {
-      createFilterGraph(frameContext, videoStreamOptions, timeBase);
-      prevFrameContext_ = frameContext;
+    if (!filterGraphContext_ || prevFiltersContext_ != filtersContext) {
+      filterGraphContext_ =
+          std::make_unique<FilterGraph>(filtersContext, videoStreamOptions);
+      prevFiltersContext_ = std::move(filtersContext);
```
NicolasHug, line 151: Same Q here about potentially unnecessarily re-creating the filtergraph context, now that we are comparing more fields than before?

src/torchcodec/_core/FilterGraph.h
```diff
@@ -0,0 +1,45 @@
...
+
+  std::string filters;
+  AVRational timeBase = {0, 0};
+  UniqueAVBufferRef hwFramesCtx;
+
+  bool operator==(const FiltersContext&);
+  bool operator!=(const FiltersContext&);
+};
```
NicolasHug, line 29: I think it's fine to keep the new `FiltersContext` struct definition in this file (`FilterGraph.h`), but it's worth adding a comment indicating that this struct is also used for `swscale`, not just filtergraph.

src/torchcodec/_core/FilterGraph.h
```diff
@@ -0,0 +1,45 @@
...
+  int outputWidth = 0;
+  int outputHeight = 0;
+  AVPixelFormat outputFormat = AV_PIX_FMT_NONE;
+
+  std::string filters;
+  AVRational timeBase = {0, 0};
+  UniqueAVBufferRef hwFramesCtx;
+
+  bool operator==(const FiltersContext&);
+  bool operator!=(const FiltersContext&);
```
NicolasHug: Probably  a NIT, but can't hurt:

```suggestion
  bool operator==(const FiltersContext&) const;
  bool operator!=(const FiltersContext&) const;
```

src/torchcodec/_core/CpuDeviceInterface.cpp
NicolasHug: Here and below, do we strictly need the use of `std::move()`? My understanding is that the FiltersContext struct contains simple types so that may not be necessary?

src/torchcodec/_core/FilterGraph.h
dvrogozh: I was wondering the same when I spotted these variables without explicit destructors. It occurs that we should not call `avfilter_free()` explicitly if filters are owned by filter graph (which is our case). They will be destroyed within a call to `avfilter_graph_free()`, see [libavfilter/avfiltergraph.c#L128](https://github.com/FFmpeg/FFmpeg/blob/2fffa01ddb777871cd53ef6c92f5c19cae8afe16/libavfilter/avfiltergraph.c#L128). And on the calls to `avfilter_graph_alloc_filter()` and `avfilter_graph_create_filter()` we get the pointer to the filter stored in the filter graph itself. So, we are good here as `UniqueAVFilterGraph` will manage filter graph for us:
https://github.com/pytorch/torchcodec/blob/d14f6423699289f8093e3fbbc3a979d559d2db5b/src/torchcodec/_core/FFMPEGCommon.h#L65-L67

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: Yes, we do need `std::move`. The reason behind is that `FiltersContext` contains `UniqueAVBufferRef hwFramesCtx` which is `std::unique_ptr`. The latter explicitly deletes copy constructor and assign operators. Usage of `std::unique_ptr` to manage `hwFramesCtx` is justified as it's `AVBufferRef` which we can't just copy as we will get issues managing references. To properly copy it's needed to call `new_ref = av_buffer_ref(hwFramesCtx)`.

src/torchcodec/_core/FilterGraph.cpp
dvrogozh: I can move it. To clarify, I cannot use existing definitions in `FFMPEGCommon.h` for deleters because of type mismatches between argument template and `av_freep`:
https://github.com/pytorch/torchcodec/blob/d14f6423699289f8093e3fbbc3a979d559d2db5b/src/torchcodec/_core/FFMPEGCommon.h#L35-L40
And `av_freep` is declared as:
```
void av_freep(void *ptr)
```
Note `T**` vs. `void*`. In our case `T=AVBufferSrcParameters`. But I can definitely move this to the ffmpeg header file.

src/torchcodec/_core/FilterGraph.h
dvrogozh: > Do we expect the `outputFormat` to be used differently in the future (maybe in the CUDA 10bit PR?) ?

Yes, it will be different for HW filters:
* For CPU, `outputFormat = AV_PIX_FMT_RGB24` (or other output format if needed)
* For CUDA, `outputFormat = AV_PIX_FMT_CUDA` (and actual format will be set via filter option: `scale_cuda=format=nv12`)
* For XPU VAAPI, `outputFormat = AV_PIX_FMT_VAAPI` (and actual format will be set via filter option: `scale_vaapi=format=rgba`)

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: Well, yes... Actually next PR will propose to move `FilterGraph` creation mechanism to `SimpleStreamDecoder` and this place will become even more weird (see #853)... I did not handle that in #853, but I don't exactly like how.

So, I thought that maybe a better option is to just separate frame contexts for SWS and `FilterGraph`. This way we address few things including your concern on starting to compare fields not specific to SWS. I've implemented this in the update to the PR I just pushed. Take a look, please.

src/torchcodec/_core/FilterGraph.cpp
dvrogozh: Moved to `FFMPEGCommon.h`.

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: Addressed by having `FiltersContext` for filter graph and `SwsFrameContext` for SWS. They have separate list of fields and separate comparison functions.

dvrogozh: As I've introduced `SwsFrameContext`, I removed `std::move` for it in SWS path. `std::move` is still needed for `FiltersContext` path.

src/torchcodec/_core/FilterGraph.h
dvrogozh: done

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: Yes, we do need `std::move`. The reason behind is that `FiltersContext` contains `UniqueAVBufferRef hwFramesCtx` which is `std::unique_ptr`. The latter explicitly deletes copy constructor and assign operators. Usage of `std::unique_ptr` to manage `hwFramesCtx` is justified as it's `AVBufferRef` which we can't just copy as we will get issues managing references. To properly copy it's needed to call `new_ref = av_buffer_ref(hwFramesCtx)`.

src/torchcodec/_core/FilterGraph.h
dvrogozh: I did not add a comment as I've introduced `SwsFrameContext` defined in CPU device interface. Take a look on this approach, please.

dvrogozh:
@NicolasHug, I've addressed your comments. One thing worth noting is that I've introduced `SwsFrameContext` as SWS indeed does not need all the fields from `FiltersContext` and having 2 separate definitions will actually simplify the code in the future PR when creation of filter graph will be moved to `SingleStreamDecoder`. Please, take a look and comment whether you are like such approach better?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -111,12 +90,27 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
...
+    // And we sometimes re-create them because it's possible for frame
+    // resolution to change mid-stream. Finally, we want to reuse the colorspace
+    // conversion objects as much as possible for performance reasons.
+    SwsFrameContext swsFrameContext;
+
+    swsFrameContext.inputWidth = avFrame->width;
+    swsFrameContext.inputHeight = avFrame->height;
+    swsFrameContext.inputFormat = frameFormat;
+    swsFrameContext.outputWidth = expectedOutputWidth;
+    swsFrameContext.outputHeight = expectedOutputHeight;
```
NicolasHug, line 106: Curious if we should prefer using [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization.html) here, instead of setting each field individually. I think I have a preference for aggregate initialization because if we add a new field to the `SwsFrameContext` struct, we'd get a loud compilation error if we forget to initialize the field (which is good).

@scotts any pref?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -132,9 +126,27 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
...
+    filtersContext.inputFormat = frameFormat;
+    filtersContext.inputAspectRatio = avFrame->sample_aspect_ratio;
+    filtersContext.outputWidth = expectedOutputWidth;
+    filtersContext.outputHeight = expectedOutputHeight;
+    filtersContext.outputFormat = AV_PIX_FMT_RGB24;
+    filtersContext.timeBase = timeBase;
+
+    std::stringstream filters;
+    filters << "scale=" << expectedOutputWidth << ":" << expectedOutputHeight;
+    filters << ":sws_flags=bilinear";
```
NicolasHug, line 144: It's unfortunate that we now have to re-create the string for every single frame in order to compare the `filtersContext` objects. Maybe eventually we should separate the concerns between filter-graph creation parameters and comparison operators. But I guess for now this is cheap enough.

src/torchcodec/_core/CpuDeviceInterface.cpp
scotts: I generally prefer objects to be fully specified on construction, rather than default-constructed and then members set one-by-one. But I don't think aggregate initialization will save use here from forgetting a field, as I think that for types with a default, it will just use that.

What I think we should do is actually just create a constructor for all of our structs. And to make sure we don't miss any members, we should remove the default constructor. (When possible. If we're creating an array of something we won't be able to do that.)

PR #838, opened 2025-08-22
Correct `seek` signature to return `int` instead of `bytes`

maksdamir:
Current examples suggest `seek` api on file-like objects should return `bytes`, while actual usage always expects `int` instead.

For example: https://docs.python.org/3/library/io.html#io.IOBase.seek

```
seek(offset, whence=os.SEEK_SET, /)
Change the stream position to the given byte offset, interpreted relative to the position indicated by whence, and return the new absolute position.
```

### DIFF
--- a/examples/decoding/file_like.py
+++ b/examples/decoding/file_like.py
@@ -215,7 +215,7 @@ class FileOpCounter:
         self.num_reads += 1
         return self._file.read(size)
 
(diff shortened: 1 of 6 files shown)


Dan-Flores (approved):
Good catch, thanks for the PR @maksdamir!

scotts:
@maksdamir, wow, great catch, thank you!

PR #847, opened 2025-08-26
Fix build

NicolasHug:
All builds started [failing](https://github.com/pytorch/torchcodec/actions/runs/17239919864/job/48914006197?pr=846) with:

```
/Users/ec2-user/runner/_work/torchcodec/torchcodec/pytorch/torchcodec/src/torchcodec/_core/custom_ops.cpp:87:47: error: must specify at least one argument for '...' parameter of variadic macro [-Werror,-Wgnu-zero-variadic-macro-arguments]
  TORCH_INTERNAL_ASSERT(tensor.is_contiguous());
                                              ^
```

Probably a BC-breaking change from torch? Anyway, we should probably rely on `TORCH_CHECK` rather than `TORCH_INTERNAL_ASSERT` regardless.

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -84,7 +84,9 @@ at::Tensor wrapDecoderPointerToTensor(
 }
 
 SingleStreamDecoder* unwrapTensorToGetDecoder(at::Tensor& tensor) {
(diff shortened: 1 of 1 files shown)


scotts:
Agreed on not using `TORCH_INTERNAL_ASSERT` - that's leftover from a long while ago.

PR #851, opened 2025-08-26
Shorter frame-mapping test name

NicolasHug:
The test currently has a really long name as it contains the entire file content. See e.g. https://github.com/pytorch/torchcodec/actions/runs/17242892367/job/48925287759#step:11:215

This PR fixes this by hard-coding the test name, which can be done through `pytes.param(..., id=NAME)`.

I think this is what's causing the failures on  https://github.com/pytorch/torchcodec/actions/runs/17242927437/job/48927121120?pr=806

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1342,7 +1342,11 @@ class TestVideoDecoder:
     @pytest.mark.parametrize(
         "custom_frame_mappings,expected_match",
         [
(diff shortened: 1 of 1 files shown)


scotts (approved):

Dan-Flores (approved):
Thanks for quickly creating this fix!

PR #856, opened 2025-08-27
Fix implicit conversion of int64 to int

NicolasHug:
Closes https://github.com/pytorch/torchcodec/issues/679

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -93,6 +93,7 @@ function(make_torchcodec_libraries
         CpuDeviceInterface.cpp
         SingleStreamDecoder.cpp
         Encoder.cpp
+        ValidationUtils.cpp
     )
 
     if(ENABLE_CUDA)
@@ -123,7 +124,6 @@ function(make_torchcodec_libraries
     set(custom_ops_sources
         AVIOTensorContext.cpp
         custom_ops.cpp
-        ValidationUtils.cpp
     )
     set(custom_ops_dependencies
         ${core_library_name}
@@ -141,7 +141,6 @@ function(make_torchcodec_libraries
     set(pybind_ops_sources
         AVIOFileLikeContext.cpp
         pybind_ops.cpp
-        ValidationUtils.cpp
     )
     set(pybind_ops_dependencies
        ${core_library_name}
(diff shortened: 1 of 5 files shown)


src/torchcodec/_core/CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ function(make_torchcodec_libraries
     set(custom_ops_sources
         AVIOTensorContext.cpp
         custom_ops.cpp
+        ValidationUtils.cpp
```
NicolasHug: Not a fan creating a new `cpp` file for this, but I'm not sure where to have it otherwise. It's needed for both the `pybind_ops` and the `custom_ops`. Happy to consider alternatives.


Also, instead of listing `ValidationUtils.cpp` for both the custom_ops and pybind_ops libs, we could just list it once as part of the core lib.

Dan-Flores: It makes sense to create a new `cpp` file for this, we may add more validations functions in the future. 

Since it is used in both libraries I think it makes sense to add `ValidationUtils.cpp` to the core lib, as the other shared libraries are also there (`SingleStreamDecoder.cpp`, `Encoder.cpp`).

Dan-Flores (approved):

src/torchcodec/_core/CMakeLists.txt
scotts: A new `cpp` for this is the best practice, as things are currently implemented. *However*, we might end up removing the `cpp` in the future because if we find ourselves implementing variants of these validation functions, we should probably instead define something like:
```c++
template <SrcT, DestT>
DestT validateIntegerConversion(SrcT value, const std::string& parameterName) {
  TORCH_CHECK(
      value >= std::numeric_limits<DestT>::min() &&
          value <= std::numeric_limits<DestT>::max(),
      parameterName,
      "=",
      value,
      " is out of range for int type.");

  return static_cast<DestT>(value);
}
```
And the definition of template functions must live in the header files. I don't think we should make this change yet because:

1. My code above may not generalize correctly! Testing it's general correctness is not worth it right now.
2. We may never need anything beyond this specific validation of types. ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).)

Yeah, I didn't do this in #830, but it was one function, and one could argue I was just being lazy. :)

PR #857, opened 2025-08-28
Prevent 3.13t Windows builds

NicolasHug:
Windows 3.13t builds don't work, which causes failures on our `main` branch (not on PRs): https://github.com/pytorch/torchcodec/actions/runs/17288016442/job/49068864699

This PR prevents those 3.13t jobs from being started. That's OK, we can do without 3.13t it for now. 3.13 is supported.

### DIFF
--- a/.github/workflows/windows_wheel.yaml
+++ b/.github/workflows/windows_wheel.yaml
@@ -36,6 +36,8 @@ jobs:
       with-rocm: disable
       with-cuda: disable
       build-python-only: "disable"
+      python-versions: '["3.10", "3.11", "3.12", "3.13"]'
+
 
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/windows_wheel.yaml
```diff
@@ -36,6 +36,8 @@ jobs:
       with-rocm: disable
       with-cuda: disable
       build-python-only: "disable"
+      python-versions: '["3.10", "3.11", "3.12", "3.13"]'
```
NicolasHug, line 39: Note that on PR, only 3.10 will be started, which is what we want. To start all jobs, we'd have to set the `ciflow/binaries/all` label as usual.

PR #862, opened 2025-09-02
Use CUDA 13.0 on CI

NicolasHug:
It seems from https://github.com/pytorch/torchcodec/actions/runs/17402821403/job/49401670135?pr=831 that test-infra stopped support 12.9 and is generating jobs for 13.0 instead, so we need to change our jobs to reflect that.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -55,7 +55,7 @@ jobs:
           # the wheel unless the label cliflow/binaries/all is present in the
           # PR.
         python-version: ['3.10']
-        cuda-version: ['12.6']
+        cuda-version: ['12.8']
         ffmpeg-version-for-tests: ['7']
     container:
       image: "pytorch/manylinux2_28-builder:cuda${{ matrix.cuda-version }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yaml
```diff
@@ -55,7 +55,7 @@ jobs:
           # the wheel unless the label cliflow/binaries/all is present in the
           # PR.
         python-version: ['3.10']
-        cuda-version: ['12.6']
+        cuda-version: ['12.8']
```
NicolasHug: Drive-by, 12.6 is still supported but the "default" CUDA version supported by pytorch is currently 12.8

Dan-Flores: Its possible that the github runner `linux.4xlarge.nvidia.gpu` does not have CUDA 12.8, and thats why this job was erroring. I've reverted it since its not related to this change, but we could leave a TODO to update the CUDA version if its something we need to change eventually.

test/utils.py
```diff
@@ -73,7 +73,7 @@ def psnr(a, b, max_val=255) -> float:
 def assert_frames_equal(*args, **kwargs):
     if sys.platform == "linux":
         if args[0].device.type == "cuda":
-            atol = 2
+            atol = 3
```
NicolasHug, line 76: I think we should be able to preserve the previous stricter with CUDA < 13 with something like:

```suggestion
            atol = 3 if cuda_version_used_for_building_torch() >= (13, 0) else 2
```

NicolasHug:
Thanks @Dan-Flores 	for the fixes! LGTM, let's close this one so you can open a new one

Issue #843, opened 2025-08-24
RuntimeError: The frame has 0 channels, expected 1. If you are hitting this, it may be because you are using a buggy FFmpeg version. FFmpeg4 is known to fail here in some valid scenarios. Try to upgrade FFmpeg?


obadx:
### 🐛 Describe the bug

I having an error trying to read single channel wave file. I tried to upgrade `ffmpeg` version but the same error remains.

Reproducing the error using [Google Colab Notebook](https://colab.research.google.com/drive/1cherOqxUL-5kK5reGX73cmvZyGExcbb6?usp=sharing):

```python
! apt install ffmpeg
! pip install torchcodec

! ffmpeg -version

! wget https://raw.githubusercontent.com/obadx/quran-muaalem/main/assets/test.wav

from torchcodec.decoders import AudioDecoder


decoder = AudioDecoder("./test.wav", sample_rate=16000, num_channels=1)
print(decoder.get_all_samples().data[0].shape)

RuntimeError: The frame has 0 channels, expected 1. If you are hitting this, it may be because you are using a buggy FFmpeg version. FFmpeg4 is known to fail here in some valid scenarios. Try to upgrade FFmpeg?

! ffmpeg -i test.wav output.mp3

from torchcodec.decoders import AudioDecoder


decoder = AudioDecoder("./output.mp3", sample_rate=16000, num_channels=1)
print(decoder.get_all_samples().data[0].shape)
# torch.Size([140320])
```
 

### Versions

```


Collecting environment information...
PyTorch version: 2.8.0+cu126
Is debug build: False
CUDA used to build PyTorch: 12.6
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.4 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version: 14.0.0-1ubuntu1.1
CMake version: version 3.31.6
Libc version: glibc-2.35

Python version: 3.12.11 (main, Jun  4 2025, 08:56:18) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-6.1.123+-x86_64-with-glibc2.35
Is CUDA available: False
CUDA runtime version: 12.5.82
CUDA_MODULE_LOADING set to: N/A
GPU models and configuration: Could not collect
Nvidia driver version: Could not collect
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.9.2.1
[...]

PR #865, opened 2025-09-02
Fix audio decoding bug with FFmpeg4 by setting channel_layout

Dan-Flores:
This PR avoids the error message in #843 when usinf FFmpeg4 by setting the missing `channel_layout` field avFrame. [FFmpeg documentation](https://ffmpeg.org/doxygen/4.4/structAVCodecParameters.html#a18e0af4b263443a6c4d2d5304dbd537e) indicates that this field may be 0 if it is unknown or unspecified, so it is set to the default layout for the number of channels present using `av_get_default_channel_layout`.

The test was updated to compare the decoded frames to the checked in samples from `sine_mono_s16.wav`. These checked in samples were generated using torchcodec with FFmpeg6.


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1277,15 +1277,6 @@ void SingleStreamDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
   int srcNumChannels = getNumChannels(streamInfo.codecContext);
-  TORCH_CHECK(
-      srcNumChannels == getNumChannels(srcAVFrame),
-      "The frame has ",
-      getNumChannels(srcAVFrame),
-      " channels, expected ",
-      srcNumChannels,
-      ". If you are hitting this, it may be because you are using "
-      "a buggy FFmpeg version. FFmpeg4 is known to fail here in some "
-      "valid scenarios. Try to upgrade FFmpeg?");
```
NicolasHug: I think we should still have this check, but perhaps relax it if `getNumChannels(srcAVFrame)` returns 0 while `getNumChannels(streamInfo.codecContext)` returns something < 1 (which is what is happening in https://github.com/pytorch/torchcodec/issues/843

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -61,7 +61,14 @@ int getNumChannels(const UniqueAVFrame& avFrame) {
...
 #else
-  return av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  int numChannels = av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  // Handle FFmpeg 4 bug where channel_layout is 0 or unset
+  // Set it to the default layout for the number of channels
+  // to allow successful initialization of SwrContext
+  if (numChannels == 0 && avFrame->channels > 0) {
+    avFrame->channel_layout = av_get_default_channel_layout(avFrame->channels);
+  }
+  return numChannels;
```
NicolasHug, line 72: I'm a bit surprised by this. We're returning `numChannels` but we know it's 0. Shouldn't we return `avFrame->channels` which is non-zero instead?

Also, can you help me understand the effect of setting `avFrame->channel_layout` - why is it necessary?

Dan-Flores: You're right,  it makes more sense to return `avFrame->channels` and keep a `TORCH_CHECK`.

> can you help me understand the effect of setting avFrame->channel_layout - why is it necessary?

It seems the `Invalid argument` error that gets thrown in `createSwrContext` in this file happens when `srcAVFrame->channel_layout` is unset / 0. I'm not sure why or when this value is unset. 

I'm also open to suggestions on where to update this field, if this function is not the right place.

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: I added back the check, and excluded the case where `getNumChannels(srcAVFrame)` returns 0 while `getNumChannels(streamInfo.codecContext)` returns a valid fallback.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -61,7 +61,15 @@ int getNumChannels(const UniqueAVFrame& avFrame) {
...
   return avFrame->ch_layout.nb_channels;
 #else
-  return av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  int numChannels = av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  // Handle FFmpeg 4 bug where channel_layout is 0 or unset
+  // Set it to the default layout for the number of channels
+  // to allow successful initialization of SwrContext
+  if (numChannels == 0 && avFrame->channels > 0) {
+    avFrame->channel_layout = av_get_default_channel_layout(avFrame->channels);
+    return avFrame->channels;
```
NicolasHug: Let's just do this instead, so we can have a single `return` statement

```suggestion
    numChannels = avFrame->channels;
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1276,16 +1276,20 @@ void SingleStreamDecoder::convertAudioAVFrameToFrameOutputOnCPU(
...
+  int srcNumChannels = getNumChannels(srcAVFrame);
+  int srcNumChannelsFromCodec = getNumChannels(streamInfo.codecContext);
+  // Use number of channels from codec if 0 returned from frame
+  if (srcNumChannels == 0 && srcNumChannelsFromCodec > 0) {
+    srcNumChannels = srcNumChannelsFromCodec;
```
NicolasHug: I wonder if we still need this `if` branch, now that we are returning the hopefully-correct `avFrame->channels` instead of the incorrect `numChannels` in `getNumChannels()`.

Can you double check if that's still needed? Perhaps we can leave the entire `TORCH_CHECK()` the way it was!

test/test_decoders.py
```diff
@@ -1682,26 +1682,25 @@ def test_downsample_empty_frame(self):
...
         assert decoder.metadata.sample_rate == asset.sample_rate
         assert decoder.metadata.sample_format == asset.sample_format
 
-        cm = (
-            pytest.raises(RuntimeError, match="The frame has 0 channels, expected 1.")
-            if get_ffmpeg_major_version() == 4
-            else contextlib.nullcontext()
+        test_frames = decoder.get_samples_played_in_range()
+        assert test_frames.data.shape[0] == decoder.metadata.num_channels
+        assert test_frames.sample_rate == decoder.metadata.sample_rate
```
NicolasHug: Let's use the term "samples" instead of frame for the stuff that comes out of the decoder. For the asset, the use of

```
reference_frames = asset.get_frame_data_by_range
```

is still correct since this is really a frame-based API.

test/test_decoders.py
```diff
@@ -1682,26 +1682,25 @@ def test_downsample_empty_frame(self):
...
+        reference_frames = asset.get_frame_data_by_range(
+            start=0, stop=1, stream_index=0
+        )
+        torch.testing.assert_close(
+            test_frames.data[0], reference_frames, atol=0, rtol=0
```
NicolasHug: I'm surprised by the use of `[0]` here - can you help me understand why that's needed?

src/torchcodec/_core/FFMPEGCommon.cpp
NicolasHug: Thanks, that makes sense!

test/test_decoders.py
Dan-Flores: The decoded audio has the shape `[1, 64000]`, where the first dimension is the number of channels (1, in the case of `SINE_MONO_S16`), and the second dimension is the samples.

The checked in tensor is returned by `get_frame_data_by_index` in `utils.py`, which returns only the samples at `index=1` for the one `stream_index`. 

So to make the comparison, the test must use only the samples from the decoded audio at `[0]`.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -61,7 +61,15 @@ int getNumChannels(const UniqueAVFrame& avFrame) {
     (LIBAVFILTER_VERSION_MAJOR == 8 && LIBAVFILTER_VERSION_MINOR >= 44)
   return avFrame->ch_layout.nb_channels;
 #else
-  return av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  int numChannels = av_get_channel_layout_nb_channels(avFrame->channel_layout);
+  // Handle FFmpeg 4 bug where channel_layout is 0 or unset
+  // Set it to the default layout for the number of channels
+  // to allow successful initialization of SwrContext
```
NicolasHug: Nit, I think the code has slightly changed since the comment was written, so this edit may be closer to what's happening (feel free to edit!)

```suggestion
  // Handle FFmpeg 4 bug where channel_layout and numChannels are 0 or unset
  // Set them based on avFrame->channels which appears to be correct
  // to allow successful initialization of SwrContext
```

test/test_decoders.py
NicolasHug: Got it, I think eventually we may want to update `get_frame_data_by_index` to always return 2D audio samples, but that's not urgent.  Using `[0]` like yo did is fine.

PR #866, opened 2025-09-03
VideoEncoder first pass, round trip test

Dan-Flores:
This PR implements the backend functions needed for video encoding. 
The main changes are in  `Encoder.cpp/h` and `test_ops.py`, the rest are scaffolding to allow the code to build / test.

Encoder.cpp implements primary functions:
* `initializeEncoder`
* `convertTensorToAVFrame`
* `encodeFrame`
* `encode` (the top level function)

One test is implemented in `test_video_encoder_test_round_trip`: 
* This test converts the `nasa_13013.mp4` to `mp4`, `mov`, and `avi`, with a resulting peak-signal to noise ratio around 30 - 40. 

Remaining items to implement are marked with `TODO-VideoEncoder`:
* Remove hardcoding of H264 codec
* Validate sample rate, similar to the AudioEncoder
* Enable selection of VideoStreamOptions data to be set or remove them if unneeded
  * height, width, bitrate, GOP size, max B-frames
* test additional video container formats

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -1,7 +1,3 @@
-#include <iostream>
-extern "C" {
-#include <libavutil/pixdesc.h>
-}
 #include <sstream>
 
 #include "src/torchcodec/_core/AVIOTensorContext.h"
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -1,3 +1,7 @@
+#include <iostream>
+extern "C" {
+#include <libavutil/pixdesc.h>
+}
 #include <sstream>
 
 #include "src/torchcodec/_core/AVIOTensorContext.h"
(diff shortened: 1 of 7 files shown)


src/torchcodec/_core/Encoder.h
```diff
@@ -57,6 +57,46 @@ class AudioEncoder {
   bool encodeWasCalled_ = false;
   int64_t lastEncodedAVFramePts_ = 0;
 };
+
```
NicolasHug, line 60: Nit: start the video encoder code after the big "Note: [Encoding loop, sample rate conversion and FIFO]"

src/torchcodec/_core/Encoder.h
```diff
@@ -57,6 +57,46 @@ class AudioEncoder {
...
+      int frameRate,
+      std::string_view fileName,
+      const VideoStreamOptions& videoStreamOptions);
+
+  void encode();
+
+ private:
+  void initializeEncoder(const VideoStreamOptions& videoStreamOptions);
+  UniqueAVFrame convertTensorToAVFrame(
+      const torch::Tensor& frameTensor,
```
NicolasHug: nit, for consistency with the rest of the codebase


```suggestion
      const torch::Tensor& frame,
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+
+  flushBuffers();
+
+  status = av_write_trailer(avFormatContext_.get());
+  TORCH_CHECK(
+      status == AVSUCCESS,
+      "Error in av_write_trailer: ",
+      getFFMPEGErrorStringFromErrorCode(status));
+
+  // close_avio();
```
NicolasHug: Maybe this should be put back? Or add a TODO?

src/torchcodec/_core/Encoder.h
```diff
@@ -57,6 +57,46 @@ class AudioEncoder {
...
+  void flushBuffers();
+  void close_avio();
+
+  UniqueEncodingAVFormatContext avFormatContext_;
+  UniqueAVCodecContext avCodecContext_;
+  int streamIndex_;
+  UniqueSwsContext swsContext_;
+
+  const torch::Tensor frames_;
+  int frameRate_;
```
NicolasHug: Nit: I suspect this will soon have to be renamed to `inFrameRate_`

test/test_ops.py
```diff
@@ -1224,5 +1226,30 @@ def test_bad_input(self, tmp_path):
...
+        decoder = create_from_file(str(file_path), seek_mode="approximate")
+        add_video_stream(decoder)
+        frames, *_ = get_frames_in_range(decoder, start=0, stop=60)
+        return frames
+
+    @pytest.mark.parametrize("format", ("mov", "mp4", "avi"))
+    # TODO-VideoEncoder: enable additional formats (mkv, webm)
+    def test_video_encoder_test_round_trip(self, tmp_path, format):
+        asset = NASA_VIDEO
+        encoded_path = str(tmp_path / f"encoder_output.{format}")
```
NicolasHug: Nit: move this line closer to where it's used. Same with `frame_rate`

test/test_ops.py
```diff
@@ -1224,5 +1226,30 @@ def test_bad_input(self, tmp_path):
...
+        source_frames = self.decode(str(asset.path)).data
+        frame_rate = 30  # Frame rate is fixed with num frames decoded
+
+        encode_video_to_file(source_frames, frame_rate, encoded_path)
+
+        # Check that PSNR for decode(encode(samples)) is above 30
+        for s_frame, d_frame in zip(source_frames, self.decode(encoded_path).data):
+            res = psnr(s_frame, d_frame)
+            assert res > 30
+
```
NicolasHug, line 1285: just a quick debugging tip that I find useful sometimes, you can save the expected and the decoded frames side-by-side as pngs with e.g. 

```py

from torchvision.io import write_png
from torchvision.utils import make_grid

def save_image(a, b, name):
    image = make_grid(torch.stack([a, b]), nrow=2).cpu()
    write_png(image, f"results/{name}.png")
```

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -406,6 +408,14 @@ OpsAudioFramesOutput get_frames_by_pts_in_range_audio(
   return makeOpsAudioFramesOutput(result);
 }
 
+void encode_video_to_file(
+    const at::Tensor& frames,
+    int64_t frame_rate,
+    std::string_view file_name) {
+  VideoStreamOptions videoStreamOptions;
+  VideoEncoder(frames, frame_rate, file_name, videoStreamOptions).encode();
```
NicolasHug: The input `frame_rate` is `int64_t` (which is good, because this is what pytorch gives us), and the `VideoEncoder`'s constructor parameter `frameRate` is an `int` (which is good too, because this is what FFmpeg needs).

But then we're implicitly casting this int64 into an int without validating its range. We should call the validation utils that were introduced in https://github.com/pytorch/torchcodec/pull/856/files , I don't think they're in your branch yet, you'll have to pull from `main`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
 
   encodeFrame(autoAVPacket, UniqueAVFrame(nullptr));
 }
+
+namespace {
+
+torch::Tensor validateFrames(const torch::Tensor& frames) {
+  TORCH_CHECK(
+      frames.dtype() == torch::kFloat32 || frames.dtype() == torch::kUInt8,
```
NicolasHug: I don't think we'll want to accept float32, at least not for our first version. We can safely enforce uint8 values here.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+      "frames must have float32 or kUInt8 dtype, got ",
+      frames.dtype());
+  TORCH_CHECK(
+      frames.dim() == 4,
+      "frames must have 4 dimensions (N, H, W, C) or (N, C, H, W), got ",
+      frames.dim());
```
NicolasHug: Let's also enforce that `shape[1] == 3`.
The error message above should slightly change to avoid mentioning (N, H, W, C) which we're enforcing.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+
+  return frames.contiguous();
```
NicolasHug, line 531: It's great that you added this - let's add a TODO to remember that we should investigate whether we can accept channels-last or non-contiguous frames (we can chat offline about what that means :D )

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+}
```
NicolasHug, line 532: All of the validations above are great: we should test for all of them :) ! But this can be a TODO.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  avFormatContext_.reset(avFormatContext);
+
+  status = avio_open(&avFormatContext_->pb, fileName.data(), AVIO_FLAG_WRITE);
+  TORCH_CHECK(
+      status >= 0,
+      "avio_open failed. The destination file is ",
+      fileName,
+      ", make sure it's a valid path? ",
+      getFFMPEGErrorStringFromErrorCode(status));
+
```
NicolasHug, line 574: Let's test both TORCH_CHECK above (file name and file path). This can be a TODO

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  avCodecContext_->bit_rate = desiredBitRate.value_or(0);
+  // TODO-VideoEncoder: Verify that frame_rate and time_base are correct
+  avCodecContext_->time_base = {1, frameRate_};
+  avCodecContext_->framerate = {frameRate_, 1};
+
+  // Store dimension order and input pixel format
+  // TODO-VideoEncoder: Remove assumption that tensor in NCHW format
+  auto sizes = frames_.sizes();
+  inPixelFormat_ =
+      (sizes[1] == 3) ? AV_PIX_FMT_GBRP : AV_PIX_FMT_GBRAP; // Planar
```
NicolasHug: Let's hardcode `AV_PIX_FMT_GBRP` (following suggestion above to assert that sizes[1] == 3). I think `AV_PIX_FMT_GBRAP` relates to alpha channel with transparence, and we don't need to support that.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  inHeight_ = sizes[2];
+  inWidth_ = sizes[3];
+
+  // Use specified dimensions or input dimensions
+  // TODO-VideoEncoder: Allow height and width to be set
+  outWidth_ = videoStreamOptions.width.value_or(inWidth_);
+  outHeight_ = videoStreamOptions.height.value_or(inHeight_);
+
+  // Use YUV420P as default output format
```
NicolasHug, line 609: Add a TODO to enable other pixel formats, I suspect we'll have to

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  TORCH_CHECK(avStream != nullptr, "Couldn't create new stream.");
+  status = avcodec_parameters_from_context(
+      avStream->codecpar, avCodecContext_.get());
+  TORCH_CHECK(
+      status == AVSUCCESS,
+      "avcodec_parameters_from_context failed: ",
+      getFFMPEGErrorStringFromErrorCode(status));
+  streamIndex_ = avStream->index;
+}
+
```
NicolasHug, line 654: Nit but regarding the file structure, it might make sense to put this function (convertTensorToAVFrame) below the public `encode()` one, next to `encodeFrame` since they're called at the same place

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  // Use YUV420P as default output format
+  outPixelFormat_ = AV_PIX_FMT_YUV420P;
+
+  // Configure codec parameters
+  avCodecContext_->codec_id = avCodec->id;
+  avCodecContext_->width = outWidth_;
+  avCodecContext_->height = outHeight_;
+  avCodecContext_->pix_fmt = outPixelFormat_;
+  avCodecContext_->time_base = {1, frameRate_};
```
NicolasHug: This is probably OK for now but I think we should try to investigate and understand what a good time-base should be. I suspect we'll generally want a much higher resolution than this. Our NASA  timebase for stream 3 (v:1 because it's the second video stream) is

```
~/dev/torchcodec (video_encoder*) » ffprobe -v quiet -select_streams v:1 -show_streams test/resources/nasa_13013.mp4 | grep time_base
time_base=1/30000
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+  UniqueAVFrame avFrame(av_frame_alloc());
+  TORCH_CHECK(avFrame != nullptr, "Failed to allocate AVFrame");
+
+  // Set output frame properties
+  avFrame->format = outPixelFormat_;
+  avFrame->width = outWidth_;
+  avFrame->height = outHeight_;
+  avFrame->pts = frameIndex;
+
+  int status = av_frame_get_buffer(avFrame.get(), 32);
```
NicolasHug: Looking at the [docs](https://ffmpeg.org/doxygen/7.0/group__lavu__frame.html#ga6b1acbfa82c79bf7fd78d868572f0ceb), I think it might be best to leave it to 0!

> Required buffer size alignment. If equal to 0, alignment will be chosen automatically for the current CPU. It is highly recommended to pass 0 here unless you know what you are doing.


```suggestion
  int status = av_frame_get_buffer(avFrame.get(), 0);
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+
+  inputFrame->format = inPixelFormat_;
+  inputFrame->width = inWidth_;
+  inputFrame->height = inHeight_;
+
+  uint8_t* tensorData = static_cast<uint8_t*>(frameTensor.data_ptr());
+
+  // TODO-VideoEncoder: Reorder tensor if in NHWC format
+  int channelSize = inHeight_ * inWidth_;
+  // Reorder RGB -> GBR for AV_PIX_FMT_GBRP or AV_PIX_FMT_GBRAP formats
```
NicolasHug: Let's add a TODO to investigate this: ideally, we should be able to pass packed RGB data as RGBRGBRGB... without having to do anything fancy (like GBR). I'm not sure. I don't fully understand this yet.

Also let's remove the ref to AV_PIX_FMT_GBRAP as I don't think we'll want to support that.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -507,4 +507,284 @@ void AudioEncoder::flushBuffers() {
...
+    TORCH_CHECK(
+        status >= 0,
+        "Error receiving packet: ",
+        getFFMPEGErrorStringFromErrorCode(status));
+
+    av_packet_rescale_ts(
+        packet.get(),
+        avCodecContext_->time_base,
+        avFormatContext_->streams[streamIndex_]->time_base);
+    packet->stream_index = streamIndex_;
```
NicolasHug: Can you help me understand why that part is needed?

NicolasHug:
On the Windows test failures with FFmpeg 6  https://github.com/pytorch/torchcodec/actions/runs/17468896896/job/49613138742?pr=866

I think we can safely just skip the test on Windows + FFmpeg 6 and make this is TODO

src/torchcodec/_core/Encoder.cpp
Dan-Flores: I was also a bit confused on the formats, here's why I think this is necessary:
* The video decoder outputs a tensor in packed RGB24 pixel format, which gets converted in `SingleStreamDecoder::maybePermuteHWC2CHW` to NCHW planar format, with the channels in RGB order.
* Rather than convert the planar data to packed data, this code sets the input pixel format to planar GBR format, and reorders the frame's data pointers to the correct channels (this is why my images were green earlier).

src/torchcodec/_core/Encoder.cpp
Dan-Flores: I believe this is leftover from debugging the frame rates in my test. Since we aren't modifying the `frame_rate` or `time_base`, I'll remove this.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Good point, it looks like `mp4` does not support the alpha channel. While some other formats do support it, we can consider adding it at a later time.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -511,4 +511,283 @@ void AudioEncoder::flushBuffers() {
...
+      frames.dtype());
+  TORCH_CHECK(
+      frames.dim() == 4,
+      "frames must have 4 dimensions (N, C, H, W), got ",
+      frames.dim());
+  TORCH_CHECK(
+      frames.sizes()[1] == 3,
+      "frame must have 3 channels (R, G, B), got ",
+      frames.sizes()[1]);
+  // TODO-VideoEncoder: Investigate if non-contiguous frames can be returned
```
NicolasHug: It's about input, not output :)

```suggestion
  // TODO-VideoEncoder: Investigate if non-contiguous frames can be accepted
```

test/test_ops.py
```diff
@@ -1224,5 +1226,62 @@ def test_bad_input(self, tmp_path):
             )
 
 
+class TestVideoEncoderOps:
+
+    def test_bad_input(self, tmp_path):
+        output_file = str(tmp_path / ".mp3")
```
NicolasHug: Probably shouldn't be mp3 for video?

src/torchcodec/_core/Encoder.h
```diff
@@ -121,3 +120,45 @@ class AudioEncoder {
...
+  void encode();
+
+ private:
+  void initializeEncoder(const VideoStreamOptions& videoStreamOptions);
+  UniqueAVFrame convertTensorToAVFrame(
+      const torch::Tensor& frame,
+      int frameIndex);
+  void encodeFrame(AutoAVPacket& autoAVPacket, const UniqueAVFrame& avFrame);
+  void flushBuffers();
+  void close_avio();
```
NicolasHug: this isn't declared but not defined - remove

src/torchcodec/_core/custom_ops.cpp
NicolasHug: I think this is still relevant

src/torchcodec/_core/Encoder.cpp
```diff
@@ -511,4 +511,283 @@ void AudioEncoder::flushBuffers() {
...
+  // Use YUV420P as default output format
+  // TODO-VideoEncoder: Enable other pixel formats
+  outPixelFormat_ = AV_PIX_FMT_YUV420P;
+
+  // Configure codec parameters
+  avCodecContext_->codec_id = avCodec->id;
+  avCodecContext_->width = outWidth_;
+  avCodecContext_->height = outHeight_;
+  avCodecContext_->pix_fmt = outPixelFormat_;
+  avCodecContext_->time_base = {1, inFrameRate_};
```
NicolasHug, line 619: This was already set above

src/torchcodec/_core/Encoder.cpp
```diff
@@ -511,4 +511,283 @@ void AudioEncoder::flushBuffers() {
...
+  inputFrame->height = inHeight_;
+
+  uint8_t* tensorData = static_cast<uint8_t*>(frame.data_ptr());
+
+  // TODO-VideoEncoder: Reorder tensor if in NHWC format
+  int channelSize = inHeight_ * inWidth_;
+  // Reorder RGB -> GBR for AV_PIX_FMT_GBRP format
+  inputFrame->data[0] = tensorData + channelSize;
+  inputFrame->data[1] = tensorData + (2 * channelSize);
+  inputFrame->data[2] = tensorData;
```
NicolasHug: I agree with your assessment on https://github.com/pytorch/torchcodec/pull/866#discussion_r2329098617, but I'm still a bit surprised there's no native for native RGBRGBRGB support in FFmpeg. It might be the case, I haven't checked, but I'd recommend leaving a TODO to investigate further

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Thanks for catching that, fixed.

src/torchcodec/_core/custom_ops.cpp
Dan-Flores: Thanks, I added a call to `validateInt64ToInt` here.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -511,4 +511,283 @@ void AudioEncoder::flushBuffers() {
...
+  inputFrame->format = inPixelFormat_;
+  inputFrame->width = inWidth_;
+  inputFrame->height = inHeight_;
+
+  uint8_t* tensorData = static_cast<uint8_t*>(frame.data_ptr());
+
+  // TODO-VideoEncoder: Reorder tensor if in NHWC format
+  int channelSize = inHeight_ * inWidth_;
+  // Reorder RGB -> GBR for AV_PIX_FMT_GBRP format
+  // TODO-VideoEncoder: Determine if FFmpeg supports RGB input format directly
```
dvrogozh: From https://ffmpeg.org/ffmpeg-codecs.html#libx264_002c-libx264rgb: "the libx264rgb encoder is the same as libx264, except it accepts packed RGB pixel formats as input instead of YUV."

In general, it's encoder specific whether it accepts RGB input or not. You need to check with each encoder separately on that. Note that there are multiple implementations for each codec as soon as you consider HW accelerated ones. The story might be different for each of them.

test/test_ops.py
```diff
@@ -1224,5 +1226,62 @@ def test_bad_input(self, tmp_path):
...
+    def decode(self, file_path) -> torch.Tensor:
+        decoder = create_from_file(str(file_path), seek_mode="approximate")
+        add_video_stream(decoder)
+        frames, *_ = get_frames_in_range(decoder, start=0, stop=60)
+        return frames
+
+    @pytest.mark.parametrize("format", ("mov", "mp4", "avi"))
+    # TODO-VideoEncoder: enable additional formats (mkv, webm)
+    def test_video_encoder_test_round_trip(self, tmp_path, format):
+        asset = NASA_VIDEO
```
NicolasHug, line 1271: Let's have a TODO to use FFmpeg's testsrc2 instead https://www.youtube.com/watch?v=dd7HvxZfIgU

NicolasHug (approved):
Thanks @Dan-Flores !

PR #869, opened 2025-09-04
Refactor context structs to use constructors

scotts:
Because we're creating this in one place, with all fields always available, we should use a constructor. Note that the current code does not yet use `FiltersContext::hwFramesCtx`, so we don't accept it in the constructor yet.


src/torchcodec/_core/FilterGraph.h
```diff
@@ -19,11 +19,24 @@ struct FiltersContext {
   int outputWidth = 0;
   int outputHeight = 0;
   AVPixelFormat outputFormat = AV_PIX_FMT_NONE;
-
   std::string filtergraphStr;
   AVRational timeBase = {0, 0};
   UniqueAVBufferRef hwFramesCtx;
 
+  FiltersContext() = default;
```
dvrogozh, line 26: Should default initializers be removed from the above then?

scotts: Unfortunately, we can't do that as they are default-initialized as members of `CpuDeviceInterface`: https://github.com/pytorch/torchcodec/blob/da3eddaa6746d093b09ad1413a42d7244686c58e/src/torchcodec/_core/CpuDeviceInterface.h#L64-L67
Before we decode anything, they start out empty. We could instead make the members unique pointers, but I think that's more complexity than it's worth.

dvrogozh: Yeah, you are right. But in this case, I just realized that I forgot to add default initializers for `SwsFrameContext` struct. Can you help to fix within this PR?

scotts: @dvrogozh, I believe I added them here: https://github.com/pytorch/torchcodec/pull/869/files#diff-454252ad2b209f4b84958d7a3a01c56e49c2cbc21f3aea12acdaa0cada3e8ee2R52

dvrogozh: You might wish to adjust a format as `AV_PIX_FMT_NONE` defined as `-1` (https://github.com/FFmpeg/FFmpeg/blob/6891314db18c96afa35f5e3765d60f4e257fc665/libavutil/pixfmt.h#L72):
```
AVPixelFormat inputFormat = AV_PIX_FMT_NONE;
```

scotts: @dvrogozh, oh, you mean for the individual fields. Yes, got it.

dvrogozh: Yes. Thank you for the change.

dvrogozh (approved):
Looks good.

PR #875, opened 2025-09-05
Assert `deviceInterface_` exists when it should

NicolasHug:


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1246,7 +1246,11 @@ FrameOutput SingleStreamDecoder::convertAVFrameToFrameOutput(
       formatContext_->streams[activeStreamIndex_]->time_base);
   if (streamInfo.avMediaType == AVMEDIA_TYPE_AUDIO) {
     convertAudioAVFrameToFrameOutputOnCPU(avFrame, frameOutput);
-  } else if (deviceInterface_) {
+  } else {
+    TORCH_CHECK(
+        deviceInterface_ != nullptr,
+        "No device interface available for video decoding. This ",
+        "shouldn't happen, please report.");
```
NicolasHug, line 1253: The new `TORCH_CHECK` is important but just as important is the change of the `else if` into a simple `else`

scotts: Can we also add `TORCH_CHECK(streamInfo.avMediaType == AVMEDIA_TYPE_VIDEO)`? I believe that is also assumed.

NicolasHug: It's true that this is assumed but I think we can skip the check in the function, because this is heavily validated upstream in all callers, e.g. by calling `validateActiveStream(...);`

No strong opinion, I can add the check in a follow-up if you think it's good to validate it here as well.

Issue #526, opened 2025-02-27
Support decoder-native transforms


scotts:
### 🚀 The feature

Decoders usually can do more than just decode. They often also implement transformations on what they decode. For example, see the list of [FFmpeg filters](https://ffmpeg.org/ffmpeg-filters.html), which is the FFmpeg term for a transformation. Several users have requested a way to apply these decoder-native transformations.

### Motivation, pitch

The main motivation is performance. We have benchmarks which show, for example, that it is much faster for FFmpeg to resize a frame while decoding than it is to pass a decoded frame to a transformation that lives outside of the decoder.

The *counter pitch* is that exposing such a functionality can be potentially dangerous for users. For example, there are many different ways one can resize a video frame, and it's important that users know which one they are using. We have observed model performance problems with resize specifically: users use a different resize algorithm during inference than they used for training, causing problems that are difficult to detect. Any design for this feature should include mitigations for this problem.

donthomasitos:
Sounds great!

scotts:
Design ideas for this feature, following the principles that:

1. Users *specify* what they want, and we translate that to what FFmpeg understands. We never just pass a string directly from users to FFmpeg.
2. We can use testing to enforce and quantify closeness of transforms to those already in the PyTorch ecosystem.

## API
We create the module `torchcodec.decoders.transform_specs`. Let's use cropping and resizing as an illustration of what could be in this module:
```
@dataclass
class Crop:
    width: int
    height: int
    x:  int
    y:  int

    def _ffmpeg_str() -> str:
        """Logic which knows how to correctly turn the 
           user spec into the exact string FFmpeg needs.
        """
        pass

@dataclass
class Resize:
    width: int
    height: int

    def _ffmpeg_str() -> str:
        pass
```
The actual specifications that FFmpeg accepts are much richer than this, and are in fact their own expression language per transformation. See, for example, the documentation for the [crop filter](https://ffmpeg.org/ffmpeg-filters.html#crop). We would only expose subsets of this functionality as needed.

Users could then pass a sequence of transformations to a decoder at initialization:
```
decoder = VideoDecoder(
    "vid.mp4",
    transforms=[
        Resize(width=640, height=480),
        Crop(width=32, height=32, x=0, y=0),
    ]
)
```
In the actual implementation, we would call the `_ffmpeg_str()` method on each transformation when setting up FFmpeg.

## The mitigation
[...]

PR #885, opened 2025-09-08
[RFC] Decoder-native transforms design

scotts:
Sketching a design and implementation plan for #526.

### DIFF
--- a/decoder_native_transforms.md
+++ b/decoder_native_transforms.md
@@ -22,7 +22,7 @@ We want to support this user-facing API:
 
 What the user is asking for, in English:
 
-1. I want to decode frames from the file `"vid.mp4".`
+1. I want to decode frames from the file `"vid.mp4"`.
 2. For each decoded frame, I want each frame to pass through the following transforms:
    1. Add or remove frames as necessary to ensure a constant 30 frames per second.
    2. Resize the frame to 640x480. Use the algorithm that is TorchVision's default.


decoder_native_transforms.md
```diff
@@ -0,0 +1,103 @@
...
+ ```python
+    decoder = VideoDecoder(
+        "vid.mp4",
+        transforms=[
+            torchcodec.transforms.FPS(
+                fps=30,
+            ),
+            torchvision.transforms.v2.Resize(
+                width=640,
+                height=480,
```
NicolasHug: Nit: this isn't critical for the purpose of this doc, but `width` and `height` aren't supported by torchvision's resize. This is the `size` parameter. Same below for RandomCrop.

decoder_native_transforms.md
```diff
@@ -0,0 +1,103 @@
...
+1. We define a new module, `torchcodec.decoders.transforms`.
+2. All transforms we define in there inherit from
+   `torchvision.transforms.v2.Transform`.
+3. We implement the mimimum needed to hook the new transforms into the
+   machinery defined above.
+
+## Open Questions:
+
+1. Is `torchcodec.transforms` the right namespace?
+2. For random transforms, when should the value be fixed?
```
NicolasHug, line 92: It will be probably be transforms dependent, but I agree with your call that in general this will be per-video. I think that ideally this would be "per-scene", but that's not in scope.

decoder_native_transforms.md
```diff
@@ -0,0 +1,103 @@
...
+3. Transforms such as Resize don't actually implement a `make_params()`
+   method. How does TorchVision get their parameters? How will TorchCodec?
```
NicolasHug, line 94: `make_params()` is mostly relevant for random transforms, either those that are randomly applied, or those that do random sampling. Resize isn't random. The relevant parameters are just stored as public attributes on the Transform object.

decoder_native_transforms.md
```diff
@@ -0,0 +1,103 @@
...
+1. We add an optional `transforms` parameter to the initialization of
+   `VideoDecoder`. It is a sequence of TorchVision Transforms.
+2. During `VideoDecoder` object creation, we walk the list, capturing two
+   pieces of information:
+   1. The transform name that the C++ layer will understand. (We will
+      have to decide if we want to just use the FFmpeg filter name
+      here, the fully resolved Transform name, or introduce a new
+      naming layer.)
+   2. The parameters in a format that the C++ layer will understand. We
+      obtain them by calling `make_params()` on the Transform object.
```
NicolasHug, line 60: Three things come to mind:

- For non-random transforms `make_transforms()` doesn't exist, the relevant attributes are just attributes on the Transform object (mentioned below as well). That's not an issue, we can easily work around that.
- `make_params()` in its [current state](https://github.com/pytorch/vision/blob/dcd1e4213394883b688d75df110839cede1193f4/torchvision/transforms/v2/_geometry.py#L842) requires the input frames to be passed (`flat_inputs`). Those are expected to be materialized tensors, which we definitely won't have by the time we need to call `make_params()` in TorchCodec. Most of the time, all of what's actually needed from `flat_inputs` are the H and W of the frames. So we should be able to get around that by slightly modifying the torchvision part, and accept H and W instead of `flat_input` as input to `make_params()` (or similar new hook). We'll probably want to keep this private for the time being.
- the logic in `make_param()` is often dependent on the H and W of the input frames, as described above. For most streams that's not an issue because H and W are constant. For the streams with variable resolution, that quickly becomes **very** hairy. Do we call `make_params()` again on the new H and W? Where? We're risking sampling new random values that may lead to nonsensical transforms for a given video!

decoder_native_transforms.md
```diff
@@ -0,0 +1,103 @@
...
+   parameter will be a vector, but whether the vector contains strings,
+   tensors, or some combination of them is TBD.
+4. The `custom_ops.cpp` and `pybind_ops.cpp` layer is responsible for turning
+   the values passed from the Python layer into transform objects that the
+   C++ layer knows about. We will have one class per transform we support.
+   Each class will have:
+   1. A name which matches the FFmpeg filter name.
+   2. One member for each supported parameter.
+   3. A virtual member function that knows how to produce a string that
+      can be passed to FFmpeg's filtergraph.
```
NicolasHug, line 71: Sounds reasonable to stick to filtergraph for this, but I wonder how this affect our current use of libswscale for color-conversion? Maybe we'll have to declare that any use of the `transforms` parameter means filtergraph is used for color-conversion, despite `libswscale` being faster according to our past benchmarks?

decoder_native_transforms.md
scotts: @NicolasHug, on the three things:

* We can work around it, but it does mean we'll probably need to do a lot of ` if isinstance(t, Resize)`, as that means there is no uniform way to discover a transforms parameters. We'll need to know that for transform X it has params Y and Z.
* Noted. I think we can start the implementation with `Resize`, and punt on these issues until we do `RandomCrop`.
* FFmpeg filters such as `crop` actually accept expressions, not just number values: https://ffmpeg.org/ffmpeg-filters.html#crop. Some are evaluated per-frame, so we may be able to use those to deal with variable resolution videos. For first versions, I also think it's reasonable to say variable resolution videos are not fully supported.

decoder_native_transforms.md
scotts: I think we can do a bit of both. Some transforms (such as `Resize`) we can support on libswscale. I assume there's a lot of more involved transforms that are not going to work on libswscale. We can add a virtual member function to this C++ class that returns a libswscale string, if possible. If all transforms are supportable on libswscale and all other its other criteria are met, we use that. If not, we have to use filtergraph.

We can also start with just filtergraph, and do libswscale as an optimization.

decoder_native_transforms.md
```diff
@@ -0,0 +1,101 @@
...
+   transforms. They are all related, but it's not clear to me what the
+   exact relationship should be.
+7. The actual string we pass to FFmepg's filtergraph comes from calling
+   the virtual member function on each transform object.
+
+For the transforms that do not exist in TorchVision, we can build on the above:
+
+1. We define a new module, `torchcodec.decoders.transforms`.
+2. All transforms we define in there inherit from
+   `torchvision.transforms.v2.Transform`.
```
Dan-Flores, line 85: This is a good solution for Design Consideration 3. Do you foresee any issues from adding a dependency on torchvision in torchcodec?

scotts: Good call out. I'd like to explore if we can avoid an explicit dependency and instead go a duck-typing route. We absolutely want to _accept_ TorchVision transforms. But it would be great if we could avoid an explicit dependency from TorchCodec to TorchVision.

scotts:
Closing this issue as this design is now reflected in the code.

PR #887, opened 2025-09-10
Add custom_frame_mappings tutorial and optimizations

Dan-Flores:
This PR adds the code that will become the `custom_frame_mappings` tutorial. 
Currently, 2 benchmarks are run using a short video (~3 minutes) and a long video (~13 minutes) :
* Initializing VideoDecoders 
* Decoding 10 frames from 10 videos

Additionally, some fixes / optimizations were added to `custom_frame_mappings` logic in `SingleStreamDecoder.cpp` to improve the performance.

The page takes 1 minute 22 seconds to execute.

Benchmark results: 
<details> 

#### Compare performance of initializing VideoDecoder with custom_frame_mappings vs seek_modes

Running benchmarks on short_video.mp4
Creating a VideoDecoder object with custom_frame_mappings:
med = 7.68ms +- 20.83
Creating a VideoDecoder object with seek_mode='exact':
med = 17.26ms +- 14.54

Running benchmarks on long_video.mp4
Creating a VideoDecoder object with custom_frame_mappings:
med = 23.70ms +- 3.39
Creating a VideoDecoder object with seek_mode='exact':
med = 60.47ms +- 7.30

#### Decode frames with custom_frame_mappings vs exact seek_mode

Running benchmarks on short_video.mp4
Decoding frames with custom_frame_mappings:
med = 34.21ms +- 11.42
Decoding frames with seek_mode='exact':
med = 39.69ms +- 3.37

Running benchmarks on long_video.mp4
Decoding frames with custom_frame_mappings:
med = 50.53ms +- 8.38
Decoding frames with seek_mode='exact':
med = 83.95ms +- 17.38

</details>

### DIFF
--- a/examples/decoding/custom_frame_mappings.py
+++ b/examples/decoding/custom_frame_mappings.py
@@ -24,7 +24,7 @@ This makes it ideal for workflows where:
 # %%
 # First, some boilerplate: we'll download a short video from the web, and
 # use ffmpeg to create a longer version by repeating it multiple times. We'll end up
-# with two videos: a short one of approximately 14 seconds and a long one of about 14 minutes.
+# with two videos: a short one of approximately 14 seconds and a long one of about 12 minutes.
 # You can ignore this part and skip below to :ref:`frame_mappings_creation`.
 
 import tempfile
@@ -48,7 +48,7 @@ with open(short_video_path, 'wb') as f:
 long_video_path = Path(temp_dir) / "long_video.mp4"
 ffmpeg_command = [
     "ffmpeg",
-    "-stream_loop", "80",  # repeat video 80 times to get a ~18 min video
+    "-stream_loop", "50",  # repeat video 50 times to get a ~12 min video
     "-i", f"{short_video_path}",
     "-c", "copy",
     f"{long_video_path}"
@@ -83,7 +83,7 @@ import json
 # Lets define a simple function to run ffprobe on a video's first stream index, then writes the results in output_json_path.
 def generate_frame_mappings(video_path, output_json_path, stream_index):
     ffprobe_cmd = ["ffprobe", "-i", f"{video_path}", "-select_streams", f"{stream_index}", "-show_frames", "-show_entries", "frame=pts,duration,key_frame", "-of", "json"]
-    print(f"Running ffprobe:\n{' '.join(ffprobe_cmd)}")
+    print(f"Running ffprobe:\n{' '.join(ffprobe_cmd)}\n")
     ffprobe_result = subprocess.run(ffprobe_cmd, check=True, capture_output=True, text=True)
     with open(output_json_path, "w") as f:
         f.write(ffprobe_result.stdout)
--- a/examples/decoding/custom_frame_mappings.py
+++ b/examples/decoding/custom_frame_mappings.py
@@ -24,7 +24,7 @@ This makes it ideal for workflows where:
 # %%
 # First, some boilerplate: we'll download a short video from the web, and
 # use ffmpeg to create a longer version by repeating it multiple times. We'll end up
-# with two videos: a short one of approximately 3 minutes and a long one of about 13 minutes.
+# with two videos: a short one of approximately 14 seconds and a long one of about 14 minutes.
 # You can ignore this part and skip below to :ref:`frame_mappings_creation`.
 
 import tempfile
@@ -32,7 +32,9 @@ from pathlib import Path
 import subprocess
 import requests
 
-url = "https://download.pytorch.org/torchaudio/tutorial-assets/stream-api/NASAs_Most_Scientifically_Complex_Space_Observatory_Requires_Precision-MP4_small.mp4"
+# Video source: https://www.pexels.com/video/dog-eating-854132/
+# License: CC0. Author: Coverr.
+url = "https://videos.pexels.com/video-files/854132/854132-sd_640_360_25fps.mp4"
 response = requests.get(url, headers={"User-Agent": ""})
 if response.status_code != 200:
     raise RuntimeError(f"Failed to download video. {response.status_code = }.")
@@ -46,7 +48,7 @@ with open(short_video_path, 'wb') as f:
 long_video_path = Path(temp_dir) / "long_video.mp4"
 ffmpeg_command = [
     "ffmpeg",
-    "-stream_loop", "3",  # repeat video 3 times to get a ~13 min video
+    "-stream_loop", "80",  # repeat video 80 times to get a ~18 min video
     "-i", f"{short_video_path}",
     "-c", "copy",
     f"{long_video_path}"
@@ -77,24 +79,27 @@ import tempfile
 from time import perf_counter_ns
 import json
 
+
+# Lets define a simple function to run ffprobe on a video's first stream index, then writes the results in output_json_path.
+def generate_frame_mappings(video_path, output_json_path, stream_index):
+    ffprobe_cmd = ["ffprobe", "-i", f"{video_path}", "-select_streams", f"{stream_index}", "-show_frames", "-show_entries", "frame=pts,duration,key_frame", "-of", "json"]
+    print(f"Running ffprobe:\n{' '.join(ffprobe_cmd)}")
+    ffprobe_result = subprocess.run(ffprobe_cmd, check=True, capture_output=True, text=True)
+    with open(output_json_path, "w") as f:
+        f.write(ffprobe_result.stdout)
+
+
 stream_index = 0
 long_json_path = Path(temp_dir) / "long_custom_frame_mappings.json"
 short_json_path = Path(temp_dir) / "short_custom_frame_mappings.json"
 
-ffprobe_cmd = ["ffprobe", "-i", f"{long_video_path}", "-select_streams", f"{stream_index}", "-show_frames", "-show_entries", "frame=pts,duration,key_frame", "-of", "json"]
-ffprobe_result = subprocess.run(ffprobe_cmd, check=True, capture_output=True, text=True)
-with open(long_json_path, "w") as f:
-    f.write(ffprobe_result.stdout)
-
-ffprobe_cmd = ["ffprobe", "-i", f"{short_video_path}", "-select_streams", f"{stream_index}", "-show_frames", "-show_entries", "frame=pts,duration,key_frame", "-of", "json"]
-ffprobe_result = subprocess.run(ffprobe_cmd, check=True, capture_output=True, text=True)
-with open(short_json_path, "w") as f:
-    f.write(ffprobe_result.stdout)
-
-sample_data = json.loads(ffprobe_result.stdout)
-print("Data structure of custom frame mappings:")
+generate_frame_mappings(long_video_path, long_json_path, stream_index)
+generate_frame_mappings(short_video_path, short_json_path, stream_index)
+with open(short_json_path) as f:
+    sample_data = json.loads(f.read())
+print("Sample of fields in custom frame mappings:")
 for frame in sample_data["frames"][:3]:
-    print(f"{frame}")
+    print(f"{frame['key_frame'] = }, {frame['pts'] = }, {frame['duration'] = }")
 
 # %%
 # .. _custom_frame_mappings_perf_creation:
@@ -103,12 +108,13 @@ for frame in sample_data["frames"][:3]:
 # --------------------------------------
 #
 # Custom frame mappings affect the **creation** of a :class:`~torchcodec.decoders.VideoDecoder`
-# object. As video length increases, the performance gain compared to exact mode increases.
+# object. As video length or resolution increases, the performance gain compared to exact mode increases.
 #
 
 import torch
 
 
+# Here, we define a benchmarking function, with the option to seek to the start of a file_like.
 def bench(f, file_like=False, average_over=50, warmup=2, **f_kwargs):
     for _ in range(warmup):
         f(**f_kwargs)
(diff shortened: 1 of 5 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -501,7 +511,7 @@ void SingleStreamDecoder::addVideoStream(
         customFrameMappings.has_value(),
         "Missing frame mappings when custom_frame_mappings seek mode is set.");
     readCustomFrameMappingsUpdateMetadataAndIndex(
-        streamIndex, customFrameMappings.value());
+        activeStreamIndex_, customFrameMappings.value());
```
Dan-Flores, line 519: This resolves a bug that raised `Runtime Error: bad_optional_access`.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -322,19 +322,30 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
       "all_frames, is_key_frame, and duration from custom_frame_mappings were not same size.");
 
+  // Allocate vectors using num frames to reduce reallocations
+  int64_t numFrames = all_frames.size(0);
+  streamInfos_[streamIndex].allFrames.reserve(numFrames);
+  streamInfos_[streamIndex].keyFrames.reserve(numFrames);
+  // Access tensor data directly rather than element wise to speed up loop below
+  auto* pts_data = all_frames.data_ptr<int64_t>();
+  auto* is_key_frame_data = is_key_frame.data_ptr<bool>();
+  auto* duration_data = duration.data_ptr<int64_t>();
```
NicolasHug: Ah, great catch. Instead of using raw pointers, let's use accessors instead, they're safer https://docs.pytorch.org/cppdocs/notes/tensor_basics.html#cpu-accessors

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -322,19 +322,30 @@ void SingleStreamDecoder::scanFileAndUpdateMetadataAndIndex() {
...
     int streamIndex,
     FrameMappings customFrameMappings) {
-  auto& all_frames = customFrameMappings.all_frames;
-  auto& is_key_frame = customFrameMappings.is_key_frame;
-  auto& duration = customFrameMappings.duration;
+  const torch::Tensor& all_frames =
+      customFrameMappings.all_frames.to(torch::kLong);
+  const torch::Tensor& is_key_frame =
+      customFrameMappings.is_key_frame.to(torch::kBool);
+  const torch::Tensor& duration = customFrameMappings.duration.to(torch::kLong);
```
NicolasHug, line 334: There's a logic mismatch between:

- what we're doing here, i.e. converting the `all_frames` tensor to Long
- and what we're doing earlier in `_read_custom_frame_mappings` where we convert the int values to float: https://github.com/pytorch/torchcodec/blob/67b4381e4c9b695f8d3c8319539791aebc150ef3/src/torchcodec/decoders/_video_decoder.py#L478-L481


So effectively what we're doing is that we're converting int to float in `_read_custom_frame_mappings` and then convert these floats back to int (actually: long!) here.

I think we should:

- force `_read_custom_frame_mappings`  to return `int64` dtype tensors for  all_frames and  duration.  is_key_frame can stay as a bool
- add a test for the dtype
- TORCH_CHECK the expected dtype here in `SingleStreamDecoder`. Obviously

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: That's surprising, `streamIndex` isn't an optional. Do you remember what was causing the issue? Should we add a test?

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,179 @@
...
+
+# %%
+# Compare performance of initializing VideoDecoder with custom_frame_mappings vs exact seek_mode
+
+
+for video_path, json_path in ((short_video_path, short_json_path), (long_video_path, long_json_path)):
+    print(f"Running benchmarks on {Path(video_path).name}")
+    print("Creating a VideoDecoder object with custom_frame_mappings JSON str from file:")
+    with open(json_path, "r") as f:
+        bench(VideoDecoder, source=video_path, stream_index=stream_index, custom_frame_mappings=(f.read()))
```
NicolasHug: I think we can simplify the benchmark a bit and only benchmark passing `f` instead of `f.read()`

Ultimately the main difference between these 2 will be the speed of the file system, I don't feel like it's super relevant in this specific tutorial which should be a fairly straightforward entry point

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,179 @@
...
+    print("Decoding frames with custom_frame_mappings JSON str from file:")
+    with open(json_path, "r") as f:
+        bench(decode_frames_from_n_videos, video_path=video_path, custom_frame_mappings=(f.read()))
+
+    print("Decoding frames with seek_mode='exact':")
+    bench(decode_frames_from_n_videos, video_path=video_path, seek_mode="exact")
+
+# %%
+# Compare performance of sampling clips from multiple videos with custom_frame_mappings vs exact seek_mode
+
```
NicolasHug: Just a suggestion, we might get away with dropping the sampling part for this tutorial, and just ref to that of the seek_mode tutorial. Ultimately the point of this section was to show that an *instanciation* speedup also leads to a sampling speed-up, but the point is already made in the other tuto, so maybe a link and a mention is enough (and would save us some CI time!)

Dan-Flores: Good point, I've replaced this section with a simple frame accuracy comparison.

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: Thank you for catching this! I've added a test and made the typing consistent. Hopefully this PR is not too large to review.

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: I suspect the issue was not with `streamIndex` variable itself, but accessing an unset value in `StreamInfo[streamIndex]` on videos with multiple streams. I can spend some time digging into this.

Dan-Flores: In the Python constructor `streamIndex` is optional. It defaults to `-1` in `custom_ops.cpp`'s `_add_video_stream`. 

`addStream` passes this value to `av_find_best_stream` as either a real stream index if provided, or as `-1`, indicating FFmpeg should choose.

So at this point in `addVideoStream`, `activeStreamIndex_` is the stream index we should be accessing, since we have already determined and stored the best stream index.

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,218 @@
...
+:class:`~torchcodec.decoders.VideoDecoder` class.
+This parameter allows you to provide pre-computed frame mapping information to
+speed up :class:`~torchcodec.decoders.VideoDecoder` instantiation, while
+maintaining the frame seeking accuracy of ``seek_mode="exact"``.
+
+This makes it ideal for workflows where:
+
+    1. Frame accuracy is critical, so :doc:`approximate mode <approximate_mode>` cannot be used
+    2. Videos can be preprocessed once and then decoded many times
+"""
```
NicolasHug, line 22: **Great** intro 👍 ! It's immediately clear to the reader whether the tutorial is relevant to them

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,218 @@
...
+    print("Creating a VideoDecoder object with seek_mode='exact':")
+    bench(VideoDecoder, source=video_path, stream_index=stream_index, seek_mode="exact")
+
+# %%
+# Performance: Frame decoding with custom frame mappings
+# ------------------------------------------------------
+#
+# Although using custom_frame_mappings only impacts the initialization speed of
+# :class:`~torchcodec.decoders.VideoDecoder`, decoding workflows
+# usually involve creating a :class:`~torchcodec.decoders.VideoDecoder` instance,
```
NicolasHug:
```suggestion
# :class:`~torchcodec.decoders.VideoDecoder`, decoding workflows
# involve creating a :class:`~torchcodec.decoders.VideoDecoder` instance,
```

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,218 @@
...
+
+ffprobe_cmd = ["ffprobe", "-i", f"{short_video_path}", "-select_streams", f"{stream_index}", "-show_frames", "-show_entries", "frame=pts,duration,key_frame", "-of", "json"]
+ffprobe_result = subprocess.run(ffprobe_cmd, check=True, capture_output=True, text=True)
+with open(short_json_path, "w") as f:
+    f.write(ffprobe_result.stdout)
+
+sample_data = json.loads(ffprobe_result.stdout)
+print("Data structure of custom frame mappings:")
+for frame in sample_data["frames"][:3]:
+    print(f"{frame}")
```
NicolasHug: On the above: consider printing the whole `ffprobe_cmd` so that the reader can see exactly what is being executed, and so they can copy/paste it.

If it's worth it, we could also refactor the ffprobe_cmd creation, subprocess.run and `f.write()` into a function that we call twice? No strong opinion.

-----

QQ what's the `side_data_list` on the output?

```
Data structure of custom frame mappings:
{'key_frame': 1, 'pts': 0, 'duration': 1001, 'side_data_list': [{'side_data_type': 'H.26[45] User Data Unregistered SEI message'}]}
{'key_frame': 0, 'pts': 1001, 'duration': 1001}
{'key_frame': 0, 'pts': 2002, 'duration': 1001}
```

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,218 @@
...
+
+    # Compare against exact seek_mode
+    print("Creating a VideoDecoder object with seek_mode='exact':")
+    bench(VideoDecoder, source=video_path, stream_index=stream_index, seek_mode="exact")
+
+# %%
+# Performance: Frame decoding with custom frame mappings
+# ------------------------------------------------------
+#
+# Although using custom_frame_mappings only impacts the initialization speed of
```
NicolasHug:
```suggestion
# Although using ``custom_frame_mappings`` only impacts the initialization speed of
```

examples/decoding/custom_frame_mappings.py
```diff
@@ -0,0 +1,218 @@
...
+# use ffmpeg to create a longer version by repeating it multiple times. We'll end up
+# with two videos: a short one of approximately 3 minutes and a long one of about 13 minutes.
+# You can ignore this part and skip below to :ref:`frame_mappings_creation`.
+
+import tempfile
+from pathlib import Path
+import subprocess
+import requests
+
+url = "https://download.pytorch.org/torchaudio/tutorial-assets/stream-api/NASAs_Most_Scientifically_Complex_Space_Observatory_Requires_Precision-MP4_small.mp4"
```
NicolasHug: QQ Do you think we could be using the same video as in the seek_mode tutorial( [url](https://docs.python.org/3/library/stdtypes.html#str) = "https://videos.pexels.com/video-files/854132/854132-sd_640_360_25fps.mp4")? 

I could be wrong but I suspect this is where the difference in execution time comes from: 2 minutes for this tutorial vs 30 seconds for seek_mode.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -476,10 +476,12 @@ def _read_custom_frame_mappings(
...
 
     frame_data = [
-        (float(frame[pts_key]), frame["key_frame"], float(frame[duration_key]))
+        (int(frame[pts_key]), frame["key_frame"], int(frame[duration_key]))
         for frame in input_data["frames"]
     ]
-    all_frames, is_key_frame, duration = map(torch.tensor, zip(*frame_data))
+    all_frames = torch.tensor([x[0] for x in frame_data], dtype=torch.int64)
+    is_key_frame = torch.tensor([x[1] for x in frame_data], dtype=torch.bool)
+    duration = torch.tensor([x[2] for x in frame_data], dtype=torch.int64)
```
NicolasHug: At this point, it's not worth creating `frame_data` anymore, since the `x[...]` indexing becomes un-pythonic. I think it'd be preferable to just define `all_frames` and the other tensors as e.g.:

```py
all_frames = torch.tensor([int(frame[pts_key]) for frame in input_data["frames"]], dtype=torch.int64)
...
```

src/torchcodec/decoders/_video_decoder.py
NicolasHug, line 1: Above in this file, in docstring description of `custom_frames_mappings`, let's add a link to the new tutorial, similarly to what was done for the `seek_mode` docstring.

NicolasHug (approved):
Thanks a lot for the great tutorial @Dan-Flores !

I made some small suggestions below, but nothing is blocking so I'll approve.

examples/decoding/custom_frame_mappings.py
Dan-Flores: Agreed on printing the `ffprobe_cmd` and refactoring into a reusable function, thanks for the suggestions!

The `side_data_list` is a field returned when accessing `-show_entries`, even if it was not a specified key. I believe its additional metadata related to the codec? 

Since it is not important (and I am unable to remove it within the ffprobe command), I'll update the print to only show the fields that we use.

examples/decoding/custom_frame_mappings.py
Dan-Flores: Yes, using the same video brought the execution time down to 23 seconds. I used a different resource with higher resolution to emphasize the performance gains, but I'll use the same video to reduce the docs execution time.

PR #889, opened 2025-09-10
Refactor pybind_ops to only deal with file like context holders

scotts:


src/torchcodec/_core/AVIOFileLikeContext.h
```diff
@@ -17,7 +17,8 @@ namespace facebook::torchcodec {
 
 // Enables uers to pass in a Python file-like object. We then forward all read
 // and seek calls back up to the methods on the Python object.
-class AVIOFileLikeContext : public AVIOContextHolder {
+class __attribute__((visibility("hidden"))) AVIOFileLikeContext
+    : public AVIOContextHolder {
```
scotts: I don't love that we need this again, but that seems to be the case. This is, I think, because now the `custom_ops.cpp` side needs to see it.

src/torchcodec/_core/AVIOFileLikeContext.h
```diff
@@ -11,13 +11,27 @@
 
 #include "src/torchcodec/_core/AVIOContextHolder.h"
 
+// pybind11 has specific visibility rules; see:
+//  https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes
+// This file is included in both pybind_ops.cpp and custom_ops.cpp. We compile
+// pybind_ops.cpp with the global visibility flag as required. On the
+// custom_ops.cpp side we don't need to do that. However, we do need to ensure
+// that this class is also seen as the same visibility.
```
NicolasHug: This is fine, but if we can, we might as well set the hidden visibility flag for the whole pybind_ops lib and remove the `VISIBILITY_HIDDEN` stuff:

```
        target_compile_options(
            ${custom_ops_library_name}
            PUBLIC
            "-fvisibility=hidden"
        )
```

This seems to work OK locally

src/torchcodec/_core/pybind_ops.cpp
```diff
@@ -26,62 +21,18 @@ namespace facebook::torchcodec {
...
-      std::make_unique<AVIOFileLikeContext>(file_like, /*isForWriting=*/true);
-
-  AudioEncoder encoder(
-      samples,
-      validateInt64ToInt(sample_rate, "sample_rate"),
-      format,
-      std::move(avioContextHolder),
-      audioStreamOptions);
-  encoder.encode();
+int64_t create_file_like_context(py::object file_like, bool is_for_writing) {
```
NicolasHug, line 36: The comment above was meant to be about "laundering into into tensors" because it was about the decoder object itself. Now we're laundering an int into an `AVIOFileLikeContext` object IIUC, not a tensor, so the comment may need to be updated.

Separately from the comment, and since we shouldn't need the intermediate tensor representation, I wonder if we could actually work around all that by "cleanly" binding the `AVIOFileLikeContext` to Python and just pass normal py::objects? I thought the whole laundering problem was due to a problem between pybind and torch tensors, but IIUC we don't need the tensor part at all anymore.

scotts: I believe py::objects need to be existing kinds of objects Python already knows about. I thought about what you're suggesting, and I think that would mean actually defining a new kind of Python to C++ object through pybind. We can do it, but it will take some work to figure out if it's worth it.

src/torchcodec/_core/AVIOFileLikeContext.h
NicolasHug: Here's what I did, I can confirm this compiles locally


```diff
diff --git a/src/torchcodec/_core/AVIOFileLikeContext.h b/src/torchcodec/_core/AVIOFileLikeContext.h
index 3146c00..fd7f534 100644
--- a/src/torchcodec/_core/AVIOFileLikeContext.h
+++ b/src/torchcodec/_core/AVIOFileLikeContext.h
@@ -11,27 +11,13 @@
 
 #include "src/torchcodec/_core/AVIOContextHolder.h"
 
-// pybind11 has specific visibility rules; see:
-//  https://pybind11.readthedocs.io/en/stable/faq.html#someclass-declared-with-greater-visibility-than-the-type-of-its-field-someclass-member-wattributes
-// This file is included in both pybind_ops.cpp and custom_ops.cpp. We compile
-// pybind_ops.cpp with the global visibility flag as required. On the
-// custom_ops.cpp side we don't need to do that. However, we do need to ensure
-// that this class is also seen as the same visibility.
-//
-// This only matters on Linux and Mac; on Windows we don't need to do anything.
-#ifdef _WIN32
-#define VISIBILITY_HIDDEN
-#else
-#define VISIBILITY_HIDDEN __attribute__((visibility("hidden")))
-#endif
-
 namespace py = pybind11;
 
 namespace facebook::torchcodec {
 
 // Enables uers to pass in a Python file-like object. We then forward all read
 // and seek calls back up to the methods on the Python object.
-class VISIBILITY_HIDDEN AVIOFileLikeContext : public AVIOContextHolder {
+class AVIOFileLikeContext : public AVIOContextHolder {
  public:
   explicit AVIOFileLikeContext(const py::object& fileLike, bool isForWriting);
 
diff --git a/src/torchcodec/_core/CMakeLists.txt b/src/torchcodec/_core/CMakeLists.txt
index 03f68f6..c45a614 100644
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -137,6 +137,14 @@ function(make_torchcodec_libraries
         "${custom_ops_dependencies}"
     )
 
+    if(NOT WIN32)
+        target_compile_options(
+            ${custom_ops_library_name}
+            PUBLIC
+            "-fvisibility=hidden"
+        )
+    endif()
+
     # 3. Create libtorchcodec_pybind_opsN.so.
     set(pybind_ops_library_name "libtorchcodec_pybind_ops${ffmpeg_major_version}")
     set(pybind_ops_sources
```

scotts: @NicolasHug, yeah, I did the same. The one difference is that I put both of them under the same `if(NOT WIN32)` and explaining comment. It breaks the convention of only doing one library at a time, but I think it helps understanding. All of the visibility stuff, with the same cause, are all in one place.

src/torchcodec/_core/pybind_ops.cpp
scotts: I improved the comment, referenced and created #896

PR #890, opened 2025-09-10
Update Mac CI jobs to macos-m2-15

Dan-Flores:


NicolasHug:
@Dan-Flores  I think we can close this one :) ?

PR #891, opened 2025-09-10
Update runner type to macos-14 in workflow

ZainRizvi:
Removing dependency on self-hosted macOS runners, in prep for migration to meta-pytorch org

### DIFF
--- a/.github/workflows/macos_wheel.yaml
+++ b/.github/workflows/macos_wheel.yaml
@@ -52,7 +52,7 @@ jobs:
       pre-script: packaging/pre_build_script.sh
       post-script: packaging/post_build_script.sh
       smoke-test-script: packaging/fake_smoke_test.py
-      runner-type: macos-m1-stable
+      runner-type: macos-14
       package-name: torchcodec
       trigger-event: ${{ github.event_name }}
       build-platform: "python-build-package"


Dan-Flores:
Thanks for the PR @ZainRizvi! I'll let @scotts review as well, but I had some questions about this change.
* Will this migration affect the availability, support, or performance of the runners?
* Will there be similar changes to the other runners (linux, windows)?

.github/workflows/macos_wheel.yaml
```diff
@@ -52,7 +52,8 @@ jobs:
       pre-script: packaging/pre_build_script.sh
       post-script: packaging/post_build_script.sh
       smoke-test-script: packaging/fake_smoke_test.py
-      runner-type: macos-m1-stable
+      runner-type: macos-14
+      setup-miniconda: true
```
Dan-Flores, line 56: Why is miniconda needed here?

ZainRizvi:
Hey @Dan-Flores this should actually give you better availability for the mac runners, since the jobs will be moving to the github hosted macos runners instead of using the self-hosted ones that we have a fixed supply of and do not autoscale. Most immediately, you'll likely notice that these runners will have much lower queue times.

Linux/windows self-hosted runners are auto-scaled and will not be changed.

.github/workflows/macos_wheel.yaml
ZainRizvi: The pytorch/test-infra/.github/workflows/build_wheels_macos.yml@main workflow expects miniconda to be pre-installed on the host.  All our self-hosted mac runners have miniconda pre-installed on them, so this flag wasn't needed.  The github hosted runners don't have it though, so we pass in this flag to install it.

PR #894, opened 2025-09-11
[WIP] Add support for pixel format parameter

scotts:
Still needs significant testing and explanation. We'll need to commit new reference frames, at the least. We may also need to create new reference videos that are encoded in RGB48.


scotts:
Closing as this is not the way we'll want to support this capability.

PR #895, opened 2025-09-11
Linux `aarch64` wheels

chizkiyahu:
Adding support for `Linux aarch64`

fixed https://github.com/pytorch/torchcodec/issues/569
and help for https://github.com/pytorch/executorch/pull/13946

needs to do before merge 

1. run `.github/workflows/build_ffmpeg.yaml`
2. upload to `s3` and update [fetch_and_expose_non_gpl_ffmpeg_libs.cmake](https://github.com/pytorch/torchcodec/blob/main/src/torchcodec/_core/fetch_and_expose_non_gpl_ffmpeg_libs.cmake)

@NicolasHug
can you help be with this ?

### DIFF
--- a/.github/workflows/build_ffmpeg.yaml
+++ b/.github/workflows/build_ffmpeg.yaml
@@ -48,6 +48,32 @@ jobs:
         mkdir -p "${artifact_dir}"
         mv ffmpeg.tar.gz "${artifact_dir}/${FFMPEG_VERSION}.tar.gz"
 
+  LGPL-Linux-aarch64:
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


NicolasHug:
Hi @Chizkiyahu , thanks for the PR - we're following up on that in https://github.com/pytorch/executorch/pull/13946#issuecomment-3280786860, I think there are faster ways to unblock executorch here

NicolasHug:
I'll close this PR @Chizkiyahu , as I think the original issue was addressed in https://github.com/pytorch/executorch/pull/13946#issuecomment-3280813023

Issue #873, opened 2025-09-05
FFmpeg 7.1 deprecated AVCodec::ch_layouts


scotts:
When compiling against FFmpeg 7.1, we get deprecation warnings because [`AVCodec`](https://ffmpeg.org/doxygen/7.1/structAVCodec.html) has deprecated the field [`ch_layouts`](https://ffmpeg.org/doxygen/7.1/structAVCodec.html#a3aa4a57d42f9b9dfc867fe3fbcad4539). We'll need to deal with it here: https://github.com/pytorch/torchcodec/blob/da3eddaa6746d093b09ad1413a42d7244686c58e/src/torchcodec/_core/FFMPEGCommon.cpp#L111-L133 One wrinkle is that this code is already dealing with FFmpeg > 4 versions switching from `channel_layout` to `ch_layouts`, so this will likely need to become a three-armed `#if`.

PR #898, opened 2025-09-12
Handle AVCodec fields deprecated in FFmpeg 7

Dan-Flores:
This PR resolves the AVCodec fields deprecated in FFmpeg 7, described in #873.
This field is replaced by [avcodec_get_supported_config](https://ffmpeg.org/doxygen/trunk/group__lavc__decoding.html#gadd58e6b0bbca99fdbc547efbaa6b0ef1), which supports retrieval of multiple [AVCodecConfig](https://ffmpeg.org/doxygen/trunk/group__lavc__decoding.html#ga9abe043ed40f3c270dff77235fcfcd0b)

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -33,21 +33,39 @@ torch::Tensor validateSamples(const torch::Tensor& samples) {
 }
 
 void validateSampleRate(const AVCodec& avCodec, int sampleRate) {
-  if (avCodec.supported_samplerates == nullptr) {
+  const int* supportedSampleRates = nullptr;
+#if LIBAVFILTER_VERSION_MAJOR >= 10 // FFmpeg >= 7
+  int numSampleRates = 0;
+  int ret = avcodec_get_supported_config(
+      nullptr,
+      &avCodec,
+      AV_CODEC_CONFIG_SAMPLE_RATE,
+      0,
+      (const void**)&supportedSampleRates,
+      &numSampleRates);
+  if (ret < 0 || supportedSampleRates == nullptr) {
+    // If we can't validate, assume it'll be fine??
+    return;
+  }
+#else
+  supportedSampleRates = avCodec.supported_samplerates;
+#endif
+
+  if (supportedSampleRates == nullptr) {
     return;
   }
 
-  for (auto i = 0; avCodec.supported_samplerates[i] != 0; ++i) {
-    if (sampleRate == avCodec.supported_samplerates[i]) {
+  for (auto i = 0; supportedSampleRates[i] != 0; ++i) {
+    if (sampleRate == supportedSampleRates[i]) {
       return;
     }
   }
   std::stringstream supportedRates;
-  for (auto i = 0; avCodec.supported_samplerates[i] != 0; ++i) {
+  for (auto i = 0; supportedSampleRates[i] != 0; ++i) {
     if (i > 0) {
       supportedRates << ", ";
     }
-    supportedRates << avCodec.supported_samplerates[i];
+    supportedRates << supportedSampleRates[i];
   }
 
   TORCH_CHECK(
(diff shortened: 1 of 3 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -33,21 +33,38 @@ torch::Tensor validateSamples(const torch::Tensor& samples) {
...
+      (const void**)&supportedSampleRates,
+      &numSampleRates);
+  if (ret < 0 || supportedSampleRates == nullptr) {
+    TORCH_CHECK(false, "Couldn't get supported sample rates from encoder.");
+  }
+#else
+  supportedSampleRates = avCodec.supported_samplerates;
+#endif
+
+  if (supportedSampleRates == nullptr) {
```
scotts, line 37: We've been trying to keep FFmpeg version ifdef-ing out of the mainline logic, and keep it all in functions defined in `FFMPEGCommon`. I think we should defined there as something like:
```c++
const int* getSupportedSampleRates(const AVCodec& avCodec);
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -73,27 +90,44 @@ static const std::vector<AVSampleFormat> preferredFormatsOrder = {
...
+      0,
+      (const void**)&supportedSampleFormats,
+      &numSampleFormats);
+  if (ret < 0 || supportedSampleFormats == nullptr) {
+    TORCH_CHECK(false, "Couldn't get supported sample formats from encoder.");
+  }
+#else
+  supportedSampleFormats = avCodec.sample_fmts;
+#endif
+
```
scotts, line 79: Ditto here about creating a function in `FFMPEGCommon`.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -56,6 +56,46 @@ int64_t getDuration(const UniqueAVFrame& avFrame) {
...
+const int* getSupportedSampleRates(const AVCodec& avCodec) {
+  const int* supportedSampleRates = nullptr;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100)
+  int numSampleRates = 0;
+  int ret = avcodec_get_supported_config(
+      nullptr,
+      &avCodec,
+      AV_CODEC_CONFIG_SAMPLE_RATE,
+      0,
+      (const void**)&supportedSampleRates,
```
scotts: Let's do a `reinterpret_cast` here; https://en.cppreference.com/w/cpp/language/reinterpret_cast.html. In general, we should never do a C-style cast, as it's not always obvious what _kind_ of cast a C-style cast will actually become: https://en.cppreference.com/w/cpp/language/explicit_cast.html

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -56,6 +56,46 @@ int64_t getDuration(const UniqueAVFrame& avFrame) {
...
+const AVSampleFormat* getSupportedOutputSampleFormats(const AVCodec& avCodec) {
+  const AVSampleFormat* supportedSampleFormats = nullptr;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100) // FFmpeg >= 7
+  int numSampleFormats = 0;
+  int ret = avcodec_get_supported_config(
+      nullptr,
+      &avCodec,
+      AV_CODEC_CONFIG_SAMPLE_FORMAT,
+      0,
+      (const void**)&supportedSampleFormats,
```
scotts: Ditt on `reinterpret_cast`.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -109,7 +149,31 @@ void setDefaultChannelLayout(UniqueAVFrame& avFrame, int numChannels) {
...
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100) // FFmpeg >= 7
+  std::stringstream supportedNumChannels;
+  const AVChannelLayout* supported_layouts = nullptr;
+  int num_layouts = 0;
+  int ret = avcodec_get_supported_config(
+      nullptr,
+      &avCodec,
+      AV_CODEC_CONFIG_CHANNEL_LAYOUT,
+      0,
+      (const void**)&supported_layouts,
```
scotts: Ditto on `reinterpret_cast`.

scotts (approved):
Let's do the `reinterpret_cast` changes before merging. Otherwise, this is great! Thanks for fixing it.

PR #899, opened 2025-09-12
Use cuda filters to support 10-bit videos

dvrogozh:
As we discussed in #853, that's the version to enabled 10-bit support in cuda device interface by managing cuda filter graph within cuda device interface. This change makes these changes:
1. Added support for 10-bit (and likely 12-bit which I did not test) via `scale_cuda` ffmpeg filter on ffmpeg `>=n5.0`
2. Added support for 10-bit via CPU fallback on ffmpeg `n4.4` (as color conversion in `scale_cuda` appeared in ffmpeg `>=n5.0`)
3. Added support to directly returned input from in CPU device interface if it does not require conversion (used in combination with CUDA `n4.4` fallback)

CC: @scotts @NicolasHug

### DIFF
--- a/src/torchcodec/_core/CpuDeviceInterface.cpp
+++ b/src/torchcodec/_core/CpuDeviceInterface.cpp
@@ -83,6 +83,24 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
   enum AVPixelFormat frameFormat =
       static_cast<enum AVPixelFormat>(avFrame->format);
 
+  // This is an early-return optimization: if the format is already what we
+  // need, and the dimensions are also what we need, we don't need to call
+  // swscale or filtergraph. We can just convert the AVFrame to a tensor.
+  if (frameFormat == AV_PIX_FMT_RGB24 &&
+      avFrame->width == expectedOutputWidth &&
+      avFrame->height == expectedOutputHeight) {
+    outputTensor = toTensor(avFrame);
+    if (preAllocatedOutputTensor.has_value()) {
+      // We have already validated that preAllocatedOutputTensor and
+      // outputTensor have the same shape.
+      preAllocatedOutputTensor.value().copy_(outputTensor);
+      frameOutput.data = preAllocatedOutputTensor.value();
+    } else {
+      frameOutput.data = outputTensor;
+    }
+    return;
+  }
+
   // By default, we want to use swscale for color conversion because it is
   // faster. However, it has width requirements, so we may need to fall back
   // to filtergraph. We also need to respect what was requested from the
@@ -159,7 +177,7 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
           std::make_unique<FilterGraph>(filtersContext, videoStreamOptions);
       prevFiltersContext_ = std::move(filtersContext);
     }
-    outputTensor = convertAVFrameToTensorUsingFilterGraph(avFrame);
+    outputTensor = toTensor(filterGraphContext_->convert(avFrame));
 
     // Similarly to above, if this check fails it means the frame wasn't
     // reshaped to its expected dimensions by filtergraph.
@@ -208,23 +226,20 @@ int CpuDeviceInterface::convertAVFrameToTensorUsingSwsScale(
   return resultHeight;
 }
 
-torch::Tensor CpuDeviceInterface::convertAVFrameToTensorUsingFilterGraph(
-    const UniqueAVFrame& avFrame) {
-  UniqueAVFrame filteredAVFrame = filterGraphContext_->convert(avFrame);
+torch::Tensor CpuDeviceInterface::toTensor(const UniqueAVFrame& avFrame) {
+  TORCH_CHECK_EQ(avFrame->format, AV_PIX_FMT_RGB24);
 
-  TORCH_CHECK_EQ(filteredAVFrame->format, AV_PIX_FMT_RGB24);
-
-  auto frameDims = getHeightAndWidthFromResizedAVFrame(*filteredAVFrame.get());
+  auto frameDims = getHeightAndWidthFromResizedAVFrame(*avFrame.get());
   int height = frameDims.height;
   int width = frameDims.width;
   std::vector<int64_t> shape = {height, width, 3};
-  std::vector<int64_t> strides = {filteredAVFrame->linesize[0], 3, 1};
-  AVFrame* filteredAVFramePtr = filteredAVFrame.release();
-  auto deleter = [filteredAVFramePtr](void*) {
-    UniqueAVFrame avFrameToDelete(filteredAVFramePtr);
+  std::vector<int64_t> strides = {avFrame->linesize[0], 3, 1};
+  AVFrame* avFrameClone = av_frame_clone(avFrame.get());
+  auto deleter = [avFrameClone](void*) {
+    UniqueAVFrame avFrameToDelete(avFrameClone);
   };
   return torch::from_blob(
-      filteredAVFramePtr->data[0], shape, strides, deleter, {torch::kUInt8});
+      avFrameClone->data[0], shape, strides, deleter, {torch::kUInt8});
 }
 
 void CpuDeviceInterface::createSwsContext(
--- a/src/torchcodec/_core/CpuDeviceInterface.h
+++ b/src/torchcodec/_core/CpuDeviceInterface.h
@@ -39,8 +39,7 @@ class CpuDeviceInterface : public DeviceInterface {
       const UniqueAVFrame& avFrame,
       torch::Tensor& outputTensor);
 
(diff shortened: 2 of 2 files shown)


src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -83,6 +83,21 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
...
+      // We have already validated that preAllocatedOutputTensor and
+      // outputTensor have the same shape.
+      preAllocatedOutputTensor.value().copy_(outputTensor);
+      frameOutput.data = preAllocatedOutputTensor.value();
+    } else {
+      frameOutput.data = outputTensor;
+    }
+    return;
+  }
+
```
scotts, line 103: Let's put a comment in explaining that this is an early-return optimization: if the format is already what we need, and the dimensions are also what we need, we don't need to call swscale or filtergraph. We can just convert the AVFrame to a tensor.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -208,23 +223,20 @@ int CpuDeviceInterface::convertAVFrameToTensorUsingSwsScale(
...
+  auto frameDims = getHeightAndWidthFromResizedAVFrame(*avFrame.get());
   int height = frameDims.height;
   int width = frameDims.width;
   std::vector<int64_t> shape = {height, width, 3};
-  std::vector<int64_t> strides = {filteredAVFrame->linesize[0], 3, 1};
-  AVFrame* filteredAVFramePtr = filteredAVFrame.release();
-  auto deleter = [filteredAVFramePtr](void*) {
-    UniqueAVFrame avFrameToDelete(filteredAVFramePtr);
+  std::vector<int64_t> strides = {avFrame->linesize[0], 3, 1};
+  AVFrame* avFrameClone = av_frame_clone(avFrame.get());
```
scotts, line 237: We weren't calling `av_frame_clone()` here before. Was that an error?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -199,12 +199,121 @@ void CudaDeviceInterface::initializeContext(AVCodecContext* codecContext) {
...
+
+  unsigned version_int = avfilter_version();
+  if (version_int < AV_VERSION_INT(8, 0, 103)) {
+    // Color conversion support ('format=' option) was added to scale_cuda from
+    // n5.0. With the earlier version of ffmpeg we have no choice but use CPU
+    // filters. See:
+    // https://github.com/FFmpeg/FFmpeg/commit/62dc5df941f5e196164c151691e4274195523e95
+    outputFormat = AV_PIX_FMT_RGB24;
+
+    filters << "hwdownload,format=" << av_pix_fmt_desc_get(actualFormat)->name;
```
scotts: We should do a nullptr check on `av_pix_fmt_desc_get()` rather than using the result unconditionally.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -199,12 +199,121 @@ void CudaDeviceInterface::initializeContext(AVCodecContext* codecContext) {
...
+  auto frameDims =
+      getHeightAndWidthFromOptionsOrAVFrame(videoStreamOptions, avFrame);
+  int height = frameDims.height;
+  int width = frameDims.width;
+
+  AVPixelFormat outputFormat;
+  std::stringstream filters;
+
+  unsigned version_int = avfilter_version();
+  if (version_int < AV_VERSION_INT(8, 0, 103)) {
```
scotts, line 242: I don't love that we're doing FFmpeg versions checks here, but we also do this in several other places in `CudaDeviceInterface.cpp`. In the rest of the C++ code, we've kept such checks hidden behind functions in `FFMPEGCommon`. Most of those, however, are about a particular field of a struct being renamed. This logic is very specific to filtergraph, and it does make sense to keep that here.

There's no action to take based on this comment, I'm just pointing out it's an awkward situation. I might end up refactoring it in some of my decoder-native transform work.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -199,12 +199,121 @@ void CudaDeviceInterface::initializeContext(AVCodecContext* codecContext) {
...
+  UniqueAVFrame avFilteredFrame;
+  if (newFiltersContext) {
+    // We need to compare the current filter context with our previous filter
+    // context. If they are different, then we need to re-create a filter
+    // graph. We create a filter graph late so that we don't have to depend
+    // on the unreliable metadata in the header. And we sometimes re-create
+    // it because it's possible for frame resolution to change mid-stream.
+    // Finally, we want to reuse the filter graph as much as possible for
+    // performance reasons.
+    if (!filterGraph_ || filtersContext_ != newFiltersContext) {
```
scotts: This is a copy-paste of the conditions from the CPU side, but `filtersContext_` and `newFiltersContext` are now pointers, not the actual objects. So `filtersContext_ != newFiltersContext` is comparing the _pointer values_, not the actual values in the context. We need `*filtersContext_ != *newFiltersContext`.

scotts (changes requested):
@dvrogozh, thank you for fixing this! There's some minor changes to make, the biggest one being I think we're mistakenly comparing pointer values instead of actual objects. But this is great, we should be able to merge after these minor changes.

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: No, that was not a mistake. The function has changed. Previously it combined 2 operations: 1) conversion of the frame with filtergraph, 2) creating a tensor. The frame converted by filtergraph lived locally inside `UniqueAVFrame` and we just released it in the end to the tensor `deleter`. I, the schema was:
```
torch::Tensor func(const UniqueAVFrame& input) {
   UniqueAVFrame output = filtergraph->convert(input);
   AVFrame* outputPtr = output.release();
   return torch::from_blob(output->data, deleter(outputPtr), ...);
}
```

In the new code I moved frame conversion with filtergraph outside of the function. And function signature has changed - it started to accept just a constant reference to a frame without knowing whether it will be further needed or not. But we still need to pass a reference to the `deleter` to make sure that tensor will unreference it once it's not needed. To achieve that we clone the frame which references the same data and pass it to the deleter. In this way caller of the function can still do something with the frame if needed, or just unreference it (what we actually are doing). So, schema now is:
```
torch::Tensor func(const UniqueAVFrame& frame) {
   AVFrame* frameClone = av_frame_clone(frame.get());
   return torch::from_blob(frameClone->data, deleter(frameClone), ...);
}
```

src/torchcodec/_core/CpuDeviceInterface.cpp
dvrogozh: done

src/torchcodec/_core/CudaDeviceInterface.cpp
dvrogozh: added

src/torchcodec/_core/CudaDeviceInterface.cpp
dvrogozh: yes, thank you for the catch. Fixed.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -83,6 +83,24 @@ void CpuDeviceInterface::convertAVFrameToFrameOutput(
...
+      // We have already validated that preAllocatedOutputTensor and
+      // outputTensor have the same shape.
+      preAllocatedOutputTensor.value().copy_(outputTensor);
+      frameOutput.data = preAllocatedOutputTensor.value();
+    } else {
+      frameOutput.data = outputTensor;
+    }
+    return;
+  }
+
```
NicolasHug, line 103: @dvrogozh QQ - do we expect frames to come out as `AV_PIX_FMT_RGB24`? I'm a bit surprised that this would be the case, I'd assume most encoded frames are in YUV or something similar?

dvrogozh: This change here in CPU device interface is due to CPU fallback in CUDA device interface to handle ffmpeg-4.4 10-bit streams:
https://github.com/dvrogozh/torchcodec/blob/f298fa7e722e9e0932813545e819e7da1a31994d/src/torchcodec/_core/CudaDeviceInterface.cpp#L242-L247
The filter graph gets setup in CUDA device interface which outputs RGB24, then the output needs to be converted to tensor. Which happens here in CPU device interface.

Above being said, I think there might be 2 cases in the future where we might see RGB24 coming out of decoders:
1. Images support and "image" codecs such as motion jpeg
2. HW decoders sometimes blend decoding and color space/scaling  into single pass for the optimization purposes.

PR #902, opened 2025-09-19
Initial C++ implementation of transforms

scotts:
Initial implementation of C++ transforms. This PR makes no changes to any Python APIs.

There is a major refactoring of all things related to frame dimensions and user-provided requests. It was necessary to rationalize that with the concept of user-provided transforms. Throughout, we apply the principle: *figure something out as soon as possible, store that decision in a variable, and then **only** reference that variable to know what to do.* We apply this principle to the output dimensions for a frame and what color conversion library to use.

Suggested way to review this PR:
1. First read **`Transform.h`** and **`Transform.cpp`**. This introduces the concept of a transform on the C++ side. We will eventually have a whole family of `Transform` subclasses. Right now, we only have one: `ResizeTransform`. Resizing is special in two respects: it can change the output resolution, and it's the only transfrom that swscale can handle. There will be other transforms that change the output resolution, such as crop. But we actually need to specially handle `ResizeTransform` so that we can use swscale if it's the only transform requested.
2. Review `SingleStreamDecoder.h`.
    * We track new state in the decoder. Notably, we maintain a vector of transforms, an optional `resizedOutputDims_` that is the dimensions of a resize if it exists, and `metadataDims_` which are the dimensions from the metadata. This is a major refactor. We used to pass around the video stream options and constantly refer to it to figure out what the output frame's dimensions should be. The video stream options no longer specify width and height; that exclusively comes from the presence of a `ResizeTransform`. We set both `resizedOutputDims_` and `metadataDims_` in `SingleStreamDecoder::addVideoStream()`. This means we no longer need to pass around the video stream options. We still give priority to the resized values, if they exist.
    * There's also what looks like other new state tracked in the decoder. That's a refactor of moving fields from `StreamInfo` to the decoder itself. It was a lingering TODO, which I chose to do here because I was adding more decoder state. It helps clarify that the new state I'm adding belongs in the decoder and not in `StreamInfo`.
4. Review `SinglestreamDecoder.cpp`. Most of the changes here are a result of all of the above. Focus on `SingleStreamDecoder::addVideoStream()`. Note that when we want to know the output dimensions, we say `resizedOutputDims_.value_or(metadataDims_)`. This replaces the functions we had before.
5. Review `DeviceInterface.h`. The big change here is that `initializeContext()` just became `initialize()` with a long list of parameters. There's a lot of stuff we were passing to `DeviceInterface::convertAVFrameToFrameOutput()` that we now pass to `initialize()`. Note that we pass `resizedOutputDims_` if they exist.
6. Review `CpuDeviceInterface.h` and `CpuDeviceInterface.cpp`. Values that we used to pass into `convertAVFrameToFrameOutput()` are now passed to initialize, and we just store them as state. However, because we need to be resilient to variable resolution streams, we still need to defer knowing the output frame's resolution and the color conversion library until we get a raw decoded frame. The difference from before is that we pre-compute what we can in `initialize()`. And instead of having the video stream options, we have `resizedOutputDims_` as an optional. We use that or the raw decoded frame's resolution.
7. Review `CudaDeviceInterface.h` and `CudaDeviceInterface.cpp`. This is both a consequence of all of the above refactors, as well as a refactoring of #899.
8. Review `custom_ops.cpp`. Right now, we have some ugly bridge code that turns `width` and `height` parameters into a `ResizeTransform`. That will eventually go away and we will accept transforms as tensors. How is beyond the scope of this PR.


MadcowD:
wow just intime for me to need it haha

MadcowD:
When do you think this might land :)?

src/torchcodec/_core/Transform.h
```diff
@@ -0,0 +1,62 @@
...
+  virtual bool isResize() const {
+    return false;
+  }
+};
+
+class ResizeTransform : public Transform {
+ public:
+  enum class InterpolationMode { BILINEAR, BICUBIC, NEAREST };
+
+  ResizeTransform(int width, int height)
```
NicolasHug: For consistency with the rest of pytorch, we should prefer height and width for the parameter order, rather than width and height. That's how frames are shaped (NCHW or NHWC), and that's also how torchvision APIs are ordered.

I understand the FFmpeg libraries (filtegraph and swscale) may be using a different order convention, but I think we should stick to that of PyTorch in our code

src/torchcodec/_core/Transform.h
```diff
@@ -0,0 +1,62 @@
...
+  // The ResizeTransform is special, because it is the only transform that
+  // swscale can handle.
+  virtual bool isResize() const {
+    return false;
+  }
+};
+
+class ResizeTransform : public Transform {
+ public:
+  enum class InterpolationMode { BILINEAR, BICUBIC, NEAREST };
```
NicolasHug: I understand we'll be validating these values against those of torchvision later, but I'd still suggest to only expose BILINEAR at this time. to keep things simpler. BICUBIC is more complex than BILINEAR in the sense that there are more ways to implement it. For NEAREST, torchvision exposes 2 of those (NEAREST and NEAREST-EXACT), I'm not sure which one it will correspond to yet.

src/torchcodec/_core/Transform.cpp
```diff
@@ -0,0 +1,67 @@
...
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#include "src/torchcodec/_core/Transform.h"
+#include <torch/types.h>
+#include "src/torchcodec/_core/FFMPEGCommon.h"
+
+namespace facebook::torchcodec {
+
+namespace {
```
NicolasHug, line 13: Nit: I have a minor preference for inlining `toFilterGraphInterpolation` inside of `getFilterGraphCpu` instead of defining it as a new function, because it is only needed in a single place currently. It makes reading the code slightly easier as it avoids jumping from place to place.

Feel free to disregard though.

(same for `toSwsInterpolation`).

src/torchcodec/_core/Transform.h
```diff
@@ -0,0 +1,62 @@
...
+  // If the transformation does not change the output frame dimensions, then
+  // there is no need to override this member function. The default
+  // implementation returns an empty optional, indicating that the output frame
+  // has the same dimensions as the input frame.
+  //
+  // If the transformation does change the output frame dimensions, then it
+  // must override this member function and return the output frame dimensions.
+  virtual std::optional<FrameDims> getOutputFrameDims() const {
+    return std::nullopt;
+  }
```
NicolasHug, line 29: QQ both for myself and for you: we should make sure the output frame dims are never "dynamic", i.e. that they never depend on the input frame's dimension. I don't think this is needed for `resize` as-is, at this time, but it would be needed if e.g. we had parametrized `resize` with a `scale` parameter, as is the case for[ torch.nn.functional.interpolate](https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.interpolate.html)

Not sure how that would impact the current design.

----

EDIT: OK, I can see now from the logic in `addVideoStream()`:

```C++
  for (auto& transform : transforms) {
    TORCH_CHECK(transform != nullptr, "Transforms should never be nullptr!");
    if (transform->getOutputFrameDims().has_value()) {
      resizedOutputDims_ = transform->getOutputFrameDims().value();
    }
    transforms_.push_back(std::unique_ptr<Transform>(transform));
  }
```

we are making a hard assumption that the output frame dims is *uniquely* determined by the last transform's parameters, and nothing else. That's probably OK for now.

src/torchcodec/_core/Frame.h
```diff
@@ -13,6 +13,15 @@
 
 namespace facebook::torchcodec {
 
+struct FrameDims {
+  int width = 0;
+  int height = 0;
+
+  FrameDims() = default;
+
+  FrameDims(int w, int h) : width(w), height(h) {}
```
NicolasHug: Similar to the comment about Resize's parameter order, I think we should use H and W instead of W and H, like it previously was.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -442,13 +443,6 @@ void SingleStreamDecoder::addStream(
   streamInfo.codecContext->thread_count = ffmpegThreadCount.value_or(0);
   streamInfo.codecContext->pkt_timebase = streamInfo.stream->time_base;
 
-  // TODO_CODE_QUALITY same as above.
-  if (mediaType == AVMEDIA_TYPE_VIDEO) {
-    if (deviceInterface_) {
-      deviceInterface_->initializeContext(codecContext);
-    }
-  }
-
```
NicolasHug, line 468: We are now initializing the `codecContext->hw_device_ctx` **after** the call to `avcodec_open2`. 

I don't know why, but I think this is a subtle breaking change that might cause the decoding to be routed to the CPU instead of the GPU in some cases. On `main`, I tried this diff:

```diff

diff --git a/src/torchcodec/_core/CudaDeviceInterface.cpp b/src/torchcodec/_core/CudaDeviceInterface.cpp
index 6a69d4f..997e596 100644
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -321,6 +321,7 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
   // graph level. For example, that's how we handle color format conversion
   // on FFmpeg 4.4 where scale_cuda did not have this supported implemented yet.
   if (avFrame->format != AV_PIX_FMT_CUDA) {
+    TORCH_CHECK(false);
     // The frame's format is AV_PIX_FMT_CUDA if and only if its content is on
     // the GPU. In this branch, the frame is on the CPU: this is what NVDEC
     // gives us if it wasn't able to decode a frame, for whatever reason.
diff --git a/src/torchcodec/_core/SingleStreamDecoder.cpp b/src/torchcodec/_core/SingleStreamDecoder.cpp
index b52556e..57cb64a 100644
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -442,6 +442,9 @@ void SingleStreamDecoder::addStream(
   streamInfo.codecContext->thread_count = ffmpegThreadCount.value_or(0);
   streamInfo.codecContext->pkt_timebase = streamInfo.stream->time_base;
 
+
+  retVal = avcodec_open2(streamInfo.codecContext.get(), avCodec, nullptr);
+  TORCH_CHECK(retVal >= AVSUCCESS, getFFMPEGErrorStringFromErrorCode(retVal));
   // TODO_CODE_QUALITY same as above.
   if (mediaType == AVMEDIA_TYPE_VIDEO) {
     if (deviceInterface_) {
@@ -449,9 +452,6 @@ void SingleStreamDecoder::addStream(
     }
   }
 
-  retVal = avcodec_open2(streamInfo.codecContext.get(), avCodec, nullptr);
-  TORCH_CHECK(retVal >= AVSUCCESS, getFFMPEGErrorStringFromErrorCode(retVal));
-
   codecContext->time_base = streamInfo.stream->time_base;
   containerMetadata_.allStreamMetadata[activeStreamIndex_].codecName =
       std::string(avcodec_get_name(codecContext->codec_id));

```

i.e. I moved the `codecContext->hw_device_ctx` initialization after the call to `avcodec_open2`, and added a `TORCH_CHECK(false)` in the CPU decoding path of the CUDA interface. And with that, I'm seeing this `TORCH_CHECK(false)` being hit on some tests, which suggests the CPU path was taken where it shouldn't:

```
...
test/test_decoders.py::TestVideoDecoder::test_getitem_int[exact-cuda-4] FAILED                                                                                         [  2%]
...
test/test_decoders.py::TestVideoDecoder::test_getitem_int[approximate-cuda-4] FAILED                                                                                   [  4%]
...
```

I can also confirm the `TORCH_CHECK(false)` is **not** hit if the initialization order is left untouched. 

This seems to only happen on tests where `num_ffmpeg_threads` is non-zero (which is even weirder). I don't understand everything, but to be safe, we might prefer keeping the initialization of the AVCodecContext order as it previously was?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -791,13 +812,16 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
-  // which by our abstract player, means that both intervals map to frame 0. By
-  // the definition of a half open interval, interval A should return no frames.
-  // Interval B should return frame 0. However, for both A and B, the individual
-  // values of the intervals will map to the same frame indices below. Hence, we
-  // need this special case below.
+  // which by our abstract player, means that both intervals map to frame 0.
+  // By the definition of a half open interval, interval A should return no
+  // frames. Interval B should return frame 0. However, for both A and B, the
+  // individual values of the intervals will map to the same frame indices
+  // below. Hence, we need this special case below.
```
NicolasHug, line 848: Q: there are a lot of cosmetic changes in this file, related to the comments line length. Were those introduced by the linter? If we can it's best to avoid those, especially on PRs as large as this one, because it makes it harder for the reviewer to focus on the load-bearing changes.
No need to go back and revert, just pointing that out for next time!

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1680,28 +1677,4 @@ double SingleStreamDecoder::getPtsSecondsForFrame(int64_t frameIndex) {
...
-}
-
-FrameDims getHeightAndWidthFromOptionsOrAVFrame(
-    const VideoStreamOptions& videoStreamOptions,
-    const UniqueAVFrame& avFrame) {
-  return FrameDims(
-      videoStreamOptions.height.value_or(avFrame->height),
-      videoStreamOptions.width.value_or(avFrame->width));
-}
-
```
NicolasHug, line 1762: Note for self and for a later review round: those functions were previously introduced in https://github.com/pytorch/torchcodec/pull/332/files#diff-f130765abad340e1c80ce0728540a357bba43f03402f72ef512bc8e786fb147eR440-R443, whose main goal was to rationalize and simplify what used to be a pretty confusing output shape inference logic.
In particular those functions were created so that they can easily be `git-grep`able, so one can figure out which logic is used where.

Need to understand if the different (3!) logics still exist, if they've changed, and if they're still easily grep-able.

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -76,4 +76,19 @@ std::unique_ptr<DeviceInterface> createDeviceInterface(
...
+  std::vector<int64_t> shape = {height, width, 3};
+  std::vector<int64_t> strides = {avFrame->linesize[0], 3, 1};
+  AVFrame* avFrameClone = av_frame_clone(avFrame.get());
+  auto deleter = [avFrameClone](void*) {
+    UniqueAVFrame avFrameToDelete(avFrameClone);
+  };
+  return torch::from_blob(
+      avFrameClone->data[0], shape, strides, deleter, {torch::kUInt8});
+}
+
```
NicolasHug, line 116: Is it worth writing a short comment explaining why we're calling `av_frame_clone()` ? I guess it's related to https://github.com/pytorch/torchcodec/pull/899/files#r2369708555, which I don't fully understand yet.

src/torchcodec/_core/CpuDeviceInterface.h
```diff
@@ -63,15 +71,48 @@ class CpuDeviceInterface : public DeviceInterface {
...
-  // Used to know whether a new FilterGraphContext or UniqueSwsContext should
-  // be created before decoding a new frame.
   SwsFrameContext prevSwsFrameContext_;
-  FiltersContext prevFiltersContext_;
+
+  // The filter we supply to filterGraph_, if it is used. The copy filter just
+  // copies the input to the output. Computationally, it should be a no-op. If
+  // we get no user-provided transforms, we will use the copy filter. Otherwise,
+  // we will construct the string from the transforms.
+  std::string filters_ = "copy";
```
NicolasHug: I'm a bit confused by this:

> If we get no user-provided transforms, we will use the copy filter

Assuming filtergraph is used, we're using it at least to do some color-conversion (YUV -> RGB), correct? What am I missing? Is the color conversion not determined by the `filter_` string?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -46,6 +46,92 @@ CpuDeviceInterface::CpuDeviceInterface(const torch::Device& device)
...
+  for (const auto& transform : transforms) {
+    if (!first) {
+      filters << ",";
+    }
+    filters << transform->getFilterGraphCpu();
+    first = false;
+  }
+  if (!transforms.empty()) {
+    filters_ = filters.str();
+  }
```
NicolasHug, line 104: Just a comment:

IIUC, there seems to be a fundamental design assumption here: that each transform is independent and contributes a `getFilterGraphCpu()` filter string that **doesn't conflict** with the other transforms. This is probably a reasonable assumption to make at this stage, but I don't know if it'll always hold in practice.

For example, if the user specifices multiple `Resize()` instances. As you noted above there's no strong reason to even do that, but that would potentially not be supported.


As a side note: intuitively I thought we'd want to move the filter string specification and validation as much as possible to Python. But I don't know that it'll be entirely possible considering the real frame dimensions are only known late?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -253,94 +252,128 @@ std::unique_ptr<FiltersContext> CudaDeviceInterface::initializeFiltersContext(
...
-          std::make_unique<FilterGraph>(*newFiltersContext, videoStreamOptions);
-      filtersContext_ = std::move(newFiltersContext);
-    }
-    avFilteredFrame = filterGraph_->convert(avInputFrame);
+  // Note that CUDA does not yet support transforms, so the only possible
+  // frame dimensions are the raw decoded frame's dimensions.
+  auto frameDims = FrameDims(avFrame->width, avFrame->height);
 
-    // If this check fails it means the frame wasn't
-    // reshaped to its expected dimensions by filtergraph.
```
NicolasHug, line 315: QQ - are we still validating this (that the frame was properly resized by filtergraph)?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -225,16 +229,11 @@ std::unique_ptr<FiltersContext> CudaDeviceInterface::initializeFiltersContext(
       reinterpret_cast<AVHWFramesContext*>(avFrame->hw_frames_ctx->data);
   AVPixelFormat actualFormat = hwFramesCtx->sw_format;
 
-  // NV12 conversion is implemented directly with NPP, no need for filters.
+  // If the frame is already in NV12 format, we don't need to do anything.
```
NicolasHug, line 262: Not from this PR, but just above, should we `TORCH_CHECK(hwFramesCtx != nullptr)` before accessing `hwFramesCtx->sw_format`?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -253,94 +252,128 @@ std::unique_ptr<FiltersContext> CudaDeviceInterface::initializeFiltersContext(
...
-        avFilteredFrame->height);
+        frameDims.width,
+        "x3, got ",
+        shape);
   }
 
-  UniqueAVFrame& avFrame = (avFilteredFrame) ? avFilteredFrame : avInputFrame;
+  // All of our CUDA decoding assumes NV12 format. We handle non-NV12 formats by
+  // converting them to NV12.
+  avFrame = maybeConvertAVFrameToNV12(avFrame);
```
NicolasHug: I think what we now have is a net improvement over the logic in `main`, but I wonder if we could clarify the logic a bit further:  IIUC `maybeConvertAVFrameToNV12`  doesn't only conver to NV12, it will actually convert to RGB24 (on the CPU!) in some cases. I don't claim I know how to address this because I can't fit all the code branches in my head at this time, but perhaps there's an opportunity to split the logic further into separate methods?

scotts:
@MadcowD, thanks for your interest! As you can see, this PR is still very much under review. This is also only the C++ side of the implementation for this feature. We'll have a few more follow-on PRs that will bridge the Python and C++ as well as expose the APIs at the Python level. All of that will take several weeks before the feature lands on main.

src/torchcodec/_core/Transform.h
scotts: Good reasons, agreed. What's most important is as much consistency as possible. I'll change everything over to (height, width).

src/torchcodec/_core/Transform.h
scotts: Agreed we need this constraint because we need to know the size of the tensors. We can easily enforce this through what we accept at various levels of the API.

My rationale with the vector of transforms is that the last transform _which has an output dimension_ is the final output dimension. Combined with the constraint above, I _think_ that will always be correct.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: Yeah, this was VSCode doing its thing. I'm not sure why it changed these lines - it might have been because I changed actual code around these lines when I changed class variables.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: It should now be two logics:

1. resized or metadata dimensions
2. resized or raw decoded frame dimensions

The moral equivalent of that comment is now in `SingleStreamDecoder.h` where we define the variables that hold the values for 1. We can also put another comment in `CpuDeviceInterface.h` where we store the resized dimensions. Because we're now always using `FrameDim` objects and not separate ints for height and width, I find using a function unnecessary, since the logic is now pretty straightforward:

1. `resizedOutputDims_.value_or(metadataDims_)`
2. `resizedOutputDims_.value_or(outputDims)` where we've defined `outputDims` locally to be the raw decoded output frame's dimensions.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: I'll dig into this.

src/torchcodec/_core/DeviceInterface.cpp
scotts: I don't fully understand either. :(

src/torchcodec/_core/CpuDeviceInterface.h
scotts: Correct, I had assumed the filter string itself solely determined the color conversion. It _can_, in that you can set up a pipeline that does color conversion. But we're defining the output to always be RGB24 here: https://github.com/pytorch/torchcodec/pull/902/files/1a078287cb8e23133b24312ff88c55086ead5536#diff-d7cd7f718c42d71619d3515b4acf2546cb34ec82ccda349dfa8be74aba60e75cR331

I tried using a literal empty filter string, but that's not valid.

src/torchcodec/_core/CpuDeviceInterface.cpp
scotts: > IIUC, there seems to be a fundamental design assumption here: that each transform is independent and contributes a getFilterGraphCpu() filter string that doesn't conflict with the other transforms. This is probably a reasonable assumption to make at this stage, but I don't know if it'll always hold in practice.

My reasoning here is that we are going to limit our users to a simple pipeline of transforms. Even if there are multiple transforms that change the size of the frame, we only need to know the _final_ output size. Interior transforms will be handled entirely by filtergraph itself. For example, I think it will be normal for us to receive something like `[Resize(480, 640), CenterCrop(32, 32)]` then the final output frame will always be (32, 32). 

> For example, if the user specifices multiple Resize() instances. As you noted above there's no strong reason to even do that, but that would potentially not be supported.

I think that is supported with the current logic; that's why we will always go to filtergraph if there are multiple transforms. My reasoning is that with a pipeline of resize transforms, the final output dimensions will always be the dimensions of the final transform.

> As a side note: intuitively I thought we'd want to move the filter string specification and validation as much as possible to Python. But I don't know that it'll be entirely possible considering the real frame dimensions are only known late?

I actually don't think the real frame dimensions are a strong reason why; those are only important when we have transforms that _don't_ change the frame dimensions. But I think of filtergraph as a C++ implementation detail. I think it's cleaner if we keep the following concerns localized:

1. Public API only knows about TorchVision transforms.
2. Core API knows the mapping of TorchVision transforms to their C++ name and parameters.
3. C++ knows what filtergraph string implements a transform.

That means that if we have some limitation of what users can provided due to filtergraph itself, we'll need to enforce that in what we accept in layers 1 and 2.

src/torchcodec/_core/CudaDeviceInterface.cpp
scotts: Yup: https://github.com/pytorch/torchcodec/pull/902/files/1a078287cb8e23133b24312ff88c55086ead5536#diff-37d8a09669d3f009b6850f6e66888b6875d805064933148fce3a637cc7694712R284-R296

What's not obvious, as I didn't call it out, is that we're not doing any resizing on CUDA yet. I added a test to check for this. We're only doing a format conversion.

src/torchcodec/_core/CpuDeviceInterface.h
NicolasHug: I see - then I think I was confused by the mention of a "no-op" in the comment above: since we're always going to do at least color-conversion, the filter graph pipeline will never be a no-op.

scotts: @NicolasHug, ah, got it. The copy filter is a no-op, but not the execution of entire filtegraph. I'll improve the comment.

src/torchcodec/_core/CudaDeviceInterface.cpp
scotts: Actually, we already do that _before_ the cast. I'll just make the check about the pointer we actually access directly.

src/torchcodec/_core/Transform.cpp
scotts: I'm going to keep them for now, even though with removing the other options they do less. When we can break things down into pure, stateless functions, I find that useful.

src/torchcodec/_core/CudaDeviceInterface.cpp
scotts: The challenge is that we're branching on the content of the frame and the version of FFmpeg. The current version is as clean as I could get it. What I think is promising is if we could find another way to do a CUDA format conversion on FFmpeg 4. I tried a bunch of attempts, but I think that should be its own effort.

scotts:
@NicolasHug, thank you for catching the bug regarding `avcodec_open2()` and setting the `AVCodecContext.hw_device_ctx`! In order to solve it, I had to change `DeviceInterface`: https://github.com/meta-pytorch/torchcodec/pull/902/files#diff-5502c9dc9edfe6912c95c36ccb012f50930fd275d2c7feec96cddf9e0cdba085R33-R44.

src/torchcodec/_core/CudaDeviceInterface.cpp
NicolasHug: Agreed this isn't trivial, and I don't know how to do simplify this ATM. Should we at least rename `maybeConvertAVFrameToNV12` into `maybeConvertAVFrameToNV12OrRGB24`? When I read "maybeConvertAVFrameToNV12", I assume the behavior is that it will either:

- convert the frame to NV12
- not do any conversion

But this function actually does a second type of conversion: to RGB24, on CPU. I think reflecting that in the name can help a bit

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -427,43 +427,44 @@ void SingleStreamDecoder::addStream(
...
   TORCH_CHECK(
       streamInfo.stream->codecpar->codec_type == mediaType,
       "FFmpeg found stream with index ",
       activeStreamIndex_,
       " which is of the wrong media type.");
 
+  deviceInterface_ = createDeviceInterface(device);
+  TORCH_CHECK(
+      deviceInterface_ != nullptr,
+      "Failed to create device interface. This should never happen, please report.");
```
NicolasHug, line 441: Nit: not from this PR but do we need to create an interface for audio streams?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -46,6 +46,95 @@ CpuDeviceInterface::CpuDeviceInterface(const torch::Device& device)
       device_.type() == torch::kCPU, "Unsupported device: ", device_.str());
 }
 
+void CpuDeviceInterface::initialize(
+    [[maybe_unused]] AVCodecContext* codecContext,
+    const AVRational& timeBase) {
+  timeBase_ = timeBase;
+}
+
```
NicolasHug, line 53: General comment on my understanding of `CpuDeviceInterface::initialize` and `CpuDeviceInterface::initializeVideo`:

- `initialize` (previously `initializeContext`) now does 2 seemingly orthogonal things:
   - it initializes the codecContext
   - it initializes the interface `timeBase_` field
- `initializeVideo` (new) initializes other fields of the interface.

Was there a particular reason to set `timeBase_` within initialize? If possible I think it would help to separate concerns here. For example we could have one method to initialize the `codecContext` itself (and nothing else), and one separate method to initialize all the interface and its fields (`timeBase_` and the rest). In https://github.com/meta-pytorch/torchcodec/pull/910/files#diff-e4fe2b8c629954b1dac78a122bd2b33664844f022aa0bf3abc72b5938a14f4e2R465 I did something very similar to your changes, where:

- `initializeContext` is unchanged and only initializes the codecContext
- `initialize` is a new method that initializes fields on the interface itself, including the `timeBase_` which I needed too. It's analogous to your `initializeVideo()`.

Here you're calling your `initializeVideo()` later than where I call my `initialize()`, but I think still would still work fine with your initialization order.


(I also noticed you moved the cuda context creation to the constructor, which I think is a good thing!)

NicolasHug:
Thanks for the great PR @scotts , made a follow-up comment above interface initialization methods, but I'll approve to unblock. There's also a minor renaming suggestion in https://github.com/meta-pytorch/torchcodec/pull/902#discussion_r2397946888 which I'm flagging here as it may be missed, since there are already a lot of comments.

src/torchcodec/_core/CpuDeviceInterface.cpp
scotts: @NicolasHug, my rationale here is:

  * `SpecificDeviceInterface::SpecificDeviceInterface()`, as in, the actual implementation's default constructor, should initialize as much as it can given that it will be initialized without any input.
  * `DeviceInterface::initialize()` is for things that are generic to decoding in general. Video and audio decoding will need whatever goes here. That's what `codecContext_` and `timeBase_` have in common.
  * `DeviceInterface::initializeVideo()` is obviously just for things specific to video. At the moment, our transforms are very video specific, although I do think there's a point where we'll want to generalize them to audio and then transforms will move to `initialize()`.
  * `DeviceInterface::initializeAudio()` is something we'll eventually implement when we move audio decoding into a device interface.

scotts: For posterity, @NicolasHug's point is about _direction of state change_. We want to differentiate between the device interface being initialized and the device interface changing the state of something else.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: No, but I personally do find it easier to maintain the invariant that `deviceInterface_` is non-null after adding a stream. This is also aligned with our eventual goal of moving audio decoding into a device interface.

PR #906, opened 2025-09-22
[torchcodec] rename deprecated sampler

scotts:
Differential Revision: D82962921


NicolasHug (approved):
LGTM modulo lint issue

scotts:
Deliberate, but not deeply considered. I want the name to be ugly and catch people's attention. *Not* adhering to normal naming standards, I think, helps do that.

PR #910, opened 2025-09-25
BETA CUDA interface: NVCUVID decoder implementation 1/N

NicolasHug:
This is the first of a long series of PR that implements our own NVCUVID decoder backend via a new **BETA CUDA interface**.

### High-level context

- **NVDEC** == hardware decoder present on an Nvidia GPU. It's literally a physical piece of silicon that can decode videos.
- **NVCUVID** == the C library that we can use to program NVDEC. Docs are [here](https://docs.nvidia.com/video-technologies/video-codec-sdk/13.0/nvdec-video-decoder-api-prog-guide/index.html#using-nvidia-video-decoder-nvdecode-api).

Currently in `main`, we support CUDA decoding by going through the [NVCUVID implementation of FFmpeg](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/nvdec.c). This makes it very easy for us, but we have limited control of the underlying resources. Specifically, caching the underlying `CUDecoder` can lead to massive performance gains, but relying on FFmpeg's NVCUVID implementation doesn't allow us to control that.

So this PR implements a separate, independent CUDA decoding path (the BETA CUDA interface), where we directly rely on NVCUVID. A lot of our implementation was inspired by [DALI's](https://github.com/NVIDIA/DALI/blob/ad0473d491be8f6a88fcf1f3235e0b7c80b04da1/dali/operators/video/frames_decoder_gpu.cc#L1).

### Design principles

As much as possible, I tried for this new decoder to require **minimal changes from our existing code-base**, and mostly just treat it purely as a new extension by creating a new `DeviceInterface`.

Crucially, we're still demuxing through FFmpeg. That is, we still rely on FFmpeg to generate the AVPacket from a stream. This is key: it means we can leave the `SingleStreamDecoder` largely untouched. All of the new decoding logic is encapsulated in a new `BetaCudaDeviceInterface` which exposes new methods, like:

- `sendPacket(AVPacket)` which is the moral equivalent of `avcodec_send_packet(AVPacket)`
- `receiveFrame(AVFrame)`, which is the moral equivalent of `avcodec_receive_frame(AVFrame)`.

This wasn't trivial, because NVCUVID isn't designed to work well with the non-blocking send/receive API of FFmpeg (see design alternatives section below).

### What is and isn't supported right now

Support is very limited ATM, and definitely buggy (don't worry, everything is private).

- h264 only
- `seek_mode="exact"` only
- index-based APIs only
- no backwards seeks

In terms of features, adding more codec support as well as approximate seek_mode will be top priority. Supporting time-based and backwards seeks is lower pri, and may never be supported. It will depend on how hard that is.

There are plenty other things that aren't supported, and a million TODOs. Also, most guards preventing you from doing bad things aren't yet in place: if you look at the new decoder the wrong way it will be angry at you and deadlock (at best). I will be working through all the features, bugs, and guards, in follow-ups.

### How to review this

- Ignore the header files in `nvcuvid_include`. They're the headers from NVCUVID. We *have* to vendor them because they're not part of the normal CUDA toolkit.
- Start with the tests to get an idea of the Python API, and of what is currently supported. You'll see that we can request the new BETA interface by passing `device="cuda:0:beta"`. That's not a legible pytorch device string, so we can consider this API to be private. It is subject  to change anyway. For now we just need a convenient way to expose this in Python, for testing. 
- Take a look at `_video_decoder.py` to further explore how the Python API is exposed. 
- Take a brief look at the changes made to the `DeviceInterface` registration in `DeviceInterface.[h, cpp]`: previously, we could only register one interface per device type (one interface for CPU, one for CUDA, etc.). Now, we can register multiple interfaces per device type with the "device_variant" key. This is how we can enable both `device="cuda"` and device="cuda:0:beta"` and have both the default and the beta CUDA interfaces. 
- Now look at `SingleStreamDecoder` and pay attention to the new extension points added to the `DeviceInterface`. You'll see `sendPacket`, `receiveFrame`, and a few other things. They mirror the existing FFmpeg APIs. Don't get too scared by the new bit-stream filtering (BSF) logic: it's just something that converts an AVPacket into an other AVPacket, with a different binary format. It's needed for some codecs. Eventually, we'll move that away from the `SingleStreamDecoder`, and encapsulate it within the interface (this is a TODO).
- Now you can start looking at the code of `BetaCudaDeviceInterface`. You're now in a whole new rabbit hole with a lot of new concepts. We're literally writing our own decoder here, and that's not something we've done yet in TorchCodec (we were always relying on FFmpeg so far). There's too much to write to describe how it works, and anything I write now will likely be obsolete within a few days, so let's go over it in our sync.

------

#### Previous design considerations (rejected)

I discussed a lot of that during meetings already, but just for ref:

- I thought about registering ourselves as a [HWAccel](https://trac.ffmpeg.org/wiki/HWAccelIntro) object. The HWAccel registration is what FFmpeg does in its [own NVCUVID backend](https://github.com/FFmpeg/FFmpeg/blob/4fbacb39440f9951904b2ebc5ad76bf8019d478a/libavcodec/nvdec_h264.c#L175-L187) (which is what we use as our current CUDA interface). There would be  a major upside of registering ourselves as a HWAcccel :  we would be able to rely on FFmpeg’s native parser and FFmpeg’s frame-reordering logic, instead of relying on NVCUVID parser and having to implement the frame recording logic ourselves. See [this](https://github.com/FFmpeg/FFmpeg/blob/4fbacb39440f9951904b2ebc5ad76bf8019d478a/libavcodec/nvdec_h264.c#L72-L114)? This is converting FFmpeg’s parser info (right side) to NVCUVID’s built-in frame info (left side).  Among other things, this frame meta-data is needed for the NVDEC hardware decoder to know which frame it should decode first when there are frame dependencies (like B-frames). If we were to implement our own HWAcccel we’d have to implement this metadata mapping ourselves, for all codecs. That’s obviously far from trivial in terms of required knowledge, but more importantly, it’s technically impossible: all of the FFmpeg parser info is [private](https://github.com/FFmpeg/FFmpeg/blob/4fbacb39440f9951904b2ebc5ad76bf8019d478a/libavcodec/nvdec_h264.c#L53). It’s not ABI stable and it relies on private headers anyway. So NVCUVID’s [claim](https://docs.nvidia.com/video-technologies/video-codec-sdk/13.0/nvdec-video-decoder-api-prog-guide/index.html#video-decoder-pipeline) that users can rely on a third-party parser instead of their own isn't really true for us. No one, except for FFmpeg themselves, can use the FFmpeg parser. We have to build our own parser (not happening), or rely on NVCUVID’s parser.

- ~`NVCUVID` is designed around callabcks which are triggered when a packet is being parsed. Crucially there is the `pfnDisplayPicture` callback which is triggered when a frame is fully decoded and ready to be displayed, **in display order** (which is *great*). Unfortunately, we cannot rely on this callback: it is triggered by the parser within a call to `cuvidParseVideoData(packet)`, which means that the only way to know that a frame is ready is to send a packet. I guess that probably makes sense in streaming applications? But that's incompatible with FFmpeg send/receive non-blocking APIs, where we should be able to query whether a frame is ready **without** having to send a packet. So if we were to rely on this callback, we'd likely need massive changes to our `SingleStreamDecoder` architecture - which is something we don't want to do. So, we can't rely on this callback, and we have to figure out the frame display order ourselves.~ EDIT: I might be wrong about this. We may be able to rely on the `pfnDisplayPicture` callback after all.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -1202,6 +1202,7 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
       if (status == AVERROR_EOF) {
         // End of file reached. We must drain the decoder
         if (useCustomInterface) {
+          // TODONVDEC P0: Re-think this. This should be simpler.
           AutoAVPacket eofAutoPacket;
           ReferenceAVPacket eofPacket(eofAutoPacket);
           eofPacket->data = nullptr;


src/torchcodec/_core/StreamOptions.h
```diff
@@ -38,6 +39,8 @@ struct VideoStreamOptions {
   std::optional<ColorConversionLibrary> colorConversionLibrary;
   // By default we use CPU for decoding for both C++ and python users.
   torch::Device device = torch::kCPU;
+  // Device variant (e.g., "default", "beta", etc.)
+  std::string_view deviceVariant = "default";
```
NicolasHug, line 43: TODO remove

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -141,12 +141,31 @@ def __init__(
...
+        # If device looks like "cuda:0:beta", make it "cuda:0" and set
+        # device_variant to "beta"
+        # TODONVDEC P2 Consider alternative ways of exposing custom device
+        # variants, and if we want this new decoder backend to be a "device
+        # variant" at all.
+        device_variant = "default"
+        if device is not None:
+            device_split = device.split(":")
+            if len(device_split) == 3:
+                device_variant = device_split[2]
```
scotts, line 155: Can we do some error checking on valid `device_variant` values? Or will that be handled entirely on the C++ side?

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -308,6 +318,7 @@ void add_video_stream(
     std::optional<std::string_view> dimension_order = std::nullopt,
     std::optional<int64_t> stream_index = std::nullopt,
     std::optional<std::string_view> device = std::nullopt,
+    std::optional<std::string_view> device_variant = std::nullopt,
```
scotts: Elsewhere, the `device_variant` has a default value of `"default"`. But here, it's defaulted to `std::nullopt`. The latter is consistent with the rest of the parameters to `add_video_stream()`, but it's not self-consistent with all other instances of `device_variant`. It might be better to just make it a non-optional parameter with a default value of `"default"`, and same with `device`.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -27,11 +45,17 @@ class DeviceInterface {
...
+    return std::nullopt;
+  };
 
   // Initialize the hardware device that is specified in `device`. Some builds
   // support CUDA and others only support CPU.
-  virtual void initializeContext(AVCodecContext* codecContext) = 0;
+  virtual void initializeContext(
+      [[maybe_unused]] AVCodecContext* codecContext) {}
+
+  virtual void initializeInterface([[maybe_unused]] AVStream* stream) {}
```
scotts, line 58: Calling out that in #902 I collapsed all device initialization in a single `DeviceInterface::initialize()` member function.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -40,20 +64,77 @@ class DeviceInterface {
...
 using CreateDeviceInterfaceFn =
     std::function<DeviceInterface*(const torch::Device& device)>;
 
+bool registerDeviceInterface(
+    const DeviceInterfaceKey& key,
+    const CreateDeviceInterfaceFn createInterface);
+
+// Backward-compatible registration function when variant is "default"
+// TODONVDEC P2 We only need this if someone in the wild has already started
+// registering their own interfaces. Ask Dmitry.
```
scotts: We're also still using this for CPU, correct? I assume that's to prevent changes to that file in this PR, and we'll eventually move it over to the new interface as well.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1123,15 +1131,26 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
...
   UniqueAVFrame avFrame(av_frame_alloc());
   AutoAVPacket autoAVPacket;
   int status = AVSUCCESS;
   bool reachedEOF = false;
+
+  // TODONVDEC P2: Instead of defining useCustomInterface and rely on if/else
+  // blocks to dispatch to the interface or to FFmpeg, consider *always*
+  // dispatching to the interface. The default implementation of the interface's
+  // receiveFrame and sendPacket could just be calling avcodec_receive_frame and
+  // avcodec_send_packet. This would make the decoding loop even more generic.
```
scotts, line 1158: Yes, agreed that's the best direction to go in. And eventually move audio in as well.

src/torchcodec/_core/BetaCudaDeviceInterface.h
```diff
@@ -0,0 +1,108 @@
...
+  UniqueCUvideodecoder decoder_;
+  CUVIDEOFORMAT videoFormat_ = {};
+
+  struct FrameBufferSlot {
+    CUVIDPARSERDISPINFO dispInfo;
+    int64_t guessedPts;
+    bool occupied = false;
+
+    FrameBufferSlot() : guessedPts(-1), occupied(false) {
+      memset(&dispInfo, 0, sizeof(dispInfo));
```
scotts: Let's call `std::memset()`; this is actually a call to `::memset()` which is the one from C headers.

src/torchcodec/_core/BetaCudaDeviceInterface.h
```diff
@@ -0,0 +1,108 @@
...
+    }
+  };
+
+  std::vector<FrameBufferSlot> frameBuffer_;
+  FrameBufferSlot* findEmptySlot();
+  FrameBufferSlot* findFrameWithExactPts(int64_t desiredPts);
```
scotts: Since we effectively have member functions that operate on `frameBuffer_`, and that appears to be mostly how we interact with it, we should probably abstract it into a `FrameBuffer` class.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+#include <mutex>
+#include <vector>
+
+#include "src/torchcodec/_core/BetaCudaDeviceInterface.h"
+
+#include "src/torchcodec/_core/DeviceInterface.h"
+#include "src/torchcodec/_core/FFMPEGCommon.h"
+#include "src/torchcodec/_core/NVDECCache.h"
+
+#include <cuda_runtime.h> // For cudaStreamSynchronize
```
scotts: I don't see `cudaStreamSynchronize` referenced anywhere - is the include still needed?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+    DeviceInterfaceKey(torch::kCUDA, "beta"),
+    [](const torch::Device& device) {
+      return new BetaCudaDeviceInterface(device);
+    });
+
+static int CUDAAPI
+pfnSequenceCallback(void* pUserData, CUVIDEOFORMAT* videoFormat) {
+  BetaCudaDeviceInterface* decoder =
+      static_cast<BetaCudaDeviceInterface*>(pUserData);
+  return static_cast<int>(decoder->streamPropertyChange(videoFormat));
```
scotts: If the callback expects an `int`, why do we define the member function to return an `unsigned char`?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+  return decoder->frameReadyForDecoding(pPicParams);
+}
+
+static UniqueCUvideodecoder createDecoder(CUVIDEOFORMAT* videoFormat) {
+  // Check decoder capabilities - same checks as DALI
+  auto caps = CUVIDDECODECAPS{};
+  caps.eCodecType = videoFormat->codec;
+  caps.eChromaFormat = videoFormat->chroma_format;
+  caps.nBitDepthMinus8 = videoFormat->bit_depth_luma_minus8;
+  CUresult result = cuvidGetDecoderCaps(&caps);
```
scotts, line 56: Do I understand correctly that `cuvidGetDecoderCaps()` fills in the NVDEC's actual capabilities and constraints into the `caps` variable?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+      caps.nMaxHeight);
+
+  TORCH_CHECK(
+      videoFormat->coded_width * videoFormat->coded_height / 256 <=
+          caps.nMaxMBCount,
+      "Video is too large (too many macroblocks). "
+      "Provided (width * height / 256): ",
+      videoFormat->coded_width * videoFormat->coded_height / 256,
+      " vs supported:",
+      caps.nMaxMBCount);
```
scotts, line 105: Any pointers to docs that talk about macroblocks? I can intuit roughly what this is, but a reference would be great. Particularly if it explains the magic 256.

src/torchcodec/_core/DeviceInterface.h
NicolasHug: Yes that's correct. I think I'll just remove this method and force the use of the new one. We've been making breaking changes to the interface lately anyway, so keeping this for BC was probably overkill.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Good catch thank you - that a remnant from a previous implem. I'll remove (although we'll very likely need it back at some point!)

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: So the callback, i.e. `pfnSequenceCallback`, must return an int.
The docs indicate that the callback may return the value of `CUVIDEOFORMAT::min_num_decode_surfaces` which is an `unsigned char`, so we have to cast at some point. I chose to cast here in this wrapper, but we could cast earlier in the `streamPropertyChange` method.

-----

The [docs](https://docs.nvidia.com/video-technologies/video-codec-sdk/13.0/nvdec-video-decoder-api-prog-guide/index.html#using-nvidia-video-decoder-nvdecode-api) about why this callback should return the `CUVIDEOFORMAT::min_num_decode_surfaces` are:

> ulMaxNumDecodeSurfaces:This is number of surfaces in parser’s DPB (decode picture buffer). This value may not be known at the parser initialization time and can be set to a dummy number like 1 to create parser object. Application must register a callback pfnSequenceCallback with the driver, which is called by the parser when the parser encounters the first sequence header or any changes in the sequence. This callback reports the minimum number of surfaces needed by parser’s DPB for correct decoding in CUVIDEOFORMAT::min_num_decode_surfaces. The sequence callback may return this value to the parser if wants to update CUVIDPARSERPARAMS::ulMaxNumDecodeSurfaces. The parser then overwrites CUVIDPARSERPARAMS::ulMaxNumDecodeSurfaces with the value returned by the sequence callback, if return value of the sequence callback is greater than 1 (see description about pfnSequenceCallback below). Therefore, for optimum memory allocation, decoder object creation should be deferred until CUVIDPARSERPARAMS::ulMaxNumDecodeSurfaces is known, so that the decoder object can be created with required number of buffers, such that CUVIDDECODECREATEINFO::ulNumDecodeSurfaces = CUVIDPARSERPARAMS::ulMaxNumDecodeSurfaces.


I've read this 100 times, and I only get more confused. But DALI does it so I trust it.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Correct yes, then we validate some of those capabilities against the input video.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: I think it refers to [this](https://en.wikipedia.org/wiki/Macroblock) which is a video encoding/decoding concept. My very very rough understanding is that a frame is divided into macroblocks (square patches) during the encoding process (https://github.com/leandromoreira/digital_video_introduction#1st-step---picture-partitioning)

Arguably the error message isn't very user friendly, but I don't think we can keep it user-friendly while keeping it informative?

src/torchcodec/decoders/_video_decoder.py
NicolasHug: There will be a loud error if no interface was registered with the provided `device_variant` string. That validation happens in C++ (there's a small test as well).

For now I'm treating this Python part pretty loosely, e.g. we're not validating against `if len(device_split) > 3` or anything similar. That's because we're not 100% sure yet that this will be how we'll want to expose this. But when we're set on the interface, we should definitely add stricter validation.

Happy to add any additional validation you had in mind in the mean time though, LMK

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+      avStream->codecpar->codec_id == AV_CODEC_ID_H264,
+      "Can only do H264 for now");
+
+  // Setup bit stream filters (BSF):
+  // https://ffmpeg.org/doxygen/7.0/group__lavc__bsf.html
+  // This is only needed for some formats, like H264 or HEVC.  TODONVDEC P1: For
+  // now we apply BSF unconditionally, but it should be optional  and dependent
+  // on codec and container.
+  const AVBitStreamFilter* avBSF = av_bsf_get_by_name("h264_mp4toannexb");
+  TORCH_CHECK(
```
scotts, line 180: How do we know `"h264_mp4toannexb"` is the one we need?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+    return AVERROR_EXTERNAL;
+  }
+  return AVSUCCESS;
+}
+
+// TODONVDEC P0: cleanup this raw pointer / reference monstruosity.
+ReferenceAVPacket* BetaCudaDeviceInterface::applyBSF(
+    ReferenceAVPacket& packet,
+    [[maybe_unused]] AutoAVPacket& filteredAutoPacket,
+    ReferenceAVPacket& filteredPacket) {
```
scotts: Cleaning this up might be as simple as just returning a reference instead. And why does it accept `filteredAutoPacket`? I don't see any implementations that use it.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+  CUVIDPARSERDISPINFO dispInfo = {};
+  dispInfo.picture_index = pPicParams->CurrPicIdx;
+  dispInfo.progressive_frame = !pPicParams->field_pic_flag;
+  dispInfo.top_field_first = pPicParams->bottom_field_flag ^ 1;
+  dispInfo.repeat_first_field = 0;
+  dispInfo.timestamp = guessedPts;
+
+  FrameBufferSlot* slot = findEmptySlot();
+  slot->dispInfo = dispInfo;
+  slot->guessedPts = guessedPts;
```
scotts, line 365: How much of a guess is this? Are we ever able to verify it, or do we have to trust it completely?

src/torchcodec/_core/NVDECCache.h
```diff
@@ -0,0 +1,102 @@
...
+
+struct CUvideoDecoderDeleter {
+  void operator()(CUvideodecoder decoder) const {
+    if (decoder) {
+      cuvidDestroyDecoder(decoder);
+    }
+  }
+};
+
+using UniqueCUvideodecoder = std::unique_ptr<void, CUvideoDecoderDeleter>;
```
scotts: Why `void` and not `CUvideodecoder`?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+  procParams.unpaired_field = dispInfo.repeat_first_field < 0;
+  CUdeviceptr framePtr = 0;
+  unsigned int pitch = 0;
+
+  // We know the frame we want was sent to the hardware decoder, but now we need
+  // to "map" it to an "output surface" before we can use its data. This is a
+  // blocking calls that waits until the frame is fully decoded and ready to be
+  // used.
+  CUresult result = cuvidMapVideoFrame(
+      static_cast<CUvideodecoder>(decoder_.get()),
```
scotts: This line is what made me realize we're declaring `UniqueCUvideodecoder` with `void` instead of `CUvideodecoder`.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+
+  avFrame->width = width;
+  avFrame->height = height;
+  avFrame->format = AV_PIX_FMT_CUDA;
+  avFrame->pts = dispInfo.timestamp; // == guessedPts
+
+  unsigned int frameRateNum = videoFormat_.frame_rate.numerator;
+  unsigned int frameRateDen = videoFormat_.frame_rate.denominator;
+  int64_t duration = static_cast<int64_t>((frameRateDen * timeBase_.den)) /
+      (frameRateNum * timeBase_.num);
```
scotts, line 456: We should probably pull the time utilities out of `SingleStreamDecoder` and put them in `FFMPEGCommon`. It's reasonable to do since they're really time utilities for FFmpeg.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+      avFrame->colorspace = AVCOL_SPC_SMPTE170M;
+      break;
+  }
+
+  avFrame->color_range =
+      videoFormat_.video_signal_description.video_full_range_flag
+      ? AVCOL_RANGE_JPEG
+      : AVCOL_RANGE_MPEG;
+
+  // Below: Ask Claude. I'm not going to even pretend.
```
scotts, line 484: Wow. Can we find a similar part of FFmpeg?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,563 @@
...
+  if (!eofSent_) {
+    CUVIDSOURCEDATAPACKET cuvidPacket = {};
+    cuvidPacket.flags = CUVID_PKT_ENDOFSTREAM;
+    CUresult result = cuvidParseVideoData(videoParser_, &cuvidPacket);
+    if (result == CUDA_SUCCESS) {
+      eofSent_ = true;
+    }
+  }
+
+  isFlushing_ = false;
```
scotts, line 512: Is there any chance of other _threads_ seeing this value, or do we need to do this because when we call `cuvidParseVideoData()`, it might trigger some of our other callbacks which then check `isFlushing_`?

src/torchcodec/_core/NVDECCache.h
```diff
@@ -0,0 +1,102 @@
...
+
+// A per-device cache for NVDEC decoders. There is one instance of this class
+// per GPU device, and it is accessed through the static GetCache() method.
+class NVDECCache {
+ public:
+  static NVDECCache& GetCache(int deviceIndex);
```
scotts: Nit: we should stick to `getCache()` for naming a function.

src/torchcodec/_core/NVDECCache.h
NicolasHug: `CUvideodecoder` is actually typedefed as a `void *` ...

But I agree this is weird. I'll make this a `UniqueCUvideodecoder =  std::unique_ptr<CUvideodecoder, CUvideoDecoderDeleter>;`, and remove the `static_cast<CUvideodecoder>`.

Now we have to update calls to e.g. `cuvidDecodePicture(*decoder_.get(), pPicParams);`, because cuvidDecodePicture accepts a raw CUvideodecoder, not a pointer to it.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: This won't always be the one we need - this is one of the many hard-coded stuff whose logic will become conditional, when we address more of the P0 TODOs (like more codec support). Eventually it will be something like this: https://github.com/NVIDIA/DALI/blob/c7539676a24a8e9e99a6e8665e277363c5445259/dali/operators/video/frames_decoder_gpu.cc#L298-L303

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: It corresponds to this part in the FFmpeg decoder: https://github.com/FFmpeg/FFmpeg/blob/bc561013c9a809a90c0d1b84413814ba612f7c44/libavcodec/nvdec.c#L550-L554

I don't understand either of these :)

We actually have a fair amount of code in torchcodec already where we manipulate both `data` and `linesize`, which TBH, I just accept as "working until proven otherwise". Some examples from `git grep -n -C 2 -e "avFrame->data" -e "avFrame->linesize`:

```
src/torchcodec/_core/CudaDeviceInterface.cpp:439:  Npp8u* yuvData[2] = {avFrame->data[0], avFrame->data[1]};
src/torchcodec/_core/CudaDeviceInterface.cpp:453:      int srcStep[2] = {avFrame->linesize[0], avFrame->linesize[1]};

src/torchcodec/_core/CpuDeviceInterface.cpp:235:  std::vector<int64_t> shape = {height, width, 3};
src/torchcodec/_core/CpuDeviceInterface.cpp:236:  std::vector<int64_t> strides = {avFrame->linesize[0], 3, 1};
src/torchcodec/_core/CpuDeviceInterface.cpp:237:  AVFrame* avFrameClone = av_frame_clone(avFrame.get());
```

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: It's the latter: the call below `cuvidParseVideoData` may trigger the `frameReadyForDecoding()`, and if that's the case while we're flushing we just want to error out. To be perfectly honest I don't understand this mechanism fully just yet - there's a P0 TODO for me to re-think and potentially simplify the flushing logic.

We're not concerned about threads here, none of that code is meant to be thread-safe.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: I don't think we have a similar logic within `SingleStreamDecoder` where we compute a duration from the frame rate and the timebase. Maybe I misunderstand?

src/torchcodec/_core/BetaCudaDeviceInterface.h
NicolasHug: I made those changes in [`0ad7370` (#910)](https://github.com/meta-pytorch/torchcodec/pull/910/commits/0ad73707a8118c323b591d50d66797e3b957de81).  Is this what you had in mind? I find it a bit overkill and boilerplate, but no strong opinion.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: It's a guess in the sense that we'll have to use some heuristics to infer a frame's pts from a packet's pts, and a packet may contain a frame, multiple frames, or just part of a frame. I think there are reliable ways to correctly infer the frames' pts, or at least in a good enough way to preserve order (that's ultimately what we care about).

For now the basic heuristic we have works on the h264 video we're testing against. I know it doesn't work on other formats, so we'll have to make it more robust.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: I reworked this logic entirely, I'll leave a new comment - I need your input on the new design.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -0,0 +1,564 @@
...
+      retVal >= AVSUCCESS,
+      "Failed to receive packet from bitstream filter: ",
+      getFFMPEGErrorStringFromErrorCode(retVal));
+
+  // Free the original packet's data which isn't needed anymore, and move the
+  // fields of the filtered packet into the original packet. The filtered packet
+  // fields are re-set by av_packet_move_ref, so when it goes out of scope and
+  // gets destructed, it's not going to affect the original packet.
+  av_packet_unref(packet.get());
+  av_packet_move_ref(packet.get(), filteredPacket.get());
```
NicolasHug: I ... I hate this, I think?

We're modifying the input `ReferenceAVPacket& packet` internal buffers **inplace**. It means that when you call `interface_->sendPacket(avPacket)` (and thus `applyBSF(avPacket)`) , the `avPacket` is modified after the call.

I think the alternative is for `applyBSF()` to **return** the new filtered packet. But that's not trivial either: this means we need to declare the new `AutoAVPacket` in the right place, to prevent it from being freed before it's used. I think we'd need to do that either:

- In `sendPacket()`, and then pass it down to applyBSF() - meh?
- or at construction of the interface, by storing the `AutoAVPacket` as a field - meh ?

CC @scotts

src/torchcodec/decoders/_video_decoder.py
scotts: We get the clear error from the C++ layer, so I think that's fine.

src/torchcodec/_core/BetaCudaDeviceInterface.h
scotts: That is what I was thinking of. I think there's value in having clear abstractions. Now that I see it all together, I also don't think this is really a "frame buffer," as I think that construct typically holds the actual memory for frames - I initially thought there would be actual `AVFrame`s in it. Its purpose is to hold a mapping from {frame pts -> CUVIDPARSERDISPINFO}. I also don't think the relative order inside the vector matters.

Given all that, I think we should use a `std::map<int64_t,  CUVIDPARSERDISPINFO>` instead of a `std::vector`, and then I don't think we'll need to implement `findEmptySlot()` and `findFrameWithExactPts()` at all. The functionality of `std::map` should cover that. We won't reuse empty slots, but I don't think that's important; we can let `std::map` worry about that sort of thing for us.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: They probably are using `unsigned char` for `CUVIDEOFORMAT::min_num_decode_surfaces` when they should be using an `uint8_t`. :/ When I see `unsigned char` I expect raw bytes, so it's kinda jarring. I'd like to shield our code from their bad choices, so I think casting their value to a known integer size as early as we can is best.

src/torchcodec/_core/NVDECCache.h
```diff
@@ -0,0 +1,104 @@
...
+ private:
+  // Cache key struct: a decoder can be reused and taken from the cache only if
+  // all these parameters match.
+  struct CacheKey {
+    cudaVideoCodec codecType;
+    unsigned width;
+    unsigned height;
+    cudaVideoChromaFormat chromaFormat;
+    unsigned int bitDepthLumaMinus8;
+    unsigned char numDecodeSurfaces;
```
scotts: Let's be explicit about integer size. I know we're getting these values from the CUDA code, but we can be explicit about what we're storing them in. For example, I'm not actually sure what integer size `width` will be here - I think that will become a `uint32_t`, but I'm not actually sure what the C++ rules are.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: Ah, yeah, I don't think we have any hope of having an informative error message here. I'm more thinking about our own understanding of where 256 came from when understanding the code. Is it coming from a limitation in the NVDEC, a limitation from the h264 format, a combination of them, or something else entirely?

src/torchcodec/_core/BetaCudaDeviceInterface.h
NicolasHug: > Its purpose is to hold a mapping from {frame pts -> CUVIDPARSERDISPINFO}.

You're right that it only does that for now, but that will quickly change. This buffer is the core of the frame reordering logic. Right now that logic is trivial (simple mapping, like you noticed), but that won't be enough for other codecs and for other scenarios, like approximate seek mode. So if we were to use a std::map, I'd have to undo it very soon.

To give you an idea, this is the moral equivalent of FFmpeg [`delayed_pic` buffer](https://github.com/FFmpeg/FFmpeg/blob/74115b017c3038b717162bb56346afb4ad937be3/libavcodec/h264dec.h#L469) , which is an array holding a bunch of info. And that's just for H264...

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: We don't do this same calculation in `SingleStreamDecoder`? My thinking was that we should try to keep all calculations with `AVRational`s in convenience functions to make sure we're doing the right thing.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: Oh, yeah, I just always though _you_ understood it! Illusion shattered. :)

src/torchcodec/_core/NVDECCache.h
scotts: > CUvideodecoder is actually typedefed as a void * ...

🤯

scotts:
@NicolasHug, forgot to say yesterday: amazing work! I think we're on the right path, this is very promising.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: From what I understand, it's a codec capability limitation, not necessarily specific to h264. The 256 constant comes from the description of that field in the header:

```
  unsigned int nMaxMBCount; /**< OUT: Max supported macroblock count
                                      CodedWidth*CodedHeight/256 must be <=
                               nMaxMBCount             */
```

src/torchcodec/_core/NVDECCache.h
NicolasHug: That's an oversight on my part!
I defined them as `unsigned int` now - is that what you wanted, or do you prefer types like `uint32_t`?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Hm, the only uses of `AVRational` I can find in `SingleStreamDecoder` is

```
double ptsToSeconds(int64_t pts, const AVRational& timeBase) {
  // To perform the multiplication before the division, av_q2d is not used
  return static_cast<double>(pts) * timeBase.num / timeBase.den;
}

int64_t secondsToClosestPts(double seconds, const AVRational& timeBase) {
  return static_cast<int64_t>(
      std::round(seconds * timeBase.den / timeBase.num));
}
```


but that's a converting ints to and from floats. Here, we're computing a duration while keeping all pts as `int`.

NicolasHug: BTW the code above will fail with a zero division error if `videoFormat_.frame_rate` is not set, which happens in our AV1 video. I'll fix that ASAP when adding AV1 support

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: Then let's make it a named constant defined at the top of the cpp file, with a comment indicating where it came from and hinting at how we might need to generalize it in the future.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1181,11 +1185,18 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
...
-        // packet.
-        status = avcodec_send_packet(
-            streamInfo.codecContext.get(),
-            /*avpkt=*/nullptr);
+        // End of file reached. We must drain the decoder
+        if (useCustomInterface) {
+          AutoAVPacket eofAutoPacket;
+          ReferenceAVPacket eofPacket(eofAutoPacket);
+          eofPacket->data = nullptr;
+          eofPacket->size = 0;
```
scotts, line 1209: We should be able to define a global, constant EOF packet that is local to this cpp file. Or if we can't do that (because maybe we need to imperatively set the fields `data` and `size` and we can't do it on object construction?) we can define a `getEoFPacket()` function which does this.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: That the only use of `AVRational` is in one place means we successfully localized it all to those functions. :) At a higher level, what I'm advocating for is for us to not have to do dimensional analysis when reading the decoder code. I can, of course, but I always have to stop and think carefully if we're multiplying and dividing correctly to get the right dimensions after the computation. Abstracting that into a function, I think, makes it more likely we'll not make a mistake and it's easier to read.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
scotts: The tension, as I understand it is:

* The member function `applyBSF(ReferenceAVPacket& )` takes a reference to a packet, and the point of that is for the reference to be immutable.
* The FFmpeg function `av_bsf_receive_packet()` needs a ["clean" packet that is freshly allocated](https://ffmpeg.org/doxygen/7.0/group__lavc__bsf.html#ga7fffb6c87b91250956e7a2367af56b38). That means we can't give it the reference we have, *and* we need to somehow override the reference we have with the result.

I think this breaks the assumptions we have with `AutoAVPacket` and `ReferenceAVPacket`. Roughly, we assumed:

* `AutoAVPacket` would be defined outside of the decoding loop. It managed allocation (`av_packet_alloc()`) and deallocation (`av_packet_free()`), but had no ability to access the underlying packet.
* `ReferenceAVPacket` would be defined inside of the decoding loop. It did not manage allocations, but was the only way to access the underlying packet. It does nothing on creation, but on destruction it calls `av_packet_unref()`.

The combination of the two made sure we only allocated one packet for all iterations of the loop, memory was always freed and we always had a valid reference to a packet. But `applyBSF()`, due to `av_bsf_receive_packet()` is the first time we need to actually _replace_ a packet. Replacing a packet breaks the model we set up above. We can break the model because we're using C++ and you can always smash the guardrails you established for yourself. :)

I had briefly (well, not so briefly; I just deleted a decent sized paragraph) considered just replacing all of that with a `std::unique_ptr`. But that's too simple; we need to manage allocations _and_ doing the unref thing. I think the best solution here is to officially bless your hack as `ReferenceAVPacket::reset(ReferenceAVPacket& other)`, inspired by [`std::unique_ptr::reset()`](https://en.cppreference.com/w/cpp/memory/unique_ptr/reset.html). Then in this function, we'll just do:
```c++
packet.reset(filteredPacket);
```

src/torchcodec/_core/NVDECCache.h
scotts: I prefer types like `uint32_t`. :)

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Makes sense - I will add a P0 TODO to move all these into FFmpegCommon. I just want to slightly delay that if that's OK, because that part will slightly change soon when I address the potential zero division issue.

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: I'll add a P0 TODO to address this. Defining a global isn't possible as you mentioned, and having `getEoFPacket()` isn't easy either because of the relationship between AutoAVPacket and ReferenceAVPacket. I think we'll be able to clean that all up by adding a new `sendEOFPacket()` to the interface, eventually.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Thank you for the feedback. Done - and added a P0 TODO to re-think that as well.

Issue #907, opened 2025-09-23
[RFC] VideoEncoder Python API


Dan-Flores:
This RFC proposes the design / parameters exposed in our VideoEncoder Python API.

The implementation aims to be simple to use without a deep understanding of FFmpeg. To achieve this, the encoder should rely on FFmpeg to select the default codec and most, if not all, other parameters similar to the FFmpeg CLI.

### Proposed API
```python
def VideoEncoder(
    frames: Tensor,
    frame_rate: int,
) -> None:
	pass


def to_file(
  dest: str
) -> None:
	pass

VideoEncoder(frames=frames, frame_rate=25).to_file("output.mov") 


def to_tensor(
  format: str
) -> torch.Tensor:
	pass

encoded_tensor = VideoEncoder(frames=frames, frame_rate=25).to_tensor("mov")
```

This exposes the following parameters:
* frame_rate, frames
  *  We cannot reasonably assume a frame_rate, so it must be provided. 

These parameters were considered, but will not be exposed:
* codec
  * We will rely on FFmpeg to select a valid default codec for the provided container format. 
* crf, bit_rate
  * Internally, we can set crf to ensure the encoder is handling the frame data correctly. Generally, we will allow FFmpeg to determine quality via bit_rate or crf based on the codec's defaults.
* gop_size, max_b_frames
  * These parameters affect frame compression and seeking performance. We can rely on the codec's default values for now.
* output_pixel_format
  * We can rely on FFmpeg to select the best pixel format using [avcodec_find_best_pix_fmt_of_list](https://www.ffmpeg.org/doxygen/5.1/group__lavc__misc__pixfmt.html#ga9e74b43a3433ccfe836814f0a6371aa0), which aims to minimize loss. For lossless encoding, this should default to YUV444 when available.

### Questions
* What would a multistream encoder look like, given this proposal and the AudioEncoder?

1. Modular approach, similar to one laid out in [Audio Encoder Design](https://github.com/NicolasHug/torchcodec/blob/encoder_design/audio_encoder_design.md#thinking-ahead)
```python
Encoder()
	.add_audio(AudioEncoder(samples, sample_rate))
	.add_video(VideoEncoder(frames, frame_rate))
	.to_file(filename)
```

2. Combined approach, similar to [torchvision's write_video](https://github.com/pytorch/vision/blob/e3b5d3a8bf5e8636462fd8bce9897bccc690b2a0/torchvision/io/video.py#L58)
```python

[...]

PR #913, opened 2025-09-30
Update Video Encoder and tests for 6 container formats

Dan-Flores:
This PR updates the VideoEncoder to support encoding for common video container formats, nearly identically to the FFmpeg CLI. 

### Changes:
Some changes are made to align with the design in #907:
* Rely on [avcodec_find_best_pix_fmt_of_list](https://www.ffmpeg.org/doxygen/5.1/group__lavc__misc__pixfmt.html#ga9e74b43a3433ccfe836814f0a6371aa0) to select the best pixel format for the default codec
* Add `crf` as an option on the C++ side to enable round trip tests

### Testing
* `test_video_encoder_round_trip`: Ensures that a video's decoded frames are the same after encoding then decoding.
  * `mov`, `mp4`, `mkv`, `webm`
* `test_video_encoder_against_ffmpeg_cli`: Ensures that the VideoEncoder frames are the same as the FFmpeg CLI.
  * `mov`, `mp4`, `avi`, `mkv`, `webm`, `flv`, `gif` 
### Testing caveats
  * The `crf` parameter is needed to test lossless encoding in the round trip test. For formats that do not support `crf`, the round trip test is not availablle.
* When lossy encoding occurs due to codec + pixel format selection, assert_close is substituted by `assert_tensor_close_on_at_least` with a lower percentage match (96-99), and a higher `atol` (2-15).

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -609,7 +609,7 @@ void VideoEncoder::initializeEncoder(
       getSupportedPixelFormats(*avCodec), // List of supported formats
       AV_PIX_FMT_GBRP, // We reorder input to GBRP currently
       0, // No alpha channel
-      0 // Discard conversion loss information
+      nullptr // Discard conversion loss information
   );
   TORCH_CHECK(outPixelFormat_ != -1, "Failed to find best pix fmt")
 


src/torchcodec/_core/Encoder.cpp
```diff
@@ -621,37 +622,51 @@ void VideoEncoder::initializeEncoder(
...
+    avCodecContext_->flags |= AV_CODEC_FLAG_QSCALE;
+    // Reuse of crf below is only intended to work in tests where crf = 0
+    // Use qmin as lower bound for best possible quality
+    int qp = videoStreamOptions.crf.value() <= avCodecContext_->qmin
+        ? avCodecContext_->qmin
+        : videoStreamOptions.crf.value();
+    avCodecContext_->global_quality = FF_QP2LAMBDA * qp;
   }
+  int status = avcodec_open2(avCodecContext_.get(), avCodec, &options);
+  av_dict_free(&options);
```
Dan-Flores, line 640: Above, the `crf` parameter is reused to set `qscale` to encode high quality videos in round-trip tests. But, the C++ function only allows `crf` to be set, not `qscale`. Since `qscale` is not needed anywhere else, I did not think it was worth including, but I am open to feedback here.

test/test_ops.py
```diff
@@ -1339,24 +1342,125 @@ def decode(self, file_path) -> torch.Tensor:
...
+            frames=source_frames, frame_rate=frame_rate, filename=encoded_path, crf=0
+        )
         round_trip_frames = self.decode(encoded_path).data
-
+        assert (
+            source_frames.shape == round_trip_frames.shape
+        ), f"Shape mismatch: source {source_frames.shape} vs round_trip {round_trip_frames.shape}"
+        assert (
+            source_frames.dtype == round_trip_frames.dtype
+        ), f"Dtype mismatch: source {source_frames.dtype} vs round_trip {round_trip_frames.dtype}"
```
NicolasHug: Just use plain assert like 

```py
assert source_frames.shape == round_trip_frames.shape
```

pytest will provide a proper error message

test/test_ops.py
```diff
@@ -1339,24 +1342,125 @@ def decode(self, file_path) -> torch.Tensor:
...
+        if ffmpeg_version == 6 or format in ("avi", "flv"):
+            assert_close = partial(assert_tensor_close_on_at_least, percentage=99)
+            atol = 15
+        else:
+            assert_close = torch.testing.assert_close
+            atol = 2
         # Check that PSNR for decode(encode(samples)) is above 30
         for s_frame, rt_frame in zip(source_frames, round_trip_frames):
             res = psnr(s_frame, rt_frame)
             assert res > 30
```
NicolasHug: I realize it's not from this PR but let's clean that up a little:

- the comment isn't needed as the code is really self explanatory (and it doesn't just do psnr validation anymore!)
- no need to store `res`


```suggestion
            atol = 2
for s_frame, rt_frame in zip(source_frames, round_trip_frames):
    assert psnr(s_frame, rt_frame) > 30
```

test/test_ops.py
```diff
@@ -1339,24 +1342,125 @@ def decode(self, file_path) -> torch.Tensor:
...
+            assert_close(s_frame, rt_frame, atol=atol, rtol=0)
+
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
```
NicolasHug, line 1425: It's duplicated
```suggestion
```

test/test_ops.py
```diff
@@ -1339,24 +1342,125 @@ def decode(self, file_path) -> torch.Tensor:
...
+            if ffmpeg_version == 4:
+                pytest.skip(
+                    "Codec for webm is not available in the FFmpeg4 installation."
+                )
+            if IS_WINDOWS and ffmpeg_version in (6, 7):
+                pytest.skip(
+                    "Codec for webm is not available in the FFmpeg6/7 installation on Windows."
+                )
+        asset = TEST_SRC_2_720P
         # Test that decode(encode(decode(asset))) == decode(asset)
```
NicolasHug: This comment should be at the top

test/test_ops.py
```diff
@@ -1339,24 +1342,125 @@ def decode(self, file_path) -> torch.Tensor:
...
+        source_frames = self.decode(str(asset.path)).data
+        frame_rate = 30
+
+        # Encode with FFmpeg CLI
+        temp_raw_path = str(tmp_path / "temp_input.raw")
+        with open(temp_raw_path, "wb") as f:
+            f.write(source_frames.permute(0, 2, 3, 1).cpu().numpy().tobytes())
+
+        ffmpeg_encoded_path = str(tmp_path / f"ffmpeg_output.{format}")
+        # Test that lossless encoding is identical
```
NicolasHug: Not sure if this comment is needed here, it looks out of place

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -85,6 +85,26 @@ const int* getSupportedSampleRates(const AVCodec& avCodec) {
   return supportedSampleRates;
 }
 
+const AVPixelFormat* getSupportedPixelFormats(const AVCodec& avCodec) {
+  const AVPixelFormat* supportedPixelFormats = nullptr;
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(61, 13, 100)
```
NicolasHug: Please add comment to specify which major FFmpeg version this correspond to

src/torchcodec/_core/Encoder.cpp
```diff
@@ -776,6 +791,14 @@ void VideoEncoder::encodeFrame(
         "Error receiving packet: ",
         getFFMPEGErrorStringFromErrorCode(status));
 
+    if (packet->duration == 0) {
+      packet->duration = 1;
+    }
+    // av_packet_rescale_ts ensures encoded frames have correct timestamps.
+    // This prevents "no more frames" errors when decoding encoded frames,
+    // https://github.com/pytorch/audio/blob/b6a3368a45aaafe05f1a6a9f10c68adc5e944d9e/src/libtorio/ffmpeg/stream_writer/encoder.cpp#L46
```
NicolasHug: https://github.com/pytorch/audio/blob/b6a3368a45aaafe05f1a6a9f10c68adc5e944d9e/src/libtorio/ffmpeg/stream_writer/encoder.cpp#L46 links to

```
    if (packet->duration == 0 && codec_ctx->codec_type == AVMEDIA_TYPE_VIDEO) {
      // 1 means that 1 frame (in codec time base, which is the frame rate)
      // This has to be set before av_packet_rescale_ts bellow.
      packet->duration = 1;
    }
```

which seems to be about the lines just above.

Is this comment at the right place? Maybe it should be a few lines above - and it should also explain why we need to set duration to 1 ?

src/torchcodec/_core/Encoder.h
```diff
@@ -154,6 +154,7 @@ class VideoEncoder {
   UniqueEncodingAVFormatContext avFormatContext_;
   UniqueAVCodecContext avCodecContext_;
   int streamIndex_ = -1;
+  AVStream* avStream_;
```
NicolasHug, line 156: IIUC we now store avStream_ mostly because we need to access `time_base`? If that's the case, then let's get rid of the `streamIndex_` field because it can now be accessed through `avStream_`

src/torchcodec/_core/Encoder.cpp
NicolasHug: Thanks for the context.

My understanding is that for those codecs that do not support crf, we set instead the qscale (quantizer scale) parameter. They both control encoding quality, but in different ways.

I think... we should avoid doing that. I don't have a good enough understanding of how these 2 parameters (and their values!) relate to each other, and I think we can punt on that for a first release of the encoder. Especially since we only really need this workaround for our round-trip test to run. It means we won't be able to do the run-trip tests on those formats, but that's OK:

- those formats aren't that popular anyway
- we should still be able to do the test against the FFmpeg CLI.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: I moved the comment to be above the `packet->duration` and `av_packet_rescale_ts` lines and updated the text to reference it as "the code below". Let me know if I misunderstood your suggestion

test/test_ops.py
```diff
@@ -1375,30 +1378,123 @@ def test_bad_input(self, tmp_path):
                 filename="./bad/path.mp3",
             )
 
-    def decode(self, file_path) -> torch.Tensor:
+    def decode(self, file_path, device="cpu") -> torch.Tensor:
```
NicolasHug: I think "cpu" is always the default, isn't it? We should be able to revert 163c5c2d4aa622ef34b90f44c50c7d902dc0db0d ?

test/test_ops.py
```diff
@@ -1375,30 +1378,123 @@ def test_bad_input(self, tmp_path):
...
+
+        ffmpeg_frames = self.decode(ffmpeg_encoded_path).data
+        encoder_frames = self.decode(encoder_output_path).data
+
+        assert ffmpeg_frames.shape[0] == encoder_frames.shape[0]
+
+        # If FFmpeg selects a codec or pixel format that uses qscale (not crf),
+        # the VideoEncoder outputs *slightly* different frames.
+        # There may be additional subtle differences in the encoder.
+        percentage = 94 if ffmpeg_version == 6 or format in ("avi") else 99
```
NicolasHug: The above works but technically, 

```
format in ("avi")
```

returns True if format is "avi" but also "a", "v", "i", "av", and "vi".


```suggestion
        percentage = 94 if ffmpeg_version == 6 or format == "avi" else 99
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -621,37 +622,39 @@ void VideoEncoder::initializeEncoder(
...
+  AVDictionary* options = nullptr;
+  if (videoStreamOptions.crf.has_value()) {
+    av_dict_set(
+        &options,
+        "crf",
+        std::to_string(videoStreamOptions.crf.value()).c_str(),
+        0);
   }
+  int status = avcodec_open2(avCodecContext_.get(), avCodec, &options);
+  av_dict_free(&options);
```
NicolasHug, line 640: I guess this is OK, if we start using `AVDictionary` in more places we should consider creating a `UniqueAVDictionary` smart pointer, like we did for the other objects e.g.

https://github.com/meta-pytorch/torchcodec/blob/e5b2eefed5d186e1232da15b1368b4561c052bae/src/torchcodec/_core/FFMPEGCommon.h#L74-L75

PR #915, opened 2025-10-01
Added proper tensor support for get_frames_at()

mollyxu:
Summary: Modified get_frames_at in _video_decoder to accept tensors and updated all downstream functions to natively accept tensors rather than converting them to lists.

Reviewed By: NicolasHug

Differential Revision: D83506846

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -602,11 +602,18 @@ FrameOutput SingleStreamDecoder::getFrameAtIndexInternal(
 }
 
 FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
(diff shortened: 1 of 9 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -776,7 +785,8 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedAt(
     frameIndices[i] = secondsToIndexLowerBound(frameSeconds);
   }
 
-  return getFramesAtIndices(frameIndices);
+  // TODO: Support tensors natively instead of a vector to avoid a copy.
```
scotts, line 788: @mollyxu, can you create an issue to track this? Specifically, we want to apply the same kind of refactoring to `get_frames_displayed_at()`.

PR #917, opened 2025-10-02
BETA CUDA interface: support for approximate mode and time-based APIs

NicolasHug:
This PR:

- (Is a lot simpler than it seems, 80% of it are just comments and tests)
- Adds support for approximate mode
- Adds support for time-based APIs.
- Drastically simplifies the logic of the BETA CUDA interface. We now rely on the NVCUVID callback which tells us when a frame is ready in display order:
  - We don't have to solve the frame reordering problem anymore, the callback is triggered in the proper order.
  - It correctly assigns the frame's PTS without us having to do any guess.

If we weren't relying on the NVCUVID callback, then we would have to solve both problems above ourselves, with codec-specific solutions. As a resut this PR also drastically simplifies future support for additional codecs - spoiler, I already added  #919 and #920 for HEVC and AV1.

In https://github.com/meta-pytorch/torchcodec/pull/910, I described this design alternative and at the time, I thought it wasn't compatible enough with our `sendPacket() / receiveFrame()` architecture. With https://github.com/meta-pytorch/torchcodec/pull/910 now merged as a minimal clean-ish skeleton of the interface, I can reason about this more clearly. And after spending a few days trying (and failing) to solve the frame-reordering problem for H264 only, I came to the conclusion that this solution, in this PR, is *well worth it*.

This new simplified design does come with a minor trade-off. I explain it in a note, in the code.

----

Why is approximate mode and time-based APIs now supported? Let's first answer: why was approximate mode and time-based APIs *not* supported before? It was because `receiveFrame(avFrame, desiredPts)` was only able to return a frame *if* we were able to find one with the exact `desiredPts`. On approximate mode, we can't guarantee that desiredPts corresponds to a frame's pts, so there generally can't be a match. Same with time-based APIs: desiredPts may not correspond to where a frame starts.

In this PR, we don't need that exact `desiredPts` matching logic anymore. But we can still guarantee that `receiveFrame` returns frames in display order, so we got approximate mode and time-based support for free.


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -41,10 +41,17 @@ pfnSequenceCallback(void* pUserData, CUVIDEOFORMAT* videoFormat) {
...
+pfnDecodePictureCallback(void* pUserData, CUVIDPICPARAMS* picParams) {
   BetaCudaDeviceInterface* decoder =
       static_cast<BetaCudaDeviceInterface*>(pUserData);
-  return decoder->frameReadyForDecoding(pPicParams);
+  return decoder->frameReadyForDecoding(picParams);
+}
+
+static int CUDAAPI
+pfnDisplayPictureCallback(void* pUserData, CUVIDPARSERDISPINFO* dispInfo) {
+  BetaCudaDeviceInterface* decoder =
```
scotts: Nit: I prefer `auto` when the expression on the right has the literal type we're getting on the left.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -211,7 +218,7 @@ void BetaCudaDeviceInterface::initializeInterface(AVStream* avStream) {
   parserParams.pUserData = this;
   parserParams.pfnSequenceCallback = pfnSequenceCallback;
   parserParams.pfnDecodePicture = pfnDecodePictureCallback;
-  parserParams.pfnDisplayPicture = nullptr;
+  parserParams.pfnDisplayPicture = pfnDisplayPictureCallback;
```
scotts, line 225: This is the key difference, correct? That is, by registering this callback, we get the new behavior and can delete all of the relevant code?

NicolasHug: yes that's correct

scotts (approved):
Amazing improvement! :)

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -329,70 +329,38 @@ void BetaCudaDeviceInterface::applyBSF(ReferenceAVPacket& packet) {
...
-  FrameBuffer::Slot* slot = frameBuffer_.findEmptySlot();
-  slot->dispInfo = dispInfo;
-  slot->guessedPts = guessedPts;
-  slot->occupied = true;
-
-  return 1;
+int BetaCudaDeviceInterface::frameReadyInDisplayOrder(
+    CUVIDPARSERDISPINFO* dispInfo) {
+  readyFrames_.push(*dispInfo);
+  return 1; // success
```
Dan-Flores, line 350: To clarify for my understanding, when the `frameReadyInDisplayOrder` callback is triggered, the parser has written the one frame that is next according to PTS in `CUVIDPARSERDISPINFO`? 

Are the function signatures for this and other callbacks defined somwhere in documentation?

NicolasHug: Yes, your understanding is correct! The `CUVIDPARSERDISPINFO` struct contains two key fields:

https://github.com/meta-pytorch/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/src/torchcodec/_core/nvcuvid_include/nvcuvid.h#L501-L509

- `timestamp`, which is the pts of the frame
- the `picture_index` field which uniquely identifies the frame. That's not the "frame index" as we have it in torchcodec, it's just an index internal to nvdec. That's what we then use here to "map" the frame:

https://github.com/meta-pytorch/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L512-L513



> Are the function signatures for this and other callbacks defined somwhere in documentation?


Not in the docs, but in the headers:

https://github.com/meta-pytorch/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/src/torchcodec/_core/nvcuvid_include/nvcuvid.h#L529-L533

Strictly speaking, this is the callaback we're defining:

https://github.com/meta-pytorch/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L49-L53


It's a pure C function that calls the corresponding method on the Interface object. We have to do this gymnastic because the pure C callbacks have no notion of the Interface object.

PR #918, opened 2025-10-03
BETA CUDA interface: Add TODOs and more explicit initialization

NicolasHug:
Small PR that adds a TODO and explicitly sets some decoder creation parameters.

Needs https://github.com/meta-pytorch/torchcodec/pull/917 first, so to review it independently please look at [`8e73bcf` (#918)](https://github.com/meta-pytorch/torchcodec/pull/918/commits/8e73bcfe0204b1b8eb69a9a379efab811fc38172)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -105,27 +112,29 @@ static UniqueCUvideodecoder createDecoder(CUVIDEOFORMAT* videoFormat) {
...
-  decoder_info.ulHeight = videoFormat->coded_height;
-  decoder_info.ulWidth = videoFormat->coded_width;
-  decoder_info.ulMaxHeight = videoFormat->coded_height;
-  decoder_info.ulMaxWidth = videoFormat->coded_width;
-  decoder_info.ulTargetHeight =
+  CUVIDDECODECREATEINFO decoderParams = {};
+  decoderParams.bitDepthMinus8 = videoFormat->bit_depth_luma_minus8;
+  decoderParams.ChromaFormat = videoFormat->chroma_format;
+  decoderParams.OutputFormat = cudaVideoSurfaceFormat_NV12;
+  decoderParams.ulCreationFlags = cudaVideoCreate_Default;
```
NicolasHug, line 116: It's not obvious from the diff because of the drive-by renaming, but these are the 2 lines that were added. They are used by default whether we explicitly set them or not (their enum value is 0 and we zero-initialize the `decoderParams`). But I think setting them explicitly makes sense, if only for the reader to know that they exist.

PR #919, opened 2025-10-03
BETA CUDA interface: H265 support

NicolasHug:
This PR adds support for H265 videos. It also creates a new `initializeBSF()` method: previously, we were hard-coding the BSF for H264. But H265 needs a different kind of BSF, and some format do not need any. So this PR allows for that more general logic.

This PR is based on  https://github.com/meta-pytorch/torchcodec/pull/917 so if it's not already merged, you can review it independently by looking at [`12c75e7` (#919)](https://github.com/meta-pytorch/torchcodec/pull/919/commits/12c75e76b21173291d72d838ec366c1c15e42f69#diff-17ebced5ec08998e4778545bbafbfbfe80fe0af466d44aea0b274161431317da)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -163,29 +181,62 @@ BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
...
-  // https://ffmpeg.org/doxygen/7.0/group__lavc__bsf.html
-  // This is only needed for some formats, like H264 or HEVC.  TODONVDEC P1: For
-  // now we apply BSF unconditionally, but it should be optional  and dependent
-  // on codec and container.
-  const AVBitStreamFilter* avBSF = av_bsf_get_by_name("h264_mp4toannexb");
+    default:
+      // No bitstream filter needed for other codecs
+      // TODONVDEC P1 MPEG4 will need one!
+      break;
+  }
```
scotts, line 267: Nit: I would prefer putting this switching logic into a function - I know we have a style difference there. :) My rationale is that I find it useful to think in terms of pure functions when I can, and this can definitely be a pure function, and then we can simply say in this scope:
```c++
std::string filterName = toFilterName(codecPar->codec_id);
```

src/torchcodec/_core/BetaCudaDeviceInterface.h
```diff
@@ -37,7 +37,9 @@ class BetaCudaDeviceInterface : public DeviceInterface {
   explicit BetaCudaDeviceInterface(const torch::Device& device);
   virtual ~BetaCudaDeviceInterface();
 
-  void initializeInterface(AVStream* stream) override;
+  void initializeInterface(
+      const AVStream* stream,
+      const UniqueDecodingAVFormatContext& avFormatCtx) override;
```
scotts, line 42: I just merged #902, so there's going to be merge conflicts. It should be easy to solve: we just rename `initializeInterface()` to `initialize()`, and keep the new param.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: IIUC we'd need `toFilterName` to return `""` when no match exist, and it would be up to the caller to check against that?

PR #920, opened 2025-10-03
BETA CUDA interface: AV1 support 

NicolasHug:
Needs #919 first.

PR #921, opened 2025-10-03
BETA CUDA interface: integrate with our existing tests and fix EOF hang

NicolasHug:
This PR lets our existing test suite check the BETA CUDA interface. The good news is that we pass the vast majority of the tests. ~The two failures, which are currently skipped with P0 TODOs, are related to end-of-stream logic. I'll follow-up with that in another PR.~ EDIT: I ended up fixing them here. We now pass all the tests!

This PR is based on https://github.com/meta-pytorch/torchcodec/pull/920, so to review it independently please look at  [`f8f0402` (#921)](https://github.com/meta-pytorch/torchcodec/pull/921/commits/f8f04028b2f840d164e76558b4fef618e4fda86e#diff-17ebced5ec08998e4778545bbafbfbfe80fe0af466d44aea0b274161431317da)


test/utils.py
```diff
@@ -27,7 +27,27 @@ def needs_cuda(test_item):
...
+    # helper meant to be used as
+    # device, device_variant = unsplit_device_str(device)
+    # when `device` comes from all_supported_devices() and may be "cuda:0:beta".
+    # It is used:
+    # - before calling `.to(device)` where device can't be "cuda:0:beta"
+    # - before calling add_video_stream(device=device, device_variant=device_variant)
+    #
+    # TODONVDEC P2: Find a less clunky way to test the BETA CUDA interface. It
+    # will ultimately depend on how we want to publicly expose it.
+    if device_str == "cuda:0:beta":
```
scotts, line 47: Does this mean that our testing will only work on CUDA device 0?

scotts (approved):

test/utils.py
NicolasHug: kinda. It's more that this line makes the assumption that we only ever test the beta interface as `cuda:0:beta` and never as `cuda:1:beta`.

We've never tested anything on `cuda:1` so far, even for the default interface, but if we ever do this will be a trivial fix.

PR #922, opened 2025-10-03
BETA CUDA interface: Cleanup EOF packet sending logic

NicolasHug:
See https://github.com/meta-pytorch/torchcodec/commit/cbfaaa7ac49f0cd2793a58ae080e91a4c757c4d1


src/torchcodec/_core/DeviceInterface.h
```diff
@@ -98,12 +86,18 @@ class DeviceInterface {
...
   }
 
+  // Send an EOF packet to flush the decoder
+  // Returns AVSUCCESS on success, or other AVERROR on failure
+  virtual int sendEOFPacket() {
+    TORCH_CHECK(
+        false, "Send EOF packet not implemented for this device interface");
+    return AVERROR(ENOSYS);
+  }
+
```
scotts, line 110: If `SingleStreamDecoder` doesn't need to call a member function, then maybe we don't need to make it a part of `DeviceInterface`? Since `sendEOFPacket()` is only called from within `BetaCudaDeviceInterface`, it feels more like an implementation detail for that device.

PR #930, opened 2025-10-04
BETA CUDA interface: properly set NVDEC stream and wait on it

NicolasHug:
We fixed a bug in https://github.com/meta-pytorch/torchcodec/pull/868 a few weeks ago for the default CUDA interface. The problem was that we needed the NPP stream to wait on the NVDEC stream.

This PR makes sure this is done as well for the BETA CUDA interface.

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -426,35 +426,36 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
     dst = allocateEmptyHWCTensor(frameDims, device_);
   }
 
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -426,35 +426,36 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
     dst = allocateEmptyHWCTensor(frameDims, device_);
   }
```
NicolasHug, line 428: Note for below: that's introducing yet another dependency relationship between the BETA interface and the default interface. All of this badness is resolved in https://github.com/meta-pytorch/torchcodec/pull/931

scotts (approved):

PR #931, opened 2025-10-04
BETA CUDA interface: separate it from the default interface

NicolasHug:
This PR is based on https://github.com/meta-pytorch/torchcodec/pull/930, so to review it, look at [`68878f2` (#931)](https://github.com/meta-pytorch/torchcodec/pull/931/commits/68878f22b36247620444b9c5efee2dcbcb3c20dd)

The PR looks scary but it's mostly just moving code around, and deleting / simplifying some parts. I added comments below to guide the review.

----

Currently in `main`, the BETA CUDA interface holds a default CUDA interface member, and uses it for the color-conversion step `convertAVFrameToFrameOutput`. This is problematic, because:

- It forces the `BetaCudaDeviceInterface` to do some fake dummy initialization of the `CudaDeviceInterface` class
- More importantly, it forces the `CudaDeviceInterface` to *know* that it may be called by `BetaCudaDeviceInterface`. This led to a bunch of `if/else` logic that is confusing, and a clear anti-pattern. The interfaces should be independent of each others.

This PR resolves this problem by:
- Removing the dependency of `BetaCudaDeviceInterface` on `CudaDeviceInterface`
- Removing the adhoc `if/else` logic in  `CudaDeviceInterface`.  `CudaDeviceInterface` doesn't need to worry about the fact that it might be called by a `BetaCudaDeviceInterface`.
- Moving common parts to both interfaces into a new `CUDACommon.cpp` file.

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -11,7 +11,6 @@
 
 #include "src/torchcodec/_core/BetaCudaDeviceInterface.h"
 
-#include "src/torchcodec/_core/CUDACommon.h"
 #include "src/torchcodec/_core/DeviceInterface.h"
 #include "src/torchcodec/_core/FFMPEGCommon.h"
 #include "src/torchcodec/_core/NVDECCache.h"
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -11,6 +11,7 @@
 
 #include "src/torchcodec/_core/BetaCudaDeviceInterface.h"
 
+#include "src/torchcodec/_core/CUDACommon.h"
 #include "src/torchcodec/_core/DeviceInterface.h"
 #include "src/torchcodec/_core/FFMPEGCommon.h"
 #include "src/torchcodec/_core/NVDECCache.h"
@@ -182,6 +183,12 @@ BetaCudaDeviceInterface::BetaCudaDeviceInterface(const torch::Device& device)
   TORCH_CHECK(g_cuda_beta, "BetaCudaDeviceInterface was not registered!");
   TORCH_CHECK(
       device_.type() == torch::kCUDA, "Unsupported device: ", device_.str());
+
+  // Initialize CUDA context with a dummy tensor
+  torch::Tensor dummyTensorForCudaInitialization = torch::empty(
+      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device_));
+
+  nppCtx_ = getNppStreamContext(device_);
 }
 
 BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
@@ -202,21 +209,13 @@ BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
     cuvidDestroyVideoParser(videoParser_);
     videoParser_ = nullptr;
   }
+
+  returnNppStreamContextToCache(device_, std::move(nppCtx_));
 }
 
 void BetaCudaDeviceInterface::initialize(
     const AVStream* avStream,
     const UniqueDecodingAVFormatContext& avFormatCtx) {
-  torch::Tensor dummyTensorForCudaInitialization = torch::empty(
-      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device_));
-
-  auto cudaDevice = torch::Device(torch::kCUDA);
-  defaultCudaInterface_ =
-      std::unique_ptr<DeviceInterface>(createDeviceInterface(cudaDevice));
-  AVCodecContext dummyCodecContext = {};
-  defaultCudaInterface_->initialize(avStream, avFormatCtx);
-  defaultCudaInterface_->registerHardwareDeviceWithCodec(&dummyCodecContext);
-
   TORCH_CHECK(avStream != nullptr, "AVStream cannot be null");
   timeBase_ = avStream->time_base;
   frameRateAvgFromFFmpeg_ = avStream->r_frame_rate;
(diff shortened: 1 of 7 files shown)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -182,6 +183,12 @@ BetaCudaDeviceInterface::BetaCudaDeviceInterface(const torch::Device& device)
   TORCH_CHECK(g_cuda_beta, "BetaCudaDeviceInterface was not registered!");
   TORCH_CHECK(
       device_.type() == torch::kCUDA, "Unsupported device: ", device_.str());
+
+  // Initialize CUDA context with a dummy tensor
+  torch::Tensor dummyTensorForCudaInitialization = torch::empty(
+      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device_));
+
+  nppCtx_ = getNppStreamContext(device_);
```
NicolasHug, line 210: Above: moved the dummy CUDA context initialization to the constructor, like how it was already done for the CudaDeviceInterface.

Also, this BETACudaInterface now needs its own NPP context. Previously, it was relying on its CudaDeviceInterface member (now removed).

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -603,15 +605,19 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
...
-  // the color conversion stuff separately.
-  defaultCudaInterface_->convertAVFrameToFrameOutput(
-      avFrame, frameOutput, preAllocatedOutputTensor);
+  validatePreAllocatedTensorShape(preAllocatedOutputTensor, avFrame);
+
+  at::cuda::CUDAStream nvdecStream =
+      at::cuda::getCurrentCUDAStream(device_.index());
+
+  frameOutput.data = convertNV12FrameToRGB(
+      avFrame, device_, nppCtx_, nvdecStream, preAllocatedOutputTensor);
```
NicolasHug, line 636: Above:

- we now call `convertNV12FrameToRGB()` for color conversion, instead of calling `defaultCudaInterface_->convertAVFrameToFrameOutput`.  `convertNV12FrameToRGB()` is now common to both interfaces.
- previously, the stream synchronization between NVDEC and NPP was done within `defaultCudaInterface_->convertAVFrameToFrameOutput()`. And `defaultCudaInterface_->convertAVFrameToFrameOutput` had to know that the NVDEC stream was: `0` for the `CudaDeviceInterface`, or the default stream for `BetaCudaDeviceInterface`. Now, each interface explicitly specifies what the NVDEC stream is by passing it down to `convertNV12FrameToRGB`, and this is now where the stream synchronization happens.
- On the TODONVDEC P2: I am doing further investigations but my current understanding is that the new BetaCudaDeviceInterface will never need to call `maybeConvertAVFrameToNV12OrRGB24` - which is a GOOD thing!

src/torchcodec/_core/CUDACommon.cpp
NicolasHug, line 1: 99% of this file is copy/pasted from what previously was inside of `CudaDeviceInterface`.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -13,15 +14,6 @@ extern "C" {
 #include <libavutil/pixdesc.h>
 }
```
NicolasHug, line 14: Below: the stuff that was removed from this file was either:

- logic that was specific for the `BetaCudaDeviceInterface`: not needed anymore now that we have removed that dependency
- moved into `CUDACommon.cpp`.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -401,127 +325,39 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
   // also need to check that the AVFrame is in AV_PIX_FMT_NV12 format (8 bits),
   // because this is what the NPP color conversion routines expect. This SHOULD
   // be enforced by our call to maybeConvertAVFrameToNV12OrRGB24() above.
```
NicolasHug, line 325: Below:

Note how we removed the `if (avFrame->hw_frames_ctx != nullptr) ...` which was needed because this could be called from the Beta interface, where `avFrame->hw_frames_ctx` would be null. Now we don't need to check for that anymore.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -401,127 +325,39 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
-    dst = allocateEmptyHWCTensor(frameDims, device_);
-  }
+  AVPixelFormat actualFormat = hwFramesCtx->sw_format;
+  TORCH_CHECK(
+      actualFormat == AV_PIX_FMT_NV12,
+      "The AVFrame is ",
+      (av_get_pix_fmt_name(actualFormat) ? av_get_pix_fmt_name(actualFormat)
+                                         : "unknown"),
+      ", but we expected AV_PIX_FMT_NV12. "
+      "That's unexpected, please report this to the TorchCodec repo.");
```
NicolasHug, line 343: Below: we would previously do the NVDEC / NPP stream synchronization here. Now, it's done a bit later inside of `convertNV12FrameToRGB`. So this portion just becomes about figuring out what the `NVDEC` stream is so we can pass it down to `convertNV12FrameToRGB`.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -401,126 +325,39 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
...
-        ", but we expected AV_PIX_FMT_NV12. "
-        "That's unexpected, please report this to the TorchCodec repo.");
-  }
-
-  torch::Tensor& dst = frameOutput.data;
-  if (preAllocatedOutputTensor.has_value()) {
-    dst = preAllocatedOutputTensor.value();
-  } else {
-    dst = allocateEmptyHWCTensor(frameDims, device_);
-  }
```
NicolasHug, line 427: This validation above was moved inside of ` convertNV12FrameToRGB`

src/torchcodec/_core/BetaCudaDeviceInterface.h
```diff
@@ -15,6 +15,8 @@
 
 #pragma once
 
+#include <npp.h>
```
scotts: This include isn't strictly needed here as `CUDACommon.h` includes it.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -11,6 +11,7 @@
 
 #include "src/torchcodec/_core/BetaCudaDeviceInterface.h"
 
+#include "src/torchcodec/_core/CUDACommon.h"
```
scotts: This include isn't strictly needed as `BetaCudaDeviceInterface.h` includes it.

PR #932, opened 2025-10-04
Added proper tensor support for get_frames_played_at()

mollyxu:
Summary:
Modified get_frames_played_at in _video_decoder to accept tensors and updated all downstream functions to natively accept tensors rather than converting them to lists.


src/torchcodec/_core/ops.py
```diff
@@ -212,6 +212,22 @@ def get_frames_at_indices(
...
+    if isinstance(timestamps, torch.Tensor):
+        # Ensure indices is the correct dtype (float64)
+        timestamps = timestamps.to(torch.float64)
+    else:
+        # Convert list to tensor for dispatch
+        try:
+            timestamps = torch.tensor(timestamps, dtype=torch.float64)
+        except (ValueError, TypeError):
+            # Type validation in C++ layer
+            pass
```
NicolasHug: I think we should try to avoid silencing exceptions here, can you share more about the kind of errors we're expecting?

NicolasHug (commented):
Thanks @mollyxu - this looks great, I do have a question about the Exception catching below. LMK your thoughts!

src/torchcodec/_core/ops.py
mollyxu: Thank you for the feedback! 
In the case where `get_frames_by_pts` is called with `timestamps` that is not of type `list[float]` but a `list[str]` like in [test_ops.py](https://github.com/mollyxu/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/test/test_decoders.py#L697C1-L698C50), there should be a type error. 

We could catch these type errors by explicitly checking that the input is of type `list[float]` in `get_frames_by_pts` or we could let the C++ layer give the error message that would look something like the following: `"torchcodec_ns::get_frames_by_pts() Expected a value of type 'Tensor' for argument 'timestamps' but instead found type 'list'.\nPosition: 1\nValue: ['bad']\nDeclaration: torchcodec_ns::get_frames_by_pts(Tensor(a!) decoder, *, Tensor timestamps) -> (Tensor, Tensor, Tensor)\nCast error details: Unable to cast ['bad'] to Tensor"`

Do you think we should handle the exception here in the python layer to provide a more informative error message?

Dan-Flores:
To ensure we do not break this functionality in the future, we could add a test case to [test_get_frames_played_at](https://github.com/meta-pytorch/torchcodec/blob/6377dfcd7da9e42d97d47a2bbc8271e2bd799107/test/test_decoders.py#L649), where we pass in a tensor and ensure the outputs are what we expect.

src/torchcodec/_core/ops.py
NicolasHug: Thanks for the details!

So IIUC, we have a test that is asserting a certain kind of user-input error, and it now fails because we now raise a different kind of user-input error.

Generally when that happens, we should just update the test itself, not the code. The idea is that the code is still fine: it's still raising an error like it's supposed to. Maybe *where* the error happens has changed, but that's not what matters. As long as the new error message is reasonably informative, that's fine.

Another principle: we do try to validate input as late as possible, but if we *know* for sure that some input is bad, we should handle it immediately. We should try to avoid passing down input that we know is going to fail.

------


In this specific case, the error message seems pretty uninformative:

```py
In [5]: torch.tensor(["bad"], dtype=torch.float64)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 torch.tensor(["bad"], dtype=torch.float64)

ValueError: too many dimensions 'str'
```

So yes, I think we should do what you suggested and handle the exception here in Python. I don't think we should explicitly assert that we have `list[float]`, because passing a `list[int]` is reasonable and should be supported. So we can be agnostic in the type of error that could happen and do something like this:

```py
        try:
            timestamps = torch.tensor(timestamps, dtype=torch.float64)
        except Exception as e:
            raise ValueError("Couldn't convert timestamps input to a tensor") from e
```

and then update the test accordingly to catch that error message.

mollyxu: Thank you for the feedback!

PR #934, opened 2025-10-05
BETA CUDA interface: clarify 10->8 bit conversions

NicolasHug:


test/test_decoders.py
```diff
@@ -1674,12 +1671,11 @@ def test_beta_cuda_interface_backwards(self, asset, seek_mode):
 
     @needs_cuda
     def test_beta_cuda_interface_small_h265(self):
-        # TODONVDEC P2 investigate why/how the default interface can decode this
-        # video.
+        # Test to illustrate current difference in behavior between the BETA and
+        # the default interface: this video isn't supported by NVDEC, but in the
+        # default interface, FFMPEG fallsback to the CPU while we don't.
```
NicolasHug, line 1676: Above is a drive-by, I was able to confirm that the default interface can decode this video because it falls back to the CPU. I will open a separate issue to discuss the CPU fallback behavior - I think we'll want to do a few changes eventually.

test/test_decoders.py
```diff
@@ -1288,18 +1288,15 @@ def test_10bit_videos(self, device, asset):
...
             #
-            # I don't remember but I suspect asset 0 is actually the one that
-            # fallsback to the CPU path on the default CUDA interface (that
-            # would make sense)
-            # We should investigate if and how we could make that fallback
-            # happen for the BETA interface.
-            pytest.skip("TODONVDEC P2 - investigate and unskip")
+            # It works on the default interface because FFmpeg fallsback to the
+            # CPU, while the BETA interface doesn't.
+            pytest.skip("Asset not supported by NVDEC")
```
NicolasHug, line 1299: Note above how we're now *not* skipping this test anymore for the `H265_10BITS` asset. Support for `H265_10BITS` isn't introduced by this PR, it was already working before, but now we know we can safely unskip it.


Also, this isn't visible in this PR and not in scope anyway, but there's something interesting: on `H265_10BITS`, the frame we decode with the BETA interface looks good, but it is slightly different than the on from the default CUDA interface. The difference is due to the fact that in the BETA interface, the 10 -> 8 bit conversion is done by NVDEC itself, while in the default interface it's done later via filtergraph in `convertAVFrameToFrameOutput()`.


So overall we should expect small differences for 10bit videos between the BETA and default interface. I will follow-up on that by creating an issue. Generally, the BETA interfaces allows for drastically simpler code (no explicit conversion needed!)

PR #938, opened 2025-10-07
Expose TorchCodecConfig.cmake

dvrogozh:
This commit exposes torchcodec core library to be used by third party modules on the C++ level. The primary purpose is to allow non-CUDA device interfaces out-of-tree implementations. There are the following major changes:

* Exposed TorchCodecConfig.cmake which defines torchcodec targets to be linked with

* Provided Python level APIs to faciliate out-of-tree device interfaces work with torchcodec:

  * `torchcodec.cmake_prefix_path` - path which points to `TorchCodecConfig.cmake` configuration
  * `torchcodec.variant` - variant of the torchcodec library which was loaded, i.e. N in libtorchcodec_core{N}.so (currently ffmpeg_major_version)
  * `torchcodec.core_library_path` - full path of the loaded torchcodec core library

* `src/torchcodec/_core/` dropped from include paths to allow using of the core library out-of-tree

`TorchCodecConfig.cmake` has 2 working modes:

* By default config works by checking available version of FFmpeg libraries via `pkg-config` and configures corresponding (single) version of torchcodec
* Altenatively, if `TORCHCODEC_FFMPEG{N}_INSTALL_PREFIX` is set (`N=4,5,6,7` - version of FFmpeg), then config defines torchcodec target corresponding to the specified FFmpeg version. Note that multiple prefixes can be specified at the same time allowing to build against few torchcodec versions at once.

Config will define `TORCHCODEC_VARIANTS` variable with value corresponding to FFmpeg major versions of available torchcodec core libraries. Further, config will also define `torchcodec::ffmpeg${N}` and `torchcodec::core${N}` targets where `N` takes values from `TORCHCODEC_VARIANTS`.

See the following repository for an actual out-of-tree device interface torchcodec plugin:
* https://github.com/dvrogozh/torchcodec-xpu

I suggest to pay attention on these:
* https://github.com/dvrogozh/torchcodec-xpu/blob/59bfef60e9b41244e804c112729c42d9176ae368/src/torchcodec_xpu/__init__.py#L44 for plugin ;loading in respect to the loaded torchcodec library
* https://github.com/dvrogozh/torchcodec-xpu/blob/59bfef60e9b41244e804c112729c42d9176ae368/src/torchcodec_xpu/CMakeLists.txt#L81 for building multiple plugin targets depending on the available ffmpeg and torchcodec versions

CC: @scotts, @NicolasHug

### DIFF
--- a/src/torchcodec/__init__.py
+++ b/src/torchcodec/__init__.py
@@ -8,7 +8,6 @@ from pathlib import Path
 
 # Note: usort wants to put Frame and FrameBatch after decoders and samplers,
 # but that results in circular import.
-from ._core import core_library_path, ffmpeg_major_version
 from ._frame import AudioSamples, Frame, FrameBatch  # usort:skip # noqa
 from . import decoders, encoders, samplers  # noqa
 
@@ -18,9 +17,11 @@ try:
 except Exception:
     pass
 
-# `torchcodec.cmake_prefix_path` is a Python-based way to programmatically
-# obtain the correct CMAKE_PREFIX_PATH value for the TorchCodec installation.
-# It can be used in a build system of a C++ application to ensure that CMake
-# can successfully find TorchCodec C++ libraries. This variable is exposed
-# as TorchCodec API.
+# cmake_prefix_path is needed for downstream cmake-based builds that use
+# torchcodec as a dependency to tell cmake where torchcodec is installed and where to find its
+# CMake configuration files.
+# Pytorch itself has a similar mechanism which we use in our setup.py!
 cmake_prefix_path = Path(__file__).parent / "share" / "cmake"
+# Similarly, these are exposed for downstream builds that use torchcodec as a
+# dependency.
+from ._core import core_library_path, ffmpeg_major_version  # usort:skip


src/torchcodec/__init__.py
```diff
@@ -14,3 +17,5 @@
     from .version import __version__  # noqa: F401
 except Exception:
     pass
+
+cmake_prefix_path = _osp.join(_osp.dirname(__file__), "share", "cmake")
```
NicolasHug: Let's add a comment to explain what this is used for, and also let's use the more modern `pathlib` tools instead of `os.path`.

src/torchcodec/_core/ops.py
```diff
@@ -47,14 +50,19 @@ def load_torchcodec_shared_libraries():
...
 
             pybind_ops_library_path = _get_extension_path(pybind_ops_library_name)
             global _pybind_ops
             _pybind_ops = _load_pybind11_module(
                 pybind_ops_module_name, pybind_ops_library_path
             )
+            global variant
+            global core_library_path
+            variant = ffmpeg_major_version
+            core_library_path = decoder_library_path
```
NicolasHug: Let's return them instead of having them as `global`.

Also, `ffmpeg_major_version` is a good name, so we can keep using it for the variable that we expose as `torchcodec.ffmpeg_major_version`

src/torchcodec/_core/ops.py
```diff
@@ -21,6 +22,8 @@
 
 _pybind_ops: Optional[ModuleType] = None
 
+variant = None
+core_library_path = None
```
NicolasHug: Can you help me understand why in 

https://github.com/dvrogozh/torchcodec-xpu/blob/59bfef60e9b41244e804c112729c42d9176ae368/src/torchcodec_xpu/__init__.py#L43

we only need to load the `libtorchcodec_core?.so` file, and not the other .so like `libtorchcodec_pybind_ops?.so` and `libtorchcodec_custom_ops?.so`

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+        )
+        set(f7_library_file_names
+          libavutil.so.59
+          libavcodec.so.61
+          libavformat.so.61
+          libavdevice.so.61
+          libavfilter.so.10
+          libswscale.so.8
+          libswresample.so.5
+        )
```
NicolasHug: Let's try to unify that with those that are already in https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/fetch_and_expose_non_gpl_ffmpeg_libs.cmake. We can create a new `ffmpeg_versions.cmake` file if that helps refactoring.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+
+        if (${libavcodec_major_version} STREQUAL "58")
+            set(ffmpeg_major_version "4")
+        elseif (${libavcodec_major_version} STREQUAL "59")
+            set(ffmpeg_major_version "5")
+        elseif (${libavcodec_major_version} STREQUAL "60")
+            set(ffmpeg_major_version "6")
+        elseif (${libavcodec_major_version} STREQUAL "61")
+            set(ffmpeg_major_version "7")
+        endif()
```
NicolasHug: Same here, the logic in this whole branch should ideally be unified with https://github.com/meta-pytorch/torchcodec/blob/75a3325f016483ae1a6af8516b804cc72da39ea6/src/torchcodec/_core/CMakeLists.txt#L276-L315

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+        return()
+    endif()
+
+    set(prefix "$ENV{${prefix}}")
+    set(incdir "${prefix}/include")
+    if (UNIX AND NOT APPLE)
+        set(libdir "${prefix}/lib")
+    else()
+        message("Skipping ${target} on non-Linux platform")
+        return()
```
NicolasHug: Let's just assert this is `UNIX AND NOT APPLE` at the top of this file. It's OK to only support linux for now.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+
+# Include directories.
+set(TORCHCODEC_INCLUDE_DIRS ${TORCHCODEC_INSTALL_PREFIX}/_core)
+set(TORCHCODEC_VARIANTS "")
+
+function(add_ffmpeg_target ffmpeg_major_version libs)
+    set(target "torchcodec::ffmpeg${ffmpeg_major_version}")
+    set(prefix "TORCHCODEC_FFMPEG${ffmpeg_major_version}_INSTALL_PREFIX")
+    if (NOT DEFINED ENV{${prefix}})
+        message("Skipping ${target} as ${prefix} is not defined")
```
NicolasHug: This should be an error I think?

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+    else()
+        message("Skipping ${target} on non-Linux platform")
+        return()
+    endif()
+
+    set(lib_paths "")
+    foreach(lib IN LISTS libs)
+        find_library(_LIB_PATH "${lib}" PATHS "${libdir}" NO_DEFAULT_PATH)
+	if (NOT _LIB_PATH)
+            message("Skipping ${target} as ${lib} is missing")
```
NicolasHug: Here as well, we'll want to error

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
```diff
@@ -0,0 +1,185 @@
...
+            return()
+        else()
+            list(APPEND lib_paths "${_LIB_PATH}")
+	endif()
+        # Removing _LIB_PATH from cache otherwise it won't be updated
+        # on the next call to find_library().
+        unset(_LIB_PATH CACHE)
+    endforeach()
```
NicolasHug: Actually, I might not completely understand what this branch does, but can we it seems to do very similar logic to what already exists in 

https://github.com/meta-pytorch/torchcodec/blob/75a3325f016483ae1a6af8516b804cc72da39ea6/src/torchcodec/_core/CMakeLists.txt#L271-L275

and in 

```
      add_ffmpeg_target(4 "${f4_library_file_names}")
      add_ffmpeg_target(5 "${f5_library_file_names}")
      add_ffmpeg_target(6 "${f6_library_file_names}")
      add_ffmpeg_target(7 "${f7_library_file_names}")
```


Can we unify these?

dvrogozh:
> On the `#include` path changes, I will have to pre-import your PR internall to verify that this isn't breaking our internal build (and to try to find internal workarounds if needed).

I've rebased the PR and resolved conflicts, but I will take a look and start address review comments later today or early next week. Please, let me know how `#include` story goes. If this change is ok (I hope so), then we may consider to make it separately to reduce the scope of this PR and simplify/focus review on the cmake aspects.

NicolasHug:
Can confirm the header `#include` changes are OK internally! Yes, happy to consider a separate diff that does just that. Thank you!

dvrogozh:
> Can confirm the header #include changes are OK internally! Yes, happy to consider a separate diff that does just that.

Thank you! Here is includes PR: #1002.

src/torchcodec/__init__.py
dvrogozh: done

src/torchcodec/_core/ops.py
dvrogozh: done

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: done, I've created `ffmpeg_versions.cmake` and moved as much common stuff there as I could.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: Ok, I've missed this one. I will need to look into it.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: Actually, this got covered after creating `ffmpeg_versions.cmake`.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: done.

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: done, see `ffmpeg_versions.cmake`

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: done for ffmpeg targets

src/torchcodec/share/cmake/TorchCodec/TorchCodecConfig.cmake
dvrogozh: Done now. This also went to `ffmpeg_versions.cmake`

src/torchcodec/_core/ops.py
dvrogozh: That's because `torchcodec_xpu` does not use any of TorchCodec Python/C++ APIs defined in `libtorchcodec_pybind_ops?.so` or `libtorchcodec_custom_ops?.so`. The only thing which `torchcodec_xpu` does is registering XPU device interface on the C++ level, i.e. calling analogue of:
https://github.com/meta-pytorch/torchcodec/blob/169484d24e63311a3891d36643818d2026677c19/src/torchcodec/_core/CudaDeviceInterface.cpp#L18-L22

Note that if we will consider to support use cases when some application wants to link against TorchCodec C++ APIs, then `libtorchcodec_pybind_ops?.so` or `libtorchcodec_custom_ops?.so` or both might be needed. In this case the `TorchCodecConfig.cmake` will need to be updated to expose these targets as well.

dvrogozh:
@NicolasHug, I've addressed all your comments and completed the 2nd iteration of the PR. Can you, please, help to review it again?

PR #940, opened 2025-10-08
Fix internal linter

NicolasHug:


src/torchcodec/_core/DeviceInterface.h
```diff
@@ -32,8 +32,8 @@ struct DeviceInterfaceKey {
 
   explicit DeviceInterfaceKey(torch::DeviceType type) : deviceType(type) {}
 
-  DeviceInterfaceKey(torch::DeviceType type, const std::string_view& var)
-      : deviceType(type), variant(var) {}
+  DeviceInterfaceKey(torch::DeviceType type, const std::string_view& variant)
+      : deviceType(type), variant(variant) {}
```
NicolasHug, line 36: Internal linter is complaining that we're doing

```Cpp
DeviceInterfaceKey(torch::kCUDA, /*variant=*/"beta"),
```

and that our `/*variant=*/` comment doesn't match the name of the parameter which is `var`.

I'm choosing to change the parameter name to from `var` to `variant`, instead of updating the comment: we use "variant" everywhere. "var" was only used for the initialization list *not* to look like `variant(variant)`, which ... it does now. 🤷‍♂️

PR #941, opened 2025-10-08
Skip CUDA AV1 tests in fbcode

NicolasHug:
These are failing internally

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -591,7 +591,7 @@ class TestVideoDecoder:
         if device == "cuda" and get_ffmpeg_major_version() == 4:
             return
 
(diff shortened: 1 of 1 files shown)


test/test_decoders.py
```diff
@@ -591,7 +591,7 @@ def test_get_frame_at_av1(self, device):
         if device == "cuda" and get_ffmpeg_major_version() == 4:
             return
 
-        if device == "cuda" and in_fbcode():
+        if "cuda" in device and in_fbcode():
```
NicolasHug, line 594: And the internal issue with CUDA and AV1 isn't new. It's not related to the BETA interface as can be seen here. We were already skipping the AV1 CUDA test for the default interface.

PR #954, opened 2025-10-10
Refactor receiveFrame and sendPacket logic to dispatch directly to interface

mollyxu:
Removed canDecodePacketDirectly() and refactored decoding loop's receiveFrame and sendPacket logic to dispatch directly to interface

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -440,8 +440,7 @@ void SingleStreamDecoder::addStream(
 
   AVCodecContext* codecContext = avcodec_alloc_context3(avCodec);
   TORCH_CHECK(codecContext != nullptr);
-  streamInfo.codecContext = SharedAVCodecContext(
-      codecContext, [](AVCodecContext* ctx) { avcodec_free_context(&ctx); });
+  streamInfo.codecContext = makeSharedAVCodecContext(codecContext);
 
   int retVal = avcodec_parameters_to_context(
       streamInfo.codecContext.get(), streamInfo.stream->codecpar);
--- a/src/torchcodec/_core/FFMPEGCommon.h
+++ b/src/torchcodec/_core/FFMPEGCommon.h
@@ -73,6 +73,12 @@ using UniqueAVCodecContext = std::unique_ptr<
     Deleterp<AVCodecContext, void, avcodec_free_context>>;
 using SharedAVCodecContext = std::shared_ptr<AVCodecContext>;
 
+// create SharedAVCodecContext with custom deleter
+inline SharedAVCodecContext makeSharedAVCodecContext(AVCodecContext* ctx) {
+  return SharedAVCodecContext(
+      ctx, Deleterp<AVCodecContext, void, avcodec_free_context>{});
+}
+
 using UniqueAVFrame =
     std::unique_ptr<AVFrame, Deleterp<AVFrame, void, av_frame_free>>;
 using UniqueAVFilterGraph = std::unique_ptr<


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1169,24 +1173,16 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
...
   bool reachedEOF = false;
 
-  // TODONVDEC P2: Instead of calling canDecodePacketDirectly() and rely on
-  // if/else blocks to dispatch to the interface or to FFmpeg, consider *always*
-  // dispatching to the interface. The default implementation of the interface's
-  // receiveFrame and sendPacket could just be calling avcodec_receive_frame and
-  // avcodec_send_packet. This would make the decoding loop even more generic.
+  // The default implementation uses avcodec_receive_frame and
+  // avcodec_send_packet, while specialized interfaces can override for
+  // hardware-specific optimizations.
```
NicolasHug, line 1181: Nicely done, this looks much nicer now. There's one more place where we could unify the calls, that I forgot to point out in the TODO:

https://github.com/meta-pytorch/torchcodec/blob/e5b2eefed5d186e1232da15b1368b4561c052bae/src/torchcodec/_core/SingleStreamDecoder.cpp#L1152-L1154

This should just become `deviceInterface_->flush();` and the default implementation in the base `DeviceInterace` class should just be `avcodec_flush_buffers(streamInfo.codecContext.get()); `.

The current code in `main` doesn't do `if (deviceInterface_->canDecodePacketDirectly()) ...` and just runs both calls - that's not intentional, it's more of a miss on my part. Really the `deviceInterface_->flush()` call is currently only relevant for the BETA interface while the `avcodec_flush_buffers(streamInfo.codecContext.get()); ` call is relevant for all other interfaces.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -126,6 +129,7 @@ class DeviceInterface {
 
  protected:
   torch::Device device_;
+  AVCodecContext* codecContext_ = nullptr; // Non-owning pointer
```
NicolasHug: I think we have 3 main options here:

- Store a raw non-owning pointer like you did
- make `codecContext_` a `shared_ptr` that is shared by both by the `DeviceInterface` and the `SingleStreamDecoder`
- Only store the `codecContext_` within  `SingleStreamDecoder`, and pass it as a reference parameter to all of the `DeviceInterface` methods. E.g. `deviceInterface_->sendPacket(packet)` would become `deviceInterface_->sendPacket(codecContext_, packet)`.

I have a slight preference for the third option because not having to deal with shared or raw pointers is... simpler. It makes the method calls slightly more ugly and we'll have to use `[[maybe_unused]]` in a few places, but I think it's worth it.

Curious if @scotts has any thought on this?

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -80,42 +80,45 @@ class DeviceInterface {
...
-        false,
-        "Send/receive packet decoding not implemented for this device interface");
-    return AVERROR(ENOSYS);
+  // Default implementation uses FFmpeg directly
+  virtual int sendPacket(ReferenceAVPacket& avPacket) {
+    if (!codecContext_) {
+      TORCH_CHECK(
+          false, "Codec context not available for default packet sending");
+      return AVERROR(EINVAL);
+    }
```
NicolasHug: Here and below I think this can just be 


```suggestion
      TORCH_CHECK(
          codecContext_ != nullptr, "Codec context not available for default packet sending");

```

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -80,42 +80,45 @@ class DeviceInterface {
...
-  //
-  //   1. sendPacket()
-  //   2. receiveFrame()
-  virtual bool canDecodePacketDirectly() const {
-    return false;
+  // Set the codec context for default FFmpeg decoding operations
+  // This must be called during initialization before using
+  // sendPacket/receiveFrame
+  virtual void setCodecContext(AVCodecContext* codecContext) {
+    codecContext_ = codecContext;
```
NicolasHug: Assuming we don't store `codecContext_` within the interface, as suggested in another comment, we won't need this method anymore.

But if we decide to still store `codecContext_` I think we can avoid creating this method and just update the existing `initialize()` method.

src/torchcodec/_core/DeviceInterface.h
scotts: I think we want to do a version of the second option where:

* `SingleStreamDecoder` has a `std::shared_ptr<AVCodecContext>`. We'll want to define something like `SharedAVCodecContext` for that, similar to what we did for `UniqueAVCodecContext`.
* `DeviceInterface` will also own a `SharedAVCodecContext`.

In discussion, I had said we should use a `std::weak<AVCodecContext>` in `DeviceInterface`, but I don't think that's the case. We always need to convert a `std::weak` pointer to a `std::shared` pointer to use it. I think it's simpler to just make it a `std::shared` pointer in both places.

I prefer this for two reasons:

1. It accurately models the object ownership.
2. It keeps the APIs simpler. I think it's cleaner to initialize objects with things they will always need to know, and only pass in the "new" stuff. We can also establish invariants, like making sure that `codecContext_` is non-null during initialization.

mollyxu: Thanks for the feedback!

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -231,8 +231,10 @@ BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
 
 void BetaCudaDeviceInterface::initialize(
     const AVStream* avStream,
-    const UniqueDecodingAVFormatContext& avFormatCtx) {
+    const UniqueDecodingAVFormatContext& avFormatCtx,
+    const SharedAVCodecContext& codecContext) {
   TORCH_CHECK(avStream != nullptr, "AVStream cannot be null");
+  codecContext_ = codecContext;
```
NicolasHug: I don't think the BETA interface currently needs to store its own `codecContext_`, so maybe it's enough to use `[[maybe_unused]] const SharedAVCodecContext& codecContext) {` and not set the `codecContext_` ?

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -80,52 +81,48 @@ class DeviceInterface {
...
   // Send an EOF packet to flush the decoder
   // Returns AVSUCCESS on success, or other AVERROR on failure
+  // Default implementation uses FFmpeg directly
   virtual int sendEOFPacket() {
-    TORCH_CHECK(
-        false, "Send EOF packet not implemented for this device interface");
-    return AVERROR(ENOSYS);
+    if (!codecContext_) {
+      TORCH_CHECK(false, "Codec context not available for EOF packet sending");
+      return AVERROR(EINVAL);
```
NicolasHug: This can just be `TORCH_CHECK(codecContext_ != nullptr, ...)` like the rest, and we can remove `return AVERROR(EINVAL);`

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -158,6 +158,15 @@ int getNumChannels(const UniqueAVCodecContext& avCodecContext) {
...
 
+int getNumChannels(const SharedAVCodecContext& avCodecContext) {
+#if LIBAVFILTER_VERSION_MAJOR > 8 || \
+    (LIBAVFILTER_VERSION_MAJOR == 8 && LIBAVFILTER_VERSION_MINOR >= 44)
+  return avCodecContext->ch_layout.nb_channels;
+#else
+  return avCodecContext->channels;
+#endif
+}
+
```
NicolasHug, line 160: We don't seem to need the previous 

```
int getNumChannels(const UniqueAVCodecContext& avCodecContext);
```

anymore. Let's remove it if we can.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -453,14 +453,19 @@ void SingleStreamDecoder::addStream(
...
+      streamInfo.codecContext.get());
   retVal = avcodec_open2(streamInfo.codecContext.get(), avCodec, nullptr);
   TORCH_CHECK(retVal >= AVSUCCESS, getFFMPEGErrorStringFromErrorCode(retVal));
 
-  codecContext->time_base = streamInfo.stream->time_base;
+  streamInfo.codecContext->time_base = streamInfo.stream->time_base;
+
+  // Initialize the device interface with the codec context
+  deviceInterface_->initialize(
+      streamInfo.stream, formatContext_, streamInfo.codecContext);
```
NicolasHug, line 464: Just pointing for myself and for other reviewers that the call to `initialize()` was moved. Which makes sense, we now pass the codecContext to `initialize()`, and the codecContext is only initialized after the call to `avcodec_open2`.

We should just make sure there was no other call to the interface method that would depend on the interface being initialized. I think this is OK, the only call was `deviceInterface_->findCodec()`, and it's pretty much a stateless check.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -441,7 +440,8 @@ void SingleStreamDecoder::addStream(
 
   AVCodecContext* codecContext = avcodec_alloc_context3(avCodec);
   TORCH_CHECK(codecContext != nullptr);
-  streamInfo.codecContext.reset(codecContext);
+  streamInfo.codecContext = SharedAVCodecContext(
+      codecContext, [](AVCodecContext* ctx) { avcodec_free_context(&ctx); });
```
NicolasHug: It'd be nice if we could have this just be:


```Cpp
streamInfo.codecContext = SharedAVCodecContext(codecContext);
```

and have the definition of `SharedAVCodecContext` specify its own destructor. This is what we do for the other smart pointers in `FFmpegCommon.h`.

(It may not be possible or too clumsy, I gave it a quick try to better illustrate what I mean, and I wasn't able to make it work!)

mollyxu: I think the deleter is not a part of type `std::shared_ptr` so there would need to be a helper function `makeSharedAVCodecContext` in `FFMPEGCommon.h` and the usage would be 

`streamInfo.codecContext = makeSharedAVCodecContext(codecContext);`

Would this be preferred?

scotts: @mollyxu, take a look at `UniqueAVCodecContext`, which is defined right above your new `SharedAVCodecContext`. That makes the deleter part of the `using` statement, which enables us to just say `UniqueAVCodecContext(ptr)` in normal code.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: Yeah, I think we're good here. My rationale with the device was to initialize it as soon as we had everything we needed to do it.

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: I looked at this again because I wanted to figure out why I couldn't make it work yesterday. It seems we can't use the same template parameter syntax as we do for [`unique_ptr`,](https://en.cppreference.com/w/cpp/memory/unique_ptr.html) because [`shared_ptr`](https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr) only accepts the deleter in the constructor. So we can't define 

```Cpp
using SharedAVCodecContext = std::shared_ptr<
   AVCodecContext,
   Deleterp<AVCodecContext, void, avcodec_free_context>>;
```

which is what I was hoping for. That's just not supported by the `shared_ptr` type.

The closest I got is to define a construction helper

```Cpp
using SharedAVCodecContext = std::shared_ptr<AVCodecContext>;
inline SharedAVCodecContext makeSharedAVCodecContext(AVCodecContext* ctx) {
  return SharedAVCodecContext(ctx, Deleterp<AVCodecContext, void, avcodec_free_context>{});
}
```

which is then used as 

```Cpp
  streamInfo.codecContext = makeSharedAVCodecContext(codecContext);
```

(When I say *I*, I mean *Claude*).

It's very close to what you already have, the main difference is that we're using our `Deleterp` template instead of a lambda. I'd still suggest to use `makeSharedAVCodecContext`, so that all our smart pointer logic is centralized within the same `FFmpegCommon.h` file.

(If we move `makeSharedAVCodecContext` to the .cpp file we can remove the `inline`, but I don't think it's worth it).

scotts: @NicolasHug, thanks for digging into this! I am resisting the temptation to dive into the C++ docs to figure out _why_ these two seemingly complementary classes have different interfaces! Not a joke. :)

NicolasHug: I was also wondering and got this:

```
  Why the difference?

  The key difference is that std::unique_ptr needs to know the deleter type at compile time because it affects the
  size of the object. But std::shared_ptr uses type erasure - it stores the deleter internally in a way that
  doesn't affect the shared_ptr type itself.

  This means:
  - std::unique_ptr<int, CustomDeleter1> and std::unique_ptr<int, CustomDeleter2> are different types
  - std::shared_ptr<int> with CustomDeleter1 and std::shared_ptr<int> with CustomDeleter2 are the same type
```

But this isn't really a great reason. This explain why `unique_ptr` *cannot* be passed the deleter to its constructor. But it doesn't explain why `shared_ptr` doesn't support have the same syntax as `unique_ptr`. :thinking:

mollyxu: Thanks for the suggestions and for taking the time to look into this!

PR #957, opened 2025-10-14
Add to_tensor support for VideoEncoder

Dan-Flores:
This PR enables encoding a video to tensor by adding a `VideoEncoder` constructor that accepts an `AVIOContextHolder` and file `format`. It is instantiated with `AVIOToTensorContext`.

Testing:
* `test_video_encoder_round_trip` is updated to test `to_tensor` method.
* The changes in `AVIOTensorContext.cpp` enable certain formats, such as `avi` and `flv` to be encoded correctly. 
  * Although we do not test these container formats in `test_video_encoder_round_trip`, without the changes, they would raise an error when attempting to decode the encoded frames, indicating an incorrectly encoded video.
  * The added `max` field allows encoders that write the header last (doing a backwards seek) to output a correctly encoded tensor.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1330,6 +1330,7 @@ class TestAudioEncoderOps:
 
 class TestVideoEncoderOps:
 
+    # TODO-VideoEncoder: Parametrize test after moving to test_encoders
     def test_bad_input(self, tmp_path):
         output_file = str(tmp_path / ".mp4")
 
@@ -1445,7 +1446,16 @@ class TestVideoEncoderOps:
             assert_close(s_frame, rt_frame, atol=atol, rtol=0)
 
     @pytest.mark.parametrize(
-        "format", ("mov", "mp4", "avi", "mkv", "webm", "flv", "gif")
+        "format",
+        (
+            "mov",
+            "mp4",
+            "avi",
+            "mkv",
+            "flv",
+            "gif",
+            pytest.param("webm", marks=pytest.mark.slow),
+        ),
     )
     def test_against_to_file(self, tmp_path, format):
         # Test that to_file and to_tensor produce the same results
@@ -1463,7 +1473,10 @@ class TestVideoEncoderOps:
         encoded_tensor = encode_video_to_tensor(source_frames, format=format, **params)
 
         torch.testing.assert_close(
-            self.decode(encoded_file).data, self.decode(encoded_tensor).data
+            self.decode(encoded_file).data,
+            self.decode(encoded_tensor).data,
+            atol=0,
+            rtol=0,
         )
 
     @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")


test/test_ops.py
```diff
@@ -1399,14 +1404,25 @@ def test_video_encoder_round_trip(self, tmp_path, format):
...
+            encoded_path = str(tmp_path / f"encoder_output.{format}")
+            encode_video_to_file(
+                frames=source_frames,
+                frame_rate=frame_rate,
+                filename=encoded_path,
+                crf=0,
+            )
+            round_trip_frames = self.decode(file_path=encoded_path).data
+        else:  # to_tensor
+            format = "matroska" if format == "mkv" else format
```
Dan-Flores: The file extension `.mkv` is not accepted as a file format. In case we need the proper format name, this line swaps `mkv` for `matroska`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -543,9 +539,15 @@ torch::Tensor validateFrames(const torch::Tensor& frames) {
...
+      avio_flush(avFormatContext_->pb);
+    }
+    if (!avioContextHolder_) {
+      if (avFormatContext_->pb->error == 0) {
+        avio_close(avFormatContext_->pb);
+      }
+      avFormatContext_->pb = nullptr;
+    }
   }
 }
```
Dan-Flores, line 553: I've updated the destructor to match the `AudioEncoder` [destructor](https://github.com/meta-pytorch/torchcodec/blob/61202b9f7f823adedaa8097943ae4f4881e5da4a/src/torchcodec/_core/Encoder.cpp#L113-L127), but it took me a bit to understand the logic. Perhaps we could move the clean up logic into the various `AVIOContextHolder` derived classes?

src/torchcodec/_core/AVIOTensorContext.cpp
```diff
@@ -75,6 +75,11 @@ int write(void* opaque, const uint8_t* buf, int buf_size) {
   uint8_t* outputTensorData = tensorContext->data.data_ptr<uint8_t>();
   std::memcpy(outputTensorData + tensorContext->current, buf, bufSize);
   tensorContext->current += bufSize;
+  // Track the maximum position written so getOutputTensor's narrow() does not
+  // truncate the file if final seek was backwards
+  if (tensorContext->current > tensorContext->max) {
+    tensorContext->max = tensorContext->current;
+  }
```
NicolasHug: Oh, this looks like a bug-fix of my previous implementation, thanks!

test/test_ops.py
```diff
@@ -1378,14 +1379,18 @@ def test_bad_input(self, tmp_path):
                 filename="./bad/path.mp3",
             )
 
-    def decode(self, file_path) -> torch.Tensor:
-        decoder = create_from_file(str(file_path), seek_mode="approximate")
+    def decode(self, file_path=None, tensor=None) -> torch.Tensor:
+        if file_path is not None:
+            decoder = create_from_file(str(file_path), seek_mode="approximate")
+        elif tensor is not None:
+            decoder = create_from_tensor(tensor, seek_mode="approximate")
```
NicolasHug: This is starting to re-implement the logic that already existing within the `VideoDecoder` class. I think it might be  better to just use `VideoDecoder` here. These tests are for testing the `_core` **encoding** capabilities, not for testing the decoding capabilities, so using `VideoDecoder` as the decoding utility is fine. This is what we do for the audio encoder tests:

https://github.com/meta-pytorch/torchcodec/blob/b084768e63d8d200b2945cf659fc930cfe72b04b/test/test_encoders.py#L106-L109

test/test_ops.py
```diff
@@ -1378,14 +1379,18 @@ def test_bad_input(self, tmp_path):
...
         add_video_stream(decoder)
         frames, *_ = get_frames_in_range(decoder, start=0, stop=60)
         return frames
 
     @pytest.mark.parametrize("format", ("mov", "mp4", "mkv", "webm"))
-    def test_video_encoder_round_trip(self, tmp_path, format):
+    @pytest.mark.parametrize("output_method", ("to_file", "to_tensor"))
```
NicolasHug: nit: this is just called "method" in the audio encoder tests.

test/test_ops.py
```diff
@@ -1399,14 +1404,25 @@ def test_video_encoder_round_trip(self, tmp_path, format):
...
+                crf=0,
+            )
+            round_trip_frames = self.decode(file_path=encoded_path).data
+        else:  # to_tensor
+            format = "matroska" if format == "mkv" else format
+            encoded_tensor = encode_video_to_tensor(
+                source_frames, frame_rate, format, crf=0
+            )
+            round_trip_frames = self.decode(tensor=encoded_tensor).data
+
```
NicolasHug, line 1432: Instead of passing frame_rate and crf to both methods, let's have those inside a `params` dict as done in the AudioEncoder tests. It will make it clearer what are common parameters (that are only set once) vs the parameters that are specific to each method.

https://github.com/meta-pytorch/torchcodec/blob/b084768e63d8d200b2945cf659fc930cfe72b04b/test/test_encoders.py#L278-L291

src/torchcodec/_core/AVIOTensorContext.h
```diff
@@ -16,6 +16,7 @@ namespace detail {
 struct TensorContext {
   torch::Tensor data;
   int64_t current;
+  int64_t max;
```
NicolasHug: Now that we have two int values it might be worth giving them slightly more descriptive names. Here's a suggestion:

```suggestion
  int64_t current_pos;
  int64_t max_pos;
```

src/torchcodec/_core/Encoder.cpp
NicolasHug: To be fair, I wrote this, and I still don't fully understand it. This was the result of various trial-and-error attempts at resolving https://github.com/meta-pytorch/torchcodec/issues/724 , which is still an open issue.

Agreed we could try to unify the video and audio encoders cleanup instead of duplicating the logic.

test/test_ops.py
NicolasHug: Thanks for the context. This is a bit unfortunate, and I think we should try to support `format="mkv"` natively. Maybe we can do the "mkv" -> "matroska" mapping ourselves within the new C++ `VideoEncoder` constructor, just before calling `avformat_alloc_output_context2`?

NicolasHug (commented):
Thanks for the PR @Dan-Flores, nice work ! Left a few comments and suggestions but this looks great

test/test_ops.py
```diff
@@ -29,6 +29,7 @@
     create_from_tensor,
     encode_audio_to_file,
     encode_video_to_file,
+    encode_video_to_tensor,
```
NicolasHug, line 32: Let's add some more test:

- let's also parametrize `test_bad_input` over the method, when relevant
- let's have a similar test to https://github.com/meta-pytorch/torchcodec/blob/b084768e63d8d200b2945cf659fc930cfe72b04b/test/test_encoders.py#L345


These tests can be easier to write (and read!) when we have the public Python `VideoEncoder` class. So if you prefer this can be left as a TODO, but we should make sure to follow up on that.

src/torchcodec/_core/AVIOTensorContext.cpp
mollyxu: I noticed that this could also be written as 
`tensorContext->max = max(tensorContext->current, tensorContext->max);`
is there a preference for using explicit if statements vs `max` function for readibility/style?

test/test_ops.py
Dan-Flores: I'll add a similar test to this PR, hopefully it is fairly readable. 
I may move some tests to `test_encoders.py` once I push a PR with the python API, since we really want to test the `VideoEncoder` entrypoint.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: I left a TODO here to follow up on this as a part of the VideoEncoder work.

src/torchcodec/_core/AVIOTensorContext.cpp
Dan-Flores: Thanks for the pointer, I will incorporate this suggestion since it conveys the same information more succinctly!

test/test_ops.py
Dan-Flores: Parametrizing `test_bad_input` without a `VideoEncoder` class is fairly messy. I'll move this test to `test_encoders.py` and clean it up. For now, I've added the case where a bad `format` is passed in.

test/test_ops.py
```diff
@@ -1422,33 +1442,49 @@ def test_video_encoder_round_trip(self, tmp_path, format):
             assert psnr(s_frame, rt_frame) > 30
             assert_close(s_frame, rt_frame, atol=atol, rtol=0)
 
+    @pytest.mark.parametrize(
+        "format", ("mov", "mp4", "avi", "mkv", "webm", "flv", "gif")
```
NicolasHug: Let's mark the webm entry as slow (if it is!), see https://github.com/meta-pytorch/torchcodec/pull/968#discussion_r2435101666 for how to do that just for the webm entry, and not to the whole test

test/test_ops.py
```diff
@@ -1422,33 +1442,49 @@ def test_video_encoder_round_trip(self, tmp_path, format):
...
+        source_frames = self.decode(TEST_SRC_2_720P.path).data
+        params = dict(frame_rate=30, crf=0)
+
+        encoded_file = tmp_path / f"output.{format}"
+        encode_video_to_file(frames=source_frames, filename=str(encoded_file), **params)
+        encoded_tensor = encode_video_to_tensor(source_frames, format=format, **params)
+
+        torch.testing.assert_close(
+            self.decode(encoded_file).data, self.decode(encoded_tensor).data
+        )
```
NicolasHug, line 1480: These are uint8 so the default `atol` will be `0` anyway, but it's good to be explicit:


```suggestion
        torch.testing.assert_close(
            self.decode(encoded_file).data, self.decode(encoded_tensor).data, rtol=0, atol=0
        )
```

test/test_ops.py
NicolasHug: That sounds good, maybe we can just write a TODO to parametrize `test_bad_input` when we move it to `test_encoders.py` ?

NicolasHug (approved):
Thanks @Dan-Flores ! Left 3 non-blocking comments (one above, two below)

 LGTM

test/test_ops.py
Dan-Flores: Thanks for the pointer, I've added the mark. The `webm` tests are in fact very slow.

I suspect its because the encoding is done in `GRB` instead of `YUV`, thanks to `avcodec_find_best_pix_fmt_of_list` choosing the least lossy format, and the `webm` encoder supports `GRB`. We will probably want to change this for efficiency. 
```
pytest --durations=25 -k "TestVideoEncoderOps"
============================================================ slowest 25 durations ============================================================
25.98s call     test/test_ops.py::TestVideoEncoderOps::test_against_to_file[webm]
18.30s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_against_ffmpeg_cli[webm]
13.53s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_round_trip[to_tensor-webm]
13.33s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_round_trip[to_file-webm]
3.67s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_against_ffmpeg_cli[gif]
3.65s call     test/test_ops.py::TestVideoEncoderOps::test_against_to_file[gif]
3.03s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_against_ffmpeg_cli[mkv]
2.78s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_against_ffmpeg_cli[mov]
2.71s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_against_ffmpeg_cli[mp4]
2.35s call     test/test_ops.py::TestVideoEncoderOps::test_against_to_file[mkv]
2.26s call     test/test_ops.py::TestVideoEncoderOps::test_against_to_file[mp4]
2.21s call     test/test_ops.py::TestVideoEncoderOps::test_against_to_file[mov]
2.19s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_round_trip[to_tensor-mp4]
2.09s call     test/test_ops.py::TestVideoEncoderOps::test_video_encoder_round_trip[to_file-mov]
```

PR #959, opened 2025-10-14
Add `set_cuda_backend` Context Manager to publicly expose the BETA CUDA Interface

NicolasHug:
Meant to be used as:

```py
with torchcodec.set_cuda_backend("beta"):
    decoder = VideoDecoder("video.mp4", device="cuda")

decoder.get_frame_at(...)
```

### DIFF
--- a/src/torchcodec/decoders/_decoder_utils.py
+++ b/src/torchcodec/decoders/_decoder_utils.py
@@ -99,5 +99,5 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
         _CUDA_BACKEND.reset(previous_state)
 
 
-def _get_current_cuda_backend() -> str:
+def _get_cuda_backend() -> str:
     return _CUDA_BACKEND.get()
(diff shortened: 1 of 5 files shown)


src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -50,3 +52,52 @@ def create_decoder(
         "read(self, size: int) -> bytes and "
         "seek(self, offset: int, whence: int) -> int methods."
     )
+
+
+# Thread-local and async-safe storage for the current CUDA backend
+_CUDA_BACKEND: contextvars.ContextVar[str] = contextvars.ContextVar(
+    "_CUDA_BACKEND", default="ffmpeg"
+)
```
Dan-Flores, line 60: Do we plan on changing the default to be the `beta` interface at some point in the future? Would we need to indicate the change to users in some way?

NicolasHug: Yes, we'll want the current "beta" to eventually become the default interface. Usually the way to give notice is that for a few versions, we'll be raising a warning if the user doesn't explicitly requests either the "beta" or the "ffmpeg" interface. Basically we'd force them to use the CM, or they get a warning. Then we do the switch and remove the warning.

This is a bit disruptive and annoying, maybe we'll decide that the new interface is just strictly better and 99.9999% compatible with the existing one, at which point we could decide not to raise any warning and just silently do the switch.

PR #960, opened 2025-10-14
Address some TODONVDEC 

NicolasHug:
This PR addresses a bunch of TODOs related to the BETA CUDA interface. More will come later. See comments below for details.

### DIFF
--- a/docs/source/api_ref_decoders.rst
+++ b/docs/source/api_ref_decoders.rst
@@ -19,6 +19,12 @@ For an audio decoder tutorial, see: :ref:`sphx_glr_generated_examples_decoding_a
     VideoDecoder
     AudioDecoder
 
+.. autosummary::
+    :toctree: generated/
+    :nosignatures:
+    :template: function.rst
+
+    set_cuda_backend
 
 .. autosummary::
     :toctree: generated/
--- a/examples/decoding/basic_cuda_example.py
+++ b/examples/decoding/basic_cuda_example.py
@@ -94,9 +94,10 @@ urllib.request.urlretrieve(
 #
 # To use CUDA decoder, you need to pass in a cuda device to the decoder.
 #
-from torchcodec.decoders import VideoDecoder
+from torchcodec.decoders import set_cuda_backend, VideoDecoder
 
-decoder = VideoDecoder(video_file, device="cuda")
+with set_cuda_backend("beta"):  # Use the BETA backend, it's faster!
+    decoder = VideoDecoder(video_file, device="cuda")
 frame = decoder[0]
 
 # %%
@@ -120,7 +121,8 @@ print(frame.data.device)
 # against equivalent results from the CPU decoders.
 timestamps = [12, 19, 45, 131, 180]
 cpu_decoder = VideoDecoder(video_file, device="cpu")
-cuda_decoder = VideoDecoder(video_file, device="cuda")
+with set_cuda_backend("beta"):
+    cuda_decoder = VideoDecoder(video_file, device="cuda")
 cpu_frames = cpu_decoder.get_frames_played_at(timestamps).data
 cuda_frames = cuda_decoder.get_frames_played_at(timestamps).data
 
--- a/src/torchcodec/decoders/_decoder_utils.py
+++ b/src/torchcodec/decoders/_decoder_utils.py
@@ -66,20 +66,29 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
 
     This context manager allows you to specify which CUDA backend implementation
     to use when creating :class:`~torchcodec.decoders.VideoDecoder` instances
-    with CUDA devices.  This is thread-safe and async-safe.
+    with CUDA devices.
 
-    Note that you still need to pass `device="cuda"` when creating the
-    :class:`~torchcodec.decoders.VideoDecoder` instance. If a CUDA device isn't
-    specified, this context manager will have no effect.
+    .. note::
+        **We recommend trying the "beta" backend instead of the default "ffmpeg"
+        backend!** The beta backend is faster, and will eventually become the
+        default in future versions. It may have rough edges that we'll polish
+        over time, but it's already quite stable and ready for adoption. Let us
+        know what you think!
 
     Only the creation of the decoder needs to be inside the context manager, the
-    decoding methods can be called outside of it.
+    decoding methods can be called outside of it. You still need to pass
+    ``device="cuda"`` when creating the
+    :class:`~torchcodec.decoders.VideoDecoder` instance. If a CUDA device isn't
+    specified, this context manager will have no effect. See example below.
+
+    This is thread-safe and async-safe.
 
     Args:
-        backend (str): The CUDA backend to use. Can be "ffmpeg" or "beta". Default is "ffmpeg".
+        backend (str): The CUDA backend to use. Can be "ffmpeg" (default) or
+            "beta". We recommend trying "beta" as it's faster!
 
     Example:
-        >>> with torchcodec.set_cuda_backend("beta"):
+        >>> with set_cuda_backend("beta"):
         ...     decoder = VideoDecoder("video.mp4", device="cuda")
         ...
         ... # Only the decoder creation needs to be part of the context manager.
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -56,6 +56,8 @@ class VideoDecoder:
             Passing 0 lets FFmpeg decide on the number of threads.
             Default: 1.
         device (str or torch.device, optional): The device to use for decoding. Default: "cpu".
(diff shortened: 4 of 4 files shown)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -216,12 +216,11 @@ BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
...
     flush();
     unmapPreviousFrame();
-    NVDECCache::getCache(device_.index())
-        .returnDecoder(&videoFormat_, std::move(decoder_));
+    NVDECCache::getCache(device_).returnDecoder(
+        &videoFormat_, std::move(decoder_));
   }
 
   if (videoParser_) {
-    // TODONVDEC P2: consider caching this? Does DALI do that?
```
NicolasHug, line 224: Answer: No, DALI doesn't cache the parser.

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -249,7 +247,7 @@ torch::Tensor convertNV12FrameToRGB(
 }
 
 UniqueNppContext getNppStreamContext(const torch::Device& device) {
-  torch::DeviceIndex nonNegativeDeviceIndex = getNonNegativeDeviceIndex(device);
+  int deviceIndex = getDeviceIndex(device);
```
NicolasHug, line 250: Going from `torch::DeviceIndex` to `int` is a fix. `cudaGetDeviceProperties` and `nppCtx->nCudaDeviceId` below both expect int, not `torch::DeviceIndex` which is `int8_t`

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -312,4 +310,21 @@ void validatePreAllocatedTensorShape(
...
+      deviceIndex >= -1 && deviceIndex < MAX_CUDA_GPUS,
+      "Invalid device index = ",
+      deviceIndex);
+
+  if (deviceIndex == -1) {
+    TORCH_CHECK(
+        cudaGetDevice(&deviceIndex) == cudaSuccess,
+        "Failed to get current CUDA device.");
+  }
+  return deviceIndex;
```
NicolasHug, line 327: Another fix: `getNonNegativeDeviceIndex()` used to map `-1` to `0`, which is incorrect. `-1` should be mapped to the "current device", which can be arbitrary and set by a context manager from Python.

src/torchcodec/_core/NVDECCache.h
```diff
@@ -68,11 +69,6 @@ class NVDECCache {
     CacheKey(const CacheKey&) = default;
     CacheKey& operator=(const CacheKey&) = default;
 
-    // TODONVDEC P2: we only implement operator< which is enough for std::map,
-    // but:
-    // - we should consider using std::unordered_map
-    // - we should consider a more sophisticated and potentially less strict
-    // cache key comparison logic
```
NicolasHug, line 75: I realized that using an `unorderd_map` would force us to define our own hash function, so that's really not worth it. The `n log(n)` complexity of `map` isn't relevant anyway considering the very small sizes we deal with.


On the potentailly less strict cache logic: I looked into more details at DALI's implementation and they do the same thing, modulo decoder re-configuration, which is tracked in another TODO. So we can remove this one.

test/utils.py
```diff
@@ -41,9 +41,6 @@ def unsplit_device_str(device_str: str) -> str:
     # It is used:
     # - before calling `.to(device)` where device can't be "cuda:0:beta"
     # - before calling add_video_stream(device=device, device_variant=device_variant)
-    #
-    # TODONVDEC P2: Find a less clunky way to test the BETA CUDA interface. It
-    # will ultimately depend on how we want to publicly expose it.
```
NicolasHug, line 46: This was addressed by https://github.com/meta-pytorch/torchcodec/pull/959 and there is another TODO to remove the legacy use of `cuda:0:beta`, which will transitively resolve this one.

test/utils.py
```diff
@@ -750,7 +747,7 @@ def sample_format(self) -> str:
 
 
 def supports_approximate_mode(asset: TestVideo) -> bool:
-    # TODONVDEC P2: open an issue about his. That's actually not related to
-    # NVDEC at all, those don't support approximate mode because they don't set
-    # a duration. CPU decoder fails too!
+    # Those are missing the `duration` field so they fail in approximate mode (on all devices).
+    # TODO: we should address this, see
+    # https://github.com/meta-pytorch/torchcodec/issues/945
```
NicolasHug, line 752: This isn't an NVDEC TODO, it's a general TODO.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -216,12 +216,11 @@ BetaCudaDeviceInterface::~BetaCudaDeviceInterface() {
     // unclear.
     flush();
     unmapPreviousFrame();
-    NVDECCache::getCache(device_.index())
-        .returnDecoder(&videoFormat_, std::move(decoder_));
+    NVDECCache::getCache(device_).returnDecoder(
+        &videoFormat_, std::move(decoder_));
```
NicolasHug, line 220: Above and everywhere else in the PR, anything related to "device index" relates to the creation of `getDeviceIndex()` which is now in `CUDACommon.cpp`, and replaces the old (buggy!) `getNonNegativeDeviceIndex`.

src/torchcodec/_core/Cache.h
```diff
@@ -95,30 +95,16 @@ class PerGpuCache {
...
-  // implementation and during initialization of CUDA and FFmpeg contexts
-  // which require non negative indices.
-  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
-  TORCH_CHECK(deviceIndex >= 0, "Device index out of range");
-  return deviceIndex;
-}
+// Forward declaration of getDeviceIndex which exists in CUDACommon.h
+// This avoids circular dependency between Cache.h and CUDACommon.cpp which also
+// needs to include Cache.h
+int getDeviceIndex(const torch::Device& device);
```
NicolasHug, line 101: Not thrilled by this, happy to consider alternatives.

We could:

- include `Cache.h` within `CUDACommon.cpp` and not bother about the potential circular issue
- Have a forward declaration of `PerGpuCache` within `CudaCommon.cpp` instead (not sure if that would work since it's a template class).

src/torchcodec/_core/Cache.h
```diff
@@ -95,30 +95,16 @@ class PerGpuCache {
...
-// grows, and there are more such functions, we should break them out into a
-// .cpp file.
-inline torch::DeviceIndex getNonNegativeDeviceIndex(
-    const torch::Device& device) {
-  torch::DeviceIndex deviceIndex = device.index();
-  // For single GPU machines libtorch returns -1 for the device index. So for
-  // that case we set the device index to 0. That's used in per-gpu cache
-  // implementation and during initialization of CUDA and FFmpeg contexts
-  // which require non negative indices.
-  deviceIndex = std::max<at::DeviceIndex>(deviceIndex, 0);
```
NicolasHug, line 112: Note: this was wrong! -1 should be mapped to the current device, not to 0.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -68,9 +65,9 @@ UniqueAVBufferRef getHardwareDeviceContext(const torch::Device& device) {
   // So we ensure the deviceIndex is not negative.
   // We set the device because we may be called from a different thread than
   // the one that initialized the cuda context.
-  cudaSetDevice(nonNegativeDeviceIndex);
+  cudaSetDevice(deviceIndex);
```
NicolasHug, line 68: Here as well: `cudaSetDevice` expects an int, not an implicitly-cast  `torch::DeviceIndex`

src/torchcodec/_core/Cache.h
scotts: Forward declarations of templates are possible, but awkward. But if `Cache.h` needs to know CUDA stuff, then it's not really generic. The fact we have a circular dependency hints to me that we're not quite modeling things correctly. I can think of two alternatives:

1. We add the ability to convert a `torch::Device` into a device-specific index to the `DeviceInterface`. That's still a little awkward, though, as we'd want to tell `PerGpuCache` the `DeviceInterface` so it would know where to get the index.
2. `PerGpuCache` accepts the device index itself, instead of a `torch::Device`. That then assumes the device logic does the conversion itself, which I think is reasonable.

I prefer 2, but that might be a bigger refactor than this PR. So I'm good with this for now, and we can file a task for option 2, if you also agree it makes sense.

NicolasHug: I think 2 is reasonable, we could align both caches to just take an index as input. Let me follow-up on that

PR #965, opened 2025-10-15
Remove legacy support of `device='cuda:0:beta'`

NicolasHug:


src/torchcodec/decoders/_video_decoder.py
```diff
@@ -151,13 +149,6 @@ def __init__(
             # TODONVDEC P2 rename 'default' into 'ffmpeg' everywhere.
             device_variant = "default"
 
-        # Legacy support for device="cuda:0:beta" syntax
-        # TODONVDEC P2: remove support for this everywhere. This will require
-        # updating our tests.
-        if device == "cuda:0:beta":
-            device = "cuda:0"
```
NicolasHug, line 155: This is the actual removal. Everything else in this PR is a consequence of that change - mostly just changing how we use the BETA backend in existing tests.

test/utils.py
```diff
@@ -26,30 +27,55 @@ def needs_cuda(test_item):
...
-    # TODONVDEC P2: Find a less clunky way to test the BETA CUDA interface. It
-    # will ultimately depend on how we want to publicly expose it.
-    if device_str == "cuda:0:beta":
+    if device_str == _CUDA_BETA_DEVICE_STR:
         return "cuda", "beta"
     else:
         return device_str, "default"
 
 
+def make_video_decoder(*args, **kwargs) -> Union[VideoDecoder, str]:
```
scotts: Return type is supposed to be a `Tuple` not `Union`, correct?

scotts (approved):
I think we need to fix the return type of `make_video_decoder`, but otherwise this is great!

PR #967, opened 2025-10-15
C++ implementation of crop transform

scotts:
Adds C++ implementation for the crop transform.

Design points worth discussing:

1. I'm always setting `exact=1` from the FFmpeg [`crop` filter](https://ffmpeg.org/ffmpeg-filters.html#crop). Based on reading, this behavior is going to be consistent with [`torchvision.transforms.v2.functional.crop`](https://docs.pytorch.org/vision/main/generated/torchvision.transforms.v2.functional.crop.html#torchvision.transforms.v2.functional.crop). Because we're eventually going to accept the TorchVision transform as the public API for crop, users won't be able to set this, so I think we have to be conservative.
2. We apply the crop in RGB space.  Note that in the references I generate from the `ffmpeg` CLI, I explicitly add `format=rgb24` before the filters. If we don't explicitly do that format conversion, the frames we get from the `ffmpeg` CLI will have been cropped in YUV space and we won't get tensor equality with TorchCodec or TorchVision's transforms. I think it would actually be faster to crop first, as then the colorspace conversion needs to do less work. But I think our ideal correctness comparison should be the result we get from passing a unchanged decoded frame to a TorchVision transform.

Implementation details:

1. I had to add TorchVision to our test environments.
2. I had to add `Transform::validate(const StreamMetadata&)` because sometimes we need to validate transforms based on the stream itself. For crop, we need to validate that the (x, y) coordinates are within the original frame's dimensions. (We could let FFmpeg throw an error instead, but that's not a great experience for our users. That will only happen upon a decode, and the error they'll get is not going to be understandable by them.) I wavered on what `validate()` should accept to do the validation, and I realized it's probably best for it to be `StreamMetadata`. If in the future what we need to validate against is not in the metadata, we should add it. I'd rather pass that around than the actual `AVStream`, and we might need to check against more than just the dimensions in the future.
3. I'm adding a new kind of reference file, which has the FFmpeg filter embedded in the file name. ~We can use a different name if we want, but it seemed simple to use the filter expression itself.~ **It looks like Windows does not like these filenames, so we'll have to do something else.**
4. I added a new test file, `test_transform_ops.py`, and moved all of the existing resize (scale) tests in there as well. My rule is that any test of the core API which sets `transform_specs` goes in the new test file. While `test_ops.py` is getting kinda long, my main motivation is that will simplify internal testing. Internally, we want to limit the places that TorchCodec and TorchVision must be used at the same time due to FFmpeg version issues.

### DIFF
--- a/.github/workflows/reference_resources.yaml
+++ b/.github/workflows/reference_resources.yaml
@@ -82,6 +82,11 @@ jobs:
           python -m pip install torch --index-url https://download.pytorch.org/whl/cpu
           python -m pip install numpy pillow pytest
 
+      - name: Install torchcodec from the wheel
+        run: |
+          wheel_path=`find pytorch/torchcodec/dist -type f -name "*.whl"`
+          echo Installing $wheel_path
+          python -m pip install $wheel_path -vvv
       - name: Check out repo
         uses: actions/checkout@v3
 


test/test_transform_ops.py
```diff
@@ -0,0 +1,271 @@
...
+            match="cannot be converted to an int",
+        ):
+            add_video_stream(decoder, transform_specs="resize, blah, 100")
+
+        with pytest.raises(
+            RuntimeError,
+            match="out of range",
+        ):
+            add_video_stream(decoder, transform_specs="resize, 100, 1000000000000")
+
```
scotts, line 209: New tests start below this line.

src/torchcodec/_core/Frame.cpp
```diff
@@ -8,6 +8,11 @@
 
 namespace facebook::torchcodec {
 
+FrameDims::FrameDims(int height, int width) : height(height), width(width) {
+  TORCH_CHECK(height > 0, "FrameDims.height must be > 0, got: ", height);
+  TORCH_CHECK(width > 0, "FrameDims.width must be > 0, got: ", width);
+}
```
scotts, line 14: Drive-by: when passing in a height and width, we should only be able to instantiate a `FrameDims` object with positive values. If we want a `FrameDims` object that has 0 for both values, that's just the default constructor. We should never have a `FrameDims` object with negative values.

src/torchcodec/_core/Transform.cpp
```diff
@@ -57,4 +57,25 @@ int ResizeTransform::getSwsFlags() const {
...
+CropTransform::CropTransform(const FrameDims& dims, int x, int y)
+    : outputDims_(dims), x_(x), y_(y) {
+  TORCH_CHECK(x_ >= 0, "Crop x position must be >= 0, got: ", x_);
+  TORCH_CHECK(y_ >= 0, "Crop y position must be >= 0, got: ", y_);
+}
+
+std::string CropTransform::getFilterGraphCpu() const {
+  return "crop=" + std::to_string(outputDims_.width) + ":" +
+      std::to_string(outputDims_.height) + ":" + std::to_string(x_) + ":" +
+      std::to_string(y_) + ":exact=1";
```
NicolasHug, line 69: On design discussion point 1: agreed this should be hardcoded to `exact=1`, and *not* a parameter. And I think this is the case regardless of what torchvision does. With `exact=0` we risk the output frames not being of the desired dimensions, and that potentially means a downstream model failing loudly because its expected input dimensions weren't respected.

I would suggest to go slightly further and not even include "exact=1" in the file names (like `nasa_13013.mp4.crop=300:200:50:35:exact=1.stream3.frame000000.pt`), because I don't think we'll be enabling exact=0.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -212,6 +212,26 @@ Transform* makeResizeTransform(
...
 }
 
+// Crop transform specs take the form:
+//
+//   "crop, <width>, <height>, <x>, <y>"
+//
+// Where "crop" is the string literal and <width>, <height>, <x> and <y> are
+// positive integers. Note that that in this spec, we are following the
+// filtergraph convention of (width, height). This makes it easier to compare it
+// against actual filtergraph strings.
```
NicolasHug: IIUC x and y are essentially the top-left coordinates of the crop, let's make that explicit (it could also be e.g. the center coordinates of the crop).

Separately, I was hoping we could use H and W everywhere for consistency. Can you share the comparison against filtergraph strings that are made easier by using the W and H order?

src/torchcodec/_core/Transform.cpp
```diff
@@ -57,4 +57,25 @@ int ResizeTransform::getSwsFlags() const {
...
+}
+
+std::optional<FrameDims> CropTransform::getOutputFrameDims() const {
+  return outputDims_;
+}
+
+void CropTransform::validate(const StreamMetadata& streamMetadata) const {
+  TORCH_CHECK(x_ <= streamMetadata.width, "Crop x position out of bounds");
+  TORCH_CHECK(y_ <= streamMetadata.height, "Crop y position out of bounds");
```
NicolasHug, line 81: Should/can we account for H and W of the crop as well? E.g.

```C
TORCH_CHECK(x_ + W < streamMetadata.width, "Crop x position out of bounds");
```

test/generate_reference_resources.py
```diff
@@ -40,13 +55,14 @@ def get_frame_by_index(video_path, frame, output_path, stream):
...
-        "-q:v",
-        "2",
+        filtergraph,
+        "-fps_mode",
+        "passthrough",
+        "-update",
+        "1",
         output_path,
     ]
+    print("===" + " ".join([str(x) for x in cmd]))
```
NicolasHug: Is this debug leftover? Just checking, maybe it's intended and OK since this is just the generation util.

test/test_transform_ops.py
```diff
@@ -0,0 +1,265 @@
...
+)
+
+from torchvision.transforms import v2
+
+from .utils import assert_frames_equal, NASA_VIDEO, needs_cuda
+
+torch._dynamo.config.capture_dynamic_output_shape_ops = True
+
+
+class TestVideoDecoderTransformOps:
```
NicolasHug: Nit, feel free to ignore as this is mostly personal preference, but I liked having all ops-related tests in a single file so we could just do `pytest test/test_ops.py`. Now we have to specify more files.

I definitely agree that separating the new tests in their own new test class is worth it. It these were still in in `test_ops.py`, we could run the transform-specific tests with just `pytest test/test_ops.py -k transform`

test/test_transform_ops.py
```diff
@@ -0,0 +1,265 @@
...
+            frame_tv = v2.functional.crop(
+                frame_full, top=35, left=50, height=200, width=300
+            )
+
+            assert frame.shape == (3, 200, 300)
+            assert frame_ref.shape == (3, 200, 300)
+            assert frame_tv.shape == (3, 200, 300)
+
+            assert_frames_equal(frame, frame_tv)
+            assert_frames_equal(frame, frame_ref)
```
NicolasHug, line 242: On the test above: I trust that you hardcoded the H and W values properly, but it might be a bit safer to set those as variables, like

```py
        W, H, X, Y = 300, 200, 50, 35
        crop_spec = f"crop, {W}, {H}, {X}, {Y}"
        crop_filtergraph = f"crop={W}:{h}:{X}:{Y}:exact=1"

....

assert frame.shape == (3, H, W)
```

test/test_transform_ops.py
```diff
@@ -0,0 +1,265 @@
...
+    get_frame_at_index,
+    get_json_metadata,
+    get_next_frame,
+)
+
+from torchvision.transforms import v2
+
+from .utils import assert_frames_equal, NASA_VIDEO, needs_cuda
+
+torch._dynamo.config.capture_dynamic_output_shape_ops = True
```
NicolasHug, line 30: I suspect we don't need this line in this file for now, nor the `os.environ["TORCH_LOGS"] = "output_code"` above, since they relate to torch.compile tests.

NicolasHug (commented):
Thanks for the PR @scotts ! I left a few comments below, this looks good.


-----

On this design discussion point:

>2. We apply the crop in RGB space. Note that in the references I generate from the ffmpeg CLI, I explicitly add format=rgb24 before the filters. If we don't explicitly do that format conversion, the frames we get from the ffmpeg CLI will have been cropped in YUV space and we won't get tensor equality with TorchCodec or TorchVision's transforms. I think it would actually be faster to crop first, as then the colorspace conversion needs to do less work. But I think our ideal correctness comparison should be the result we get from passing a unchanged decoded frame to a TorchVision transform.

I don't remember having an explicit discussion about our inclusion criteria, so let me share my personal thoughts on this. When implementing a new TC transform, I think we should consider these two questions: 

- **A. Perf: Is the TC transform faster than the TV transform?** We want the answer to be *yes*. If the answer is *no* then it's not worth implementing in TC. Users can just use TV.
- **B. Correctness: Can a *model* tell the difference between a TV transform and its TC counterpart?** We want the answer to be *no*. We know for example that for `resize`, we'll have to be careful about this. The easiest way to ensure that is to check that the TC output is exactly equal to the TV output - that's what this PR is doing, and this guarantees correctness with 100% certainty. In general though: 
  - we won't always be able to check strict equality with TV.
  - strict equality with TV is a sufficient, but not always necessary condition to validate correctness. For example, the PIL and Tensor backend of TV's `resize` slightly differ in terms of output (usually within 1-pixel difference for bilinear mode, and a bit more for bicubic mode), but they're still un-distinguishable for models. I suspect this will also be the case between TV's resize and TC's resize. And maybe, it's OK for `crop` to have slightly different output resuts, as long as models can't tell the difference.

-----

Back to our discussion point now: should we apply `crop` in YUV or in RGB?

I could be wrong but I think applying the crop in RGB will have the exact same perf as TC's crop. And if that's the case then we're not meeting criteria A. I think it's worth benchmarking this a bit, and I can help with that if needed.

We know that applying `crop` in RGB meets criteria B, so if we choose to apply it in YUV we'll have to check that it's still correct. We won't have strict equality with TV, but as I mentioned, this won't always be possible (or necessary!) anyway. Maybe we can have psnr checks with very strict tolerance. And we should try to validate filtergraph's algorithms (e.g. see if there's any kind of fancy filtering going on, which might be a problem) - that's also something I can help with.

src/torchcodec/_core/custom_ops.cpp
scotts: @NicolasHug, I'm fine with (h, w). It just means that in the tests and in the resource generation, we'll have strings that look like:
```python
        crop_spec = "crop, 200, 300, 50, 35"   # note: h=200, w=300
        crop_filtergraph = "crop=300:200:50:35:exact=1". # note: w=300, h=200
```
Comments can cover it. We'll have to deal with this awkwardness *somewhere*. I don't have a strong preference where, and I think it's probably good to keep all of TorchCodec in (h, w).

scotts:
@NicolasHug, thanks for the discussion!

> I could be wrong but I think applying the crop in RGB will have the exact same perf as TC's crop. And if that's the case then we're not meeting criteria A. I think it's worth benchmarking this a bit, and I can help with that if needed.

I actually think criteria A is too strict, as it's strictly phrased about CPU performance and only considers transforms in isolation. I think we need to also consider:

1. **Overall memory usage.** Even if crop (and resize) take the same amount of time as decoder-native transforms versus TorchVision transforms, they're going to use less memory overall. That is, if we have a video with 4K frames and we use TorchVision to do the crop or resize, we're going to return a 4K tensor. If we're cropping or resizing to something like (100, 100), then we're only going to return a (100, 100) tensor. That's quite a big difference. That's a lot less memory that FFmpeg will have to return to us in the `AVFrame` and much less memory that will be owned by the Python layer. I think we're efficient in getting our data into tensors, but once a tensor is owned by the Python layer, it's going to governed by the garbage collector. If we can just return a much smaller tensor to begin with, the garbage collector in the Python layer will be under much less pressure. The less bytes we have to shuttle around, the better overall performance will be.
2. **Some transforms will enable others.** If users want to do a pipeline that is basically `[ExpensiveTransform, Crop]`, but `Crop` is not decoder-native, then they can't do `ExpensiveTransform` decoder-natively even if it's available.

Regarding cropping in RGB versus YUV space, this is actually a general question that will apply to all transforms. What I think I'd like to go with now is: we default to doing the transform in RGB space for now, as that means we are more likely to be able to have bit-for-bit equality with TorchVision transforms. Whether our users realize it or not, I also think that's what they'll expect. If we get requests for applying the transforms in YUV space, we could provide an API to request it. (Either a decoder level option, or some special transform which would allow users to have some of their transforms in YUV, some in RGB.)

src/torchcodec/_core/Transform.cpp
scotts: @NicolasHug, how strongly do you prefer for `exact=1` to not be in the filename? Taking it out of the filename will require some special-handling per-filter. Right now we're able to just use a `sanitize_filtergraph_expression()` function that just naively turns all special characters into underscores. Knowing that a part of the expression should be taken out will not be generic and we'll need to do it differently for each kind of filter.

test/generate_reference_resources.py
```diff
@@ -12,9 +12,15 @@
 import torch
 from PIL import Image
 
+from .utils import sanitize_filtergraph_expression
```
scotts, line 15: Note that pulling in code from `utils.py` means we have to run this as a module, and since `utils.py` imports general TorchCodec stuff, we do a full build and install for the `reference_resources.yaml` workflow. I think that's a fair trade-off to ensure that the unit tests and the resource generation are using the same logic for writing and loading files.

src/torchcodec/_core/Transform.cpp
NicolasHug: Not strongly at all, the ease of use of `sanitize_filtergraph_expression()` sounds like a very good reason to keep `exact=1` in the name - thanks for looking into it!

test/generate_reference_resources.py
NicolasHug: Do we expect to pull in more things from `utils.py`? If it's just for `sanitize_filtergraph_expression`, I would have a slight preference for just duplicating it in both files, because the `python -m` stuff isn't immediately obvious. But that's fine too.

scotts:
@NicolasHug, yes, I anticipate unifying `generate_reference_resources.py` with `utils.py` more. Right now, they agree on the actual file names because they both happen to have the same string. I'd like to make `generate_reference_resources.py` actually use the file definitions in there.

PR #971, opened 2025-10-16
Indicate FFmpeg8 support in docs

Dan-Flores:
Since support for FFmpeg8 was added for Mac and Linux, we need to update our readme and build error msg to reflect that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ format you want. Refer to Nvidia's GPU support matrix for more details
 
 1. Install FFmpeg with NVDEC support.
    TorchCodec with CUDA should work with FFmpeg versions in [4, 7] on all platforms,
-   and FFmpeg version 8 is supported on Mac and Linux.
+   and FFmpeg version 8 is supported on Linux.
 
    If FFmpeg is not already installed, or you need a more recent version, an
    easy way to install it is to use `conda`:


src/torchcodec/_core/ops.py
```diff
@@ -69,7 +69,7 @@ def load_torchcodec_shared_libraries():
     raise RuntimeError(
         f"""Could not load libtorchcodec. Likely causes:
           1. FFmpeg is not properly installed in your environment. We support
-             versions 4, 5, 6 and 7.
+             versions 4, 5, 6, 7, and 8.
```
NicolasHug: thanks @Dan-Flores , here and below let's clarify that Windows doesn't support 8 yet.

README.md
```diff
@@ -147,7 +147,8 @@ format you want. Refer to Nvidia's GPU support matrix for more details
 [here](https://developer.nvidia.com/video-encode-and-decode-gpu-support-matrix-new).
 
 1. Install FFmpeg with NVDEC support.
-   TorchCodec with CUDA should work with FFmpeg versions in [4, 7].
+   TorchCodec with CUDA should work with FFmpeg versions in [4, 7] on all platforms,
+   and FFmpeg version 8 is supported on Mac and Linux.
```
NicolasHug: This is the part about GPU support, so Mac is not relevant here

```suggestion
   and FFmpeg version 8 is supported on Linux.
```

Issue #943, opened 2025-10-08
GPU -> CPU fallback: difference and alignment between the BETA and default interfaces


NicolasHug:
When it's not capable of decoding a video with NVDEC, the default interface (from FFmpeg) falls back to the CPU decoding path. This can happen in various scenarios, all related to a codec configuration that isn't supported by NVDEC:

- when decoding a 10 bit video NVDEC ([for example](https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/test/test_decoders.py#L1298-L1305))
- when decoding a video that's too small, which isn't supported by NVDEC ([example](https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/test/test_decoders.py#L1685))

In the BETA interface, however, we don't fall back to the CPU. We [error loudly](https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L64-L122).

----

Below are just my current thoughts, this is obviously open for discussion.

I think we will want to fallback on the CPU for both default and BETA interfaces. Otherwise, we're essentially putting the fallback responsibility on the user, which isn't a good UX. The user wants the video to be decoded. They don't want to deal with errors.

I also think we'll want to fallback on  the CPU **only** for the decoding part. Currently, when we fallback on the CPU, we also perform the [color-conversion stage (YUV -> RGB) ](https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/src/torchcodec/_core/CudaDeviceInterface.cpp#L299-L307) on the CPU. But it would make sense to do the color-conversion on the GPU. Presumably, it'd be faster. And it would also optimize the CPU -> memory transfer, since the YUV frame is smaller than the RGB frame.

Both interfaces should have the exact same behavior: both to fallback on the CPU, and run the color conversion on the GPU. For both interfaces, this is a somewhat BC-breaking behavior change. But I don't think anyone would complain.

Dan-Flores:
> The user wants the video to be decoded
> Both interfaces should have the exact same behavior: both to fallback on the CPU

[...]

PR #977, opened 2025-10-16
Fallback to CPU when video isn't supported by NVDEC

NicolasHug:
Currently in our BETA CUDA Interface, we error loudly if we find that NVDEC doesn't support the input video.

In this PR, we do not error loudly anymore: we fall back to the CPU. This aligns the behavior of the BETA Interface with that of the FFmpeg interface, since FFmpeg was doing the fallback internally already (see https://github.com/meta-pytorch/torchcodec/issues/943).

There are some key design choices and which are better explain in context, so I'll leave some comments below.

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -95,7 +95,7 @@ static UniqueCUvideodecoder createDecoder(CUVIDEOFORMAT* videoFormat) {
   return UniqueCUvideodecoder(decoder, CUvideoDecoderDeleter{});
 }
 
-std::optional<cudaVideoChromaFormat> mapChromaFormat(
+std::optional<cudaVideoChromaFormat> validateChromaSupport(
     const AVPixFmtDescriptor* desc) {
   // Return the corresponding cudaVideoChromaFormat if supported, std::nullopt
   // otherwise.
@@ -165,7 +165,7 @@ bool nativeNVDECSupport(const SharedAVCodecContext& codecContext) {
     return false;
   }
 
-  auto chromaFormat = mapChromaFormat(desc);
+  auto chromaFormat = validateChromaSupport(desc);
   if (!chromaFormat.has_value()) {
     return false;
   }


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -53,74 +53,6 @@ pfnDisplayPictureCallback(void* pUserData, CUVIDPARSERDISPINFO* dispInfo) {
 }
 
 static UniqueCUvideodecoder createDecoder(CUVIDEOFORMAT* videoFormat) {
```
NicolasHug, line 55: All checks below have been moved to the new `bool nativeNVDECSupport(const SharedAVCodecContext& codecContext);`

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -189,10 +147,69 @@ cudaVideoCodec validateCodecSupport(AVCodecID codecId) {
...
     // case AV_CODEC_ID_VC1:
     //   return cudaVideoCodec_VC1;
-    default: {
-      TORCH_CHECK(false, "Unsupported codec type: ", avcodec_get_name(codecId));
-    }
+    default:
+      return std::nullopt;
+  }
+}
+
```
NicolasHug, line 154: These are the same checks as before, but with a crucial difference: the source of truth that we use to get the stream properties is now FFmpeg's `AVCodecContext`. Previously, it was NVCUVID's `CUVIDEOFORMAT`.
I think that's fine. I can't imagine NVCUVID being more accurate than FFmpeg.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -618,6 +669,21 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
...
+    // Transfer CPU frame to GPU
+    if (preAllocatedOutputTensor.has_value()) {
+      preAllocatedOutputTensor.value().copy_(cpuFrameOutput.data);
+      frameOutput.data = preAllocatedOutputTensor.value();
+    } else {
+      frameOutput.data = cpuFrameOutput.data.to(device_);
+    }
+    return;
+  }
+
```
NicolasHug, line 686: Above is essentially the same as the FFmpeg interface logic:

https://github.com/meta-pytorch/torchcodec/blob/3827dfe2d6da2fee91811d1d0c8996120a899320/src/torchcodec/_core/CudaDeviceInterface.cpp#L250-L290

There are still minor differences that we should eventually align, as part of https://github.com/meta-pytorch/torchcodec/issues/943

test/test_decoders.py
```diff
@@ -1740,15 +1739,20 @@ def test_set_cuda_backend(self):
             assert _get_cuda_backend() == "beta"
 
         def assert_decoder_uses(decoder, *, expected_backend):
+            # TODO: This doesn't work anymore after
+            # https://github.com/meta-pytorch/torchcodec/pull/977
+            # We need to define a better way to assert which backend a decoder
+            # is using.
+            return
```
NicolasHug, line 1746: It's OK to deactivate this test for now, we know it's working fine.

I think we should have a private way to query the `VideoDecoder` object so that it can tell us which interace / backend it's currently using. I will follow-up on this.

test/test_decoders.py
```diff
@@ -1701,20 +1701,19 @@ def test_beta_cuda_interface_backwards(self, asset, seek_mode):
...
         with set_cuda_backend("beta"):
-            dec = VideoDecoder(H265_VIDEO.path, device="cuda")
-        with pytest.raises(
-            RuntimeError,
-            match="Video is too small in at least one dimension. Provided: 128x128 vs supported:144x144",
-        ):
-            dec.get_frame_at(0)
+            beta = VideoDecoder(H265_VIDEO.path, device="cuda").get_frame_at(0)
+
+        torch.testing.assert_close(ffmpeg.data, beta.data, rtol=0, atol=0)
```
NicolasHug, line 1716: Above: we used to assert an error, now we just assert that the video can be decoded properly. Strictly speaking, we're not asserting that there is a fallback. (That will probably come in follow-ups).

There's another important thing to check, that we're actually not enforcing in the tests: we need to make sure that we don't have false positives, i.e. that we're not falling back to the CPU when we don't need to. To check that, I added a `TORCH_CHECK(false)` in the CPU fallback path. All the tests are passing, except for these two:

```
FAILED test/test_decoders.py::TestVideoDecoder::test_get_key_frame_indices[cuda:beta] - RuntimeError: Falling back to CPU decoder.
FAILED test/test_decoders.py::TestVideoDecoder::test_beta_cuda_interface_cpu_fallback - RuntimeError: Falling back to CPU decoder.
```

This is the expected result: these two tests are using `H265_VIDEO`, which is the only video that needs a fallback.

I will try to follow-up with a an official and robust way to assert the backend / fallback of the `VideoDecoder` object (see other comment below)

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -232,6 +249,19 @@ void BetaCudaDeviceInterface::initialize(
...
+        cpuFallback_ != nullptr, "Failed to create CPU device interface");
+    cpuFallback_->initialize(avStream, avFormatCtx, codecContext);
+    cpuFallback_->initializeVideo(
+        VideoStreamOptions(),
+        {},
+        /*resizedOutputDims=*/std::nullopt);
+    // We'll always use the CPU fallback from now on, so we can return early.
+    return;
+  }
+
```
NicolasHug, line 264: Above: we create the fallback CPU interface *only* if we need it.

I'm a bit sad that the BETA Interface needs to store a CPU interface. I have TODOs in other places basically saying "An interface shouldn't rely on another interface".

But, I'm not sure we have a better way to implement this fallback mechanism, so I think our principle should be something less strict along those lines: "It's OK for interface A to use interface B, but interface B shouldn't need to worry about being called by interface A". I think we're respecting this principle so far.

The main design alternative would be for the fallback to be implemented within `SingleStreamDecoder` rather than in the interface itself. But I'm not sure we want this.
Specifically for this fallabck, we'd have to create a third hybrid interface (the fallback one), that runs all the decoding on the CPU, but runs the color conversion (`convertAVFrameToFrameOutput`) on GPU. Sounds like a hassle.

scotts: > "It's OK for interface A to use interface B, but interface B shouldn't need to worry about being called by interface A". I think we're respecting this principle so far.

Agreed and agreed.

In my previous refactoring of the FFmpeg CUDA device interface, I believe I had always created the CPU device for fallback. But in that device, I also think we only knew if we would need the fallback when we discovered that FFmpeg had decoded on the frame on the CPU and then we needed to do CPU color conversion. *If* we could discover earlier in the FFmpeg CUDA device interface, then doing this pattern would be good. But I doubt we can

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -618,6 +669,21 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
...
+    cpuFallback_->convertAVFrameToFrameOutput(avFrame, cpuFrameOutput);
+
+    // Transfer CPU frame to GPU
+    if (preAllocatedOutputTensor.has_value()) {
+      preAllocatedOutputTensor.value().copy_(cpuFrameOutput.data);
+      frameOutput.data = preAllocatedOutputTensor.value();
+    } else {
+      frameOutput.data = cpuFrameOutput.data.to(device_);
+    }
+    return;
```
scotts, line 684: We do this dance in at least one other place in the FFmpeg CUDA device. Maybe a part of #943 could also be trying to pull this into a utils function.

scotts (approved):

PR #979, opened 2025-10-17
Check return value of cudaSetDevice

NicolasHug:
Addresses an internal lint warning.

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -60,12 +60,10 @@ UniqueAVBufferRef getHardwareDeviceContext(const torch::Device& device) {
 
   // Create hardware device context
   c10::cuda::CUDAGuard deviceGuard(device);
-  // Valid values for the argument to cudaSetDevice are 0 to maxDevices - 1:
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -60,12 +60,10 @@ UniqueAVBufferRef getHardwareDeviceContext(const torch::Device& device) {
...
   c10::cuda::CUDAGuard deviceGuard(device);
-  // Valid values for the argument to cudaSetDevice are 0 to maxDevices - 1:
-  // https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g159587909ffa0791bbe4b40187a4c6bb
-  // So we ensure the deviceIndex is not negative.
   // We set the device because we may be called from a different thread than
   // the one that initialized the cuda context.
-  cudaSetDevice(deviceIndex);
+  TORCH_CHECK(
+      cudaSetDevice(deviceIndex) == cudaSuccess, "Failed to set CUDA device");
   AVBufferRef* hardwareDeviceCtxRaw = nullptr;
```
NicolasHug, line 67: Drive-by: I removed the comment above because it's now out of place, the `deviceIndex` was set to a proper value just above (currently line 54)


Separately, I'm not even sure we need the `deviceGuard` and the call to `cudaSetDevice`. I we should rely on PyTorch for that kind of stuff. Maybe it'll be worth revisiting in the future, for now, it's safe to keep as-is.

PR #984, opened 2025-10-18
Refactor gen resources

scotts:
Two refactors:

1. Resource generation, `generate_reference_resources.py`, is now more deeply integrated with our testing utils. This allows us to greatly simplify the resource generation, and it keeps the actual definitions in `utils.py`. That is, the video we define in `utils.py` defines the filename, and it has the logic for how to name reference frames.
2. Moves around definitions in `utils.py` so that we group all video and audio test files together.

### DIFF
--- a/test/utils.py
+++ b/test/utils.py
@@ -392,7 +392,10 @@ class TestVideo(TestContainerFile):
         if stream_index is None:
             stream_index = self.default_stream_index
 
-        tensor_file_path = f"{self.get_base_path_by_index(idx, stream_index=stream_index, filters=filters)}.pt"
+        base_path = self.get_base_path_by_index(
+            idx, stream_index=stream_index, filters=filters
+        )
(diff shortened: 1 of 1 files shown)


scotts:
@NicolasHug, good question. While we generated all of these resources originally with FFmpeg 4, they're actually stable across all versions we test against. For example, I've been using FFmpeg 6 for all of my testing the past several months, and when I re-run the resource generation script, everything comes out clean - we're bit-for-bit compatible with what we originally checked in.

This may seem surprising at first, but if we expect TorchCodec to be bit-for-bit compatible across versions, it's not surprising if the `ffmpeg` CLI is as well. I think that at the least, we'd want to know if our resource generation script was giving different results across FFMpeg versions.

I think we do have a comment somewhere about using FFmpeg 4. We should probably delete it, if we haven't already.

PR #987, opened 2025-10-20
Add `_core._get_backend_details()` utility for testing and debugging

NicolasHug:
This PR adds a `_core` utility to get details about the current backend / interface being used. It reports:

- the name of the interface (CPU, FFmpeg CUDA, Beta CUDA)
- whether a CPU fallback is happening for FFmpeg CUDA and Beta CUDA.

This utility is needed for testing, see resolved TODO in our tests. I also expect it will also be useful for debugging purposes, e.g. when working on https://github.com/meta-pytorch/torchcodec/issues/943. We might want to eventually expose it publicly, so that users know whether a fallback is happening or not.

ATM, the returned value is a string with no particular structure. I think that's fine for now. Everything is private. We can think of defining something more structured later (keeping in mind that our custom ops don't have much flexibility in terms of what can be returned).

The new functionality is exposed as:

- `_core._get_backend_details(decoder: torch.Tensor) -> str`
- `std::string SingleStreamDecoder::getInterfaceDetails()`
- `std::string DeviceInterface::getDetails()`

Happy to bikeshed on the names.

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -355,7 +355,7 @@ std::optional<const AVCodec*> CudaDeviceInterface::findCodec(
 }
 
 std::string CudaDeviceInterface::getDetails() {
-  // Note: for this interface specificaly the fallback is only known after a
+  // Note: for this interface specifically the fallback is only known after a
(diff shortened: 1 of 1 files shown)


scotts:
There's at least one other thing we may want to expose such a private function for for testing: color space library. We have tests that exercise the different color space libraries, but we can't know for sure from the outside which one we're using.

PR #988, opened 2025-10-20
Use FPS in C++ layer

mollyxu:
Use FPS in c++ layer if number of frames is missing


mollyxu:
blocked by #1009

mollyxu:
Addressed in #1021

Issue #945, opened 2025-10-08
Some videos' `duration` is missing, and they can't work with approximate mode


NicolasHug:
Some videos, like these ones

https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/test/utils.py#L752-L756

simply *do not* work with approximate mode. They're all missing the `duration_from_header` metadata because `avStream->duration` is invalid here:

https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/src/torchcodec/_core/SingleStreamDecoder.cpp#L121-L124

and they end up failing with:

https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/src/torchcodec/decoders/_video_decoder.py#L420-L424

which isn't a particularly helpful error message, especially since `ERROR_REPORTING_INSTRUCTIONS` says:

https://github.com/meta-pytorch/torchcodec/blob/9fa4fd1e8456b5a540b016e21683155956533262/src/torchcodec/decoders/_decoder_utils.py#L15-L18

----

We should, at the very least, provide a more sensible error message. But we should also probably try to figure what we can do with those videos to support approximate mode. `ffprobe` is able to report the duration of those videos: we should try to see if they're able to report it without scanning. 

And if really, we conclude that they can't support approximate mode, we should probably automatically fallback to exact mode (CC @scotts  as this was your idea!)

mollyxu:
I ran the following commands to check the following videos (`AV1_VIDEO, TEST_SRC_2_720P_VP9, TEST_SRC_2_720P_VP8`) and their duration. It seems like while they are missing stream-level duration, they have container-level duration. Could we potentially use container-level duration as a fallback for missing stream-level duration?

stream-level:
`ffprobe -v error -show_entries stream=duration -of default=noprint_wrappers=1:nokey=1 {file-path}`
container-level:
`ffprobe -v error -show_entries format=duration -of default=noprint_wrappers=1:nokey=1 {file-path}`

NicolasHug:
> Could we potentially use container-level duration as a fallback for missing stream-level duration?

Yes!
Thanks for looking into this @mollyxu , let's use the container duration as a fallback

mollyxu:
#989

Issue #923, opened 2025-10-03
C++ layer should use FPS if number of frames is missing


scotts:
When we report metadata at the Python layer, if we are in approximate mode and the number of frames is missing from the header, we estimate the number of frames using the FPS:
https://github.com/meta-pytorch/torchcodec/blob/401901e56349c0a901c26908e3c1ad3dc4819f98/src/torchcodec/_core/_metadata.py#L141-L157
However, in C++, we don't do the same:
https://github.com/meta-pytorch/torchcodec/blob/401901e56349c0a901c26908e3c1ad3dc4819f98/src/torchcodec/_core/SingleStreamDecoder.cpp#L1591-L1603
We should make the C++ layer have the same behavior as the Python layer. A downside here is that we don't have a test for this behavior. We've tested missing metadata in the Python layer by just overwriting values with `None`. To really test the C++ layer, we should get a video which is actually missing the number of frames from its metadata.

mollyxu:
Addressed in #1021

Issue #125, opened 2024-07-30
Align C++ names with Python names


NicolasHug:
Some of the names we use in C++ are slightly different from the ones we expose in Python. Currently the main ones are visible from the metadata:

https://github.com/pytorch/torchcodec/blob/9e3a2196d7361e86f02f5b12430477c3e872898c/src/torchcodec/decoders/_core/_metadata.py#L162-L169

We should align the C++ names with the Python ones.

scotts:
As raise din #440, we could expand this issues scope to something like "Move metadata source of truth to C++." Basically, we make decisions right now in the Python level about what kind of metadata is available, and what to return when it's not. We could simplify everything and do this only at the C++ level, and just report it back to the Pyton level. This would potentially also involve deprecating the JSON intermediary.

scotts:
@NicolasHug, this is done, right?

NicolasHug:
I addressed a fairly big chunk of that already, but we still have a fair amount of discrepancy in the metadata fields. And then there's also https://github.com/pytorch/torchcodec/issues/125#issuecomment-2590932332

PR #989, opened 2025-10-20
Fallback to container duration in approximate mode

mollyxu:
Fallback to container duration in approximate mode when stream duration is unavailable
#945

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -100,6 +100,26 @@ void SingleStreamDecoder::initializeDecoder() {
       "Failed to find stream info: ",
       getFFMPEGErrorStringFromErrorCode(status));
 
+  if (formatContext_->duration > 0) {
+    AVRational defaultTimeBase{1, AV_TIME_BASE};
+    containerMetadata_.durationSecondsFromHeader =
+        ptsToSeconds(formatContext_->duration, defaultTimeBase);
+  }
+
+  if (formatContext_->bit_rate > 0) {
+    containerMetadata_.bitRate = formatContext_->bit_rate;
+  }
+
+  int bestVideoStream = getBestStreamIndex(AVMEDIA_TYPE_VIDEO);
+  if (bestVideoStream >= 0) {
+    containerMetadata_.bestVideoStreamIndex = bestVideoStream;
+  }
+
+  int bestAudioStream = getBestStreamIndex(AVMEDIA_TYPE_AUDIO);
+  if (bestAudioStream >= 0) {
+    containerMetadata_.bestAudioStreamIndex = bestAudioStream;
+  }
+
   for (unsigned int i = 0; i < formatContext_->nb_streams; i++) {
     AVStream* avStream = formatContext_->streams[i];
     StreamMetadata streamMetadata;
@@ -149,34 +169,10 @@ void SingleStreamDecoder::initializeDecoder() {
       containerMetadata_.numAudioStreams++;
     }
 
+    streamMetadata.durationSecondsFromContainer =
+        containerMetadata_.durationSecondsFromHeader;
+
     containerMetadata_.allStreamMetadata.push_back(streamMetadata);
-  }
-
-  if (formatContext_->duration > 0) {
-    AVRational defaultTimeBase{1, AV_TIME_BASE};
-    containerMetadata_.durationSecondsFromHeader =
-        ptsToSeconds(formatContext_->duration, defaultTimeBase);
-  }
-
-  if (formatContext_->bit_rate > 0) {
-    containerMetadata_.bitRate = formatContext_->bit_rate;
-  }
-
-  int bestVideoStream = getBestStreamIndex(AVMEDIA_TYPE_VIDEO);
-  if (bestVideoStream >= 0) {
-    containerMetadata_.bestVideoStreamIndex = bestVideoStream;
-  }
-
-  int bestAudioStream = getBestStreamIndex(AVMEDIA_TYPE_AUDIO);
-  if (bestAudioStream >= 0) {
-    containerMetadata_.bestAudioStreamIndex = bestAudioStream;
-  }
-
-  if (containerMetadata_.durationSecondsFromHeader.has_value()) {
-    for (auto& streamMetadata : containerMetadata_.allStreamMetadata) {
-      streamMetadata.durationSecondsFromContainer =
-          containerMetadata_.durationSecondsFromHeader;
-    }
   }
 
   if (seekMode_ == SeekMode::exact) {
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -44,7 +44,8 @@ class StreamMetadata:
     from the actual frames if a :term:`scan` was performed. Otherwise we
     fall back to ``duration_seconds_from_header``. If that value is also None,
     we instead calculate the duration from ``num_frames_from_header`` and
-    ``average_fps_from_header``.
+    ``average_fps_from_header``. If all of those are unavailable, we fall back
+    to the container-level ``duration_seconds_from_header``.
     """
     begin_stream_seconds: Optional[float]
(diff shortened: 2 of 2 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -158,6 +158,16 @@ void SingleStreamDecoder::initializeDecoder() {
         ptsToSeconds(formatContext_->duration, defaultTimeBase);
   }
 
+  // Use container duration as fallback for streams missing stream duration
+  if (containerMetadata_.durationSecondsFromHeader.has_value()) {
+    for (auto& streamMetadata : containerMetadata_.allStreamMetadata) {
+      if (!streamMetadata.durationSecondsFromHeader.has_value()) {
+        streamMetadata.durationSecondsFromHeader =
+            containerMetadata_.durationSecondsFromHeader;
```
scotts: Rather than assigning the container's duration to the stream's duration, I think we should do something like:
```c++
if (!streamMetadata.durationSectonsFromHeader.has_value()) {
  return containerMetadata_.durationSecondsFromHeader;
}
```
I think that keeps it clear where we're getting our information from. So far, we've kept that explicit in our existing metadata fields.

One wrinkle here is that the metadata we report to users in Python is [`VideoStreamMetadata`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/_metadata.py#L49). That is, it's explicitly the *stream* metadata. We don't have a hierarchically clean way to add the container metadata there. So while it's kinda odd, maybe we should add a field for `duration_seconds_from_container` on the Python side.

To be clear, my current thinking is that:

1. In C++, we fall back to the container metadata where appropriate. This may be awkward because we don't pass around the container metadata, so we may want to add a `durationSecondsFromContainer` field to `StreamMetadata`.
2. In Python, we should add a field for `duration_seconds_from_container` to `StreamMetadata`, and we have logic to fall back to it in `duration_seconds`.

@NicolasHug, @Dan-Flores, I'm curious to hear your thoughts as well.

NicolasHug: @scotts  can you help me understand where you think we should add this?

```Cpp
if (!streamMetadata.durationSectonsFromHeader.has_value()) {
  return containerMetadata_.durationSecondsFromHeader;
}
```

-----

I agree with your general point that we should avoid setting the stream metadata to the container metadata.

We might be able to avoid defining a new `*_from_container` attribute on the `StreamMetadata`: the [`ContainerMetadata`](https://github.com/meta-pytorch/torchcodec/blob/75a3325f016483ae1a6af8516b804cc72da39ea6/src/torchcodec/_core/_metadata.py#L205) metaclass exists in Python. We just don't expose it currently. 

The original error happens here:

https://github.com/meta-pytorch/torchcodec/blob/75a3325f016483ae1a6af8516b804cc72da39ea6/src/torchcodec/decoders/_video_decoder.py#L413-L418

and this is a place where we do know the `ContainerMetadata`. Maybe we can just fallback to it in that function only (I haven't checked that this would solve the original issue). And, whether we want to expose `ContainerMetadata` is kind of a separate issue.


----

As a side note, reasoning about all these fallbacks hurts my brain. We have fallbacks in the Python `StreamMetadata` class, in the C++ `initializeDecoder()`, and in the C++ helpers like `getNumFrames`. It might be time we address our long-time issue of centralizing all of the fallback logic in C++ (https://github.com/meta-pytorch/torchcodec/issues/125#issuecomment-2590932332)

mollyxu: @scotts @NicolasHug Thank you for the feedback! 

I tried implementing fallback at line 413-418 in `_video_decoder.py` but I think that `num_frames` would be none. This is due to `metadata.num_frames` not having the container duration when it computes `num_frames` using `duration_seconds_from_header` when `num_frames` is missing from the video.

scotts: @NicolasHug, 

> @scotts can you help me understand where you think we should add this?
>```
> if (!streamMetadata.durationSectonsFromHeader.has_value()) {
>   return containerMetadata_.durationSecondsFromHeader;
> }
>```

Right now, there's only one place we really use the duration's value in C++: [`getMaxSeconds()`](https://github.com/meta-pytorch/torchcodec/blob/28cec519b259c0e7b1282f3c54839ddbfd3c51f0/src/torchcodec/_core/SingleStreamDecoder.cpp#L1592-L1604). That would become:
```
std::optional<double> SingleStreamDecoder::getMaxSeconds(
    const StreamMetadata& streamMetadata) {
  switch (seekMode_) {
    case SeekMode::custom_frame_mappings:
    case SeekMode::exact:
      return streamMetadata.endStreamPtsSecondsFromContent.value();
    case SeekMode::approximate: {
      //
      // Fallback to container duration if stream does not have one
      //
      if (!streamMetadata.durationSecondsFromHeader.has_value()) {
        return containerMetadata_.durationSecondsFromHeader;
      }
      return std::nullopt;
    }
    default:
      TORCH_CHECK(false, "Unknown SeekMode");
  }
}
```
Agreed that now may be the time to consolidate all of this fall back logic in C++. The fallback logic in Python is not straight-forward to understand because it's close to being a state machine. We're trying to be as tolerant of missing values as possible, which means that value `a` may try to fallback to value `b` in some instances, but value `b` may also try to fallback to value `a` in other instances. Implementing this logic in one place, for all uses, will at least make sure that we're consistent across Python and C++.

Part of why this logic is convoluted in Python is that the Python metadata class has no knowledge of seek modes. Instead, it has fallback logic exclusively based on what values are present.

mollyxu:
Thank you all for the discussion and feedback on this PR. 

To summarize our findings, the current state of our fallback logic is scattered across both C++ and Python. Trying to resolve the specific error in issue #945 by modifying `_video_decoder.py` still won't allow us to decode the videos in #945 because `num_frames` is `None` in those videos. This speaks to the greater issue that our current fallback logic in Python is close to being a state machine. It's best to consolidate our fallback logic all in C++ via a refactor before tackling this issue (#945) and related issues ( #923 )

mollyxu:
blocked by #1009

mollyxu:
Since #1009 is resolved, I reimplemented falling back to container mode with the refactored code.

I tried following the design decisions that were discussed in this thread, mainly 
- adding a field (`durationSecondsFromContainer`) to streamMetadata that makes it clear where we are getting the data from 
- keeping the fallback logic in Metadata.cpp

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -172,6 +172,13 @@ void SingleStreamDecoder::initializeDecoder() {
     containerMetadata_.bestAudioStreamIndex = bestAudioStream;
   }
 
+  if (containerMetadata_.durationSecondsFromHeader.has_value()) {
+    for (auto& streamMetadata : containerMetadata_.allStreamMetadata) {
+      streamMetadata.durationSecondsFromContainer =
+          containerMetadata_.durationSecondsFromHeader;
+    }
+  }
```
scotts, line 176: I think we can simplify all this with a little refactoring. Let's take all of the pre-existing container metadata logic and do that _before_ we do the per-stream for loop. Then, when looking at each stream, we can just do:
```c++
streamMetadata.durationSecondsFromContainer = containerMetadata_.durationSecondsFromHeader;
```
Note that we don't need to check anything; it's optionals on both side.

The principle here is that it's more efficient and simpler to loop over something once. The efficiency bit is minor here, as we do this once at startup, and most videos will probably have less than 10 streams. But it's usually a lot easier to understand code that loops over something once.

scotts:
@mollyxu, this is so much cleaner with your recent refactoring! Let's do the refactoring I mentioned, and let's also update the doc string to tell users that we also fallback to the container metadata if all of the other metadata is missing: https://github.com/meta-pytorch/torchcodec/blob/1ea235aa6b71bc9d6ce070d4b849dc60974082b5/src/torchcodec/_core/_metadata.py#L41-L48

PR #990, opened 2025-10-21
Create Python API for VideoEncoder

Dan-Flores:
This PR creates a simple `VideoEncoder` class, and updates several tests to utilize the `VideoEncoder.to_file` pattern.

* `test_bad_input_parameterized` is the `VideoEncoder` equivalent to the [AudioEncoder test](https://github.com/meta-pytorch/torchcodec/blob/fe738a6c0bff09c0d3f262e5b6a58d4cf3c785c1/test/test_encoders.py#L150) that ensures general error checking occurs, while test_bad_input tests method specific errors.

* `test_contiguity`: is the `VideoEncoder` equivalent to the [AudioEncoder test](https://github.com/meta-pytorch/torchcodec/blob/fe738a6c0bff09c0d3f262e5b6a58d4cf3c785c1/test/test_encoders.py#L410) to ensure contiguous and non-contiguous tensors can be encoded, and are encoded equivalently.

### DIFF
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -588,7 +588,7 @@ class TestVideoEncoder:
             getattr(encoder, method)(**valid_params)
 
         with pytest.raises(
-            ValueError, match=r"Expected 3D or 4D frames, got frames.shape = torch.Size"
+            ValueError, match=r"Expected 4D frames, got frames.shape = torch.Size"
         ):
             encoder = VideoEncoder(
                 frames=torch.zeros(10),


test/test_encoders.py
```diff
@@ -564,3 +564,117 @@ def write(self, data):
...
+    @pytest.mark.parametrize("method", ("to_file", "to_tensor", "to_file_like"))
+    def test_contiguity(self, method, tmp_path):
+        # Ensure that 2 sets of video frames with the same pixel values are encoded
+        # in the same way, regardless of their memory layout. Here we encode 2 equal
+        # frame tensors, one is contiguous while the other is non-contiguous.
+
+        num_frames, channels, height, width = 5, 3, 64, 64
+        contiguous_frames = (
+            (torch.rand(num_frames, channels, height, width) * 255)
+            .to(torch.uint8)
```
NicolasHug: Just use `torch.randint(0, 256, size=(num_frames, channels, height, width), dtype=torch.uint8).contiguous()`

test/test_encoders.py
```diff
@@ -564,3 +564,117 @@ def write(self, data):
...
+        )
+        assert contiguous_frames.is_contiguous()
+
+        # Create non-contiguous frames by permuting, calling contiguous to update memory layout,
+        # then permuting back to the initial order
+        non_contiguous_frames = (
+            contiguous_frames.permute(0, 3, 2, 1).contiguous().permute(0, 3, 2, 1)
+        )
+        assert non_contiguous_frames.stride() != contiguous_frames.stride()
+        assert not non_contiguous_frames.is_contiguous()
```
NicolasHug, line 649: This is good, but you should be able to check for this which is stricter:

```
assert  non_contiguous_frames.is_contiguous(memory_format=torch.channels_last)
```

Dan-Flores: So far, I have assumed frames will be in `NCHW` format. This shape is retained after the permutations, so I do not believe `channels_last` applies here?

NicolasHug: The frames are always of `NCHW` **shape** . But internally they can have any arbitrary memory layout (format). These are mostly orthogonal concepts, i.e. frames can have `NCHW` shape while still being represented as `NHWC` layout (channels last).

I realize now the `non_contiguous_frames` you created are indeed *not* channels last. But we can force them to be by changing to:

```py
contiguous_frames.permute(0, 2, 3, 1).contiguous().permute(0, 3, 1, 2)
```

and you should see that `assert  non_contiguous_frames.is_contiguous(memory_format=torch.channels_last)` is passing now. Note how we're now permuting the channels dimension from `N*HW` to `NHW*`, and back.

Your test was still correct BTW - I just think it's good to actually explicitly check for `channels_last` because this is how frames may arrive.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class VideoEncoder:
+    """A video encoder.
+
+    Args:
+        frames (``torch.Tensor``): The frames to encode. This must be a 4D
+            tensor of shape ``(N, C, H, W)`` where N is the number of frames,
+            C is 3 channels (RGB), H is height, and W is width.
+            A 3D tensor of shape ``(C, H, W)`` is also accepted as a single RGB frame.
```
NicolasHug: >             A 3D tensor of shape ``(C, H, W)`` is also accepted as a single RGB frame.


Q - Do we need to support that? I'm wondering if it makes a lot of sense to just encode a single image as a video. I suspect this was made to mimic the AudioEncoder behavior but that was a different use-case. In the AudioEncoder we want to allow for 1D audio to be supported as it's still a valid waveform. But I don't think we need to treat a single frame as a valid video.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -0,0 +1,97 @@
...
+            Values must be uint8 in the range ``[0, 255]``.
+        frame_rate (int): The frame rate to use when encoding the
+            **input** ``frames``.
```
NicolasHug: My interpretation of this description is that this parameter actually defines the the frame rate of the encoded output, because of the "frame rate to use when encoding" part.

I think it might be less ambiguous as

> The frame rate of the **input** ``frames``. This is *not* the frame rate of the encoded output.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -0,0 +1,97 @@
...
+    def to_file(
+        self,
+        dest: Union[str, Path],
+    ) -> None:
+        """Encode frames into a file.
+
+        Args:
+            dest (str or ``pathlib.Path``): The path to the output file, e.g.
+                ``video.mp4``. The extension of the file determines the video
+                format and container.
```
NicolasHug: What's the distinction between "format" and "container" here? I would just use one or the other?

Dan-Flores: The terms are used interchangeably, so I included both here to be understandable to users of both terms. Let me know if that is actually more confusing that just naming one term.

I might have added the distinction after discovering the format `mkv` is defined as a `matroska` container. I have not encountered other formats where these are different.

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: I'm not sure what the use case is for encoding an image as a video, but since FFmpeg allows encoding an image to video, I believe we can retain this functionality for a relatively low cost.

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: We use this variable in the `Encoder.cpp`: `avCodecContext_->framerate = {inFrameRate_, 1};`. 
Perhaps the naming is confusing (should be `inputFrameRate` rather than `inFrameRate`), but is it not accurate to say this defines the frame rate of the encoded output?

NicolasHug: Good point, you're right, `frame_rate` and `inFrameRate` currently define *both* the frame rate of the input *and* the frame rate of the encoded output.

But in a future implementation, we might let the user choose to encode the output with a different frame rate. If/when that happens, then this `frame_rate` parameter will only be about the *input* frame rate, while the `frame_rate` parameter that we'll add to the decoding method (`to_file(frame_rate=...)`) will define the output frame rate. It will be similar to what we do in the `AudioEncoder` with the `sample_rate` parameters: 

- one is for the input, in the constructor: https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/encoders/_audio_encoder.py#L18-L20 
- one is for the output, in the methods: https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/encoders/_audio_encoder.py#L66-L67


My original docstring suggestion was definitely wrong (thanks for catching!) but I'd still suggest to clarify that this is really the *input* frame rate. And, for now, it also happens to be what we use for the output frame rate:

> The frame rate of the input frames. Also defines the encoded output frame rate.

src/torchcodec/encoders/_video_encoder.py
NicolasHug: I feel like saying

> determines the video format and container.

may suggest to some readers that "format" and "container" refer to different things. Especially because a) "format" is sometimes used to refer to the codec, and b) in the methods we only have one single "format" parameter and we never mention the "container".

How about using the term "container format" everywhere in the docstrings? I think that is the actual correct technical term for that ([wikipedia](https://en.wikipedia.org/wiki/Container_format)):

> Notable examples of container formats include [...] formats used for multimedia playback (such as [Matroska](https://en.wikipedia.org/wiki/Matroska), [MP4](https://en.wikipedia.org/wiki/MPEG-4_Part_14), and [AVI](https://en.wikipedia.org/wiki/Audio_Video_Interleave)).

 The parameter can still be `format` - I think this makes sense for consistency with the AudioEncoder, but in the docstrings of all methods we could clarify that `format` corresponds to the "container format"?

src/torchcodec/encoders/_video_encoder.py
NicolasHug: > I'm not sure what the use case is for encoding an image as a video

I'm not sure either - let's not do it then :). One principle we have is that we should only implement things that we *know* are useful. In doubt, it's best not to expose it, it makes our life easier as maintainers. Of course if users start to explicitly request this feature, we may reconsider.

In the near future we will start implementing proper image encoders in torchcodec, and another guiding principle is that  *there should be one-- and preferably only one --obvious way to do it.* ([ref](https://peps.python.org/pep-0020/#the-zen-of-python)). If we were to support 3D tensors (images) into the `VideoEncoder` then we'd potentially expose two ways to do the same thing.

It doesn't mean we'll prevent users from doing it if they really want to: a user can still just call `unsqueeze(0)` on their 3D tensor and pass it. It will work. But we don't need to explicitly support 3D tensors on our side.

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: Thanks for clarifying this - I'll make a note to add `frame_rate` to the encoding methods (`to_file`, etc) next.

mscheltienne:
Hello, thanks for this great addition, I'm looking forward to using it. Speaking of which, seeing the ~monthly release cadence; do you have any estimate on when the next release `0.9.0` including this feature will be available?

PR #992, opened 2025-10-21
CPU fallback: do color-conversion on GPU.

NicolasHug:
Towards https://github.com/meta-pytorch/torchcodec/issues/943

This PR speeds up the BETA Cuda interface when we need to fallback on the CPU. The idea is simple: we do the color conversion on the GPU instead of doing them on the CPU. This has 2 benefits:

- the color conversion is faster since it runs on GPU
- the CPU -> GPU transfer is faster since we now transfer a smaller YUV frame instead of a bigger RGB frame.

**Before**

```
Decode Frame on CPU (fallback) -> YUV to RGB Conversion **on CPU** -> Send bigger RGB frame from CPU to GPU
```

**Now**

```
Decode Frame on CPU (fallback) -> Send smaller YUV frame from CPU to GPU -> YUV to RGB conversion **on GPU**
```

This is for the BETA interface only. I'll handle the FFmpeg interface as a follow-up.

### Benchmarks

Benchmarks show 1.6X speed up on 1080p frames. We should expect the larger speed-ups for larger resolutions. I used this snippet.

<details>

```py


import torch
import argparse
from pathlib import Path
from torchcodec.decoders import VideoDecoder, set_cuda_backend
from joblib import Parallel, delayed


    for _ in range(warmup):
        f(*args, **kwargs)

    times = []
    for _ in range(num_exp):
        start = perf_counter_ns()
        f(*args, **kwargs)
        end = perf_counter_ns()
        times.append(end - start)


    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    return med


def decode_one_video(video_path):
       decoder = VideoDecoder(str(video_path), device="cuda:0", seek_mode="approximate")
    indices = torch.arange(len(decoder))
    decoder.get_frames_at(indices)

    torch.cuda.synchronize()


parser.add_argument("video_path")
args = parser.parse_args()

times = bench(decode_one_video, video_path=args.video_path, warmup=1, num_exp=10)
report_stats(times)

```

</details>

It's impossible to benchmark the new and old strategies together, since we need recompilation. Also I had to add modify our code to enforce the CPU fallback to be activated on videos that would otherwise not be falling back (basically just changed `if (!nativeNVDECSupport(codecContext))...` to `if(true)`)

<details>

```
#  OLD
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 543.16ms +- 56.39
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 533.95ms +- 55.50
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 659.17ms +- 56.99
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 549.27ms +- 53.50
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 538.66ms +- 34.52

#  NEW
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 318.23ms +- 8.21
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 319.42ms +- 11.30
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 325.05ms +- 9.74
~/dev/torchcodec-cuda (fallback-colorconversion*) » python benchmark_fallback.py h264_1080.mp4                           nicolashug@nicolashug-fedora-PW0H326Y
med = 325.94ms +- 12.64
```

</details>

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -729,6 +729,9 @@ UniqueAVFrame BetaCudaDeviceInterface::transferCpuFrameToGpuNV12(
       convertedHeight == height, "sws_scale failed for CPU->NV12 conversion");
 
   int ySize = width * height;
+  TORCH_CHECK(
+      ySize % 2 == 0,
+      "Y plane size must be even. Please report on TorchCodec repo.");
   int uvSize = ySize / 2; // NV12: UV plane is half the size of Y plane
   size_t totalSize = static_cast<size_t>(ySize + uvSize);
 
@@ -767,6 +770,9 @@ UniqueAVFrame BetaCudaDeviceInterface::transferCpuFrameToGpuNV12(
       "Failed to copy Y plane to GPU: ",
       cudaGetErrorString(err));
 
+  TORCH_CHECK(
+      height % 2 == 0,
+      "height must be even. Please report on TorchCodec repo.");
   err = cudaMemcpy2D(
       gpuFrame->data[1],
       gpuFrame->linesize[1],


src/torchcodec/_core/CpuDeviceInterface.cpp
NicolasHug, line 1: As can be seen above we are now using `swscale` in `BetaCudaInterface.cpp` to do the `YUV -> NV12` conversion. So I had to extract out all the `swscale` logic away from `CpuDeviceInteface.cpp` and put it into `FFmpegCommon.cpp` so it can be reused across the two interfaces. Almost everything below this comment can be ignored and treated as copy/pasting code around. Just pay attention to the test at the bottom.

test/test_decoders.py
```diff
@@ -1709,11 +1709,23 @@ def test_beta_cuda_interface_cpu_fallback(self):
...
+            beta_dec = VideoDecoder(H265_VIDEO.path, device="cuda")
+
+        assert (
+            _core._get_backend_details(beta_dec._decoder)
+            == "Beta CUDA Device Interface. Using CPU fallback."
+        )
+        beta_frame = beta_dec.get_frame_at(0)
 
-        torch.testing.assert_close(ffmpeg.data, beta.data, rtol=0, atol=0)
+        assert psnr(ref_frames.data.cpu(), beta_frame.data.cpu()) > 25
```
Dan-Flores: Does comparing frames on GPU vs CPU change floating point precision, or is there some other reason to move the frames here?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -665,38 +671,157 @@ void BetaCudaDeviceInterface::flush() {
...
+  TORCH_CHECK(cpuFrame != nullptr, "CPU frame cannot be null");
+
+  int width = cpuFrame->width;
+  int height = cpuFrame->height;
+
+  // intermediate NV12 CPU frame. It's not on the GPU yet.
+  UniqueAVFrame nv12CpuFrame(av_frame_alloc());
+  TORCH_CHECK(nv12CpuFrame != nullptr, "Failed to allocate NV12 CPU frame");
+
+  nv12CpuFrame->format = AV_PIX_FMT_NV12;
```
Dan-Flores, line 697: Is it accurate to say that NV12 very similar to `AV_PIX_FMT_YUV420P` (uses YUV, and has 4:2:0 chroma subsampling), but we use NV12 here because that is the format the NPP library requires? As explained in [this comment](https://github.com/meta-pytorch/torchcodec/blob/75a3325f016483ae1a6af8516b804cc72da39ea6/src/torchcodec/_core/CudaDeviceInterface.cpp#L294-L295)

NicolasHug: Yes that's exactly right. NV12 would contain the exact same values as `AV_PIX_FMT_YUV420P`, just ordered a bit differently.

test/test_decoders.py
NicolasHug: there will be a small difference due to floating point precision, but that wasn't the reason.
In fact... there was no good reason to call `.cpu()`, that was probably the result of copy/pasting from my previous `write_png` debug. I removed it, thanks for catching!

src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -250,4 +251,30 @@ AVFilterContext* createBuffersinkFilter(
     AVFilterGraph* filterGraph,
     enum AVPixelFormat outputFormat);
 
+struct SwsFrameContext {
```
scotts, line 254: A refactoring we should probably do is pull all sws logic into its own class in separate `.h` and `.cpp` files, similar to what we have with `FilterGraph`. But not in this PR.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -665,38 +671,157 @@ void BetaCudaDeviceInterface::flush() {
...
+      cpuFrame->linesize,
+      0,
+      height,
+      nv12CpuFrame->data,
+      nv12CpuFrame->linesize);
+  TORCH_CHECK(
+      convertedHeight == height, "sws_scale failed for CPU->NV12 conversion");
+
+  int ySize = width * height;
+  int uvSize = ySize / 2; // NV12: UV plane is half the size of Y plane
```
scotts, line 735: Integer rounding is okay? This is implicitly a floor operation.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -665,38 +671,157 @@ void BetaCudaDeviceInterface::flush() {
...
+      "Failed to copy Y plane to GPU: ",
+      cudaGetErrorString(err));
+
+  err = cudaMemcpy2D(
+      gpuFrame->data[1],
+      gpuFrame->linesize[1],
+      nv12CpuFrame->data[1],
+      nv12CpuFrame->linesize[1],
+      width,
+      height / 2,
```
scotts, line 782: Ditto for integer rounding - okay here?

test/test_decoders.py
```diff
@@ -1709,11 +1709,23 @@ def test_beta_cuda_interface_cpu_fallback(self):
...
+            beta_dec = VideoDecoder(H265_VIDEO.path, device="cuda")
+
+        assert (
+            _core._get_backend_details(beta_dec._decoder)
+            == "Beta CUDA Device Interface. Using CPU fallback."
+        )
+        beta_frame = beta_dec.get_frame_at(0)
 
-        torch.testing.assert_close(ffmpeg.data, beta.data, rtol=0, atol=0)
+        assert psnr(ref_frames.data, beta_frame.data) > 25
```
scotts, line 1728: How did you choose 25? I ask because I'm probably going to have to do something similar with decoder-native transforms.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Thanks for catching this, I forgot to look into it. Will report back.

test/test_decoders.py
NicolasHug: This is just the higher psnr value I could get that would pass the test. It's not great, 25 isn't a very good PSNR. But visually the frames are OK. The difference is mainly due to color-conversion algorithms being different on the GPU and on the CPU. Eventually, when the FFmpeg interface follows the same fallback mechanism, both frames will be exactly equal.

We should still try to have more CPU vs GPU tests for these code-paths though.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: I decided to go the easy route and just `TORCH_CHECK` for evenness. I think frame dimension are always expected to be even anyway.

PR #999, opened 2025-10-27
Add copyright header to 7 files in /packaging

Dan-Flores:


mollyxu:
Thanks for adding these copyright headers! I noticed that [packaging/build_ffmpeg.bat](https://github.com/meta-pytorch/torchcodec/blob/main/packaging/build_ffmpeg.bat) might also need a copyright header

Dan-Flores:
Updated to add copyright header to [packaging/build_ffmpeg.bat](https://github.com/meta-pytorch/torchcodec/blob/main/packaging/build_ffmpeg.bat?rgh-link-date=2025-10-27T19%3A50%3A36Z), but remove it from fake_smoke_test.py where it is not required.

PR #1002, opened 2025-10-27
Drop src/torchcodec/_core/ when including headers from _core

dvrogozh:
Follow up for https://github.com/meta-pytorch/torchcodec/pull/938#issuecomment-3450972100.

To facilitate development of out-of-tree torchcodec features (such as non-cuda gpu plugins) we need to make torchcodec headers includable by 3rd party projects. This means dropping `src/torchcodec/_core/` prefix when torchcodec self-includes headers.

CC: @scotts @NicolasHug

### DIFF
--- a/packaging/build_ffmpeg.bat
+++ b/packaging/build_ffmpeg.bat
@@ -1,3 +1,9 @@
+:: Copyright (c) Meta Platforms, Inc. and affiliates.
+:: All rights reserved.
+::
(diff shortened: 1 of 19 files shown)


dvrogozh:
Ah, I forgot to run lint locally. Sorry for that and thank you for the fix @NicolasHug.

NicolasHug (approved):
Thanks @dvrogozh !

PR #1003, opened 2025-10-27
Decoder-native resize public implementation

scotts:
Public API for decoder-native resize. The implementation in this PR accepts _both_ `torchvision.transforms.v2.Resize` and a newly defined `torchcodec.transforms.Resize`.

In #526, I had [initially proposed](https://github.com/meta-pytorch/torchcodec/issues/526#issuecomment-2688571431) _not_ using TorchVision transforms, and instead coming up with TorchCodec specific versions. @NicolasHug [proposed](https://github.com/meta-pytorch/torchcodec/issues/526#issuecomment-2690449903) that we accept TorchVision transforms, and that's what I followed up with in my design in #885.

After discussing the previous iteration of this PR, we agreed we wanted to see what it would look like to accept both. Having implemented this, I agree it's the right thing to do:

1. We now don't need to require TorchVision, even when using the decoder-native feature.
2. We have a natural place to document the behavior of each decoder-native transform that we accept, and what its limitations are compared to the TorchVision version of that transform.
3. We have a more principled mechanism of enforcing how TorchVision transforms map to decoder-native semantics. We still have to dig into the TorchVision object to get the info we need, but the `torchcodec.transforms` class is a clear representation in code of what is supported. In the old PR, that mapping was buried in the logic that turned the TorchVision transform directly into the specification string the core API needs.

Four points worth discussing:

1. I made the base class for all TorchCodec defined decoder-native transforms to be `DecoderTransform`. I think it would be confusing if it was just `Transform`, and `DecoderNativeTransform` seems both too long and too obscure.
2. I made the module path `torchcodec.transforms` instead of `torchcodec.decoder_transforms`. That's almost counter to point 1, but I think that there's less chance of confusion with the module path.
3. Should it be `DecoderResize` instead of just `Resize`?
4. The type annotation that users will see only mentions accepting `torchcodec.transforms.DecoderTransform`. It does not mention the TorchVision transforms or `nn.Module`. The _text_ of the docstring will say it, and I think that's enough?

### DIFF
--- a/docs/source/api_ref_transforms.rst
+++ b/docs/source/api_ref_transforms.rst
@@ -1,8 +1,8 @@
-.. _samplers:
+.. _transforms:
 
-===================
+=====================
 torchcodec.transforms
-===================
+=====================
 
 .. currentmodule:: torchcodec.transforms
 
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -70,10 +70,8 @@ class VideoDecoder:
         transforms (sequence of transform objects, optional): Sequence of transforms to be
             applied to the decoded frames by the decoder itself, in order. Accepts both
             :class:`~torchcodec.transforms.DecoderTransform` and
-            `torchvision.transforms.v2.Transform <https://docs.pytorch.org/vision/stable/transforms.html#v2-api-reference-recommended>`_
-            objects. All transforms are applied
-            in the ouput pixel format and colorspace. Read more about this parameter in:
-            TODO_DECODER_TRANSFORMS_TUTORIAL.
+            :class:`~torchvision.transforms.v2.Transform`
+            objects. Read more about this parameter in: TODO_DECODER_TRANSFORMS_TUTORIAL.
         custom_frame_mappings (str, bytes, or file-like object, optional):
             Mapping of frames to their metadata, typically generated via ffprobe.
             This enables accurate frame seeking without requiring a full video scan.
@@ -477,7 +475,7 @@ def _convert_to_decoder_transforms(
                     " DecoderTransform. TorchCodec also accept TorchVision "
                     "v2 transforms, but TorchVision is not installed."
                 )
-            if isinstance(transform, v2.Resize):
+            elif isinstance(transform, v2.Resize):
                 converted_transforms.append(Resize._from_torchvision(transform))
             else:
                 raise ValueError(
@@ -513,7 +511,7 @@ def _make_transform_specs(
         return ""
 
     transforms = _convert_to_decoder_transforms(transforms)
-    return ";".join([t._make_params() for t in transforms])
+    return ";".join([t._make_transform_spec() for t in transforms])
 
 
 def _read_custom_frame_mappings(
--- a/src/torchcodec/transforms/_decoder_transforms.py
+++ b/src/torchcodec/transforms/_decoder_transforms.py
@@ -6,6 +6,7 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
+from types import ModuleType
 from typing import Sequence
 
 from torch import nn
@@ -20,34 +21,41 @@ class DecoderTransform(ABC):
     should be both faster and more memory efficient than receiving normally
     decoded frames and applying the same kind of transform.
 
-    Most `DecoderTransform` objects have a complementary transform in TorchVision,
-    specificially in
-    `torchvision.transforms.v2 <https://docs.pytorch.org/vision/stable/transforms.html#v2-api-reference-recommended>`_.
-    For such transforms, we ensure that:
+    Most ``DecoderTransform`` objects have a complementary transform in TorchVision,
+    specificially in `torchvision.transforms.v2 <https://docs.pytorch.org/vision/stable/transforms.html>`_. For such transforms, we
+    ensure that:
 
       1. The names are the same.
       2. Default behaviors are the same.
-      3. The parameters for the `DecoderTransform` object are a subset of the
-         TorchVision transform object.
+      3. The parameters for the ``DecoderTransform`` object are a subset of the
+         TorchVision :class:`~torchvision.transforms.v2.Transform` object.
       4. Parameters with the same name control the same behavior and accept a
          subset of the same types.
       5. The difference between the frames returned by a decoder transform and
-         the complementary TorchVision transform are small.
-
-    All decoder transforms are applied in the output pixel format and colorspace.
+         the complementary TorchVision transform are such that a model should
+         not be able to tell the difference.
     """
 
     @abstractmethod
-    def _make_params(self) -> str:
+    def _make_transform_spec(self) -> str:
         pass
+
+
+def import_torchvision_transforms_v2() -> ModuleType:
+    try:
+        from torchvision.transforms import v2
+    except ImportError as e:
+        raise RuntimeError(
+            "Cannot import TorchVision; this should never happen, please report a bug."
+        ) from e
+    return v2
 
 
 @dataclass
 class Resize(DecoderTransform):
     """Resize the decoded frame to a given size.
 
-    Complementary TorchVision transform:
-    `torchvision.transforms.v2.Resize <https://docs.pytorch.org/vision/stable/generated/torchvision.transforms.v2.Resize.html#torchvision.transforms.v2.Resize>`_.
+    Complementary TorchVision transform: :class:`~torchvision.transforms.v2.Resize`.
     Interpolation is always bilinear. Anti-aliasing is always on.
 
     Args:
@@ -57,13 +65,13 @@ class Resize(DecoderTransform):
 
     size: Sequence[int]
 
-    def _make_params(self) -> str:
+    def _make_transform_spec(self) -> str:
         assert len(self.size) == 2
         return f"resize, {self.size[0]}, {self.size[1]}"
 
     @classmethod
     def _from_torchvision(cls, resize_tv: nn.Module):
-        from torchvision.transforms import v2
+        v2 = import_torchvision_transforms_v2()
 
         assert isinstance(resize_tv, v2.Resize)
 
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -209,6 +209,7 @@ autosummary_generate = True
 intersphinx_mapping = {
     "python": ("https://docs.python.org/3/", None),
     "torch": ("https://pytorch.org/docs/stable/", None),
+    "torchvision": ("https://docs.pytorch.org/vision/stable/", None),
     "numpy": ("https://numpy.org/doc/stable/", None),
     "PIL": ("https://pillow.readthedocs.io/en/stable/", None),
     "matplotlib": ("https://matplotlib.org/stable/", None),


src/torchcodec/_core/Transform.cpp
```diff
@@ -42,7 +42,7 @@ int toSwsInterpolation(ResizeTransform::InterpolationMode mode) {
 std::string ResizeTransform::getFilterGraphCpu() const {
   return "scale=" + std::to_string(outputDims_.width) + ":" +
       std::to_string(outputDims_.height) +
-      ":sws_flags=" + toFilterGraphInterpolation(interpolationMode_);
+      ":flags=" + toFilterGraphInterpolation(interpolationMode_);
```
scotts, line 33: From the [FFmpeg docs](https://ffmpeg.org/ffmpeg-filters.html#Filtergraph-syntax-1):

> Libavfilter will automatically insert [`scale`](https://ffmpeg.org/ffmpeg-filters.html#scale) filters where format conversion is required. It is possible to specify swscale flags for those automatically inserted scalers by prepending `sws_flags=flags;` to the filtergraph description.

Whereas `flags` is the [specific parameter to `scale`](https://ffmpeg.org/ffmpeg-filters.html#Options-2). They end up being semantically equivalent, but it's more clear to use the scale option here.

test/test_transform_ops.py
```diff
@@ -22,15 +22,69 @@
...
+        expected_shape = (NASA_VIDEO.get_num_color_channels(), height, width)
+        resize_filtergraph = f"scale={width}:{height}:flags=bilinear"
+        decoder_resize = VideoDecoder(NASA_VIDEO.path, transforms=[v2.Resize(size=(height, width))])
+        for frame_index in [17, 230, 389]:
+            frame_resize = decoder_resize[frame_index]
+            frame_ref = NASA_VIDEO.get_frame_data_by_index(frame_index, filters=resize_filtergraph)
+
+            assert frame_resize.shape == expected_shape
+            assert frame_ref.shape == expected_shape
+            assert_frames_equal(frame_resize, frame_ref)
```
scotts: Currently fails. Still investigating.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -103,6 +103,7 @@ def __init__(
         dimension_order: Literal["NCHW", "NHWC"] = "NCHW",
         num_ffmpeg_threads: int = 1,
         device: Optional[Union[str, torch_device]] = "cpu",
+        transforms: List[Any] = [],  # TRANSFORMS TODO: what is the user-facing type?
```
scotts: Discussion point 1: If we accept TorchVision transforms, and we want to lazily load TorchVision, what type do we advertise here? We can easily explain that we accept a TorchVision transform in the docs, but what should we put in the type annotation?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -432,6 +436,22 @@ def _get_and_validate_stream_metadata(
...
+        if isinstance(transform, v2.Resize):
+            if len(transform.size) != 2:
+                raise ValueError(
+                    f"Resize transform must have a (height, width) pair for the size, got {transform.size}."
+                )
+            transform_specs.append(f"resize, {transform.size[0]}, {transform.size[1]}")
+        else:
+            raise ValueError(f"Unsupported transform {transform}.")
+    return ";".join(transform_specs)
+
```
scotts, line 515: Discussion point 2: This is what we'll have to do with TorchVision transforms at the moment. We'll need special handling for each transform, looking into its internals to get what we need and enforce decoder-native limitations.

In the future, we can change TorchVision transforms to have an API so that we can get what we need in a generic way. But for now, we'll need to do something like this.

src/torchcodec/decoders/_video_decoder.py
NicolasHug: It should probably be either `Any` or `nn.Module`, which is the base class of all torchvision v2 transforms, and something users are familiar with since this is the core building block of any pytorch model.

NicolasHug: Note: it can be dangerous to use *mutable* values as parameter defaults, because this `[]` object now becomes a global value that can be changed across invocations. If anywhere within `VideoDecoder` we (as developers) were to write something that mutates `tranforms` inplace, we'd be changing the default value:

```py
In [1]: def f(transforms=[]): print(transforms); transforms.append("hello")

In [2]: f()
[]

In [3]: f()
['hello']

In [4]: f()
['hello', 'hello']
```

Best to use `None` or a non-mutable like `tuple()` (we'll probably want to accept sequences, not just lists)

src/torchcodec/decoders/_video_decoder.py
NicolasHug: I'm still undecided on whether we should accept TV transforms or not (ironic, I know), but I think this is totally OK.

And I think we'll need that level of coupling anyway, even if we were to write our own TC transforms. Echoing what you wrote:

> If we were to [...] create ]TorchCodec-specific user specification API, we'd want to make sure that its semantics match that of TorchVision. That is, if we had torchcodec.transforms.Resize(height=x, width=y), we'd want to make sure its semantics matched torchvision.transforms.v2.Resize(size=(x,y)). In that specific example, we'd want to make sure that both default to bilinear interpolation. Extrapolating that specific example across all transforms we want to support, we'd basically be creating mirror version of what TorchVision has. That seems silly, since it's more for users to understand and more for us to maintain.


Basically, that coupling between TC and TV will have to exist either in the code (as in this PR), or in our heads as API designers.



-----

Side note, slightly related: *if* we're going to have our own TC transforms, I think we'll want their API to exactly match (or be a strict subset of) the TV transforms. E.g. we'd have `torchcodec.transforms.Resize(size=...)` instead of `torchcodec.transforms.Resize(height=..., width=...)` ?

src/torchcodec/decoders/_video_decoder.py
scotts: Oh, that solves the problem nicely: it can definitely be `nn.Module`.

src/torchcodec/decoders/_video_decoder.py
scotts: @NicolasHug, I came to same conclusion as:
> Side note, slightly related: if we're going to have our own TC transforms, I think we'll want their API to exactly match (or be a strict subset of) the TV transforms. E.g. we'd have `torchcodec.transforms.Resize(size=...)` instead of `torchcodec.transforms.Resize(height=..., width=...)` ?

At which point, I don't think we've really gained anything by having them separate. And users will probably also start asking, hey, can you just accept the TorchVision ones? I also just realized a new counter-point, which I'll put up in the summary as counter point 3.

mypy.ini
```diff
@@ -4,3 +4,4 @@ files = src/torchcodec
 show_error_codes = True
 pretty = True
 allow_redefinition = True
+follow_untyped_imports = True
```
scotts, line 7: I was getting linting errors like: https://github.com/meta-pytorch/torchcodec/actions/runs/19157614790/job/54761644331

Which points to docs which recommend the above change: https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -66,6 +67,10 @@ class VideoDecoder:
             probably is. Default: "exact".
             Read more about this parameter in:
             :ref:`sphx_glr_generated_examples_decoding_approximate_mode.py`
+        transforms (sequence of transform objects, optional): Sequence of transforms to be
+            applied to the decoded frames by the decoder itself, in order. Accepts both
+            ``torchcodec.transforms.DecoderTransform`` and ``torchvision.transforms.v2.Transform``
```
NicolasHug: For it to render as links in the docs:

```suggestion
            :class:`~torchcodec.transforms.DecoderTransform` and :class:`~torchvision.transforms.v2.Transform`
```

We should also create a doc page for the transforms!

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -66,6 +67,10 @@ class VideoDecoder:
...
+            objects. All transforms are applied in the ouput pixel format and colorspace.
```
NicolasHug: Do we want to document this behavior? It seems binding, and we discussed that we may want to reserve the right to change the underlying implementation provided the output are still valid?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -104,6 +109,7 @@ def __init__(
         num_ffmpeg_threads: int = 1,
         device: Optional[Union[str, torch_device]] = "cpu",
         seek_mode: Literal["exact", "approximate"] = "exact",
+        transforms: Optional[Sequence[DecoderTransform]] = None,
```
NicolasHug: On point 4:

> The type annotation that users will see only mentions accepting torchcodec.transforms.DecoderTransform. It does not mention the TorchVision transforms or nn.Module. The text of the docstring will say it, and I think that's enough?


One consequence of not accepting `nn.Module` is that  `mypy` and other type-checkers will complain when they see:

```py
VideoDecoder(..., transforms=[v2.Resize(...)])
```

which is a valid input.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+from typing import Sequence
+
+
+@dataclass
+class DecoderTransform(ABC):
+    """Base class for all decoder transforms.
+
+    A DecoderTransform is a transform that is applied by the decoder before
+    returning the decoded frame. The implementation does not live in TorchCodec
+    itself, but in the underyling decoder. Applying DecoderTransforms to frames
```
NicolasHug: > The implementation does not live in TorchCodec itself, but in the underyling decoder

I think this might potentially be confusing to users? As far as they're concerned, this is a torchcodec thing?

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+    should be both faster and more memory efficient than receiving normally
+    decoded frames and applying the same kind of transform.
+
+    Most DecoderTransforms have a complementary transform in TorchVision,
+    specificially in torchvision.transforms.v2. For such transforms, we ensure
```
NicolasHug: Nit: add URL

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+    @abstractmethod
+    def make_params(self) -> str:
+        pass
+
+
+@dataclass
+class Resize(DecoderTransform):
+    """Resize the decoded frame to a given size.
+
+    Complementary TorchVision transform: torchvision.transforms.v2.Resize.
```
NicolasHug:
```suggestion
    Complementary TorchVision transform: :class:`~torchvision.transforms.v2.Resize~.
```

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,58 @@
...
+    All DecoderTranforms are applied in the output pixel format and colorspace.
+    """
+
+    @abstractmethod
+    def make_params(self) -> str:
+        pass
+
+
+@dataclass
+class Resize(DecoderTransform):
```
NicolasHug, line 55: > Should it be DecoderResize instead of just Resize

Agreed it should be `Resize`,  I think the API of the `DecoderNative` transforms should be a strict subset of the `v2` ones. That includes parameters and their values, but also the class names (consider adding this to the description above? :) )

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -432,6 +441,83 @@ def _get_and_validate_stream_metadata(
...
+    converted_transforms = []
+    for transform in transforms:
+        if not isinstance(transform, DecoderTransform):
+            if not tv_available:
+                raise ValueError(
+                    f"The supplied transform, {transform}, is not a TorchCodec "
+                    " DecoderTransform. TorchCodec also accept TorchVision "
+                    "v2 transforms, but TorchVision is not installed."
+                )
+            if isinstance(transform, v2.Resize):
```
NicolasHug: I think this fails if `tv_available` is False? Because `v2` wouldn't exist

EDIT ah no that's probably fine because of the `if not tv_available:` check above.

NicolasHug: Makes me thing we should have a dummy job where we don't install TV that ensures TC still works fine...

test/test_transform_ops.py
```diff
@@ -34,7 +36,89 @@
...
 
-torch._dynamo.config.capture_dynamic_output_shape_ops = True
+
+class TestPublicVideoDecoderTransformOps:
+    @pytest.mark.parametrize(
+        "height_scaling_factor, width_scaling_factor",
+        ((1.5, 1.31), (0.5, 0.71), (0.7, 1.31), (1.5, 0.71), (1.0, 1.0), (2.0, 2.0)),
+    )
+    @pytest.mark.parametrize("video", [NASA_VIDEO, TEST_SRC_2_720P])
+    def test_resize_torchvision(
```
NicolasHug, line 46: Can you confirm that this test was mostly unchanged, you just added `            assert_frames_equal(frame_resize_tv, frame_resize)` ?

test/test_transform_ops.py
```diff
@@ -34,7 +36,89 @@
...
+            if height_scaling_factor < 1 or width_scaling_factor < 1:
+                # Antialias only relevant when down-scaling!
+                with pytest.raises(AssertionError, match="Expected at least"):
+                    assert_tensor_close_on_at_least(
+                        frame_resize, frame_tv_no_antialias, percentage=99, atol=1
+                    )
+                with pytest.raises(AssertionError, match="Tensor-likes are not close"):
+                    torch.testing.assert_close(
+                        frame_resize, frame_tv_no_antialias, rtol=0, atol=6
+                    )
```
Dan-Flores, line 107: I realize this is not a new test, but I am curious - is the intent of this section of the test to document that when antialias is disabled via a torchvision transform, we expect the frames to be different from torchcodec frames?

test/test_transform_ops.py
scotts: @NicolasHug, yup, exactly.

test/test_transform_ops.py
scotts: @Dan-Flores, yes, exactly.

src/torchcodec/decoders/_video_decoder.py
scotts: We want to reserve the right to change the underlying implementation, but we may not be able to easily change when we apply the transform with respect to the colorspace conversion. That fact is, I think, implied by what we consider to be our reference: a fully decoded frame passed to a TorchVision transform. In that scenario, the transform is always applied after the colorspace conversion.

Then I think the questions are:

1. _Do_ we want to document that we consider passing untransformed frames to TorchVision transforms as our reference? I think we do, because I think that's implied by accepting the TorchVision transforms, and it's a easy way to explain the feature to users.
2. Is when the transform is applied useful to users? I thought it was, but if it's of little value, we could potentially just not talk about it.

Given how far away the tolerances were when TorchCodec applied the transform in YUV, but TorchVision applied them in RGB, I think that if we ever changed this behavior, it would have to be an option.

src/torchcodec/decoders/_video_decoder.py
scotts: That's a good point. I'm happy just putting in `nn.Module`.

src/torchcodec/transforms/_decoder_transforms.py
scotts: Fair. I'm trying to figure out how to communicate the difference between these and TorchVision's transforms, but maybe we don't have to.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -66,6 +67,13 @@ class VideoDecoder:
             probably is. Default: "exact".
             Read more about this parameter in:
             :ref:`sphx_glr_generated_examples_decoding_approximate_mode.py`
+        transforms (sequence of transform objects, optional): Sequence of transforms to be
+            applied to the decoded frames by the decoder itself, in order. Accepts both
+            :class:`~torchcodec.transforms.DecoderTransform` and
+            `torchvision.transforms.v2.Transform <https://docs.pytorch.org/vision/stable/transforms.html#v2-api-reference-recommended>`_
+            objects. All transforms are applied
```
scotts: This and all other references to TorchVision transforms use hard links. I don't think we can get proper Sphinx references when it's in a different project.

src/torchcodec/decoders/_video_decoder.py
scotts: On a job which doesn't have TorchVision installed: I agree we need to do something here, but I'd like to punt on this for now. The current testing file imports TorchVision unconditionally. I think we'll want to separate out the tests that require TorchVision from those that don't so that we can test both behaviors, but that will require different `.py` files. I'd like to deal with that in its own PR.

I actually started to add a step in the current linux wheel test that did not install TorchVision when I realized this.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+        size: (sequence of int): Desired output size. Must be a sequence of
+            the form (height, width).
+    """
+
+    size: Sequence[int]
+
+    def _make_params(self) -> str:
+        assert len(self.size) == 2
+        return f"resize, {self.size[0]}, {self.size[1]}"
+
```
scotts, line 71: Note this class method below is new. Because I'm trying to exhaustively catch all of the `v2.Resize` options we don't support, the code for turning a `v2.Resize` into a `torchcodec.transforms.Resize` got more involved. Extrapolated across more transforms, this kind of logic would end up dominating the code in `_video_decoder.py`. By make this a private class method, we can put all logic related to what in `v2.Resize` we support and how to turn a `v2.Resize` into a `torchcodec.transforms.Resize` in one place.

Also, to state it explicit, `_from_torchvision()` and `_make_params()` are private methods so they're not publicly documented. Users shouldn't need to know about them.

docs/source/api_ref_transforms.rst
```diff
@@ -0,0 +1,17 @@
+.. _samplers:
```
NicolasHug:
```suggestion
.. _transforms:
```

docs/source/api_ref_transforms.rst
```diff
@@ -0,0 +1,17 @@
...
+
+===================
+torchcodec.transforms
+===================
```
NicolasHug: We're getting warnings when the length don't match


```suggestion
=====================
torchcodec.transforms
=====================

```

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+@dataclass
+class DecoderTransform(ABC):
+    """Base class for all decoder transforms.
+
+    A *decoder transform* is a transform that is applied by the decoder before
+    returning the decoded frame.  Applying decoder transforms to frames
+    should be both faster and more memory efficient than receiving normally
+    decoded frames and applying the same kind of transform.
+
+    Most `DecoderTransform` objects have a complementary transform in TorchVision,
```
NicolasHug: Annoyingly, single backticks in `rst` means italics. I think you wanted those to be code, like in markdown? For that we need double backticks (there are other instances of single backticks below and maybe in other files?)


```suggestion
    Most ``DecoderTransform`` objects have a complementary transform in TorchVision,
```

src/torchcodec/decoders/_video_decoder.py
NicolasHug: Yes, we can punt on this. I'm hoping we can do something very simple regarding testing: keep all but one test job using torchvision, and just have one small CI job that doesn't install TV and just runs a few tests, basically just insuring TV is an optional dependency. I'd like to avoid separating tests in different files just for that - we may have more than one optional dependency and that quickly becomes untractable.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -432,6 +444,78 @@ def _get_and_validate_stream_metadata(
...
+    converted_transforms: list[DecoderTransform] = []
+    for transform in transforms:
+        if not isinstance(transform, DecoderTransform):
+            if not tv_available:
+                raise ValueError(
+                    f"The supplied transform, {transform}, is not a TorchCodec "
+                    " DecoderTransform. TorchCodec also accept TorchVision "
+                    "v2 transforms, but TorchVision is not installed."
+                )
+            if isinstance(transform, v2.Resize):
```
NicolasHug: Nit, I think I would have been less surprised by `v2` being actually optional if this were `elif`.

```suggestion
            elif isinstance(transform, v2.Resize):
```

src/torchcodec/decoders/_video_decoder.py
NicolasHug: We can, we'll just need to add a torchvision entry here:

https://github.com/meta-pytorch/torchcodec/blob/b35005da9c8ea9e6d0d7b173442b8f2e9267b27d/docs/source/conf.py#L209-L215

Feel free to leave that as follow-up / open an issue.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+
+    size: Sequence[int]
+
+    def _make_params(self) -> str:
+        assert len(self.size) == 2
+        return f"resize, {self.size[0]}, {self.size[1]}"
+
+    @classmethod
+    def _from_torchvision(cls, resize_tv: nn.Module):
+        from torchvision.transforms import v2
```
NicolasHug: I'd suggest the following:

```py
try:
    from torchvision.transforms import v2
except ImportError from e:
    raise RuntimeError("Couldn't find TorchVision - this should never happen, please report a bug") from e
```

This should probably be in a helper function, reused across classes. My goal here is mainly to help the reader (us, the devs) understand that this code-path is only expected to be run when TV is already available. Otherwise, the plain import makes it look like `v2` could be a hard dep.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -0,0 +1,85 @@
...
+    Args:
+        size: (sequence of int): Desired output size. Must be a sequence of
+            the form (height, width).
+    """
+
+    size: Sequence[int]
+
+    def _make_params(self) -> str:
+        assert len(self.size) == 2
+        return f"resize, {self.size[0]}, {self.size[1]}"
```
NicolasHug: Can we call it something else than `_make_params`? 

`make_params` exists for the v2 transforms, but it does something quite different.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -66,6 +67,13 @@ class VideoDecoder:
             probably is. Default: "exact".
             Read more about this parameter in:
             :ref:`sphx_glr_generated_examples_decoding_approximate_mode.py`
+        transforms (sequence of transform objects, optional): Sequence of transforms to be
+            applied to the decoded frames by the decoder itself, in order. Accepts both
+            :class:`~torchcodec.transforms.DecoderTransform` and
+            `torchvision.transforms.v2.Transform <https://docs.pytorch.org/vision/stable/transforms.html#v2-api-reference-recommended>`_
+            objects. All transforms are applied
+            in the ouput pixel format and colorspace. Read more about this parameter in:
```
NicolasHug: Following up on https://github.com/meta-pytorch/torchcodec/pull/1003#discussion_r2511552241

> Do we want to document that we consider passing untransformed frames to TorchVision transforms as our reference? I think we do, because I think that's implied by accepting the TorchVision transforms, and it's a easy way to explain the feature to users.

Agreed, we should document and claim that TV is our ref. I think we have slightly different understandings of what we mean by "TV is our ref", your definition is slightly stricter than mine (see below). 

> Is when the transform is applied useful to users? I thought it was, but if it's of little value, we could potentially just not talk about it.

I don't think it adds a lot of value to document, as I don't know if that's a questions users are even asking themselves. But I could be wrong and I don't feel strongly  about it. What I'm slightly more concerned about the comment is that it seems like a contract, and I suspect we may want to relax that behavior in the future. E.g. for crop, we might want to apply it in YUV space instead of RGB if it's faster and if models can't notice the difference.

To me, when we say "TV is our ref", it means "this transforms has the same behavior as the TV transform *as far as models are concerned*". It's not strictly about bitwise equality (we'll never have that). It's only about whether the models can tell the difference. We know they can tell the difference for resize's interpolation mode. But if they can't tell the difference for (e.g.) crop being applied before or after color-conversion, I think we could  allow ourselves to make that change of behavior. That allows us more freedom to potentially enable higher perf gains in the future. 

None of my comments above are blocking. We can go ahead as-is. I'm happy that for once, I am not the one insisting on strictness :D

NicolasHug (approved):

src/torchcodec/transforms/_decoder_transforms.py
scotts: I saw them render as italics, and I just though, "Oh, Sphinx makes code just italics? Okay..." :)

src/torchcodec/decoders/_video_decoder.py
scotts: @NicolasHug, that's all fair, and I also think it's fair to err on the side of explaining less about the implementation. If folks start asking about it, we can revisit.

>  But if they can't tell the difference for (e.g.) crop being applied before or after color-conversion, I think we could allow ourselves to make that change of behavior. That allows us more freedom to potentially enable higher perf gains in the future.

Based on what I did with crop and resize, I actually think that is likely to be the case everywhere: applying the transform in YUV versus RGB will be noticeable by the model. But we can easily punt on that determination by just not saying anything about it. If it becomes something folks ask about, we may need to make it an explicit option, in which case we'll document behavior.

src/torchcodec/transforms/_decoder_transforms.py
scotts: At the moment, there is only one place where it's a bug if we can't find TorchVision, so I'll keep this as not-a-function for now. The other place we import TorchVision dynamically, we are fine if it's not there.

scotts: Ohhh, right, but we're going to have one of these for each transform. I'll pull into a function now, and it should just live in this file.

PR #1004, opened 2025-10-28
Update docs to indicate support for FFmpeg 8 on all platforms

Dan-Flores:
In addition, the installation commands now find `FFmpeg<9` as suggested in #861.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Start by installing the **nightly** build of PyTorch following the
 Then, the easiest way to install the rest of the dependencies is to run:
 
 ```bash
-conda install cmake pkg-config pybind11 "ffmpeg<9" -c conda-forge
+conda install cmake pkg-config pybind11 "ffmpeg" -c conda-forge
 ```
 
 ### Clone and build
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -30,7 +30,7 @@ Start by installing the **nightly** build of PyTorch following the
 Then, the easiest way to install the rest of the dependencies is to run:
 
 ```bash
-conda install cmake pkg-config pybind11 "ffmpeg<8" -c conda-forge
+conda install cmake pkg-config pybind11 "ffmpeg<9" -c conda-forge
```
NicolasHug: Here and below, let's remove the bound, FFmpeg 9 isn't out and probably won't be anytime soon, and using `ffmpeg<9` may suggest to some users that we don't support the latest FFmpeg - which we do!

The intent behind `ffmpeg<8` was that for a short period of time, FFmpeg 8 was released and we didn't support it, so users would potentially get a conflict. But there's no risk of conflict ATM since FFmpeg 9 doesn't exist.

```suggestion
conda install cmake pkg-config pybind11 "ffmpeg" -c conda-forge
```

PR #1008, opened 2025-10-29
Enable CUDA device for video encoder

Dan-Flores:
This PR enables encoding frames on GPU device in VideoEncoder. 

Previously, `DeviceInterface` and its subclasses only contained methods related to decoding. For now, we are going to break that rule and add two encoding specific methods: `convertTensorToAVFrame`, and `setupHardwareFrameContext`.
* This is necessary to isolate CUDA headers to CUDA specific files.
* I will propose options to decide where these functions belong longterm.

The constructor for `VideoEncoder` is unchanged. 
Inspired by other torch operators, we will determine use the device the `frames` tensor` is on (thanks for the suggestion @NicolasHug!) :

```python
frames = (torch.randint(0, 256, size=(5, 3, 64, 64), dtype=torch.uint8).to("cuda:0")
encoder = VideoEncoder(frames, output.mp4) # will encode on GPU
```
The code path in `Encoder.cpp` is unchanged on CPU. 
On GPU, a member variable `deviceInterface_` will be initialized and used to call GPU encoding functions.
* In `convertTensorToAVFrame`, there are several TODOs to extend it. Right now, it only handles limited range color, and always encodes frames in the pixel format `AV_PIX_FMT_NV12` (essentially `yuv420p`).

### DIFF
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -17,6 +17,7 @@ from .utils import (
     assert_tensor_close_on_at_least,
     get_ffmpeg_major_version,
     get_ffmpeg_minor_version,
+    in_fbcode,
     IS_WINDOWS,
     NASA_AUDIO_MP3,
     needs_ffmpeg_cli,
@@ -831,7 +832,19 @@ class TestVideoEncoder:
 
     @pytest.mark.parametrize("method", ("to_file", "to_tensor", "to_file_like"))
     @pytest.mark.parametrize(
-        "device", ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
+        "device",
+        (
+            "cpu",
+            pytest.param(
+                "cuda",
+                marks=[
+                    pytest.mark.needs_cuda,
+                    pytest.mark.skipif(
+                        in_fbcode(), reason="NVENC not available in fbcode"
+                    ),
+                ],
+            ),
+        ),
     )
     def test_contiguity(self, method, tmp_path, device):
         # Ensure that 2 sets of video frames with the same pixel values are encoded


src/torchcodec/_core/Encoder.cpp
```diff
@@ -724,6 +736,10 @@ VideoEncoder::VideoEncoder(
 
 void VideoEncoder::initializeEncoder(
     const VideoStreamOptions& videoStreamOptions) {
+  if (videoStreamOptions.device.is_cuda()) {
```
NicolasHug: I was hopiing we wouldn't need to support a `device` parameter anywhere. Any reason we can't just rely on the input frames device?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -333,7 +335,7 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
 // we have to do this because of an FFmpeg bug where hardware decoding is not
 // appropriately set, so we just go off and find the matching codec for the CUDA
 // device
-std::optional<const AVCodec*> CudaDeviceInterface::findCodec(
```
NicolasHug, line 336: Was this renaming necessary? Could it be left-out for later?

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  at::cuda::CUDAStream currentStream =
+      at::cuda::getCurrentCUDAStream(device_.index());
+
+  facebook::torchcodec::convertRGBTensorToNV12Frame(
+      tensor, avFrame, device_, nppCtx_, currentStream);
+
+  // Set color properties to FFmpeg defaults
+  avFrame->colorspace = AVCOL_SPC_SMPTE170M; // BT.601
+  avFrame->color_range = AVCOL_RANGE_MPEG; // Limited range
```
NicolasHug: Does the FFmpeg CLI support these options for the nvenc backend? If they do, we might want to support them too - potentially as a follow-up

src/torchcodec/_core/Encoder.cpp
```diff
@@ -873,9 +909,25 @@ void VideoEncoder::encode() {
...
+  inHeight_ = static_cast<int>(tensor.sizes()[1]);
+  inWidth_ = static_cast<int>(tensor.sizes()[2]);
+
+  // For now, reuse input dimensions as output dimensions
+  outWidth_ = inWidth_;
+  outHeight_ = inHeight_;
+
+  // Input format is RGB planar (AV_PIX_FMT_GBRP after channel reordering)
+  inPixelFormat_ = AV_PIX_FMT_GBRP;
+
```
NicolasHug: The above changes look copy/pasted from what's already in `initializeEncoder`. Are they needed? If they are, can you help me understand why?

src/torchcodec/_core/GpuEncoder.h
```diff
@@ -0,0 +1,58 @@
...
+    return device_;
+  }
+
+ private:
+  torch::Device device_;
+  UniqueAVBufferRef hardwareDeviceCtx_;
+  UniqueNppContext nppCtx_;
+
+  void initializeHardwareContext();
+  void setupHardwareFrameContext(AVCodecContext* codecContext);
```
NicolasHug: This isn't used anywhere

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -246,6 +261,68 @@ torch::Tensor convertNV12FrameToRGB(
   return dst;
 }
 
+void convertRGBTensorToNV12Frame(
```
NicolasHug: This doesn't need to be in `CUDACommon`, it's only used in GpuEncoder.cpp so it can be defined there.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,194 @@
...
+    [[maybe_unused]] AVPixelFormat targetFormat,
+    int frameIndex,
+    AVCodecContext* codecContext) {
+  TORCH_CHECK(tensor.is_cuda(), "GpuEncoder requires CUDA tensors");
+  TORCH_CHECK(
+      tensor.dim() == 3 && tensor.size(0) == 3,
+      "Expected 3D RGB tensor (CHW format), got shape: ",
+      tensor.sizes());
+
+  return convertRGBTensorToNV12Frame(tensor, frameIndex, codecContext);
```
NicolasHug: There's little value in calling (and defining) `GpuEncoder::convertRGBTensorToNV12Frame`. We're not re-using it anywhere else. So let's just have all the `GpuEncoder::convertRGBTensorToNV12Frame` logic *and* the `facebook::torchcodec::convertRGBTensorToNV12Frame` here in this `GpuEncoder::convertTensorToAVFrame` method. It significantly simplifies the interfaces.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -16,9 +16,18 @@ class VideoEncoder:
             C is 3 channels (RGB), H is height, and W is width.
             Values must be uint8 in the range ``[0, 255]``.
         frame_rate (float): The frame rate of the **input** ``frames``. Also defines the encoded **output** frame rate.
+        device (str or torch.device, optional): The device to use for encoding. Default: "cpu".
+            If you pass a CUDA device, frames will be encoded on GPU.
+            Note: The "beta" CUDA backend is not supported for encoding.
```
NicolasHug: There's no need to mention that, the `set_cuda_backend` selector is currently only defined in `torchcodec.decoders` so it's only clearly about decoding.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: As per our offline discussion, I've updated this PR to not have an explicit `device` param, and instead use whichever device the `frames` Tensor is on.

src/torchcodec/_core/CudaDeviceInterface.cpp
Dan-Flores: Yes, renaming made more sense when this file had both `findDecoder` and `findEncoder` functions. Since `findEncoder` will be isolated to `GpuEncode.cpp`, I'll remove this change.

src/torchcodec/_core/GpuEncoder.cpp
Dan-Flores: FFmpeg CLI does, I'll add a TODO.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Thanks for catching this, deleted in the latest update.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -748,7 +764,12 @@ void VideoEncoder::initializeEncoder(
     TORCH_CHECK(
         avFormatContext_->oformat != nullptr,
         "Output format is null, unable to find default codec.");
+    // Try to find a hardware-accelerated encoder if not using CPU
```
NicolasHug: This comment looks out of place, the line below isn't about GPU only. It also doesn't add much value so it can be removed.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -536,6 +538,15 @@ torch::Tensor validateFrames(const torch::Tensor& frames) {
...
       frames.sizes()[1]);
+  if (device.type() != torch::kCPU) {
+    TORCH_CHECK(
+        frames.is_cuda(),
+        "When using CUDA encoding (device=",
+        device.str(),
+        "), frames must be on a CUDA device. Got frames on ",
+        frames.device().str(),
+        ". Please move frames to a CUDA device: frames.to('cuda')");
+  }
```
NicolasHug: The error above should be an internal bug: if the frames are on CUDA while the device parameter is not, it means it wasn't set properly in the custom ops. So this should be an "internal bug" error message rather than a user-facing thing, since we assume users aren't using the C++ APIs.

But, I think we should push the logic from https://github.com/meta-pytorch/torchcodec/pull/1008/files#r2565059217 further: we don't need a `device` parameter at all, anywhere. Having a device parameter duplicates the source of truth of the device and leads to potential bugs (like the one above that the `TORCH_CHECK` is preventing). So I think we shouldn't set `options.device` at all for encoding, and always rely on the frames. We can add a comment in `src/torchcodec/_core/StreamOptions.h` to indicate that.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -647,6 +647,7 @@ void encode_video_to_file(
     std::optional<std::string_view> preset = std::nullopt,
     std::optional<std::vector<std::string>> extra_options = std::nullopt) {
   VideoStreamOptions videoStreamOptions;
+  videoStreamOptions.device = frames.device();
```
NicolasHug: I think we don't need to do that, it duplicates the source of truth of the device. See longer comment below.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -748,7 +764,12 @@ void VideoEncoder::initializeEncoder(
     TORCH_CHECK(
         avFormatContext_->oformat != nullptr,
         "Output format is null, unable to find default codec.");
+    // Try to find a hardware-accelerated encoder if not using CPU
     avCodec = avcodec_find_encoder(avFormatContext_->oformat->video_codec);
+    if (gpuEncoder_) {
+      avCodec = gpuEncoder_->findEncoder(avFormatContext_->oformat->video_codec)
+                    .value_or(avCodec);
+    }
```
NicolasHug: Is there actual value in having `findEncoder`? Did you notice any problem if we didn't use it? If not, then let's remove it. 

If yes, then let's rename it into `findCodec` and add a comment simlar to this one

https://github.com/meta-pytorch/torchcodec/blob/392bab3cb619d3bfdb6e9f3a5b705425eb7c1133/src/torchcodec/_core/CudaDeviceInterface.cpp#L332-L336

and also indicate that this `findCodec` function exists for similar reasons as the one I linked to.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -820,6 +841,14 @@ void VideoEncoder::initializeEncoder(
         videoStreamOptions.preset.value().c_str(),
         0);
   }
+
+  // Register the hardware device context with the codec
+  // context before calling avcodec_open2().
```
NicolasHug: Is it *necessary* to do so? If yes the comment should explain *why* - otherwise it can be removed.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+int getFlagsAVHardwareDeviceContextCreate() {
+#if LIBAVUTIL_VERSION_INT >= AV_VERSION_INT(58, 26, 100)
+  return AV_CUDA_USE_CURRENT_CONTEXT;
+#else
+  return 0;
+#endif
+}
+
+// Redefinition from CudaDeviceInterface.cpp anonymous namespace
+// TODO-VideoEncoder: unify device context creation, add caching to encoder
```
NicolasHug: Agreed on the TODOs. Seeing this makes me think we will probably want to follow-up quickly and implement `GpuEncoder.cpp` as part of `CudaDeviceInterface.cpp`. But this is better done as a follow-up considering this encoder itself is already a big change.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+      defaultLimitedRangeRgbToNv12,
+      *nppCtx_);
+
+  TORCH_CHECK(
+      status == NPP_SUCCESS,
+      "Failed to convert RGB to NV12: NPP error code ",
+      status);
+
+  // Validate CUDA operations completed successfully
+  cudaError_t memCheck = cudaGetLastError();
```
NicolasHug: I don't think we've used that before. Is it neede, considering we check the error status of all operations? If not, or if we're not certain we need it, let's remove it.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+      tensor.dim() == 3 && tensor.size(0) == 3,
+      "Expected 3D RGB tensor (CHW format), got shape: ",
+      tensor.sizes());
+  UniqueAVFrame avFrame(av_frame_alloc());
+  TORCH_CHECK(avFrame != nullptr, "Failed to allocate AVFrame");
+
+  avFrame->format = AV_PIX_FMT_CUDA;
+  avFrame->width = static_cast<int>(tensor.size(2));
+  avFrame->height = static_cast<int>(tensor.size(1));
+  avFrame->pts = frameIndex;
```
NicolasHug: the 4 lines above, let's add a TODO to unify this with the logic that's already in the CPU version of `convertTensorToAVFrame`. We should make sure frames use the exact same code path for setting the PTS, width and height, with the same sources of truth.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  int ret = av_hwframe_get_buffer(
+      codecContext ? codecContext->hw_frames_ctx : nullptr, avFrame.get(), 0);
+  TORCH_CHECK(
+      ret >= 0,
+      "Failed to allocate hardware frame: ",
+      getFFMPEGErrorStringFromErrorCode(ret));
+
+  // Validate that avFrame was properly allocated with CUDA memory
```
NicolasHug: Not much added value, remove comment - beware of claude' verbosity :)

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+  TORCH_CHECK(
+      avFrame != nullptr && avFrame->data[0] != nullptr,
+      "avFrame must be pre-allocated with CUDA memory");
+
+  // Convert CHW to HWC for NPP processing
+  int height = static_cast<int>(tensor.size(1));
+  int width = static_cast<int>(tensor.size(2));
+  torch::Tensor hwcFrame = tensor.permute({1, 2, 0}).contiguous();
+
+  // Get current CUDA stream for NPP operations
```
NicolasHug: same

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+  at::cuda::CUDAStream currentStream =
+      at::cuda::getCurrentCUDAStream(device_.index());
+
+  // Setup NPP context with current stream
```
NicolasHug: same

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+  nppCtx_->hStream = currentStream.stream();
+  cudaError_t cudaErr =
+      cudaStreamGetFlags(nppCtx_->hStream, &nppCtx_->nStreamFlags);
+  TORCH_CHECK(
+      cudaErr == cudaSuccess,
+      "cudaStreamGetFlags failed: ",
+      cudaGetErrorString(cudaErr));
+
+  // Always use FFmpeg's default behavior: BT.601 limited range
```
NicolasHug: this comment seems out of place. Since I'm seeing a lot of comments like that, I'll make a general request: please go over all existing comments in this PR and review / edit / fix / delete those that need it.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+      getFFMPEGErrorStringFromErrorCode(ret));
+
+  // Validate that avFrame was properly allocated with CUDA memory
+  TORCH_CHECK(
+      avFrame != nullptr && avFrame->data[0] != nullptr,
+      "avFrame must be pre-allocated with CUDA memory");
+
+  // Convert CHW to HWC for NPP processing
+  int height = static_cast<int>(tensor.size(1));
+  int width = static_cast<int>(tensor.size(2));
```
NicolasHug: 1. make sure to use the same height and width that were used  to set the AVframe properties - use the same variables!
2. define those where they're needed.

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+}
+
+void GpuEncoder::registerHardwareDeviceWithCodec(AVCodecContext* codecContext) {
+  TORCH_CHECK(
+      hardwareDeviceCtx_, "Hardware device context has not been initialized");
+  TORCH_CHECK(codecContext != nullptr, "codecContext is null");
+  codecContext->hw_device_ctx = av_buffer_ref(hardwareDeviceCtx_.get());
+}
+
+void GpuEncoder::setupEncodingContext(AVCodecContext* codecContext) {
```
NicolasHug: This needs a comment to explain what this function does, and why.

My understranding is that it's mostly setting the `AVHWFramesContext` field of the AVFrame. And that's needed because this is how FFmpeg knows to allocate frames on the GPU's memory. I could be wrong, check me on this. This should be documented :)

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+      tensor.sizes());
+  UniqueAVFrame avFrame(av_frame_alloc());
+  TORCH_CHECK(avFrame != nullptr, "Failed to allocate AVFrame");
+
+  avFrame->format = AV_PIX_FMT_CUDA;
+  avFrame->width = static_cast<int>(tensor.size(2));
+  avFrame->height = static_cast<int>(tensor.size(1));
+  avFrame->pts = frameIndex;
+
+  int ret = av_hwframe_get_buffer(
```
NicolasHug: Add a note here that we're letting FFmpeg allocate the CUDA memory. I think we should explore allocating the memory with pytorch instead, so that we can automatically rely on  pytorch's CUDA memory allocator, which should be more efficient. There could be a TODO to investigate how to do that (this is related to my comment about setupEncodingContext above).

src/torchcodec/_core/GpuEncoder.cpp
```diff
@@ -0,0 +1,242 @@
...
+
+  // Validate CUDA operations completed successfully
+  cudaError_t memCheck = cudaGetLastError();
+  TORCH_CHECK(
+      memCheck == cudaSuccess,
+      "CUDA error detected: ",
+      cudaGetErrorString(memCheck));
+
+  // TODO-VideoEncoder: Enable configuration of color properties, similar to
+  // FFmpeg Set color properties to FFmpeg defaults
```
NicolasHug:
```suggestion
  // FFmpeg. Below are the FFmpeg defaults
```

test/test_encoders.py
```diff
@@ -825,7 +832,7 @@ def encode_to_tensor(frames):
                 dest = str(tmp_path / "output.mp4")
                 VideoEncoder(frames, frame_rate=30).to_file(dest=dest, **common_params)
                 with open(dest, "rb") as f:
-                    return torch.frombuffer(f.read(), dtype=torch.uint8)
+                    return torch.frombuffer(f.read(), dtype=torch.uint8).clone()
```
NicolasHug: why was that needed?

test/test_encoders.py
```diff
@@ -1259,3 +1266,109 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+            )  # Set rate control mode for AV1        else:
+        ffmpeg_cmd.extend(["-qp", str(qp)])  # Use lossless qp for other codecs
+        ffmpeg_cmd.extend([ffmpeg_encoded_path])
+
+        # Will this prevent CI from treating test as failed if NVENC is not available?
+        try:
+            subprocess.run(ffmpeg_cmd, check=True, capture_output=True)
+        except subprocess.CalledProcessError as e:
+            if b"No NVENC capable devices found" in e.stderr:
+                pytest.skip("NVENC not available on this system")
```
NicolasHug: Add a TODO to make sure our CI never ever skips those tests. I.e. we should have a mechanism in place that makes sure our CI fails here, instead of skipping these tests. This should be the first follow-up of this PR.

test/test_encoders.py
```diff
@@ -1259,3 +1266,109 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
         assert metadata["profile"].lower() == expected_profile
         assert metadata["color_space"] == colorspace
         assert metadata["color_range"] == color_range
+
+    @pytest.mark.needs_cuda
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
+    @pytest.mark.parametrize("pixel_format", ("nv12", "yuv420p"))
```
NicolasHug: I'm surprised to see this. I thought nvenc only supports NV12 output. Is that not the case?

src/torchcodec/_core/GpuEncoder.cpp
Dan-Flores: Would an example of this be [allocateEmptyHWCTensor](https://github.com/meta-pytorch/torchcodec/blob/392bab3cb619d3bfdb6e9f3a5b705425eb7c1133/src/torchcodec/_core/Frame.cpp#L25)?

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Let's delete it. The intention was to find a hardware enabled encoder if one was not specified, but as far as I can tell, FFmpeg CLI does not support that for encoding, only for decoding via the `-hwaccel` flag.

src/torchcodec/_core/GpuEncoder.cpp
Dan-Flores: Thanks for the feedback, I added a comment and renamed it to `setupHardwareFrameContext` to be more clear.

src/torchcodec/_core/GpuEncoder.cpp
Dan-Flores: I'm not certain and it has not caught any errors, lets delete it

test/test_encoders.py
Dan-Flores: This was an attempt to fix a warning on this test, but it does not prevent the warning, I'll clean it up.
<details>

```

test/test_encoders.py::TestVideoEncoder::test_contiguity[cuda-to_file]
  /home/dev/torchcodec/test/test_encoders.py:835: UserWarning: The given buffer is not writable, and PyTorch does not support non-writable tensors. 
This means you can write to the underlying (supposedly non-writable) buffer using the tensor. 
You may want to copy the buffer to protect its data or make it writable before converting it to a tensor. 
This type of warning will be suppressed for the rest of this program. (Triggered internally at /pytorch/torch/csrc/utils/tensor_new.cpp:1581.)
    return torch.frombuffer(f.read(), dtype=torch.uint8)
```
</details>

test/test_encoders.py
Dan-Flores: Thanks for pointing this out, I took another look at my code:
* `h264_nvenc` supports multiple output pixel formats: `yuv420p nv12 p010le yuv444p p016le nv16 ... `

* The bottleneck is that `GpuEncoder::convertTensorToAVFrame` always uses `nppiRGBToNV12_8u_ColorTwist32f_C3P2R_Ctx`, which only handles `nv12`. 

* Since `nv12` and `yuv420p` do the same chroma subsampling, the results appeared to be correct. 

I'll add a TODO to enable utilizing the user's selected pixel formats. There are [other nvidia](https://docs.nvidia.com/cuda/npp/image_color_conversion.html#rgbtoyuv) functions we can use based on the target pixel format, or I could investigate using filtergraph's `scale_cuda` to handle conversion, as is done in [maybeConvertAVFrameToNV12OrRGB24](https://github.com/meta-pytorch/torchcodec/blob/2311422cf808ff08f65e33f60a304136fa9e9bc2/src/torchcodec/_core/CudaDeviceInterface.cpp#L147).

src/torchcodec/_core/GpuEncoder.cpp
NicolasHug: Correct,  in `allocateEmptyHWCTensor` we pre-allocate the output tensors with pytorch, and thus with pytorch's CUDA memory allocator. But with the decoder, it's more of a necessity than an optimization: we want to output tensors, so of course we have to use pytorch to allocate those. We wouldn't want or be able to do that via FFmpeg.

In our case here, with the encoder, relying on pytorch for the allocation isn't a necessity: FFmpeg can do that. It's more of an optimization.

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+
+    @pytest.mark.needs_cuda
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
+    @pytest.mark.parametrize(
+        "format_codec",
+        [
+            ("mov", "h264_nvenc"),
+            ("mp4", "hevc_nvenc"),
+            ("avi", "h264_nvenc"),
+            # ("mkv", "av1_nvenc"), # av1_nvenc is not supported on CI
```
NicolasHug, line 1340: That's OK for now but add a TODO that we should re-enable it when not in CI. We should be able to have an `in_CI()` helper just like we do for `in_fbcode()`. Torchvision has: https://github.com/pytorch/vision/blob/6b56de1cc83386025f2bad87abf608077d1853f7/test/common_utils.py#L28

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+                file_like=file_like,
+                format=format,
+                codec=codec,
+                pixel_format=pixel_format,
+                extra_options=encoder_extra_options,
+            )
+            encoder_output = file_like.getvalue()
+        else:
+            raise ValueError(f"Unknown method: {method}")
+
```
NicolasHug, line 1412: After we have encoded the frame, let's assert with ffprobe that the pixel format is NV12 (and any other relevant properties)

src/torchcodec/_core/StreamOptions.h
```diff
@@ -41,6 +41,8 @@ struct VideoStreamOptions {
       ColorConversionLibrary::FILTERGRAPH;
 
   // By default we use CPU for decoding for both C++ and python users.
+  // Note: For video encoding, device is determined by the location of the input
+  // frame tensor.
```
NicolasHug:
```suggestion
  // Note: This is not used for video encoding, because device is determined by the device of the input
  // frame tensor.
```

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
         assert metadata["profile"].lower() == expected_profile
         assert metadata["color_space"] == colorspace
         assert metadata["color_range"] == color_range
+
+    @pytest.mark.needs_cuda
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
+    @pytest.mark.parametrize(
+        "format_codec",
+        [
+            ("mov", "h264_nvenc"),
```
NicolasHug, line 1336: IIUC, we get an error when not explicitly passing the `codec`. That's OK for this PR, but not a great UX. On CPU we don't require the user to pass the `codec`, so ideally it should work the same on GPU. Let's add a TODO to enable that, I think that should be one of the first follow-ups.

I acknowledge it might be related to the `*find_codec*` function that I suggested we remove earlier, in a previous comment. I also understand that the FFmpeg CLi requires the codec, but there should be no reason we have to :)

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+            # ("mkv", "av1_nvenc"), # av1_nvenc is not supported on CI
+        ],
+    )
+    @pytest.mark.parametrize("method", ("to_file", "to_tensor", "to_file_like"))
+    # TODO-VideoEncoder: Enable additional pixel formats ("yuv420p", "yuv444p")
+    def test_nvenc_against_ffmpeg_cli(self, tmp_path, format_codec, method):
+        # Encode with FFmpeg CLI using nvenc codecs
+        format, codec = format_codec
+        device = "cuda"
+        pixel_format = "nv12"
```
NicolasHug: We only support nv12 for now. We should error if the user tries to pass anything else.

Also note that the tests are passing when I remove `pixel_format` from both the FFmpeg CLI and our API calls - that's good, and it suggests we should remove it from this test as well. As long as we assert with ffprobe that the output is nv12, we're good (see other suggestion below)

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
         assert metadata["profile"].lower() == expected_profile
         assert metadata["color_space"] == colorspace
         assert metadata["color_range"] == color_range
+
+    @pytest.mark.needs_cuda
+    @pytest.mark.skipif(in_fbcode(), reason="ffmpeg CLI not available")
```
NicolasHug: Let's mark that with @mollyxu 's  `@needs_ffmpeg_cli` instead.

test/test_encoders.py
```diff
@@ -1291,3 +1303,105 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+        ffmpeg_cmd.extend([ffmpeg_encoded_path])
+
+        # TODO-VideoEncoder: Ensure CI does not skip this test, as we know NVENC is available.
+        try:
+            subprocess.run(ffmpeg_cmd, check=True, capture_output=True)
+        except subprocess.CalledProcessError as e:
+            if b"No NVENC capable devices found" in e.stderr:
+                pytest.skip("NVENC not available on this system")
+            else:
+                raise
```
NicolasHug: On the CI TODO above: I think the easiest way to address it is to remove the whole `try/except` block and to never skip this test. In the GitHub CI, it should always pass. In fbcode, it's consistently skipped anyway. I now think we should do that here in this PR rather than leaving it out as a TODO (sorry for suggesting a TODO earlier :) )

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -138,6 +138,18 @@ class DeviceInterface {
     return "";
   }
 
+  // Function used for video encoding, only implemented in CudaDeviceInterface.
+  virtual std::optional<UniqueAVFrame> convertTensorToAVFrame(
+      [[maybe_unused]] const torch::Tensor& tensor,
+      [[maybe_unused]] int frameIndex,
+      [[maybe_unused]] AVCodecContext* codecContext) {
+    return std::nullopt;
+  }
```
NicolasHug: This is all good, a few suggestions:

- let's explicitly reflect in the name that it's about encoding and about CUDA. Something like `convertCUDATensorToCUDAAVFrameForEncoding`. The name is ugly but that's kind of on purpose. We should add a TODO here to re-consider the use of encoding via the device interface
- Let's add a comment above indicating that these are here because we need the interface "plumbing" for not including CUDA-specific dependencies in CPU builds (related to TODO mentioned above)
- Let's not return an optional. I don't thikn we need to?
- Let's call TORCH_CHECK(false) here in this implementation. No one should ever call that except the CUDA interface.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -138,6 +138,18 @@ class DeviceInterface {
...
+
+  // Function used for video encoding, only implemented in CudaDeviceInterface.
+  virtual void setupHardwareFrameContext(
+      [[maybe_unused]] AVCodecContext* codecContext) {}
```
NicolasHug: Similar comments: 

- rename to something like `setupHardwareFrameContextForEncoding`
- use TORCH_CHECK(false).

test/test_encoders.py
Dan-Flores: I agree on the worse UX there, I'll add a TODO.

src/torchcodec/_core/DeviceInterface.h
Dan-Flores: Thanks for these suggestions, I've implemented them in my latest commit.

test/test_encoders.py
Dan-Flores: This is a little tricky. FFmpeg CLI defaults to `pix_fmt=gbrp` which does no chroma subsampling, but somehow the  the frame assertions pass against our `VideoEncoder` which defaults to `pix_fmt=yuv420p`. 
For now, I'll keep specifying `pix_fmt`, and investigate and address this once we enable other pixel formats.

test/test_encoders.py
Dan-Flores: I added the assert below and learned that ffmpeg does not distinguish `nv12` from `yuv420p` in video metadata. I hope to gain a better understanding of this once we enable alternate pixel formats.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -362,4 +362,117 @@ std::string CudaDeviceInterface::getDetails() {
...
+  TORCH_CHECK(
+      ret >= 0,
+      "Failed to allocate hardware frame: ",
+      getFFMPEGErrorStringFromErrorCode(ret));
+
+  TORCH_CHECK(
+      avFrame != nullptr && avFrame->data[0] != nullptr,
+      "avFrame must be pre-allocated with CUDA memory");
+
+  torch::Tensor hwcFrame = tensor.permute({1, 2, 0}).contiguous();
```
NicolasHug, line 421: This creates a copy, add a TODO that we should look into avoiding that if possible.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -362,4 +362,117 @@ std::string CudaDeviceInterface::getDetails() {
...
+} // namespace
+
+UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
+    const torch::Tensor& tensor,
+    int frameIndex,
+    AVCodecContext* codecContext) {
+  TORCH_CHECK(
+      tensor.dim() == 3 && tensor.size(0) == 3,
+      "Expected 3D RGB tensor (CHW format), got shape: ",
+      tensor.sizes());
```
NicolasHug, line 389: Also check the tensor device is on CUDA.

test/test_encoders.py
NicolasHug: I think ffmpeg doesn't differentiate much between yuv420p and nv12 in the contexts we're using them in. They're the same value as you know, just different layout in memory. Maybe nv12 is actually aliased to yuv420p for our usages? I'm not sure, it doesn't matter much right now I agree.

So, from that:

> FFmpeg CLI defaults to pix_fmt=gbrp which does no chroma subsampling, but

I understand that we'll have to expliciltly pass `-pix_fmt nv12` to the command line.

However, we don't need to explicitly pass `pixel_format="nv12"` to the `VideoEncoder`, so we should remove it. We only have to pass it to ffmpeg.

More importantly, I also noticed that passing `pixel_format="yuv444p"` or any other valid value *still* encodes into `yuv420p`, and that qualifies as a bug :). It's OK to only support nv12/yuv420p for now with CUDA encoding, but what's not OK is to pretend we support other pixel formats by not erroring on other valid values. We should:

- enforce that the user leaves `pixel_format` to its default (i.e. `None`) when using the CUDA encoder. We have already made `None` default to nv12 in the code (you did that in `convertCUDATensorToAVFrameForEncoding`). If you want to be extra nice you can also accept the `"nv12"` string, but I don't think it's worth it right now. I think we should just enforce None because that's the current default.
- Add a test to enforce the error message.
- Leave `pixel_format` to its default (None) i.e. not specify it in this test (we still have to pass it to ffmpeg).


-----

> somehow the the frame assertions pass against our VideoEncoder which defaults to pix_fmt=yuv420p

I think that's because we are asserting that the *decoded output* are roughly the same. They're not encoded to the same format, but once decoded, they look similar. It sounds OK to me.

Issue #926, opened 2025-10-03
Generalize `DeviceInterface` to include audio decoding


scotts:
Currently, our `DeviceInterface` is implicitly a video device interface. Inside `SingleStreamDecoder`, we dispatch differently based on audio and video: https://github.com/meta-pytorch/torchcodec/blob/6d72f115c4fbbcb8ed62cac43f5f1cfe2bd5f692/src/torchcodec/_core/SingleStreamDecoder.cpp#L1322-L1327
Ideally, we'd like to turn that line into just:
```c++
deviceInterface_->convertAVFrameToFrameOutput(
        avFrame, frameOutput, preAllocatedOutputTensor);
```
That is, we handle audio and video the same. In order to do that, we need to somehow get [`SingleStreamDecoder::convertAudioAVFrameToFrameOutputOnCPU`](https://github.com/meta-pytorch/torchcodec/blob/6d72f115c4fbbcb8ed62cac43f5f1cfe2bd5f692/src/torchcodec/_core/SingleStreamDecoder.cpp#L1331) into a device interface. Design questions:

1. Should we extend `CpuDeviceInterface` to handle audio, in which case we'd dispatch inside of it?
2. Or should we keep `CpuDeviceInterface` to be just video only, and create a new kind of device interface that is just audio?

I think 1 might be the better option, but I'm not sure. I'm confident we'll only ever do audio decoding on the CPU.

mollyxu:
Closing with PR #1010

PR #1010, opened 2025-10-29
Generalize DeviceInterface to include audio decoding

mollyxu:
Generalize DeviceInterface to include audio decoding so that there will be a single dispatch for video and audio in SingleStreamDecoder.cpp
#926

### DIFF
--- a/src/torchcodec/_core/CpuDeviceInterface.cpp
+++ b/src/torchcodec/_core/CpuDeviceInterface.cpp
@@ -4,7 +4,7 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include "src/torchcodec/_core/CpuDeviceInterface.h"
+#include "CpuDeviceInterface.h"
 
 namespace facebook::torchcodec {
 namespace {
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -4,7 +4,7 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include "src/torchcodec/_core/SingleStreamDecoder.h"
+#include "SingleStreamDecoder.h"
 #include <cstdint>
 #include <cstdio>
 #include <iostream>
--- a/src/torchcodec/_core/AVIOContextHolder.cpp
+++ b/src/torchcodec/_core/AVIOContextHolder.cpp
@@ -4,7 +4,7 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include "src/torchcodec/_core/AVIOContextHolder.h"
+#include "AVIOContextHolder.h"
 #include <torch/types.h>
 
 namespace facebook::torchcodec {
--- a/src/torchcodec/_core/AVIOContextHolder.h
+++ b/src/torchcodec/_core/AVIOContextHolder.h
@@ -6,7 +6,7 @@
 
 #pragma once
 
-#include "src/torchcodec/_core/FFMPEGCommon.h"
+#include "FFMPEGCommon.h"
 
 namespace facebook::torchcodec {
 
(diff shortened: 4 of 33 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1025,7 +1028,7 @@ AudioFramesOutput SingleStreamDecoder::getFramesPlayedInRangeAudio(
         (stopPts <= lastDecodedAvFrameEnd);
   }
 
-  auto lastSamples = maybeFlushSwrBuffers();
+  auto lastSamples = deviceInterface_->maybeFlushAudioBuffers();
```
NicolasHug, line 1031: great job on renaming this to `maybeFlushAudioBuffers`. The name of the interface method should be agnostic to the underlying implementation now, so it's good that `swr` isn't in the name anymore 👍

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -123,12 +144,10 @@ ColorConversionLibrary CpuDeviceInterface::getColorConversionLibrary(
 // TODO: Figure out whether that's possible!
 // Dimension order of the preAllocatedOutputTensor must be HWC, regardless of
 // `dimension_order` parameter. It's up to callers to re-shape it if needed.
-void CpuDeviceInterface::convertAVFrameToFrameOutput(
+void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
     UniqueAVFrame& avFrame,
     FrameOutput& frameOutput,
     std::optional<torch::Tensor> preAllocatedOutputTensor) {
-  TORCH_CHECK(initialized_, "CpuDeviceInterface was not initialized.");
-
```
NicolasHug, line 131: Q - was it intentional to remove this check?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -114,6 +120,21 @@ ColorConversionLibrary CpuDeviceInterface::getColorConversionLibrary(
   }
 }
 
+void CpuDeviceInterface::convertAVFrameToFrameOutput(
+    UniqueAVFrame& avFrame,
+    FrameOutput& frameOutput,
+    AVMediaType mediaType,
```
NicolasHug: We're adding the new `AVMediaType`  to all `convertAVFrameToFrameOutput` methods now. It works, but IIUC it's only really necessary for the Cpu interface, not for the other ones.

Alternatively, we could remove this parameter, and instead store `AVMediaType avMediaType_` field on the interface, which gets initialized when calling either `initializeVideo()` or `initializeAudio()` ?

I think it will be semantically the same thing, and may help keep the API surface low?

Curious about your thoughts @mollyxu and @Dan-Flores @scotts ?

NicolasHug (commented):
Well done @mollyxu . Left some minor comments but this looks great

src/torchcodec/_core/CpuDeviceInterface.cpp
scotts: @NicolasHug, good call, that's aligned with the recent refactors I've done: if you can learn something during initialization and store it, do that. The code here is @mollyxu following my advice from yesterday, and I didn't think about your suggestion. :)

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1293,136 +1296,11 @@ FrameOutput SingleStreamDecoder::convertAVFrameToFrameOutput(
...
-  // When sample rate conversion is involved, swresample buffers some of the
-  // samples in-between calls to swr_convert (see the libswresample docs).
-  // That's because the last few samples in a given frame require future
-  // samples from the next frame to be properly converted. This function
-  // flushes out the samples that are stored in swresample's buffers.
-  auto& streamInfo = streamInfos_[activeStreamIndex_];
-  if (!swrContext_) {
-    return std::nullopt;
-  }
-  auto numRemainingSamples = // this is an upper bound
```
Dan-Flores, line 1402: nit: I think this comment still applies, lets keep it

src/torchcodec/_core/CpuDeviceInterface.cpp
mollyxu: I renamed the original `convertAVFrameToFrameOutput` to `convertVideoAVFrameToFrameOutput`. Now `convertVideoAVFrameToFrameOutput` is called by `convertAVFrameToFrameOutput` in line 133. This check now exists at line 128

NicolasHug (approved):
Thanks @mollyxu

PR #1011, opened 2025-10-30
Remove old method declarations

NicolasHug:
Those have been moved to the CPUDeviceInterface a while ago, we don't need them on the SingleStreamDecoder anymore.


Dan-Flores (approved):
Thanks for cleaning this up!

scotts:
I am surprised C++ allows classes to have declarations without implementations as long as no one calls them, and I feel like I am surprised by this on a regular basis. :(

Issue #1009, opened 2025-10-29
Refactor metadata fallback logic


mollyxu:
Our fallback logic for missing metadata is currently scattered across both C++ and Python. 

https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/_core/SingleStreamDecoder.cpp#L1579-L1591

While the general logic looks very similar, in Python, the metadata class has no knowledge of seek modes. This leads to us relying on our knowledge of which values are present as seen in `metadata.begin_stream_seconds` and other metadata. It can create scenarios where value `a` may try to fallback to value `b` in some instances, but value `b` may also try to fallback to value `a` in other instances.

https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/decoders/_video_decoder.py#L406-L411

https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/_core/_metadata.py#L116-L126

From our discussions, we want to consolidate the logic all in C++ then expose it to the Python layer. Some options we were considering are pybind11 vs custom_ops.

Would love to hear more thoughts on this!

NicolasHug:
I haven't thought too much about the C++ fallback logic itself, but regarding the C++ -> Python mapping (pybind vs custom_ops), I was assuming we'd keep it mostly as it currently is? I.e. using an ugly custom-built json string in custom_ops?

> https://github.com/meta-pytorch/torchcodec/blob/28b03462ca77f2a06e3de03ae862bc1ac5053efb/src/torchcodec/_core/custom_ops.cpp#L750

scotts:
@NicolasHug, I forgot we're returning everything in a JSON string! That simplifies things. With that in mind, my current thinking of what we want to do is:

1. We already represent the metadata in C++, but it's currently a struct with no member functions: [`StreamMetadata`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/Metadata.h#L21).
2. It can stay a struct, but we should add member functions that mirror the property methods we have in Python. That is, we should have a `getNumFrames()` method that returns `numFramesFromContent` or `numFramesFromHeader` as appropriate, just like in Python.
3. In C++, we should use these member functions to access the values at any point that we want to do any fallback.
[...]

PR #1021, opened 2025-11-04
refactor metadata fallback logic

mollyxu:
Consolidate Metadata Fallback Logic in C++
==========================================================
Related issue: #1009 

Our metadata fallback logic was previously scattered across both C++ and Python layers, making it confusing and potentially led to circular dependencies.

**In C++**, we had seek-mode-aware logic:

https://github.com/meta-pytorch/torchcodec/blob/c552b60b195db733caa2b83f3c9d3abc4bec71e6/src/torchcodec/_core/SingleStreamDecoder.cpp#L1456-L1467

**In Python**, we had fallback logic without seek mode awareness:

https://github.com/meta-pytorch/torchcodec/blob/c552b60b195db733caa2b83f3c9d3abc4bec71e6/src/torchcodec/_core/_metadata.py#L116-L126

This separation meant:

*   Python code couldn't make seek-mode-aware decisions
*   Fallback logic was duplicated and could diverge

Changes Made
------------

### 1. Added Computed Methods to C++ `StreamMetadata`

Created seek-mode-aware methods in [`**Metadata.cpp**`](command:code-compose.open?%5B%22%2Fhome%2Fmollyxu%2Ftorchcodec%2Fsrc%2Ftorchcodec%2F_core%2FMetadata.cpp%22%2Cnull%5D "/home/mollyxu/torchcodec/src/torchcodec/_core/Metadata.cpp"):

*   `getDurationSeconds(SeekMode)`
*   `getBeginStreamSeconds(SeekMode)`
*   `getEndStreamSeconds(SeekMode)`
*   `getNumFrames(SeekMode)`
*   `getAverageFps(SeekMode)`

These methods encapsulate all fallback logic with proper seek mode handling.

### 2. Moved `SeekMode` Enum

Moved `SeekMode` from `SingleStreamDecoder` class to `Metadata.h`  to make it accessible to `StreamMetadata` methods.

### 3. Updated C++ Custom Ops Layer

Modified `custom_ops.cpp` to call the new computed methods and pass computed values to Python via JSON serialization

### 4. Simplified Python Metadata Classes

Updated `_metadata.py`

*   Removed all `@property` methods with fallback logic
*   Added simple fields to receive computed values from C++

### 5. Updated Tests

**Removed Python tests** `test_metadata.py`

*   `test_num_frames_fallback` - fallback logic now in C++
*   `test_calculate_num_frames_using_fps_and_duration` - calculated in C++
*   `test_duration_seconds_fallback` - fallback logic now in C++
*   `test_calculate_duration_seconds_using_fps_and_num_frames` - calculated in C++

**Added C++ tests** in `MetadataTest.cpp`

*   `NumFramesFallbackPriority`
*   `CalculateNumFramesUsingFpsAndDuration`
*   `DurationSecondsFallback`
*   `CalculateDurationSecondsUsingFpsAndNumFrames`

### DIFF
--- a/src/torchcodec/_core/Metadata.cpp
+++ b/src/torchcodec/_core/Metadata.cpp
@@ -98,18 +98,19 @@ std::optional<int64_t> StreamMetadata::getNumFrames(SeekMode seekMode) const {
 std::optional<double> StreamMetadata::getAverageFps(SeekMode seekMode) const {
   switch (seekMode) {
     case SeekMode::custom_frame_mappings:
-    case SeekMode::exact:
-      if (getNumFrames(seekMode).has_value() &&
+    case SeekMode::exact: {
+      auto numFrames = getNumFrames(seekMode);
+      if (numFrames.has_value() &&
           beginStreamPtsSecondsFromContent.has_value() &&
-          endStreamPtsSecondsFromContent.has_value() &&
-          (beginStreamPtsSecondsFromContent.value() !=
-           endStreamPtsSecondsFromContent.value())) {
-        return static_cast<double>(
-            getNumFrames(seekMode).value() /
-            (endStreamPtsSecondsFromContent.value() -
-             beginStreamPtsSecondsFromContent.value()));
+          endStreamPtsSecondsFromContent.has_value()) {
+        double duration = endStreamPtsSecondsFromContent.value() -
+            beginStreamPtsSecondsFromContent.value();
+        if (duration != 0.0) {
+          return static_cast<double>(numFrames.value()) / duration;
+        }
       }
       return averageFpsFromHeader;
+    }
     case SeekMode::approximate:
       return averageFpsFromHeader;
     default:
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -801,6 +801,69 @@ std::string get_stream_json_metadata(
 
   std::map<std::string, std::string> map;
 
+  if (streamMetadata.durationSecondsFromHeader.has_value()) {
+    map["durationSecondsFromHeader"] =
+        std::to_string(*streamMetadata.durationSecondsFromHeader);
+  }
+  if (streamMetadata.bitRate.has_value()) {
+    map["bitRate"] = std::to_string(*streamMetadata.bitRate);
+  }
+  if (streamMetadata.numFramesFromContent.has_value()) {
+    map["numFramesFromContent"] =
+        std::to_string(*streamMetadata.numFramesFromContent);
+  }
+  if (streamMetadata.numFramesFromHeader.has_value()) {
+    map["numFramesFromHeader"] =
+        std::to_string(*streamMetadata.numFramesFromHeader);
+  }
+  if (streamMetadata.beginStreamSecondsFromHeader.has_value()) {
+    map["beginStreamSecondsFromHeader"] =
+        std::to_string(*streamMetadata.beginStreamSecondsFromHeader);
+  }
+  if (streamMetadata.beginStreamPtsSecondsFromContent.has_value()) {
+    map["beginStreamSecondsFromContent"] =
+        std::to_string(*streamMetadata.beginStreamPtsSecondsFromContent);
+  }
+  if (streamMetadata.endStreamPtsSecondsFromContent.has_value()) {
+    map["endStreamSecondsFromContent"] =
+        std::to_string(*streamMetadata.endStreamPtsSecondsFromContent);
+  }
+  if (streamMetadata.codecName.has_value()) {
+    map["codec"] = quoteValue(streamMetadata.codecName.value());
+  }
+  if (streamMetadata.width.has_value()) {
+    map["width"] = std::to_string(*streamMetadata.width);
+  }
+  if (streamMetadata.height.has_value()) {
+    map["height"] = std::to_string(*streamMetadata.height);
+  }
+  if (streamMetadata.sampleAspectRatio.has_value()) {
+    map["sampleAspectRatioNum"] =
+        std::to_string((*streamMetadata.sampleAspectRatio).num);
+    map["sampleAspectRatioDen"] =
+        std::to_string((*streamMetadata.sampleAspectRatio).den);
+  }
+  if (streamMetadata.averageFpsFromHeader.has_value()) {
+    map["averageFpsFromHeader"] =
+        std::to_string(*streamMetadata.averageFpsFromHeader);
+  }
+  if (streamMetadata.sampleRate.has_value()) {
+    map["sampleRate"] = std::to_string(*streamMetadata.sampleRate);
+  }
+  if (streamMetadata.numChannels.has_value()) {
+    map["numChannels"] = std::to_string(*streamMetadata.numChannels);
+  }
+  if (streamMetadata.sampleFormat.has_value()) {
+    map["sampleFormat"] = quoteValue(streamMetadata.sampleFormat.value());
+  }
+  if (streamMetadata.mediaType == AVMEDIA_TYPE_VIDEO) {
+    map["mediaType"] = quoteValue("video");
+  } else if (streamMetadata.mediaType == AVMEDIA_TYPE_AUDIO) {
+    map["mediaType"] = quoteValue("audio");
+  } else {
+    map["mediaType"] = quoteValue("other");
+  }
+
   // Check whether content-based metadata is available for this stream.
   // In exact mode: content-based metadata exists for all streams.
   // In approximate mode: content-based metadata does not exist for any stream.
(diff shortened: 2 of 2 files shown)


src/torchcodec/_core/_metadata.py
```diff
@@ -38,14 +38,22 @@ class StreamMetadata:
     stream_index: int
     """Index of the stream that this metadata refers to (int)."""
 
+    # Computed fields (computed in C++ with fallback logic)
+    duration_seconds: Optional[float]
+    """Duration of the stream in seconds. Computed in C++ with fallback logic:
+    tries to calculate from content if scan was performed, otherwise falls back
+    to header values."""
```
scotts: Since the doc strings are actually documentation for users, let's not mention that the fallback occurs in C++. We'll want to keep the old explanation of "it's value x_from_header in this situation, and x_from_content in this other situation." But agreed that we can now mention the scan instead of `None`. However, when we mention the scan in the doc string, let's make sure we use `:term:scan` so that we get a link to the definition of scan in our [glossary](https://meta-pytorch.org/torchcodec/stable/glossary.html). (Get an example from existing doc strings, not my comment. The real think you want has backticks that screw up code rendering.)

The comment is fine, as that's for us. We'll want a reminder that the fallback logic is implemented in C++.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1441,43 +1445,17 @@ int64_t SingleStreamDecoder::getPts(int64_t frameIndex) {
...
-    case SeekMode::custom_frame_mappings:
-    case SeekMode::exact:
-      return streamMetadata.numFramesFromContent.value();
-    case SeekMode::approximate: {
-      return streamMetadata.numFramesFromHeader;
-    }
-    default:
-      TORCH_CHECK(false, "Unknown SeekMode");
-  }
+  return streamMetadata.getNumFrames(seekMode_);
```
scotts: I think we can just remove `SingleStreamDecoder::getNumFrames()`, `SingleStreamDecoder::getMinSeconds()` and `std::optional<double> SingleStreamDecoder::getMaxSeconds()`. I think they're only called from within `SingleStreamDecoder`, and those calls can just be replaced with the calls to the metadata object.

test/test_decoders.py
```diff
@@ -899,6 +899,11 @@ def test_get_frames_with_missing_num_frames_metadata(
             "mediaType": "video",
             "numFramesFromHeader": None,
             "numFramesFromContent": None,
+            "beginStreamSeconds": 0.0,
+            "durationSeconds": 13.013,
+            "endStreamSeconds": 13.013,
+            "numFrames": int(29.97003 * 13.013),  # Calculated from fps * duration
+            "averageFps": 29.97003,
```
Dan-Flores: I think we can delete this test, since this fallback calculation has a new test in `MetadataTest.cpp` at https://github.com/meta-pytorch/torchcodec/pull/1021/files#diff-4520a23eabe0a9806c4734ddd72f2eb9d1e49edc63d6e8ed8d7cce86d3c86f26R50-R63.

src/torchcodec/_core/_metadata.py
```diff
@@ -38,14 +38,21 @@ class StreamMetadata:
...
+
     def __repr__(self):
         s = self.__class__.__name__ + ":\n"
         for field in dataclasses.fields(self):
             s += f"{SPACES}{field.name}: {getattr(self, field.name)}\n"
         return s
 
 
-@dataclass
+@dataclass(repr=False)
```
Dan-Flores: For my understanding, why is `repr=False` needed here?

src/torchcodec/_core/_metadata.py
```diff
@@ -38,14 +38,21 @@ class StreamMetadata:
     stream_index: int
     """Index of the stream that this metadata refers to (int)."""
 
+    # Computed fields (computed in C++ with fallback logic)
+    duration_seconds: Optional[float]
+    """Duration of the stream in seconds. Tries to calculate from content
```
scotts: In the original doc strings, we specifically referenced the variable names - I think we should preserve that.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,110 @@
...
+  switch (seekMode) {
+    case SeekMode::custom_frame_mappings:
+    case SeekMode::exact:
+      // In exact mode, use the scanned content value
+      if (endStreamPtsSecondsFromContent.has_value() &&
+          beginStreamPtsSecondsFromContent.has_value()) {
+        return endStreamPtsSecondsFromContent.value() -
+            beginStreamPtsSecondsFromContent.value();
+      }
+      return std::nullopt;
```
scotts: One of the benefits of using the seek mode is that we can know that a scan should have been performed, and the metadata from the scan should definitely be there. I think we should instead make this:
```c++
  case SeekMode:exact:
    return endStreamPtsSecondsFromContent.value() -
            beginStreamPtsSecondsFromContent.value();
```
That is, we don't check if the value is present, and we don't return `nullopt`. The unconditional call to `.value()` will throw an exception if we somehow get into a situation where we're in exact seek mode and there is no value. But, that is what we want. Such a situation should never happen, and we want an exception thrown.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,110 @@
...
+        return durationSecondsFromHeader.value();
+      }
+      if (numFramesFromHeader.has_value() && averageFpsFromHeader.has_value() &&
+          averageFpsFromHeader.value() != 0.0) {
+        return static_cast<double>(numFramesFromHeader.value()) /
+            averageFpsFromHeader.value();
+      }
+      return std::nullopt;
+  }
+  return std::nullopt;
```
scotts: Every switch statement that is intended to be exhaustive should end with:
```c++
    default:
      TORCH_CHECK(false, "Unknown SeekMode");
  }
```
And then we don't need to do a `return std::nullopt`. The compiler is smart enough to figure out that all paths through the function either return a value or result in an exception.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+double StreamMetadata::getBeginStreamSeconds(SeekMode seekMode) const {
+  switch (seekMode) {
+    case SeekMode::custom_frame_mappings:
+    case SeekMode::exact:
+      if (beginStreamPtsSecondsFromContent.has_value()) {
+        return beginStreamPtsSecondsFromContent.value();
+      }
```
scotts, line 49: Similar to above, we should just say `return beginStreamPtsSecondsFromContent.value();` without checking if it has the value. If there's no value, we want the exception thrown.

This applies to all of the functions below as well: in exact mode, let's just access the value without checking.

src/torchcodec/_core/_metadata.py
mollyxu: I originally wrote `repr=False` so that `VideoStreamMetadata` wouldn't generate a default `__repr__` method that would override the custom `__repr__ ` inhereited from `StreamMetadata`. I realized that I should follow the pattern we use for `AudioStreamMetadata` and do something like the following instead
 
```
    def __repr__(self):
        return super().__repr__()
```

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,110 @@
...
+      return 0.0;
+    case SeekMode::approximate:
+      return 0.0;
```
scotts: I think this should be:
```c++
  case SeekMode::approximate:
    if (beginStreamPtsSecondsFromHeader.has_value()) {
      return beginStreamPtsSecondsFromHeader.value();
    }
    return 0.0.;
```
That is, in approximate mode, we should return the beginning we read from the header if it's there. If it's not there, just return 0.0.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,110 @@
...
+}
+
+std::optional<double> StreamMetadata::getEndStreamSeconds(
+    SeekMode seekMode) const {
+  switch (seekMode) {
+    case SeekMode::custom_frame_mappings:
+    case SeekMode::exact:
+      if (endStreamPtsSecondsFromContent.has_value()) {
+        return endStreamPtsSecondsFromContent.value();
+      }
```
scotts, line 68: We should just say `return endStreamPtsSecondsFromContent.value();`. We always set it when we scan, so we don't need to fall back to the duration.

src/torchcodec/_core/Metadata.cpp
mollyxu: Would you recommend modifying the logic such that `exact` mode is accessed directly but we keep the checks for `custom_frame_mappings`? Previously the Python logic didn't have knowledge of seek mode so `custom_frame_mappings` was oftentimes grouped with `exact` because scans are performed.

In `custom_frame_mappings` a scan only happens to specific streams upon `add_video_stream()` is called (ie. `test_get_metadata` in `test_metadata.py`). Removing the checks would lead to cases in which not all streams in the container would have scanned metadata. 

Or should `custom_frame_mappings` that's missing the scanned information behave the same as `approximate` mode?

scotts:
@mollyxu, this is great! I think we can further refine the implementations of the fallback logic, but this is definitely the way. Thank you for working on this! :)

mollyxu:
Thanks for the review! 

To summarize, from my understanding, we currently store the metadata of all streams in a video (including streams that we are not decoding on). 

- The refactoring did not affect `exact` mode because we did a scan on the entire container and have the content metadata for all streams. 
- This was also not a problem in `approximate` mode because we are using fallback logic with checks. 
- However, with the way we are currently handling `custom_frame_mappings`, it would only have stream metadata for the stream that we are decoding on. This would lead to `bad optional access` on the other streams. 

In my latest commit, I tried to expose the `activeStreamIndex` in `SingleStreamDecoder` so that in `custom_ops.py` we can handle the stream that we are decoding on differently from the other streams for seekmode = `custom_frame_mappings`. 

Another potential approach that we could take is not modifying `custom_ops.py` but handling the case of `custom_frame_mapings` in a similar way as our `approximate` mode, falling back to default values after not finding values. This approach might overlook some underlying issues with how `custom_frame_mapings` should behave. 

I would love to hear everyone's thoughts on these approaches and the general state of our metadata retrieval!

src/torchcodec/_core/Metadata.cpp
Dan-Flores: We only allow one stream to be added and scanned via `add_video_stream`, and I believe (but do correct me if I am mistaken!) when we access any stream metadata, its from the stream currently stored as `activeStreamIndex_`. This stream index is updated by the `custom_frame_mappings` equivalent "scan" function.

> Removing the checks would lead to cases in which not all streams in the container would have scanned metadata.

Is it possible to access the stream metadata of a stream other than `activeStreamIndex_`? If not, it should be safe to assume that the metadata will be present, and treat it as exact mode.

mollyxu: > when we access any stream metadata, its from the stream currently stored as activeStreamIndex_

I think in `get_container_metadata.py` in `_metadata.py` we call `get_stream_json_metadata` for each stream in the container. It is at this step that we are triggering the logic in `Metadata.cpp` for `custom_frame_mappings` that doesn't exist for streams that are at the `activeStreamIndex_`

https://github.com/meta-pytorch/torchcodec/blob/afd5aba80e680378b69be9eb35b3d981c3887eff/src/torchcodec/_core/_metadata.py#L259-L260

The error seems to come not from accessing a non-active stream but rather from how we store metadata.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,123 @@
...
+  switch (seekMode) {
+    case SeekMode::custom_frame_mappings:
+    case SeekMode::exact:
+      TORCH_CHECK(
+          numFramesFromContent.has_value(), "Missing numFramesFromContent");
+      return numFramesFromContent.value();
+    case SeekMode::approximate: {
+      if (numFramesFromContent.has_value()) {
+        return numFramesFromContent.value();
+      }
```
scotts: In approximate mode, `numFramesFromContent` should never have a value, so we should be able to remove the check above.

scotts:
@mollyxu, I actually think this approach might be the best way to do it.

Looking more, we expose [`get_container_metadata()`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/_metadata.py#L250), and I think that's useful. It's a convenient way for people to inspect all metadata we can retrieve, both at the container level and all streams. And it's in there that we're hitting problems, because when `seek_index==custom_frame_mappings`, we don't have certain metadata for our non-active streams. During decoding, this doesn't matter because we are only concerned with the active stream. It's only when we're getting metadata from all streams we hit this.

The alternatives that I can think of:

1. Change `get_container_metadata()` so that it only returns stream metadata from the active index, if there is one. That's actually a breaking change, so I'd rather not do it for that reason alone. But it also removes some useful functionality.
2. Implement *different* member functions in `StreamMetadata` for decoding versus metadata retrieval. That goes against some of the point of this refactor, which is to do this logic in one place.
3. Make the `StreamMetadata` member functions aware of the active stream, and behave differently for each case. This is basically the same thing as option 2, except we're putting all of the logic in one function. In option 2, the caller would be responsible for knowing which function to call when.

Tagging in @NicolasHug in case he has an opinion.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -796,72 +796,161 @@ std::string get_stream_json_metadata(
...
-  if (streamMetadata.sampleFormat.has_value()) {
-    map["sampleFormat"] = quoteValue(streamMetadata.sampleFormat.value());
-  }
-  if (streamMetadata.mediaType == AVMEDIA_TYPE_VIDEO) {
-    map["mediaType"] = quoteValue("video");
-  } else if (streamMetadata.mediaType == AVMEDIA_TYPE_AUDIO) {
-    map["mediaType"] = quoteValue("audio");
+  if ((seekMode == SeekMode::custom_frame_mappings &&
+       stream_index == activeStreamIndex) ||
+      (seekMode != SeekMode::custom_frame_mappings)) {
```
scotts: This condition is unusual enough that we should have a comment explaining why we need to do this. Roughly:

  * In `exact` mode, we have content-based metadata from all streams. Our fallback logic assumes its there, and it is.
  * In `approximate` mode, we have no content-based metadata from any stream. Our fallback logic assumes it is not there.
  * In `custom_frame_mappings`, we only have content-based metadata from the active stream. That's the only stream we received any information on. Our fallback logic assumes it is there. This works during decoding, but when getting metadata from the non-active stream, things break.

src/torchcodec/_core/_metadata.py
```diff
@@ -87,103 +102,27 @@ class VideoStreamMetadata(StreamMetadata):
...
 
     def __repr__(self):
-        s = super().__repr__()
-        s += f"{SPACES}duration_seconds: {self.duration_seconds}\n"
-        s += f"{SPACES}begin_stream_seconds: {self.begin_stream_seconds}\n"
-        s += f"{SPACES}end_stream_seconds: {self.end_stream_seconds}\n"
-        s += f"{SPACES}num_frames: {self.num_frames}\n"
-        s += f"{SPACES}average_fps: {self.average_fps}\n"
-        return s
+        return super().__repr__()
```
scotts, line 125: Since this looks like the default now, do we even need to define this method anymore?

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -796,72 +796,169 @@ std::string get_stream_json_metadata(
...
+  // In exact mode: content-based metadata exists for all streams.
+  // In approximate mode: content-based metadata does not exist for any stream.
+  // In custom_frame_mappings: content-based metadata exists only for the active
+  // stream.
+  // Our fallback logic assumes content-based metadata is available.
+  // It is available for decoding on the active stream, but would break
+  // when getting metadata from non-active streams.
+  if ((seekMode != SeekMode::custom_frame_mappings) ||
+      (seekMode == SeekMode::custom_frame_mappings &&
+       stream_index == activeStreamIndex)) {
```
scotts, line 877: We should be able to factor out all of values that are common to both cases, such as `durationSecondsFromHeader` and `mediaType`. That way, it's more clear what the difference between the two cases is.

src/torchcodec/_core/Metadata.cpp
```diff
@@ -0,0 +1,120 @@
...
+    default:
+      TORCH_CHECK(false, "Unknown SeekMode");
+  }
+}
+
+std::optional<double> StreamMetadata::getAverageFps(SeekMode seekMode) const {
+  switch (seekMode) {
+    case SeekMode::custom_frame_mappings:
+    case SeekMode::exact:
+      if (getNumFrames(seekMode).has_value() &&
```
scotts: Nit: let's define:
```c++
auto numFrames = getNumFrames(seekMode);
```
And use it inside the expressions.

src/torchcodec/_core/_metadata.py
mollyxu: I think without defining this method, the child class will generate a `__repr__()` that will override the custom `__repr__()` we had in the parent class. 

The child class generated __repr__'s formatting will be 
```
VideoStreamMetadata(duration_seconds_from_header=13.013, begin_stream_seconds_from_header=0.0, bit_rate=128783.0, ...)
```

Rather than 
```
VideoStreamMetadata:
  duration_seconds_from_header: 13.013
  begin_stream_seconds_from_header: 0.0
  bit_rate: 128783.0
  ...
```

Alternatively, to prevent this, we could also use
```
@dataclass(repr=False)
```

scotts (approved):
Let's address my two new comments before merging, but approving to unblock. This is great, I'm happy we've finally cleaned this up! :)

mollyxu:
Thanks for the reviews!

src/torchcodec/_core/_metadata.py
scotts: Hmmm. Reading up on it, I think that if we do `repr=False`, then we won't get one at all! I think the current definition is fine.

PR #1025, opened 2025-11-06
Suport device=None

### LINKED ISSUE #993: Support `device=None`
We should support `VideoDecoder(..., device=None)`. 

`device=None` should mean "use the current device", and the "current device" is usually the CPU but can also be set by (e.g.) the `with torch.device(...)` context manager.

device=None should probably become the default eventually, even if it's a  very, very small BC-breaking change.

mollyxu:
Support `device=None` in VideoDecoder
-----------------------------------------------

Fixes #993

### Summary

This PR implements support for `device=None` in `VideoDecoder`, allowing it to use PyTorch's "current device" as determined by context managers and global settings. `device=None` is now the default parameter value.

### Backward Compatibility

*   Existing code with explicit `device="cpu"` or `device="cuda"` works exactly as before
*   **Only change**: The default device will no longer be `cpu` but rather PyTorch's device context if set.

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -144,7 +144,6 @@ class VideoDecoder:
         if num_ffmpeg_threads is None:
             raise ValueError(f"{num_ffmpeg_threads = } should be an int.")
 
-        # Handle device=None by using the current default device
         if device is None:
             device = str(torch.get_default_device())
         elif isinstance(device, torch_device):
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -396,8 +396,8 @@ class TestVideoDecoder:
         ],
     )
     def test_device_none_default_device(self, device_str):
-        # device=None should respect both torch.device() context manager
-        # and torch.set_default_device()
+        # VideoDecoder defaults to device=None, which should respect both
+        # torch.device() context manager and torch.set_default_device().
 
         # Test with context manager
         with torch.device(device_str):


test/test_decoders.py
```diff
@@ -388,6 +388,31 @@ def test_device_instance(self):
...
+            assert decoder[0].device.type == device_str
+
+        # Test with set_default_device
+        original_device = torch.get_default_device()
+        try:
+            torch.set_default_device(device_str)
+            decoder = VideoDecoder(NASA_VIDEO.path)
+            assert decoder[0].device.type == device_str
+        finally:
+            torch.set_default_device(original_device)
```
Dan-Flores, line 414: Why do we use `try` / `finally` here? Does pytest not automatically clean up the torch default device after testing?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -143,7 +144,10 @@ def __init__(
         if num_ffmpeg_threads is None:
             raise ValueError(f"{num_ffmpeg_threads = } should be an int.")
 
-        if isinstance(device, torch_device):
+        # Handle device=None by using the current default device
```
scotts: Nit: the comment doesn't tell us anything that's not already in the code. I think we can remove the comment and it will actually be more readable. My comments tend to fall in one of three uses:

1. What the code is doing is not obvious - a confusing condition, or an API call that is not named well, or some gnarly math.
2. The code is surprising - the code itself might be obvious, but why it's here, or it has some non-obvious implications elsewhere.
3. Section headers for a long sequence of operations.

test/test_decoders.py
```diff
@@ -388,6 +388,31 @@ def test_device_instance(self):
...
+    @pytest.mark.parametrize(
+        "device_str",
+        [
+            "cpu",
+            pytest.param("cuda", marks=pytest.mark.needs_cuda),
+        ],
+    )
+    def test_device_none_default_device(self, device_str):
+        # device=None should respect both torch.device() context manager
+        # and torch.set_default_device()
```
scotts: What we're actually testing here is the default device behavior of `VideoDecoder`. Yes, it currently defaults to `device=None`, but that's not actually prescribed by the test. I think the test as-is is fine, but in the comment we should connect the default behavior of `VideoDecoder` to `device=None`.

test/test_decoders.py
mollyxu: I think since `set_default_device()` modifies the global state, pytest won't automatically clean it up after testing

scotts (approved):
Looks great! Approved to unblock, let's update the comments before merging.

PR #1027, opened 2025-11-06
Add pixel_format to VideoEncoder API 

Dan-Flores:
This PR updates VideoEncoder API to accept an optional `pixel_format`. 
* Remove `avcodec_find_best_pix_fmt_of_list`, instead validate and use provided pixel format.
	* If no `pixel_format` provided, use default pixel format. Often this is `yuv420p`, but not always (ex. `gif` codec).
* Error message appears similar to FFmpeg's internal logged error to help users: 
```
       RuntimeError: Unknown pixel format: invalid_pix_fmt
       Supported pixel formats for libx264: yuv420p yuvj420p yuv422p yuvj422p yuv444p yuvj444p nv12 nv16 nv21 yuv420p10le yuv422p10le yuv444p10le nv20le gray gray10le
```

### DIFF
--- a/src/torchcodec/_core/ops.py
+++ b/src/torchcodec/_core/ops.py
@@ -321,8 +321,8 @@ def encode_video_to_file_abstract(
     frames: torch.Tensor,
     frame_rate: int,
     filename: str,
-    crf: Optional[int],
-    pixel_format: Optional[str],
+    crf: Optional[int] = None,
+    pixel_format: Optional[str] = None,
 ) -> None:
     return
 
@@ -332,8 +332,8 @@ def encode_video_to_tensor_abstract(
     frames: torch.Tensor,
     frame_rate: int,
     format: str,
-    crf: Optional[int],
-    pixel_format: Optional[str],
+    crf: Optional[int] = None,
+    pixel_format: Optional[str] = None,
 ) -> torch.Tensor:
     return torch.empty([], dtype=torch.long)
 


src/torchcodec/_core/Encoder.cpp
```diff
@@ -635,15 +639,20 @@ void VideoEncoder::initializeEncoder(
...
+        outPixelFormat_ != AV_PIX_FMT_NONE,
+        "Unknown pixel format: ",
+        videoStreamOptions.pixelFormat.value());
+  } else {
+    const AVPixelFormat* formats = getSupportedPixelFormats(*avCodec);
+    // If pixel formats are undefined for some reason, try yuv420p
+    outPixelFormat_ = (formats && formats[0] != AV_PIX_FMT_NONE)
+        ? formats[0]
+        : AV_PIX_FMT_YUV420P;
+  }
```
Dan-Flores, line 684: `getSupportedPixelFormats` is not guaranteed to return any formats. If the user does not specify a format and we find none, I think we should try to use the broadly supported `yuv420p`, rather than error out.

test/test_ops.py
```diff
@@ -1274,16 +1267,18 @@ def test_against_to_file(self, tmp_path, format, method):
             "avi",
             "mkv",
             "flv",
-            "gif",
```
Dan-Flores, line 1277: `gif` only supports rgb pixel formats, this test is now focused on the more common `yuv` formats.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -35,29 +35,38 @@ def __init__(self, frames: Tensor, *, frame_rate: int):
     def to_file(
         self,
         dest: Union[str, Path],
+        *,
+        pixel_format: Optional[str] = None,
```
NicolasHug, line 39: Good job on making this a keyword only params 👍

src/torchcodec/_core/Encoder.cpp
NicolasHug: Agreed, this makes sense and that's similar to what we do for the audio encoder when we can't validate:

https://github.com/meta-pytorch/torchcodec/blob/8e615e39850d6e76868dbc0993b71aac653ce92c/src/torchcodec/_core/Encoder.cpp#L85-L87

src/torchcodec/_core/Encoder.cpp
```diff
@@ -635,15 +669,17 @@ void VideoEncoder::initializeEncoder(
...
-      0, // No alpha channel
-      nullptr // Discard conversion loss information
-  );
-  TORCH_CHECK(outPixelFormat_ != -1, "Failed to find best pix fmt")
+  if (videoStreamOptions.pixelFormat.has_value()) {
+    outPixelFormat_ =
+        validatePixelFormat(*avCodec, videoStreamOptions.pixelFormat.value());
+  } else {
+    const AVPixelFormat* formats = getSupportedPixelFormats(*avCodec);
+    // Use first listed pixel format as default.
```
NicolasHug: Before we were using `avcodec_find_best_pix_fmt_of_list`, now our heuristic is to return the first format in the list. Do I understand correctly that the reason for this change is that you have empirically observe that the first in the list is often `yuv420p`, which is a good default?

I think it makes sense, just confirming my understanding. IT might be worth making that very explicit through a comment explaining that `yuv420p` is often the first entry

NicolasHug (approved):
Nicely done @Dan-Flores , thank you!

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Yes, `yuv420p` is often the first in the list, and FFmpeg's [avcodec_default_get_format](https://www.ffmpeg.org/doxygen/4.0/decode_8c_source.html#l01087) uses the same heuristic. 

I'll add a comment here to mention this as well

test/test_encoders.py
```diff
@@ -629,6 +629,30 @@ def test_bad_input(self, tmp_path):
...
+            RuntimeError,
+            match=r"Unknown pixel format: invalid_pix_fmt[\s\S]*Supported pixel formats.*yuv420p",
+        ):
+            getattr(encoder, method)(**valid_params, pixel_format="invalid_pix_fmt")
+
+        with pytest.raises(
+            RuntimeError,
+            match=r"Specified pixel format rgb24 is not supported[\s\S]*Supported pixel formats.*yuv420p",
+        ):
+            getattr(encoder, method)(**valid_params, pixel_format="rgb24")
```
scotts, line 654: I guess it's personal taste, but I think this test might be simpler and more clear if we did the more basic thing:
```python
with pytest.raises(
    RuntimeError,
    match=r"Unknown pixel format: invalid_pix_fmt[\s\S]*Supported pixel formats.*yuv420p",
):
    encoder.to_file(str(tmp_path / "output.mp4", pixel_format="invalid_pix_fmt")

with pytest.raises(
    RuntimeError,
    match=r"Unknown pixel format: invalid_pix_fmt[\s\S]*Supported pixel formats.*yuv420p",
):
    encoder.to_tensor(format="mp4", pixel_format="invalid_pix_fmt")

with pytest.raises(
    RuntimeError,
    match=r"Unknown pixel format: invalid_pix_fmt[\s\S]*Supported pixel formats.*yuv420p",
):
    encoder.to_file_like(file_like=io.BytesIO(), format="mp4", pixel_format="invalid_pix_fmt")

...

```

src/torchcodec/_core/ops.py
```diff
@@ -319,6 +322,7 @@ def encode_video_to_file_abstract(
     frame_rate: int,
     filename: str,
     crf: Optional[int],
+    pixel_format: Optional[str],
```
scotts: Missing `= None`?

src/torchcodec/_core/ops.py
```diff
@@ -329,6 +333,7 @@ def encode_video_to_tensor_abstract(
     frame_rate: int,
     format: str,
     crf: Optional[int],
+    pixel_format: Optional[str],
```
scotts: Missing `= None`?

test/test_encoders.py
Dan-Flores: I agree that this is a bit confusing, but I reused it as the AudioEncoder tests also use this pattern to test across encoding methods. 

https://github.com/meta-pytorch/torchcodec/blob/dc86a8ce83ec6aab954ad4850cc33c3e3898aaf0/test/test_encoders.py#L193-L194

In terms of taste, I would prefer to use this pattern and reduce code duplication. Alternatively, I don't think this test *needs* to be parametrized across encoding methods, since the error will always be hit in `VideoEncoder::initializeEncoder`.

src/torchcodec/_core/ops.py
Dan-Flores: Perhaps - I'm not sure if they are needed for the `@register_fake` annotated functions. I'll add them in case.

test/test_encoders.py
NicolasHug: I don't have a strong pref on the style. The compact parametrized form is a pattern we use across all existing encode tests (both audio and video), but I agree it is harder to understand at first sight.

We should however test all 3 public methods: we want to make sure all these errors are properly raised by all methods, and the fact that they all rely on the same `VideoEncoder::initializeEncoder` is largely an implementation detail. If we were to change the underlying implementation (unlikely, I agree), we'd risk changing this behavior and we wouldn't notice.

PR #1031, opened 2025-11-07
Add crf to VideoEncoder API

Dan-Flores:
Since `crf` was already utlized in the C++ layer, this PR adds `crf` to the python API, and moves the tests from `test_ops.py` to `test_encoders.py`.

### Validation
The function `validateNumericOption` lets us validate an argument if its `AVOption` has `min` and `max` fields. This error checking is applied to `crf` to improve our error message. 
FFmpeg's output error message:
```
RuntimeError: avcodec_open2 failed: Result too large
```
To our own message:
```
RuntimeError: crf=-10 is out of valid range [-1, 3.40282e+38] for this codec. For more details, run 'ffmpeg -h encoder=libx264'

RuntimeError: crf=-10 is out of valid range [0, 63] for this codec. For more details, run 'ffmpeg -h encoder=libsvtav1'

RuntimeError: crf=-10 is out of valid range [-1, 63] for this codec. For more details, run 'ffmpeg -h encoder=libvpx-vp9'
```

### Testing
The tests are updated to use the python API encoding pattern:
```python3
# Previous ops pattern:
encode_video_to_file(
            frames=source_frames,
            frame_rate=frame_rate,
            filename=encoder_output_path,
            pixel_format=pixel_format,
            crf=crf,
        )
```
```python3
# Updated python pattern:
encoder = VideoEncoder(frames=source_frames, frame_rate=frame_rate)
encoder.to_file(dest=encoder_output_path, pixel_format=pixel_format, crf=crf)
```

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -570,17 +570,16 @@ AVPixelFormat validatePixelFormat(
   TORCH_CHECK(false, errorMsg.str());
 }
 
-void validateNumericOption(
+void validateDoubleOption(
     const AVCodec& avCodec,
     const char* optionName,
-    int value) {
-  // First determine if codec's private class is defined
+    double value) {
   if (!avCodec.priv_class) {
     return;
   }
   const AVOption* option = av_opt_find2(
-      // The obj arg must be converted from const AVClass* const* to non-const
-      // void* First cast to remove const, then cast to void*
+      // Convert obj arg from const AVClass* const* to non-const void*
+      // First cast to remove const, then cast to void*
       const_cast<void*>(static_cast<const void*>(&avCodec.priv_class)),
       optionName,
       nullptr,
@@ -739,7 +738,7 @@ void VideoEncoder::initializeEncoder(
   // Apply videoStreamOptions
   AVDictionary* options = nullptr;
   if (videoStreamOptions.crf.has_value()) {
-    validateNumericOption(*avCodec, "crf", videoStreamOptions.crf.value());
+    validateDoubleOption(*avCodec, "crf", videoStreamOptions.crf.value());
     av_dict_set(
         &options,
         "crf",
--- a/src/torchcodec/encoders/_video_encoder.py
+++ b/src/torchcodec/encoders/_video_encoder.py
@@ -37,7 +37,7 @@ class VideoEncoder:
         dest: Union[str, Path],
         *,
         pixel_format: Optional[str] = None,
-        crf: Optional[int] = None,
+        crf: Optional[Union[int, float]] = None,
     ) -> None:
         """Encode frames into a file.
 
@@ -47,7 +47,7 @@ class VideoEncoder:
                 container format.
             pixel_format (str, optional): The pixel format for encoding (e.g.,
                 "yuv420p", "yuv444p"). If not specified, uses codec's default format.
-            crf (int, optional): Constant Rate Factor for encoding quality. Lower values
+            crf (int or float, optional): Constant Rate Factor for encoding quality. Lower values
                 mean better quality. Valid range depends on the encoder (commonly 0-51).
                 Defaults to None (which will use encoder's default).
         """
@@ -64,7 +64,7 @@ class VideoEncoder:
         format: str,
         *,
         pixel_format: Optional[str] = None,
-        crf: Optional[int] = None,
+        crf: Optional[Union[int, float]] = None,
     ) -> Tensor:
         """Encode frames into raw bytes, as a 1D uint8 Tensor.
 
@@ -73,7 +73,7 @@ class VideoEncoder:
                 "mkv", "avi", "webm", "flv", etc.
             pixel_format (str, optional): The pixel format to encode frames into (e.g.,
                 "yuv420p", "yuv444p"). If not specified, uses codec's default format.
-            crf (int, optional): Constant Rate Factor for encoding quality. Lower values
+            crf (int or float, optional): Constant Rate Factor for encoding quality. Lower values
                 mean better quality. Valid range depends on the encoder (commonly 0-51).
                 Defaults to None (which will use encoder's default).
 
@@ -94,7 +94,7 @@ class VideoEncoder:
         format: str,
         *,
         pixel_format: Optional[str] = None,
-        crf: Optional[int] = None,
+        crf: Optional[Union[int, float]] = None,
     ) -> None:
         """Encode frames into a file-like object.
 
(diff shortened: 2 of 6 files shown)


src/torchcodec/encoders/_video_encoder.py
```diff
@@ -46,27 +47,35 @@ def to_file(
...
         """Encode frames into raw bytes, as a 1D uint8 Tensor.
 
         Args:
             format (str): The container format of the encoded frames, e.g. "mp4", "mov",
-            "mkv", "avi", "webm", "flv", or "gif"
+                "mkv", "avi", "webm", "flv", etc.
             pixel_format (str, optional): The pixel format to encode frames into (e.g.,
                 "yuv420p", "yuv444p"). If not specified, uses codec's default format.
+            crf (int, optional): Constant Rate Factor for encoding quality. Lower values
+                mean better quality. Valid range depends on the encoder (commonly 0-51).
```
scotts, line 77: Is it ever valid to be less than 0?

scotts:
I think this is great! We should add some tests for invalid crf values, both less than 0 (which I think is always invalid?), values we know are outside of the range for a given codec and the wrong type. It's fine if these result in exceptions on the C++ side, but we want to make sure users get a clean Python exception and not a segfault.

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: I believe -1 is valid and is equivalent to leaving `crf` unset. Otherwise, no negative values are valid.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -46,27 +47,35 @@ def to_file(
...
         *,
         pixel_format: Optional[str] = None,
+        crf: Optional[int] = None,
     ) -> Tensor:
         """Encode frames into raw bytes, as a 1D uint8 Tensor.
 
         Args:
             format (str): The container format of the encoded frames, e.g. "mp4", "mov",
-            "mkv", "avi", "webm", "flv", or "gif"
+                "mkv", "avi", "webm", "flv", etc.
```
NicolasHug, line 73: Q - why remove "gif"? Do we not support it anymore?

test/test_encoders.py
```diff
@@ -700,3 +705,234 @@ def encode_to_tensor(frames):
...
+            round_trip_frames = self.decode(file_like.getvalue()).data
+        else:
+            raise ValueError(f"Unknown method: {method}")
+
+        assert source_frames.shape == round_trip_frames.shape
+        assert source_frames.dtype == round_trip_frames.dtype
+
+        for s_frame, rt_frame in zip(source_frames, round_trip_frames):
+            assert psnr(s_frame, rt_frame) > 30
+            torch.testing.assert_close(s_frame, rt_frame, atol=2, rtol=0)
```
NicolasHug: This seems to be [failing for webm](https://github.com/meta-pytorch/torchcodec/actions/runs/19244212593/job/55015186996?pr=1031), you might need to use the previous logic

```
        # If FFmpeg selects a codec or pixel format that does lossy encoding, assert 99% of pixels
        # are within a higher tolerance.
        if ffmpeg_version == 6:
            assert_close = partial(assert_tensor_close_on_at_least, percentage=99)
            atol = 15
        else:
            assert_close = torch.testing.assert_close
            atol = 3 if format == "webm" else 2
```

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: We do not test explicitly for it anymore, but it still works. I mostly wanted to amend the docstring to make it seem less like a finalized, exhaustive list of supported formats.

test/test_encoders.py
Dan-Flores: Thanks for the reminder - it seems I applied the webm tolerance to the wrong test. We can simply use `atol = 3 if format == "webm" else 2` on the round_trip_test, though I'm not sure why `webm` needs this special handling.

test/test_encoders.py
```diff
@@ -605,6 +610,13 @@ def test_bad_input_parameterized(self, tmp_path, method):
             )
             getattr(encoder, method)(**valid_params)
 
+        with pytest.raises(RuntimeError, match=r"crf=-10 is out of valid range"):
+            encoder = VideoEncoder(
+                frames=torch.zeros((5, 3, 64, 64), dtype=torch.uint8),
+                frame_rate=30,
+            )
+            getattr(encoder, method)(**valid_params, crf=-10)
+
```
Dan-Flores, line 619: This is the only new test case - all other tests are copied over from `test_ops.py`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -568,6 +569,44 @@ AVPixelFormat validatePixelFormat(
...
+      AV_OPT_SEARCH_FAKE_OBJ,
+      nullptr);
+  // If the option was not found, let FFmpeg handle it later
+  if (!option) {
+    return;
+  }
+  if (option->type == AV_OPT_TYPE_INT || option->type == AV_OPT_TYPE_INT64 ||
+      option->type == AV_OPT_TYPE_FLOAT || option->type == AV_OPT_TYPE_DOUBLE) {
+    TORCH_CHECK(
+        value >= option->min && value <= option->max,
```
NicolasHug, line 596: This is comparing an `int` (value) to a `double` (min and max), we should cast.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -568,6 +569,44 @@ AVPixelFormat validatePixelFormat(
   }
   TORCH_CHECK(false, errorMsg.str());
 }
+
+void validateNumericOption(
+    const AVCodec& avCodec,
+    const char* optionName,
+    int value) {
+  // First determine if codec's private class is defined
+  if (!avCodec.priv_class) {
```
NicolasHug, line 577: are we OK to use `priv_class`? I.e. is it meant to be "private"?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -568,6 +569,44 @@ AVPixelFormat validatePixelFormat(
   }
   TORCH_CHECK(false, errorMsg.str());
 }
+
+void validateNumericOption(
+    const AVCodec& avCodec,
+    const char* optionName,
+    int value) {
+  // First determine if codec's private class is defined
```
NicolasHug: Nit: this comment doesn't add a lot of value, it can be removed

src/torchcodec/_core/Encoder.cpp
```diff
@@ -568,6 +569,44 @@ AVPixelFormat validatePixelFormat(
   }
   TORCH_CHECK(false, errorMsg.str());
 }
+
+void validateNumericOption(
```
NicolasHug: For now this expects an `int`, not a float or double. Let's reflect that in the name. We may define this with a template later to be truely about generic numeric options.


```suggestion
void validateIntOption(
```

src/torchcodec/_core/Encoder.cpp
Dan-Flores: In the `avcodec.h`, `priv_class` is defined in the section for public fields: https://www.ffmpeg.org/doxygen/2.0/libavcodec_2avcodec_8h_source.html

src/torchcodec/_core/Encoder.cpp
Dan-Flores: This comment led me to realize that codecs can implement 'crf' as a `double` or an `int`. 
I'll update the PR to accept either type, and treat it as a double in the C++, so this casting will not be necessary.

PR #1039, opened 2025-11-11
Simplify seek skipping logic

NicolasHug:
This is a minor simplification that I hope will make https://github.com/meta-pytorch/torchcodec/pull/1028 easier to understand.

I'm trying to explain in plain english what the simplification is doing, but anything I come up with is a lot more complicated than just looking at the code. I hope it makes sense.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -1100,15 +1100,16 @@ I    P     P    P    I    P    P    P    I    P    P    I    P    P    I    P
 */
 bool SingleStreamDecoder::canWeAvoidSeeking() const {
   const StreamInfo& streamInfo = streamInfos_.at(activeStreamIndex_);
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
scotts, line 1107: I believe this can now be just `return true`. Should the comment above it also be updated?

NicolasHug: I think the comment about only needing to seek when backwards is relevant.

Logically yes, we could just return true here now, but that would make the logic of this `if` block dependent on the existence of the previous one. I.e. if we were to change it to true and later remove the new

```C
  if (!cursorWasJustSet_) {
    return true;
  }
```

then we'd be changing the behavior and that wouldn't be trivial to notice.

Actually, I might move the new condition *below* the audio one to avoid just that.

PR #1040, opened 2025-11-12
test PR for release 0.9

Dan-Flores:


.github/workflows/docs.yaml
NicolasHug: Make sure to rely on the `release/2.9` branch of `test-infra` for all these yaml files. You'll also need to update `test-infra-ref: main`.

pyproject.toml
```diff
@@ -17,7 +17,7 @@ Documentation = "https://pytorch.org/torchcodec/stable/index.html"
 version = {file = "version.txt"}
 
 [build-system]
-requires = ["setuptools>=61.0"]
+requires = ["setuptools>=69.0"]
```
NicolasHug, line 20: were the changes above needed? I'm surprise we don't need them in `main`?

Dan-Flores: The `license-files` changes were needed to resolve [this error](https://github.com/meta-pytorch/torchcodec/actions/runs/19584544167/job/56090355784?pr=1040) on python 3.14. On review, I am not certain that requires a newer version of setuptools, though.

NicolasHug: This is fishy (and sounds like a regression). We established the `license-files = ["LICENSE"]` way in https://github.com/meta-pytorch/torchcodec/pull/789

This is either a bug in setuptools, or we're doing something wrong here.

NicolasHug (approved):

PR #1042, opened 2025-11-12
Add preset to VideoEncoder API

Dan-Flores:
This PR adds preset as an option in the VideoEncoder API, and is build on the CRF changes in #1031.
To review just these changes, see 9eee77fca79e0b907403e46f006780e8f3062cfc

`preset` is a readable string in libx264, so the docstring suggests those strings. Occasionally (as in the case of the `libsvtav1` codec), `preset` can be an `int`, so both types are accepted.

To test, test_video_encoder_against_ffmpeg_cli is updated to try combinations of pixel_format, crf, and preset.

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -96,6 +96,7 @@ function(make_torchcodec_libraries
         Encoder.cpp
         ValidationUtils.cpp
         Transform.cpp
+        Metadata.cpp
     )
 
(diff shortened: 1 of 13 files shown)


Dan-Flores:
> Did you empirically observe the effect of this parameter, e.g. does using `preset="slow"` produce a smaller output file than using `preset="fast"`?

Yes, I originally had a test for this but felt it had potential to be flaky.
FFmpeg's verbose logs also indicate the preset is being consumed correctly.

PR #1038, opened 2025-11-11
Add codec selection to VideoEncoder API

Dan-Flores:
This PR adds `codec` as an option in the VideoEncoder API.

It accepts codec specs or codec implems, similar to [FFmpeg's find_codec](https://github.com/FFmpeg/FFmpeg/blob/master/fftools/ffmpeg_opt.c#L804-L835) function by using `avcodec_find_encoder_by_name` or `avcodec_descriptor_get_by_name` + `avcodec_find_encoder`.

PR #1050, opened 2025-11-13
Add codec options to VideoEncoder API

Dan-Flores:
This PR adds the `codec_options` dictionary arg to enable all remaining codec arguments. 
The Python API accepts a `Dict[str, str]`, converts it to a flattened `List[str]` to pass to `custom_ops.py`, which rebuilds the dict in `unflattenCodecOptions()`.

#### Validation
To validate numeric options passed into `codec_options`, `validateDoubleOption` is refactored into `tryToValidateCodecOption`:
  * Check the type for each option
  * If numeric, attempt to convert provided value to a double
  * Validate value is in valid range
  * Do no validation for non-numeric options

#### Testing
The error handling done in `tryToValidateCodecOption` is documented in `test_codec_options_errors`.

Via local testing, I also checked that:
* `sortCodecOptions` properly finds and sets format options correctly
* Complex args such as `{"-x264-params": "keyint=120:bframes=3"}` are utilized

### DIFF
--- a/src/torchcodec/encoders/_video_encoder.py
+++ b/src/torchcodec/encoders/_video_encoder.py
@@ -74,7 +74,7 @@ class VideoEncoder:
             crf=crf,
             preset=preset,
             extra_options=[
-                x for k, v in (extra_options or {}).items() for x in (k, str(v))
+                str(x) for k, v in (extra_options or {}).items() for x in (k, v)
             ],
         )
 
@@ -106,7 +106,7 @@ class VideoEncoder:
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
             extra_options (dict[str, Any], optional): A dictionary of additional
-                encoder options to pass, e.g. ``{"preset": "slow", "tune": "film"}``.
+                encoder options to pass, e.g. ``{"qp": 5, "tune": "film"}``.
                 Values will be converted to strings before passing to the encoder.
 
         Returns:
@@ -122,7 +122,7 @@ class VideoEncoder:
             crf=crf,
             preset=preset_value,
             extra_options=[
-                x for k, v in (extra_options or {}).items() for x in (k, str(v))
+                str(x) for k, v in (extra_options or {}).items() for x in (k, v)
             ],
         )
 
@@ -160,7 +160,7 @@ class VideoEncoder:
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
             extra_options (dict[str, Any], optional): A dictionary of additional
-                encoder options to pass, e.g. ``{"preset": "slow", "tune": "film"}``.
+                encoder options to pass, e.g. ``{"qp": 5, "tune": "film"}``.
                 Values will be converted to strings before passing to the encoder.
         """
         preset = str(preset) if isinstance(preset, int) else preset
@@ -174,6 +174,6 @@ class VideoEncoder:
             crf=crf,
             preset=preset,
             extra_options=[
-                x for k, v in (extra_options or {}).items() for x in (k, str(v))
+                str(x) for k, v in (extra_options or {}).items() for x in (k, v)
             ],
         )
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -1092,9 +1092,9 @@ class TestVideoEncoder:
         dest = str(tmp_path / f"output.{format}")
 
         VideoEncoder(frames=frames, frame_rate=30).to_file(dest=dest, codec=codec_spec)
-        actual_codec_spec = self._get_video_metadata(dest, fields=["codec_name"]).get(
+        actual_codec_spec = self._get_video_metadata(dest, fields=["codec_name"])[
             "codec_name"
-        )
+        ]
         assert actual_codec_spec == codec_spec
 
     @pytest.mark.skipif(
@@ -1131,15 +1131,11 @@ class TestVideoEncoder:
         )
 
         assert (
-            self._get_video_metadata(spec_output, fields=["codec_name"]).get(
-                "codec_name"
-            )
+            self._get_video_metadata(spec_output, fields=["codec_name"])["codec_name"]
             == codec_spec
         )
         assert (
-            self._get_video_metadata(impl_output, fields=["codec_name"]).get(
-                "codec_name"
-            )
+            self._get_video_metadata(impl_output, fields=["codec_name"])["codec_name"]
             == codec_spec
         )
 
@@ -1175,6 +1171,7 @@ class TestVideoEncoder:
             fields=["profile", "color_space", "color_range"],
         )
         # Validate profile (case-insensitive, baseline is reported as "Constrained Baseline")
-        assert profile in metadata.get("profile", "").lower()
-        assert metadata.get("color_space") == colorspace
-        assert metadata.get("color_range") == color_range
+        expected_profile = "constrained baseline" if profile == "baseline" else profile
+        assert metadata["profile"].lower() == expected_profile
+        assert metadata["color_space"] == colorspace
+        assert metadata["color_range"] == color_range


src/torchcodec/encoders/_video_encoder.py
```diff
@@ -110,6 +137,7 @@ def to_file_like(
         pixel_format: Optional[str] = None,
         crf: Optional[Union[int, float]] = None,
         preset: Optional[Union[str, int]] = None,
+        codec_options: Optional[Dict[str, str]] = None,
```
NicolasHug: Could we make it `Dict[str, Any]`? I think there's value in allowing `"some_param":16` instead of the un-pythonic `"some_param: "16"`. We just need to call `str()` on all the values, which is OK.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -130,6 +158,8 @@ def to_file_like(
                 encoding speed and compression. Valid values depend on the encoder (commonly
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
+            codec_options (dict[str, str], optional): A dictionary of codec-specific
+                options to pass to the encoder, e.g. ``{"preset": "slow", "tune": "film"}``.
```
NicolasHug: We already have `preset` as a built-in parameter so it might be confusing to document it here.

src/torchcodec/_core/Encoder.h
```diff
@@ -154,6 +160,10 @@ class VideoEncoder {
 
  private:
   void initializeEncoder(const VideoStreamOptions& videoStreamOptions);
+  void sortCodecOptions(
+      const std::map<std::string, std::string>& codecOptions,
+      AVDictionary** codecDict,
+      AVDictionary** formatDict);
```
NicolasHug: This seems like it could be a pure function in an anonymous namespace rather than a method?

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -1,12 +1,31 @@
...
-from typing import Optional, Union
+from typing import Dict, Optional, Union
 
 import torch
 from torch import Tensor
 
 from torchcodec import _core
 
 
+def _flatten_codec_options(codec_options: Optional[Dict[str, str]]) -> Optional[list]:
```
NicolasHug: Nit: write that at the bottom like the rest of the helpers?

test/test_encoders.py
```diff
@@ -689,6 +689,45 @@ def test_pixel_format_errors(self, method, tmp_path):
...
+                {"direct-pred": "a"},
+                "Option direct-pred expects a numeric value but got 'a'",
+            ),
+            ({"tune": "not_a_real_tune"}, "avcodec_open2 failed: Invalid argument"),
+            (
+                {"tune": "10"},
+                "avcodec_open2 failed: Invalid argument",
+            ),
+        ],
+    )
```
NicolasHug, line 746: Nice job on the parametrization above

src/torchcodec/_core/Encoder.cpp
```diff
@@ -737,13 +773,19 @@ void VideoEncoder::initializeEncoder(
 
   // Apply videoStreamOptions
   AVDictionary* options = nullptr;
+  if (videoStreamOptions.codecOptions.has_value()) {
+    // Validate all codec options before setting them
```
NicolasHug: Nit, I think we can remove this comment, it doesn't add much value

src/torchcodec/_core/Encoder.cpp
```diff
@@ -586,24 +586,36 @@ void validateDoubleOption(
...
+          optionName,
+          "=",
+          numericValue,
+          " is out of valid range [",
+          option->min,
+          ", ",
+          option->max,
+          "] for this codec. For more details, run 'ffmpeg -h encoder=",
+          avCodec.name,
+          "'");
```
NicolasHug, line 609: I think this is OK for now, I suspect `min` and `max` may not always be set on for all parameters, in which case we may error out when we shouldn't? We'll know if / when we get user reports about that. Let's keep it as-is for now and see if we need to revisit in the future.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -737,13 +773,19 @@ void VideoEncoder::initializeEncoder(
 
   // Apply videoStreamOptions
   AVDictionary* options = nullptr;
+  if (videoStreamOptions.codecOptions.has_value()) {
+    // Validate all codec options before setting them
+    for (const auto& [key, value] : videoStreamOptions.codecOptions.value()) {
+      tryToValidateCodecOption(*avCodec, key.c_str(), value);
+    }
+    sortCodecOptions(
+        videoStreamOptions.codecOptions.value(), &options, &formatOptions_);
```
NicolasHug: Let's rename these variables so that it's clearer what they refer to:
```suggestion
        videoStreamOptions.codecOptions.value(), &avCodecOptions, &avFormatOptions_);
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -775,7 +817,8 @@ void VideoEncoder::encode() {
   TORCH_CHECK(!encodeWasCalled_, "Cannot call encode() twice.");
   encodeWasCalled_ = true;
 
-  int status = avformat_write_header(avFormatContext_.get(), nullptr);
+  int status = avformat_write_header(avFormatContext_.get(), &formatOptions_);
+  av_dict_free(&formatOptions_);
```
NicolasHug: Let's free it in the destructor rather than here, it's slightly less surprising.

This makes me realize: we should avoid calling `av_dict_free` and define a unique_ptr type with RAII semantics on AVDictionary, like we do for all the other FFmpeg types.

Let's open an issue to follow-up on that.

src/torchcodec/_core/ops.py
```diff
@@ -216,6 +216,7 @@ def encode_video_to_file_like(
     crf: Optional[Union[int, float]] = None,
     pixel_format: Optional[str] = None,
     preset: Optional[str] = None,
+    codec_options: Optional[list[str]] = None,
```
NicolasHug: I'm starting to wonder if that's the right name for this parameter. IIUC, this can also be used to set the AVFormatContext parameters, right? So it's not just related to the codec itself?

Maybe this could be `extra_options` or something like that? CC @scotts @mollyxu

test/test_encoders.py
```diff
@@ -689,6 +689,45 @@ def test_pixel_format_errors(self, method, tmp_path):
...
+            valid_params = dict(file_like=io.BytesIO(), format="mp4")
+        else:
+            raise ValueError(f"Unknown method: {method}")
+
+        with pytest.raises(
+            RuntimeError,
+            match=error,
+        ):
+            getattr(encoder, method)(**valid_params, codec_options=codec_options)
+
```
NicolasHug, line 766: Would be good to add a test where we pass some specific codec options and verify through ffprobe that the option was respected - very similar to the test you added in https://github.com/meta-pytorch/torchcodec/pull/1038

src/torchcodec/_core/Encoder.cpp
```diff
@@ -685,6 +697,30 @@ VideoEncoder::VideoEncoder(
...
+        key.c_str(),
+        nullptr,
+        0,
+        AV_OPT_SEARCH_CHILDREN | AV_OPT_SEARCH_FAKE_OBJ,
+        nullptr);
+    if (fmtOpt) {
+      av_dict_set(formatDict, key.c_str(), value.c_str(), 0);
+    } else {
+      // Default to codec option (includes AVCodecContext + encoder-private)
+      // validateCodecOption(*avCodecContext_->codec, key.c_str(), value);
```
NicolasHug: Remove:

```suggestion
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -685,6 +697,30 @@ VideoEncoder::VideoEncoder(
   initializeEncoder(videoStreamOptions);
 }
 
+void VideoEncoder::sortCodecOptions(
+    const std::map<std::string, std::string>& codecOptions,
+    AVDictionary** codecDict,
+    AVDictionary** formatDict) {
```
NicolasHug: This interface is OK for now but we should consider changing it once we use RAII types for the dics (see my other follow-up suggestion).

src/torchcodec/_core/Encoder.cpp
```diff
@@ -685,6 +697,30 @@ VideoEncoder::VideoEncoder(
...
+  // Search AVFormatContext's AVClass for options
```
NicolasHug: Let's remove this comment, it doesn't add much value. Let's also document what this function is doing: it takes some options as input and sorts them into codec options and format options, which are returned into two separate dicts.

src/torchcodec/_core/ops.py
scotts: Yeah, `extra_options` is probably better.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: We only access `min`/`max` on numeric parameters, my expectation (hope) it is populated for those.

src/torchcodec/encoders/_video_encoder.py
Dan-Flores: I probably don't need a function to do this in Python, I'll remove the function altogether.

test/test_encoders.py
Dan-Flores: I added `test_extra_options_utilized` that checks `profile`, `colorspace`, `color_range`. `ffprobe` has a limited number of fields we can verify.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Opened https://github.com/meta-pytorch/torchcodec/issues/1053 to follow up on this, thanks for the suggestion

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -110,6 +121,9 @@ def to_tensor(
             pixel_format=pixel_format,
             crf=crf,
             preset=preset_value,
+            extra_options=[
+                x for k, v in (extra_options or {}).items() for x in (k, str(v))
+            ],
```
NicolasHug, line 126: I was going to suggest something like that instead of the previous `_flatten_codec_options` but refrained as I thought it might be too crazy lol.

I like it, I'd just suggest the following, which is a tiiiiiny bit safer, and makes it more obvious that everything in the list is a string.


```suggestion
            extra_options=[
                str(x) for k, v in (extra_options or {}).items() for x in (k, v)
            ],
```

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -144,6 +159,9 @@ def to_file_like(
                 encoding speed and compression. Valid values depend on the encoder (commonly
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
+            extra_options (dict[str, Any], optional): A dictionary of additional
+                encoder options to pass, e.g. ``{"preset": "slow", "tune": "film"}``.
+                Values will be converted to strings before passing to the encoder.
```
NicolasHug, line 164: Make sure to align the docs (this one still have preset), same with the one below

test/test_encoders.py
```diff
@@ -1048,7 +1092,9 @@ def test_codec_parameter_utilized(self, tmp_path, format, codec_spec):
         dest = str(tmp_path / f"output.{format}")
 
         VideoEncoder(frames=frames, frame_rate=30).to_file(dest=dest, codec=codec_spec)
-        actual_codec_spec = self._get_codec_spec(dest)
+        actual_codec_spec = self._get_video_metadata(dest, fields=["codec_name"]).get(
+            "codec_name"
+        )
```
NicolasHug: here and everywhere, use the  `[]` syntax instead of `.get()`. We usually `get()` only when we want to specify a fallback value, which I don't think is intended here.

test/test_encoders.py
```diff
@@ -1084,9 +1130,51 @@ def test_codec_spec_vs_impl_equivalence(self, tmp_path, codec_spec, codec_impl):
...
+                "colorspace": colorspace,
+                "color_range": color_range,
+            },
+        )
+        metadata = self._get_video_metadata(
+            output_path,
+            fields=["profile", "color_space", "color_range"],
+        )
+        # Validate profile (case-insensitive, baseline is reported as "Constrained Baseline")
+        assert profile in metadata.get("profile", "").lower()
```
NicolasHug: I'd suggest the following, because using `in` is less strict and can sometimes be surprisingly easy to pass, like `assert "" in "abc"`


```suggestion
        expected_profile = "constrained baseline" if profile == "baseline" else profile
        assert metadata["profile"].lower() == expected_profile
```

NicolasHug (approved):
Nice work @Dan-Flores !

PR #1056, opened 2025-11-17
Use std::move to reduce copies in VideoEncoder

Dan-Flores:
This PR resolves lints and other miscellaneous fixes: 
* use `std::move` on `codec`, `pixel_format`
* update `codec` to `std::string_view` rather than `std::string`
* remove unused exception in `tryToValidateCodecOption()`
* Add `None` as default in `ops.py`
* Move `extra_options` to be keyword param in `video_encoder.py`

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -8,7 +8,6 @@
 #include <cstdint>
 #include <sstream>
 #include <string>
-#include <utility>
 #include "AVIOFileLikeContext.h"
 #include "AVIOTensorContext.h"
 #include "Encoder.h"


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -648,8 +649,8 @@ at::Tensor encode_video_to_tensor(
     std::optional<std::vector<std::string>> extra_options = std::nullopt) {
   auto avioContextHolder = std::make_unique<AVIOToTensorContext>();
   VideoStreamOptions videoStreamOptions;
-  videoStreamOptions.codec = codec;
-  videoStreamOptions.pixelFormat = pixel_format;
+  videoStreamOptions.codec = std::move(codec);
```
NicolasHug, line 651: Can you share the lint warning related to that? I'm surprised we need `std::move` for codec and pixel_format but not for preset?

Made me realize we should make this a `string_view` like the rest, instead of a `string`. Same below in the other function.

Dan-Flores: This comes from clang-tidy:
`parameter 'codec' is passed by value and only copied once; consider moving it to avoid unnecessary copies`.
Perhaps because `preset` can be an int, the lint is not raised? 

> we should make this a string_view like the rest, instead of a string

Thanks for catching this, I'll update this PR

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <sstream>
 #include <string>
+#include <utility>
```
NicolasHug: Can you double check whether that's needed?

Issue #1057, opened 2025-11-17
Video Encoder notes


NicolasHug:
- let's parametrize https://github.com/meta-pytorch/torchcodec/blob/17164fd160e11f0a3734b4c28604c9fb2690bbc8/test/test_encoders.py#L922 over the other encoding methods (to_file_like, to_tensor)
- Let's call `self._get_video_metadata` in this test to ensure the expected metadata is the same. We should explicitly check for `duration`!
- `frame_rate` should accept a float. Right now it expects an `int` and errors on floats, but fps come in floats, like in:
```py
import torch
from torchcodec.decoders import VideoDecoder
from torchcodec.encoders import VideoEncoder

dec = VideoDecoder("test/resources/nasa_13013.mp4")
print(dec.metadata)  # average_fps=29.xxxxx

frames = dec.get_frames_in_range(0, -1)
enc = VideoEncoder(frames.data, frame_rate=dec.metadata.average_fps))  # this currently raises an error
```

Let's add a test making sure we can pass a float. The test can just decode and encode `nasa_13013.mp4` as done above, and assert that the encoded duration is the same as the original.

Dan-Flores:
Completed in #1058 and #1061

PR #1058, opened 2025-11-17
Improve VideoEncoder test against FFmpeg CLI

Dan-Flores:
This PR resolves the first two bullet points in #1057:
* parametrize `test_video_encoder_against_ffmpeg_cli` over the other encoding methods (`to_file_like`, `to_tensor`)
* call `self._get_video_metadata` in this test to ensure the expected metadata is the same

### DIFF
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -992,9 +992,6 @@ class TestVideoEncoder:
                 preset=preset,
             )
             encoder_frames = self.decode(encoded_output).data
-            # Write tensor to file to check with ffprobe
-            with open(encoder_output_path, "wb") as f:
-                f.write(encoded_output.numpy().tobytes())
         elif method == "to_file_like":
             file_like = io.BytesIO()
             encoder.to_file_like(
@@ -1005,9 +1002,6 @@ class TestVideoEncoder:
                 preset=preset,
             )
             encoder_frames = self.decode(file_like.getvalue()).data
-            # Write file_like to file to check with ffprobe
-            with open(encoder_output_path, "wb") as f:
-                f.write(file_like.getvalue())
         else:
             raise ValueError(f"Unknown method: {method}")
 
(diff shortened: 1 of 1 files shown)


test/test_encoders.py
```diff
@@ -967,26 +970,45 @@ def test_video_encoder_against_ffmpeg_cli(self, tmp_path, format, encode_params)
...
+            raise ValueError(f"Unknown method: {method}")
 
         assert ffmpeg_frames.shape[0] == encoder_frames.shape[0]
 
-        # If FFmpeg selects a codec or pixel format that uses qscale (not crf),
-        # the VideoEncoder outputs *slightly* different frames.
-        # There may be additional subtle differences in the encoder.
-        percentage = 94 if ffmpeg_version == 6 or format == "avi" else 99
+        # MPEG codec used for avi format does not accept CRF
+        percentage = 94 if format == "avi" else 99
```
Dan-Flores, line 1011: I had difficulty returning identical frames for `avi` with the default codec, `mpeg4`. We could consider choosing a specific codec for this test to enforce quality, but I think its ok to keep the relaxed exactness check in this case.

test/test_encoders.py
```diff
@@ -996,6 +1024,19 @@ def test_video_encoder_against_ffmpeg_cli(self, tmp_path, format, encode_params)
...
+        ffmpeg_metadata = self._get_video_metadata(
+            ffmpeg_encoded_path,
+            fields=fields,
+        )
+        encoder_metadata = self._get_video_metadata(
+            encoder_output_path,
+            fields=fields,
+        )
+        for key in ffmpeg_metadata:
+            assert ffmpeg_metadata[key] == encoder_metadata[key]
```
NicolasHug: I think we should be able to just `assert ffmpeg_metadata == encoder_metadata` ?

test/test_encoders.py
```diff
@@ -967,26 +970,51 @@ def test_video_encoder_against_ffmpeg_cli(self, tmp_path, format, encode_params)
...
+            encoded_output = encoder.to_tensor(
+                format=format,
+                pixel_format=pixel_format,
+                crf=crf,
+                preset=preset,
+            )
+            encoder_frames = self.decode(encoded_output).data
+            # Write tensor to file to check with ffprobe
+            with open(encoder_output_path, "wb") as f:
+                f.write(encoded_output.numpy().tobytes())
```
NicolasHug: I think it's OK to only validate `to_file` against ffprobe, that's what we do for the audio encoder and it makes this test a tiny bit easier to follow. No strong opinion. If you remove it from here, make sure to remove it from `to_file_like` as well.

https://github.com/meta-pytorch/torchcodec/blob/22bcf4ddc01416b25103d0e73c62b464a7fdc201/test/test_encoders.py#L353-L354

PR #1060, opened 2025-11-18
Refactor order of getting metadata and adding a stream

scotts:
I've thought this was strange for a long time now - on main, in the public `VideoDecoder` and `AudioDecoder`, we add a stream _before_ getting the metadata. This was not the originally intended order, as evidenced by some of the error checking we do: https://github.com/meta-pytorch/torchcodec/blob/22bcf4ddc01416b25103d0e73c62b464a7fdc201/src/torchcodec/decoders/_video_decoder.py#L409-L414
We should never hit that error condition, as before we call it, we add the stream. And if the video file has no best video stream, the C++ layer would have thrown before we ever had a chance to reach this condition. I feel that it's more natural to do things in the order in this PR: first get the metadata from the file, then add the stream if the metadata is valid.

The reason why I'm doing this now is that this should simplify the decoder-native transforms. We'll want to know a video stream's height and width when pre-processing the transforms before adding a stream. And that means getting that metadata before adding a stream. In the C++ layer, this does mean accessing values in the headers in `initializeDecoder()` through `AVCodecParameters` that we didn't before.

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -198,6 +198,34 @@ SeekMode seekModeFromString(std::string_view seekMode) {
   }
 }
 
+void writeFallbackBasedMetadata(
+    std::map<std::string, std::string>& map,
+    const StreamMetadata& streamMetadata,
+    SeekMode seekMode) {
+  auto durationSeconds = streamMetadata.getDurationSeconds(seekMode);
+  if (durationSeconds.has_value()) {
+    map["durationSeconds"] = std::to_string(durationSeconds.value());
+  }
+
+  auto numFrames = streamMetadata.getNumFrames(seekMode);
+  if (numFrames.has_value()) {
+    map["numFrames"] = std::to_string(numFrames.value());
+  }
+
+  double beginStreamSeconds = streamMetadata.getBeginStreamSeconds(seekMode);
+  map["beginStreamSeconds"] = std::to_string(beginStreamSeconds);
+
+  auto endStreamSeconds = streamMetadata.getEndStreamSeconds(seekMode);
+  if (endStreamSeconds.has_value()) {
+    map["endStreamSeconds"] = std::to_string(endStreamSeconds.value());
+  }
+
+  auto averageFps = streamMetadata.getAverageFps(seekMode);
+  if (averageFps.has_value()) {
+    map["averageFps"] = std::to_string(averageFps.value());
+  }
+}
+
 int checkedToPositiveInt(const std::string& str) {
   int ret = 0;
   try {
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -524,6 +524,10 @@ void SingleStreamDecoder::addVideoStream(
   auto& streamInfo = streamInfos_[activeStreamIndex_];
   streamInfo.videoStreamOptions = videoStreamOptions;
 
+  // This metadata was already set in initializeDecoder() from the
+  // AVCodecParameters that are part of the AVStream. But we consider the
+  // AVCodecContext to be more authoritative, so we use that for our decoding
+  // stream.
```
NicolasHug: From what I understand, the AVCodecContext fields were set to those of the AVCodecParameters when we called [avcodec_parameters_to_context](https://ffmpeg.org/doxygen/4.4/codec__par_8c_source.html#l00147) just above in `addStream`:

https://github.com/meta-pytorch/torchcodec/blob/22bcf4ddc01416b25103d0e73c62b464a7fdc201/src/torchcodec/_core/SingleStreamDecoder.cpp#L462-L463

I think it's best to remove the lines below and trust that `avcodec_parameters_to_context` is doing what we expect it to do. Right now, we are setting the `streamMetadata` in a lot of different places and it makes it harder to reason about.

scotts: Oh, good call! Yup, I'm happy to remove more code. :)

src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -180,6 +180,7 @@ const AVPixelFormat* getSupportedPixelFormats(const AVCodec& avCodec);
 
 int getNumChannels(const UniqueAVFrame& avFrame);
 int getNumChannels(const SharedAVCodecContext& avCodecContext);
+int getNumChannels(const AVCodecParameters* codecpar);
```
NicolasHug, line 183: We may not need the one above anymore, I'm not sure.

NicolasHug:
oh, I approved but I wonder if the [docs failure](https://github.com/meta-pytorch/torchcodec/actions/runs/19470203429/job/55718214261?pr=1060) is real :thinking: 

 >      File "/__w/_temp/conda_environment_19470203429/lib/python3.10/site-packages/torchcodec/decoders/_video_decoder.py", line 423, in _get_and_validate_stream_metadata
        raise ValueError(
    ValueError: The minimum pts value in seconds is unknown. 
    This should never happen. Please report an issue following the steps in

src/torchcodec/_core/FFMPEGCommon.h
scotts: We're using it in `CPUDeviceInterface`: https://github.com/meta-pytorch/torchcodec/blob/22bcf4ddc01416b25103d0e73c62b464a7fdc201/src/torchcodec/_core/CpuDeviceInterface.cpp#L325
And: https://github.com/meta-pytorch/torchcodec/blob/22bcf4ddc01416b25103d0e73c62b464a7fdc201/src/torchcodec/_core/CpuDeviceInterface.cpp#L416

Maybe we could instead go to the metadata, but I _think_ it's better to get it through the codec context there. I think it's possible for the codec context to have more accurate info here while decoding. At least I believe that is the case for video dimensions, as we can have variable resolution streams. I'm unsure if that's also true for the number of channels for audio. If you can say definitely that you know it's okay to just use the header-based metadata in both of these places, I can make the change. But if you're unsure, let's handle that later - I can create an issue for follow-up.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -917,30 +945,28 @@ std::string get_stream_json_metadata(
...
+    // the fallback logic itself is tricky and not worth it for this corner
+    // case. So we hardcode in approximate mode.
+    //
+    // TODO: This hacky behavior is only necessary because the custom frame
+    //       mapping is supplied in SingleStreamDecoder::addVideoStream() rather
+    //       than in the constructor. And it's supplied to addVideoStream() and
+    //       not the constructor because we need to know the stream index. If we
+    //       can encode the relevant stream indices into custom frame mappings
+    //       itself, then we can put it in the constructor.
+    writeFallbackBasedMetadata(map, streamMetadata, SeekMode::approximate);
```
NicolasHug, line 969: I understand these workarounds are needed right now, but I have a really hard time cleanly reasoning about all the comments above.

We might eventually want to revisit the existence of `addStream`? Maybe we should just have a constructor, just like we do in Python. I think all the "add stream" logic is mainly a relic of when the decoder was potentially thought to be a multi-stream decoder, but it seems like it's hurting us now 

We still want to enable existing use-cases of users getting metadata without having to scan, but I'm pretty sure we can support that by passing approximate mode (that's what we do from the public Python APIs).

scotts: @NicolasHug, you're right, that's actually the cleanest resolution here: there's no value any more in differentiating the constructor from adding a stream. Created #1064 for follow-up.

mollyxu: To double check my understanding, the expected behaviour for custom frame mappings is 

- like exact mode for the active stream
- like approximate mode for non active streams

The reason we have to make this distinction with these conditions is because we don't know the active stream index during construction (since we `addStream` separately from the constructor). Once we consolidate `addStream` into the constructor, we would be able to get rid of all the conditions and just call  
```writeFallbackBasedMetadata(map, streamMetadata, seekMode);```

NicolasHug: Your understanding is correct @mollyxu !

PR #1063, opened 2025-11-19
Add video encoding tutorial doc

Dan-Flores:


examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+Encoding video frames with VideoEncoder
+=======================================
+
+In this example, we'll learn how to encode video frames to a file or to raw
+bytes using the :class:`~torchcodec.encoders.VideoEncoder` class.
+"""
+
+# %%
+# First, we'll download a video and decode some frames to tensors.
+# These will be the input to the VideoEncoder. For more details on decoding,
```
NicolasHug: Make sure to use ``` :class:`~torchcodec.encoders.VideoEncoder` ``` everywhere.

It should link to the docstring page. Right now it doesn't, because you need to add it to https://github.com/meta-pytorch/torchcodec/blob/main/docs/source/api_ref_encoders.rst?plain=1

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+        data=encoded_bytes.numpy().tobytes(),
+        embed=True,
+        width=640,
+        height=360,
+        mimetype="video/mp4",
+    )
+
+
+# Video source: https://www.pexels.com/video/adorable-cats-on-the-lawn-4977395/
+# License: CC0. Author: Altaf Shah.
```
NicolasHug: I think we can still use it, but I don't see the license being explicitly CC0.

```suggestion
# Author: Altaf Shah.
```

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+url = "https://videos.pexels.com/video-files/4977395/4977395-hd_1920_1080_24fps.mp4"
+
+response = requests.get(url, headers={"User-Agent": ""})
+if response.status_code != 200:
+    raise RuntimeError(f"Failed to download video. {response.status_code = }.")
+
+raw_video_bytes = response.content
+
+decoder = VideoDecoder(raw_video_bytes)
+frames = decoder[:60]  # Get first 60 frames
```
NicolasHug: Let's use get_frames_in_range instead, it's more efficient, and we want users to use the most efficient decoding methods.

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+encoded_frames = encoder.to_tensor(format="mp4")
+play_video(encoded_frames)
+
+# %%
+#
+# Now that we have encoded data, we can decode it back to verify the
+# round-trip encode/decode process works as expected:
+
+decoder_verify = VideoDecoder(encoded_frames)
+decoded_frames = decoder_verify[:]
```
NicolasHug: same here

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+            "-show_entries",
+            "stream=codec_name",
+            "-of",
+            "default=noprint_wrappers=1:nokey=1",
+            output,
+        ],
+        capture_output=True,
+        text=True,
+    )
+    print(f"Codec used in {output}: {result.stdout.strip()}")
```
NicolasHug: Great section above. The only issue is that it pollutes the codespace with `libx264_encoded.mp4` and `hevc_encoded.mp4`. Let's use temporary files instead, with e.g. https://docs.python.org/3/library/tempfile.html

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+decoded_frames = decoder_verify[:]
+
+print(f"Re-decoded video: {decoded_frames.shape = }")
+print(f"Original frames: {frames.shape = }")
+
+# %%
+# Codec Selection
+# ---------------
+#
+# The ``codec`` parameter specifies which video codec to use for encoding.
```
NicolasHug: we could start this section by indicating that by default, the codec is selected automatically based on the container format, for example "mp4" tends to default to h264 (I think? Please check me on this)

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+#
+
+# High quality (low CRF)
+high_quality_output = encoder.to_tensor(format="mp4", codec="libx264", crf=0)
+play_video(high_quality_output)
+
+# %%
+# Low quality (high CRF)
+low_quality_output = encoder.to_tensor(format="mp4", codec="libx264", crf=50)
+play_video(low_quality_output)
```
NicolasHug, line 213: well done, it's really cool to visually see the effect it has on quality!

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+
+# %%
+# Preset
+# ------
+#
+# The ``preset`` parameter controls the tradeoff between encoding speed and file compression.
+# Faster presets encode faster but produce larger files, while slower
+# presets take more time to encode but result in better compression.
+#
+# For example, with the commonly used H.264 codec, ``libx264`` presets include:
```
NicolasHug: We could link to https://trac.ffmpeg.org/wiki/Encode/H.264#a2.Chooseapresetandtune specifically here since it's about h264

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+#
+# - ``"ultrafast"`` (fastest), ``"fast"``, ``"medium"`` (default), ``"slow"``, ``"veryslow"`` (slowest, best compression).
+#
+# .. note::
+#
+#     Not all codecs support the ``presets`` option. Use ``ffmpeg -h encoder=<codec_name>``
+#     to check available options for your selected codec.
+#
+
+import os
```
NicolasHug: I think we can use pathlib instead https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat

scotts:
Let's add a reference to this tutorial in `index.rst` in the "Encoding" section.

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+
+encoder = VideoEncoder(frames=frames, frame_rate=frame_rate)
+
+# %%
+# Encoding to file, bytes, or file-like
+# -------------------------------------
+#
+# :class:`~torchcodec.encoders.VideoEncoder` supports encoding frames into a
+# file via the :meth:`~torchcodec.encoders.VideoEncoder.to_file` method, to
+# file-like objects via the :meth:`~torchcodec.encoders.VideoEncoder.to_filelike`
```
scotts:
```suggestion
# file-like objects via the :meth:`~torchcodec.encoders.VideoEncoder.to_file_like`
```

examples/encoding/video_encoding.py
scotts: Slicing actually calls `get_frames_in_range()`: https://github.com/meta-pytorch/torchcodec/blob/1ea235aa6b71bc9d6ce070d4b849dc60974082b5/src/torchcodec/decoders/_video_decoder.py#L193-L203

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+#
+# Now that we have encoded data, we can decode it back to verify the
+# round-trip encode/decode process works as expected:
+
+decoder_verify = VideoDecoder(encoded_frames)
+decoded_frames = decoder_verify[:]
+
+print(f"Re-decoded video: {decoded_frames.shape = }")
+print(f"Original frames: {frames.shape = }")
+
```
scotts, line 98: I think this is an excellent place to explain that the `format` parameter selects the default codec - we can also briefly explain the difference between, say, an mp4 video file and the actual codec used to decode and encode the video streams in that file. If this is well explained in any externall FFmpeg docs, we can link to those as well.

That then sets us up for the next section, as the natural next question a reader may have is, what if I don't want the default codec?

At the end of the "Codec Selection" section, we should give some guidance on when to just use `format` and when to specify `codec` as well. Nothing elaborate, just a sentence or two. I think that will go a long way to informing our about the relationship between these two options.

examples/encoding/video_encoding.py
scotts: @NicolasHug, made a similar comment to what I did above. :) Doing it here or in the previous section are both great to me.

examples/encoding/video_encoding.py
scotts: Agreed, I came here to say the same thing. :)

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,262 @@
...
+# %%
+# Extra Options
+# -------------
+#
+# The ``extra_options`` parameter accepts a dictionary of codec-specific options
+# that would normally be set via FFmpeg command-line arguments. This enables
+# control of encoding settings beyond the common parameters.
+#
+# For example, some potential extra options for the commonly used H.264 codec, ``libx264`` include:
+# For example, with , ``libx264``:
```
scotts: Looks like the second "For example" is from before adding the first.

examples/encoding/video_encoding.py
Dan-Flores: Thanks for the suggestions, I added some brief guidance on `codec` vs `format` at the end.

I drafted text to explain the difference between container-format and codec, but I am worried it dilutes the "Codec Selection" section with text that is not specific to the API. I would be happy to add a link, but I was not able to find useful FFmpeg docs on this subject.

examples/encoding/video_encoding.py
Dan-Flores: I added an intro here explaining the default behavior, this way all codec related text is under the same header. I added the `mp4` -> `h264` example, as it is often the case in my experience.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -51,18 +51,23 @@ def to_file(
...
+                See :ref:`crf` for details.
             preset (str or int, optional): Encoder option that controls the tradeoff between
                 encoding speed and compression. Valid values depend on the encoder (commonly
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
+                See :ref:`preset` for details.
             extra_options (dict[str, Any], optional): A dictionary of additional
                 encoder options to pass, e.g. ``{"qp": 5, "tune": "film"}``.
                 Values will be converted to strings before passing to the encoder.
+                See :ref:`extra_options` for details.
```
Dan-Flores, line 69: I did not edit the docstrings, as I expect they can be useful as a quick reference to valid values. I am open to suggestions on this, though.

examples/encoding/video_encoding.py
scotts: I think this works well. My one follow-up suggestion is to connect the sentence about codec selection with the default. Something like, "If you want a codec other than the default, use the `codec` parameter." Followed by the explanation of what it is.

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -51,18 +51,23 @@ def to_file(
             codec (str, optional): The codec to use for encoding (e.g., "libx264",
                 "h264"). If not specified, the default codec
                 for the container format will be used.
+                See :ref:`codec_selection` for details.
             pixel_format (str, optional): The pixel format for encoding (e.g.,
                 "yuv420p", "yuv444p"). If not specified, uses codec's default format.
+                See :ref:`pixel_format` for details.
             crf (int or float, optional): Constant Rate Factor for encoding quality. Lower values
                 mean better quality. Valid range depends on the encoder (commonly 0-51).
```
NicolasHug: Here and for the other docstrings, I think we should rather say


```suggestion
                mean better quality. Valid range depends on the encoder (e.g.  0-51 for libx264).
```

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -51,18 +51,23 @@ def to_file(
...
                 Defaults to None (which will use encoder's default).
+                See :ref:`crf` for details.
             preset (str or int, optional): Encoder option that controls the tradeoff between
                 encoding speed and compression. Valid values depend on the encoder (commonly
```
NicolasHug: Here and in the other methods. I suggest this because it wasn't immediately obvious to me what "compression" meant in this case.

```suggestion
                encoding speed and compression (output size). Valid values depend on the encoder (commonly
```

src/torchcodec/encoders/_video_encoder.py
```diff
@@ -51,18 +51,23 @@ def to_file(
...
                 a string: "fast", "medium", "slow"). Defaults to None
                 (which will use encoder's default).
+                See :ref:`preset` for details.
             extra_options (dict[str, Any], optional): A dictionary of additional
                 encoder options to pass, e.g. ``{"qp": 5, "tune": "film"}``.
                 Values will be converted to strings before passing to the encoder.
```
NicolasHug: Let's remove this, I think this is an implementation detail.

src/torchcodec/encoders/_video_encoder.py
NicolasHug: Docstrings look good, I left minor suggestions above. I agree it is super valuable to have short descriptions of valid values in there.

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,287 @@
...
+        ],
+        capture_output=True,
+        text=True,
+    )
+    print(f"Codec used in {name}: {result.stdout.strip()}")
+
+# %%
+# For most cases, you can simply specify the format parameter and let the FFmpeg select the default codec.
+# However, specifying the codec parameter is useful to select a particular codec implementation
+# (``libx264`` vs ``libx265``) or to have more control over the encoding behavior.
```
NicolasHug: No strong opinion but I think this last part could be left out. The intro to this section (starting with "By default, the codec ...") is really good and covers all what is necessary IMO

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,287 @@
...
+#
+# CRF (Constant Rate Factor)
+# --------------------------
+#
+# The ``crf`` parameter controls video quality, where lower values produce higher quality output.
+#
+# For example, with the commonly used H.264 codec, ``libx264``:
+#
+# - Values range from 0 (lossless) to 51 (worst quality)
+# - Values 17 or 18 are conisdered visually lossless, and the default is 23.
```
NicolasHug:
```suggestion
# - Values 17 or 18 are considered visually lossless, and the default is 23.
```

examples/encoding/video_encoding.py
```diff
@@ -0,0 +1,287 @@
...
+# ------
+#
+# The ``preset`` parameter controls the tradeoff between encoding speed and file compression.
+# Faster presets encode faster but produce larger files, while slower
+# presets take more time to encode but result in better compression.
+#
+# For example, with the commonly used H.264 codec, ``libx264`` presets include:
+#
+# - ``"ultrafast"`` (fastest), ``"fast"``, ``"medium"`` (default), ``"slow"``, ``"veryslow"`` (slowest, best compression).
+# - See additional details in the `H.264 Video Encoding Guide <https://trac.ffmpeg.org/wiki/Encode/H.264#a2.Chooseapresetandtune>`_.
```
NicolasHug: Definitely a nitpick, but I think we don't need bullet points here.

NicolasHug (approved):
Looks great, thank you! Left some minor comments that are easy to address but LGTM

examples/encoding/video_encoding.py
Dan-Flores: Reading it back it is a bit repetitive. Since its the only text after a code block, I'll remove it for consistency.

PR #1065, opened 2025-11-20
Add performance tips tutorial

mollyxu:
Consolidate the following performance tips in docs
1. **Batch APIs** - Decode multiple frames at once
2. **Approximate Mode & Keyframe Mappings** - Trade accuracy for speed
3. **Multi-threading** - Parallelize decoding across videos or chunks
4. **CUDA Acceleration** - Use GPU decoding for supported formats

Also updated docs on cuda example to reflect cpu fallback (#943)

### DIFF
--- a/.github/workflows/linux_cuda_aarch64_wheel.yaml
+++ b/.github/workflows/linux_cuda_aarch64_wheel.yaml
@@ -56,3 +56,80 @@ jobs:
       architecture: aarch64
       build-platform: "python-build-package"
       build-command: "BUILD_AGAINST_ALL_FFMPEG_FROM_S3=1 ENABLE_CUDA=1 python -m build --wheel -vvv --no-isolation"
+
+  install-and-test:
+    runs-on: linux.arm64.2xlarge
+    container:
+      image: pytorch/manylinuxaarch64-builder:cuda12.6
+    env:
+      cuda_version_without_periods: "126"
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.10']
+        ffmpeg-version-for-tests: ['4.4.2', '5.1.2', '6.1.1', '7.0.1', '8.0']
+    needs: build
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: meta-pytorch_torchcodec__${{ matrix.python-version }}_cu${{ env.cuda_version_without_periods }}_aarch64
+          path: pytorch/torchcodec/dist/
+      - name: Setup conda env
+        uses: conda-incubator/setup-miniconda@v3
+        with:
+          auto-update-conda: true
+          # Using miniforge instead of miniconda ensures that the default
+          # conda channel is conda-forge instead of main/default. This ensures
+          # ABI consistency between dependencies:
+          # https://conda-forge.org/docs/user/transitioning_from_defaults/
+          miniforge-version: latest
+          activate-environment: test
+          python-version: ${{ matrix.python-version }}
+      - name: Update pip
+        run: python -m pip install --upgrade pip
+      - name: Install PyTorch
+        run: |
+          ${CONDA_RUN} python -m pip install --pre torch torchvision --index-url https://download.pytorch.org/whl/nightly/cu${{ env.cuda_version_without_periods }}
+      - name: Install torchcodec from the wheel
+        run: |
+          wheel_path=`find pytorch/torchcodec/dist -type f -name "*.whl"`
+          echo Installing $wheel_path
+          python -m pip install $wheel_path -vvv
+
+      - name: Check out repo
+        uses: actions/checkout@v3
+      - name: Install ffmpeg, post build
+        run: |
+          # Ideally we would have checked for that before installing the wheel,
+          # but we need to checkout the repo to access this file, and we don't
+          # want to checkout the repo before installing the wheel to avoid any
+          # side-effect. It's OK.
+          source packaging/helpers.sh
+          assert_ffmpeg_not_installed
+
+          conda install "ffmpeg=${{ matrix.ffmpeg-version-for-tests }}" -c conda-forge
+          ffmpeg -version
+          echo LD_LIBRARY_PATH=$CONDA_PREFIX/lib:/usr/local/cuda/lib64/:${LD_LIBRARY_PATH} >> $GITHUB_ENV
+
+      - name: Install test dependencies
+        run: |
+          # Ideally we would find a way to get those dependencies from pyproject.toml
+          python -m pip install numpy pytest pillow
+
+      - name: Delete the src/ folder just for fun
+        run: |
+          # The only reason we checked-out the repo is to get access to the
+          # tests. We don't care about the rest. Out of precaution, we delete
+          # the src/ folder to be extra sure that we're running the code from
+          # the installed wheel rather than from the source.
+          # This is just to be extra cautious and very overkill because a)
+          # there's no way the `torchcodec` package from src/ can be found from
+          # the PythonPath: the main point of `src/` is precisely to protect
+          # against that and b) if we ever were to execute code from
+          # `src/torchcodec`, it would fail loudly because the built .so files
+          # aren't present there.
+          rm -r src/
+          ls
+      - name: Run Python tests
+        run: |
+          pytest --override-ini="addopts=-v" test
--- a/benchmarks/decoders/benchmark_transforms.py
+++ b/benchmarks/decoders/benchmark_transforms.py
@@ -5,14 +5,11 @@ from time import perf_counter_ns
 
 import torch
 from torch import Tensor
-from torchcodec._core import add_video_stream, create_from_file, get_frames_by_pts
 from torchcodec.decoders import VideoDecoder
 from torchvision.transforms import v2
 
-DEFAULT_NUM_EXP = 20
 
-
-def bench(f, *args, num_exp=DEFAULT_NUM_EXP, warmup=1) -> Tensor:
+def bench(f, *args, num_exp, warmup=1) -> Tensor:
 
     for _ in range(warmup):
         f(*args)
@@ -45,37 +42,55 @@ def report_stats(times: Tensor, unit: str = "ms", prefix: str = "") -> float:
 
 
 def torchvision_resize(
-    path: Path, pts_seconds: list[float], dims: tuple[int, int]
-) -> None:
-    decoder = create_from_file(str(path), seek_mode="approximate")
-    add_video_stream(decoder)
-    raw_frames, *_ = get_frames_by_pts(decoder, timestamps=pts_seconds)
-    return v2.functional.resize(raw_frames, size=dims)
+    path: Path, pts_seconds: list[float], dims: tuple[int, int], num_threads: int
+) -> Tensor:
+    decoder = VideoDecoder(
+        path, seek_mode="approximate", num_ffmpeg_threads=num_threads
+    )
+    raw_frames = decoder.get_frames_played_at(pts_seconds)
+    transformed_frames = v2.Resize(size=dims)(raw_frames.data)
+    assert len(transformed_frames) == len(pts_seconds)
+    return transformed_frames
 
 
 def torchvision_crop(
-    path: Path, pts_seconds: list[float], dims: tuple[int, int], x: int, y: int
-) -> None:
-    decoder = create_from_file(str(path), seek_mode="approximate")
-    add_video_stream(decoder)
-    raw_frames, *_ = get_frames_by_pts(decoder, timestamps=pts_seconds)
-    return v2.functional.crop(raw_frames, top=y, left=x, height=dims[0], width=dims[1])
+    path: Path, pts_seconds: list[float], dims: tuple[int, int], num_threads: int
+) -> Tensor:
+    decoder = VideoDecoder(
+        path, seek_mode="approximate", num_ffmpeg_threads=num_threads
+    )
+    raw_frames = decoder.get_frames_played_at(pts_seconds)
+    transformed_frames = v2.CenterCrop(size=dims)(raw_frames.data)
+    assert len(transformed_frames) == len(pts_seconds)
+    return transformed_frames
 
 
-def decoder_native_resize(
-    path: Path, pts_seconds: list[float], dims: tuple[int, int]
-) -> None:
-    decoder = create_from_file(str(path), seek_mode="approximate")
-    add_video_stream(decoder, transform_specs=f"resize, {dims[0]}, {dims[1]}")
-    return get_frames_by_pts(decoder, timestamps=pts_seconds)[0]
+def decoder_resize(
+    path: Path, pts_seconds: list[float], dims: tuple[int, int], num_threads: int
+) -> Tensor:
+    decoder = VideoDecoder(
+        path,
+        transforms=[v2.Resize(size=dims)],
+        seek_mode="approximate",
+        num_ffmpeg_threads=num_threads,
+    )
+    transformed_frames = decoder.get_frames_played_at(pts_seconds).data
+    assert len(transformed_frames) == len(pts_seconds)
+    return transformed_frames.data
 
 
-def decoder_native_crop(
-    path: Path, pts_seconds: list[float], dims: tuple[int, int], x: int, y: int
-) -> None:
-    decoder = create_from_file(str(path), seek_mode="approximate")
-    add_video_stream(decoder, transform_specs=f"crop, {dims[0]}, {dims[1]}, {x}, {y}")
-    return get_frames_by_pts(decoder, timestamps=pts_seconds)[0]
+def decoder_crop(
+    path: Path, pts_seconds: list[float], dims: tuple[int, int], num_threads: int
+) -> Tensor:
+    decoder = VideoDecoder(
+        path,
+        transforms=[v2.CenterCrop(size=dims)],
+        seek_mode="approximate",
+        num_ffmpeg_threads=num_threads,
+    )
+    transformed_frames = decoder.get_frames_played_at(pts_seconds).data
+    assert len(transformed_frames) == len(pts_seconds)
+    return transformed_frames
 
 
 def main():
@@ -84,8 +99,26 @@ def main():
     parser.add_argument(
         "--num-exp",
         type=int,
-        default=DEFAULT_NUM_EXP,
+        default=5,
         help="number of runs to average over",
+    )
+    parser.add_argument(
+        "--num-threads",
+        type=int,
+        default=1,
+        help="number of threads to use; 0 means FFmpeg decides",
+    )
+    parser.add_argument(
+        "--total-frame-fractions",
+        nargs="+",
+        type=float,
+        default=[0.005, 0.01, 0.05, 0.1],
+    )
+    parser.add_argument(
+        "--input-dimension-fractions",
+        nargs="+",
+        type=float,
+        default=[0.5, 0.25, 0.125],
     )
 
     args = parser.parse_args()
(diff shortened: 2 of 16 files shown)


examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+# Overview
+# --------
+#
+# When decoding videos with TorchCodec, several techniques can significantly
+# improve performance depending on your use case. This guide covers:
+#
+# 1. **Batch APIs** - Decode multiple frames at once
+# 2. **Approximate Mode & Keyframe Mappings** - Trade accuracy for speed
+# 3. **Multi-threading** - Parallelize decoding across videos or chunks
+# 4. **CUDA Acceleration (BETA)** - Use GPU decoding for supported formats
```
NicolasHug:
```suggestion
# 4. **CUDA Acceleration** - Use GPU decoding for supported formats
```

Sharing a bit more context: *CUDA Acceleration* itself isn't a beta feature: we've supported CUDA decoding with stable APIs for a while now. What is Beta is the "new Beta interface for CUDA decoding" that we introduced in 0.8, but the "previous CUDA interface" is still stable. I'll share more below in the dedicated CUDA section

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+#
+# We'll explore each technique and when to use it.
+
+# %%
+# 1. Use Batch APIs When Possible
+# --------------------------------
+#
+# If you need to decode multiple frames at once, it is faster when using the batch methods. TorchCodec's batch APIs reduce overhead and can leverage
+# internal optimizations.
```
NicolasHug: Nit: here it might be useful to explicitly say that the batch methods are faster than the single-frame decoding methods - e.g. `get_frames_at()` is faster than calling `get_frame_at()` multiple times.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+#
+#     - :ref:`sphx_glr_generated_examples_decoding_custom_frame_mappings.py`
+
+# %%
+# 3. Multi-threading for Parallel Decoding
+# -----------------------------------------
+#
+# When decoding multiple videos or decoding a large number of frames from a single video, there are a few parallelization strategies to speed up the decoding process:
+#
+# - **FFmpeg-based parallelism** - Using FFmpeg's internal threading capabilities
```
NicolasHug: Let's add that this will involve intra-frame parallelism, i.e. the parallelization happens over individual frames, not across frames.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+# - **Multiprocessing** - Distributing work across multiple processes
+# - **Multithreading** - Using multiple threads within a single process
+
```
NicolasHug, line 123: Let's add that multi-processing and multi-threading can be used either to decode multiple videos in parallel, or to decode a single long video in parallel by splitting it into chunks - that's what the tutorial referenced below does.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+# %%
+# .. note::
+#
+#     For complete examples comparing
+#     sequential, ffmpeg-based parallelism, multi-process, and multi-threaded approaches, see:
+#
+#     - :ref:`sphx_glr_generated_examples_decoding_parallel_decoding.py`
+
+# %%
+# 4. BETA: CUDA Acceleration
```
NicolasHug:
```suggestion
# 4. CUDA Acceleration
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+# :class:`~torchcodec.decoders.VideoDecoder` creation, particularly for long
+# videos, but may result in slightly less accurate seeking in some cases.
+#
+#
+# **Which mode should you use:**
+#
+# - If you care about exactness of frame seeking, use “exact”.
+# - If you can sacrifice exactness of seeking for speed, which is usually the case when doing clip sampling, use “approximate”.
+# - If your videos don’t have variable framerate and their metadata is correct, then “approximate” mode is a net win: it will be just as accurate as the “exact” mode while still being significantly faster.
+# - If your size is small enough and we’re decoding a lot of frames, there’s a chance exact mode is actually faster.
```
NicolasHug: Above:

This is a good description. I think we can be more nuanced about when to recommend approximate, e.g. we should try to clearly articulate the last 3 bullet points which are currently slightly overlapping and contradictory (we now know that approximate won't always be "a net win").
 
That's on me: I need to first have a clear understanding of why approximate mode is sometimes slower, and I'll need to update the approximate mode tutorial with more detailed recommendations.

I won't be able to do that in the next few days, so to unblock yourself I think you can just remove the claims about approximate being strictly superior ( bullet points 2 and 3), and the more generic reco could be something like

> If the video is long and you're only decoding a small amount of frames, approximate mode should be faster. 

It's not super actionable for users but I hope the dedicated tutorial I'll edit will be more precise.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+# By default, TorchCodec uses ``seek_mode="exact"``, which performs a scan when
+# the decoder is created to build an accurate internal index of frames. This
+# ensures frame-accurate seeking but takes longer for decoder initialization,
+# especially on long videos.
+
+# %%
+# **Approximate Mode**
+# ~~~~~~~~~~~~~~~~~~~~
+#
+# Setting ``seek_mode="approximate"`` skips the initial scan and relies on the
```
NicolasHug: Nit

```suggestion
# Setting ``seek_mode="approximate"`` skips the initial :term:`scan` and relies on the
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,159 @@
...
+#
+# **When NOT to use:**
+#
+# - You need bit-exact results
+# - Small resolution videos and the PCI-e transfer latency is large
+# - GPU is already busy and CPU is idle
+#
+# **Performance impact:** CUDA decoding can significantly outperform CPU decoding,
+# especially for high-resolution videos and when combined with GPU-based transforms.
+# Actual speedup varies by hardware, resolution, and codec.
```
NicolasHug, line 164: I think it's good to have those bullet points here. They overlap with what is already in the CUDA decoding tutorial, and I think we'll want to remove them from there and have them here instead.

Eventually we'll also want to update the CUDA tutorial to explain to users how to check whether they're falling back to the CPU.

Mainly here in this tutorial, I think we should insist on one thing (as the main point): users should be using the Beta interface with

```py
    dec = VideoDecoder("file.mp4", device="cuda")
```

mollyxu:
Thanks for the feedback!

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,170 @@
...
+# If you need to decode multiple frames at once, the batch methods are faster than calling single-frame decoding methods multiple times.
+# For example, :meth:`~torchcodec.decoders.VideoDecoder.get_frames_at` is faster than calling :meth:`~torchcodec.decoders.VideoDecoder.get_frame_at` multiple times.
+# TorchCodec's batch APIs reduce overhead and can leverage internal optimizations.
+#
+# **Key Methods:**
+#
+# - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_at` for specific indices
+# - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_in_range` for ranges
+# - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_at` for timestamps
+# - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_in_range` for time ranges
```
Dan-Flores, line 53: Maybe it would be more clear to group the similar functions here? 
For example, we could add two small headers to group index based vs timestamp based retrieval:

`For index based frame retrieval:` 
- `get_frames_at`
- `get_frames_in_range`

`For timestamp based frame retrieval:`
...

Dan-Flores:
Let's update `docs/source/index.rst` so this tutorial appears on the main `index.html` page (similar to [these changes](https://github.com/meta-pytorch/torchcodec/pull/1063/files#diff-4eec6cec5f6fab1548b85433ea8ca81315ae165db4b7f84019f287df9015699f))

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,170 @@
...
+# - Decoding large resolution videos
+# - Large batch of videos saturating the CPU
+# - GPU-intensive pipelines with transforms like scaling and cropping
+# - CPU is saturated and you want to free it up for other work
+#
+# **When NOT to use:**
+#
+# - You need bit-exact results
+# - Small resolution videos and the PCI-e transfer latency is large
+# - GPU is already busy and CPU is idle
```
Dan-Flores, line 160: super nit: this is a personal writing style preference, but within a section lets consistently use either active or passive voice. For example, we could remove "you" from the first bullet point, and instead use the passive voice: "bit-exact results are needed"

scotts: Let's consistently use active voice. :)

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,175 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+====================================
+Performance Tips and Best Practices
```
svekars: Probably a more descriptive title would help with discoverability : TorchCodec Performance Tips and Best Practices. 
Also adding meta directive at the top should help as well. SOmething like:
```
.. meta::
   :description:  Learn how to optimize TorchCodec video decoding performance with batch APIs, approximate seeking, multi-threading, and CUDA acceleration.
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,175 @@
...
+#         decoder = VideoDecoder("file.mp4", device="cuda")
+#
+
+# %%
+# .. note::
+#
+#     For installation instructions, detailed examples, and visual comparisons
+#     between CPU and CUDA decoding, see:
+#
+#     - :ref:`sphx_glr_generated_examples_decoding_basic_cuda_example.py`
```
svekars: I would also add a conclusion:
```
# Conclusion
# ----------
#
# TorchCodec offers multiple performance optimization strategies, each suited to
# different scenarios. Use batch APIs for multi-frame decoding, approximate mode
# for faster initialization, parallel processing for high throughput, and CUDA
# acceleration for GPU-intensive workflows.
#
# The best results often come from combining techniques. Profile your specific
# use case and apply optimizations incrementally, using the benchmarks in the
# linked examples as a guide.
#
# For more information, see: list where to go?
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,175 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
+====================================
+Performance Tips and Best Practices
+====================================
```
svekars: It's not required but there is a template, you could look into as well: https://github.com/pytorch/tutorials/blob/main/beginner_source/template_tutorial.py

mollyxu: Thanks for the feedback!

examples/decoding/basic_cuda_example.py
```diff
@@ -161,3 +139,22 @@ def plot_cpu_and_cuda_frames(cpu_frames: torch.Tensor, cuda_frames: torch.Tensor
...
 print(f"{mean_abs_diff=}")
 print(f"{max_abs_diff=}")
+
+
+# %%
+# Checking for CPU Fallback
+# -------------------------------------
+#
+# In some cases, CUDA decoding may fall back to CPU decoding. This can happen
+# when the video codec or format is not supported by the NVDEC hardware decoder.
```
NicolasHug:
```suggestion
# In some cases, CUDA decoding may fall back to CPU decoding. This can happen
# when the video codec or format is not supported by the NVDEC hardware decoder, or when NVCUVID wasn't found.
```

examples/decoding/basic_cuda_example.py
```diff
@@ -161,3 +139,22 @@ def plot_cpu_and_cuda_frames(cpu_frames: torch.Tensor, cuda_frames: torch.Tensor
...
+# to help you detect when this fallback occurs.
+#
+# You can access the fallback status via the
+# :attr:`~torchcodec.decoders.VideoDecoder.cpu_fallback` attribute:
+
+with set_cuda_backend("beta"):
+    decoder = VideoDecoder(video_file, device="cuda")
+
+# Check and print the CPU fallback status
+print(decoder.cpu_fallback)
```
NicolasHug: On the new section above: this is great. Let's move it just above the "Visualizing Frames" section.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+# frame indices and actual keyframe locations. This allows you to speed up :class:`~torchcodec.decoders.VideoDecoder`
+# instantiation while maintaining the frame seeking accuracy of ``seek_mode="exact"``
+#
+# **When to use:**
+#
+# - Frame accuracy is critical, so you cannot use approximate mode
+# - You can preprocess videos once and then decode them many times
+#
+# **Performance impact:** Enables consistent, predictable performance for repeated
+# random access without the overhead of exact mode's scanning.
```
NicolasHug: I'm not sure "random access" is really relevant here. For perf, this is less about frame access than it is about decoder initialization

```suggestion
# **Performance impact:** speeds up decoder instantiation, similarly to ``seek_mode="approximate"``.
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+#
+#     - :ref:`sphx_glr_generated_examples_decoding_custom_frame_mappings.py`
+
+# %%
+# 3. Multi-threading for Parallel Decoding
+# -----------------------------------------
+#
+# When decoding multiple videos or decoding a large number of frames from a single video, there are a few parallelization strategies to speed up the decoding process:
+#
+# - **FFmpeg-based parallelism** - Using FFmpeg's internal threading capabilities for intra-frame parallelism, where parallelization happens within individual frames rather than across frames
```
NicolasHug:
```suggestion
# - **FFmpeg-based parallelism** - Using FFmpeg's internal threading capabilities for intra-frame parallelism, where parallelization happens within individual frames rather than across frames. For that, use the `num_ffmpeg_threads` parameter of the :class:`~torchcodec.decoders.VideoDecoder`
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+#
+# - Decoding multiple frames
+
+# %%
+# .. note::
+#
+#     For complete examples with runnable code demonstrating batch decoding,
+#     iteration, and frame retrieval, see:
+#
+#     - :ref:`sphx_glr_generated_examples_decoding_basic_example.py`
```
NicolasHug: We can make the rendering slightly more compact by avoiding bullet points (here and everwhere else in these "notes"). I think it flows a bit better:


```suggestion
#     iteration, and frame retrieval, see :ref:`sphx_glr_generated_examples_decoding_basic_example.py`
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+# TorchCodec supports GPU-accelerated decoding using NVIDIA's hardware decoder
+# (NVDEC) on supported hardware. This keeps decoded tensors in GPU memory,
+# avoiding expensive CPU-GPU transfers for downstream GPU operations.
+#
+# **When to use:**
+#
+# - Decoding large resolution videos
+# - Large batch of videos saturating the CPU
+# - GPU-intensive pipelines with transforms like scaling and cropping
+# - CPU is saturated and you want to free it up for other work
```
NicolasHug: Let's remove these two. The first one isn't very precise, and the second one is redundant with another entry just above.

```suggestion
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+#
+# **When NOT to use:**
+#
+# - You need bit-exact results
```
NicolasHug:
```suggestion
# - You need bit-exact results with CPU decoding
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+# - Small resolution videos and the PCI-e transfer latency is large
+# - GPU is already busy and CPU is idle
+#
+# **Performance impact:** CUDA decoding can significantly outperform CPU decoding,
+# especially for high-resolution videos and when combined with GPU-based transforms.
```
NicolasHug: The "transforms" stuff is slightly misleading and might become even more misleading soon when we actually release native transforms - but they'll be CPU-only for a bit.

```suggestion
# **Performance impact:** CUDA decoding can significantly outperform CPU decoding,
# especially for high-resolution videos and when decoding a lot of frames.
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+# Actual speedup varies by hardware, resolution, and codec.
+
+# %%
+# **Recommended Usage for Beta Interface**
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# .. code-block:: python
+#
+#     with set_cuda_backend("beta"):
+#         decoder = VideoDecoder("file.mp4", device="cuda")
```
NicolasHug: Let's move this section above, it should be the first one we see (before the "when to use" section"). Let's also make it slightly more obvious:

```suggestion
# %%
# **Recommended:  use the Beta Interface!!**
# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
#
# We recommend you use the new "beta" CUDA interface which is significantly faster than the previous one, and supports the same features:
#
# .. code-block:: python
#
#     with set_cuda_backend("beta"):
#         decoder = VideoDecoder("file.mp4", device="cuda")
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+# video codec or format is not supported by NVDEC. You can detect this using
+# the :attr:`~torchcodec.decoders.VideoDecoder.cpu_fallback` attribute:
+#
+# .. code-block:: python
+#
+#     decoder = VideoDecoder("file.mp4", device="cuda")
+#     decoder[0]  # Decode at least one frame first (for FFmpeg backend)
+#
+#     # Print detailed fallback status
+#     print(decoder.cpu_fallback)
```
NicolasHug: Let's use the beta interface here - we really want users to use that as the default now. Since we're using beta, we don't need to decode a frame first. That will be documented just below in your great note.

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+#     - :ref:`sphx_glr_generated_examples_decoding_basic_cuda_example.py`
+
+# %%
+# Conclusion
+# ----------
+#
+# TorchCodec offers multiple performance optimization strategies, each suited to
+# different scenarios. Use batch APIs for multi-frame decoding, approximate mode
+# for faster initialization, parallel processing for high throughput, and CUDA
+# acceleration for GPU-intensive workflows.
```
NicolasHug:
```suggestion
# acceleration to offload the CPU.
```

examples/decoding/performance_tips.py
```diff
@@ -0,0 +1,222 @@
...
+#
+# The best results often come from combining techniques. Profile your specific
+# use case and apply optimizations incrementally, using the benchmarks in the
+# linked examples as a guide.
+#
+# For more information, see:
+#
```
NicolasHug, line 207: Related to my other comment above, the list below should definitely be kept as a bullet list!

PR #1068, opened 2025-11-20
Remove tests from collection

mollyxu:
## Remove skipped tests from collection in internal environment

Prevent internal CI failures from skipped tests by excluding skipped tests from collection in internal environment.

TorchCodec tests were flaky / broken internal because:
- Tests were using runtime `pytest.skip()` calls, so they were collected but then skipped during execution

#### 1. Added `@needs_ffmpeg_cli` marker to tests that need ffmpeg-cli
- `test_ops.py::TestVideoDecoderOps::test_seek_mode_custom_frame_mappings`
- `test_ops.py::TestVideoDecoderOps:: test_seek_mode_custom_frame_mappings_fails `
- `test_encoders.py::TestAudioEncoder::test_against_cli`
- `test_encoders.py::TestAudioEncoder::test_video_encoder_against_ffmpeg_cli`
- `test_encoders.py::TestAudioEncoder::test_codec_parameter_utilized`
- `test_encoders.py::TestAudioEncoder::test_codec_spec_vs_impl_equivalence`
- `test_encoders.py::TestAudioEncoder::test_extra_options_utilized`
- `test_decoders.py::TestVideoDecoder::test_custom_frame_mappings_json_and_bytes`
- `test_decoders.py::TestVideoDecoder::test_custom_frame_mappings_init_fails`

#### 2. Converted runtime skips to collection-time skips

Replaced runtime `pytest.skip()` calls with `pytest.param(..., marks=pytest.mark.skipif(...))` for:
- FFmpeg version compatibility checks (e.g., FFmpeg 4 incompatibility with wav files)
- Platform-specific codec availability (e.g., webm codec unavailable on Windows with certain FFmpeg versions)

#### 3. Updated `conftest.py` to exclude tests internally

Modified `pytest_collection_modifyitems` hook to exclude tests from collection when running internally if they have:
- `@needs_ffmpeg_cli` marker
- `@pytest.mark.skip()` decorator
- `@pytest.mark.skipif()` decorator

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,7 +30,7 @@ Start by installing the **nightly** build of PyTorch following the
 Then, the easiest way to install the rest of the dependencies is to run:
 
 ```bash
-conda install cmake pkg-config pybind11 "ffmpeg<8" -c conda-forge
+conda install cmake pkg-config pybind11 "ffmpeg" -c conda-forge
 ```
 
 ### Clone and build
--- a/README.md
+++ b/README.md
@@ -107,16 +107,16 @@ ffmpeg -f lavfi -i \
    `torch` and `torchcodec`.
 
 2. Install FFmpeg, if it's not already installed. Linux distributions usually
-   come with FFmpeg pre-installed. TorchCodec supports major FFmpeg versions
-   in [4, 7] on all platforms, and FFmpeg version 8 is supported on Mac and Linux.
+   come with FFmpeg pre-installed. TorchCodec supports supports all major FFmpeg versions
+   in [4, 8].
 
    If FFmpeg is not already installed, or you need a more recent version, an
    easy way to install it is to use `conda`:
 
    ```bash
-   conda install "ffmpeg<8"
+   conda install "ffmpeg"
    # or
-   conda install "ffmpeg<8" -c conda-forge
+   conda install "ffmpeg" -c conda-forge
    ```
 
 3. Install TorchCodec:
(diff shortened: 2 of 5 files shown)


test/test_encoders.py
```diff
@@ -356,17 +365,30 @@ def test_against_cli(
...
+        ],
+    )
     @pytest.mark.parametrize("method", ("to_tensor", "to_file_like"))
     def test_against_to_file(
         self, asset, bit_rate, num_channels, format, tmp_path, method
     ):
-        if get_ffmpeg_major_version() == 4 and format == "wav":
-            pytest.skip("Swresample with FFmpeg 4 doesn't work on wav files")
-        if IS_WINDOWS and get_ffmpeg_major_version() <= 5 and format == "mp3":
-            # TODO: https://github.com/pytorch/torchcodec/issues/837
```
scotts, line 367: Is it possible to preserve the comment pointing to an issue?

mollyxu: Do you think line 371 is a good place to move this TODO?

test/conftest.py
```diff
@@ -28,6 +33,15 @@ def pytest_collection_modifyitems(items):
         # 'needs_cuda' mark, and the ones with device == 'cpu' won't have the
         # mark.
         needs_cuda = item.get_closest_marker("needs_cuda") is not None
+        needs_ffmpeg_cli = item.get_closest_marker("needs_ffmpeg_cli") is not None
+        has_skip_marker = item.get_closest_marker("skip") is not None
+        has_skipif_marker = item.get_closest_marker("skipif") is not None
```
NicolasHug, line 38: QQ @mollyxu , if we use 

```py
@pytest.mark.skipif(cond, ...)
```

and `cond` is `False`, does the test still have the mark? I just want to make sure we're not skipping tests that shouldn't be skipped. It might be worth adding a small comment :) !

mollyxu: Sorry I missed this comment! Thanks for catching this - the marker will still apply when `cond` is `False`. I made a follow-up PR to address this #1155

PR #1070, opened 2025-11-21
Implement RandomCrop transform

scotts:
Implements `torchcodec.transforms.RandomCrop` and also accepts `torchvision.transforms.v2.RandomCrop`. The key difference between this capability and `Resize` is that we need to:

1. Compute a random location in the image to crop.
2. And it must match exactly what TorchVision does.

Short version of how we accomplish this:

1. If you give us the TorchVision object, we call `make_params()` on it to get the computed location.
2. If you don't, we do the same calculation in TorchCodec. We'll need to use testing and code review to make sure these stay aligned.

Working on this transform also made me realize that `DecoderTransform` and its subclasses are not dataclasses. I initially thought they would just be bags of values. But they're growing to have significant methods and internal state not exposed to users. In a follow-up PR, I'll refactor these into normal classes, much like the TorchVision versions. I felt that was too disruptive to do in this PR.

### DIFF
--- a/src/torchcodec/transforms/_decoder_transforms.py
+++ b/src/torchcodec/transforms/_decoder_transforms.py
@@ -58,7 +58,7 @@ class DecoderTransform(ABC):
                 we want that to still work.
 
                 Note: This method is the moral equivalent of TorchVision's
-                `Transformer.make_params()`.
+                `Transform.make_params()`.
 
         Returns:
             str: A string which contains the spec for the transform that the


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -270,8 +283,8 @@ Transform* makeCropTransform(
       "cropTransformSpec must have 5 elements including its name");
   int height = checkedToPositiveInt(cropTransformSpec[1]);
   int width = checkedToPositiveInt(cropTransformSpec[2]);
-  int x = checkedToPositiveInt(cropTransformSpec[3]);
-  int y = checkedToPositiveInt(cropTransformSpec[4]);
+  int x = checkedToNonNegativeInt(cropTransformSpec[3]);
+  int y = checkedToNonNegativeInt(cropTransformSpec[4]);
```
scotts, line 287: The location (0, 0) is a valid image location. 🤦

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -66,28 +72,167 @@ class Resize(DecoderTransform):
...
+        if len(self.size) != 2:
+            raise ValueError(
+                f"RandomCrop's size must be a sequence of length 2, got {self.size}. "
+                "This should never happen, please report a bug."
+            )
+
+        if self._top is None or self._left is None:
+            # TODO: It would be very strange if only ONE of those is None. But should we
+            #       make it an error? We can continue, but it would probably mean
+            #       something bad happened. Dear reviewer, please register an opinion here:
```
Dan-Flores: I agree it would appear something bad happened in this case. 

But when calling this function, do we expect `_top` or `_left` to have any value? My understanding is that these fields are only set when `_make_transform_spec` is called, which is only called once per `DecoderTransform`.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -482,21 +488,39 @@ def _convert_to_decoder_transforms(
...
+                input_dims = transform_tc._get_output_dims(input_dims)
+                converted_transforms.append(transform_tc)
             else:
                 raise ValueError(
                     f"Unsupported transform: {transform}. Transforms must be "
                     "either a TorchCodec DecoderTransform or a TorchVision "
                     "v2 transform."
                 )
         else:
+            input_dims = transform._get_output_dims(input_dims)
```
Dan-Flores: Is `_get_output_dims` is only used in this function for validation? I believe there are TODOs to move validation to the constructor which is great, but I do not understand the the returned value `input_dims` here.

src/torchcodec/transforms/_decoder_transforms.py
scotts: It depends on if `RandomCrop` was created via a TorchVision `RandomCrop` or instantiated directly. If created by a TorchVision `RandomCrop` in `_from_torchvision()`, then `_top` and `_left` should have values. If created directly, then both should not have values, in which case we have to do our random logic.

It has occurred to me that maybe we don't need to call `RandomCrop.make_params()` in `_from_torchvision()`. Maybe we should just always set these values in `_make_transform_spec()`.

src/torchcodec/decoders/_video_decoder.py
scotts: It's not actually used for validation. Think of the transforms as a pipeline: `A -> B -> C -> D.` Each stage may change the dimensions of the frame. We need to track the frame dimensions as we move through the pipeline because some transforms need to know the dimensions of the input frame. `RandomCrop` is one such transform: in order to randomly determine a location to crop, it needs to know the input frame dimensions to know the bounds to pass to the random number generator.

The dimensions that `A` receives come from the originally decoded frame, which we can get from the metadata. But the dimensions for `B` are actually the output of `A`! That extends to each transform in the pipeline.

This probably deserves a comment. :)

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -482,21 +488,39 @@ def _convert_to_decoder_transforms(
...
             converted_transforms.append(transform)
 
     return converted_transforms
 
 
 def _make_transform_specs(
     transforms: Optional[Sequence[Union[DecoderTransform, nn.Module]]],
+    input_dims: Tuple[Optional[int], Optional[int]],
+    dimension_order: Literal["NCHW", "NHWC"],
```
NicolasHug: That might be personal preference, but the `_make_transform_specs` function  seems like a shallow wrapper around `_convert_to_decoder_transforms`, and it duplicates both function signature (with types) and the docstring. It might be simpler to just inline it either in `VideoDecoder.__init__`, or in `_convert_to_decoder_transforms` ?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -482,21 +488,39 @@ def _convert_to_decoder_transforms(
...
+                transform_tc = Resize._from_torchvision(transform)
+                input_dims = transform_tc._get_output_dims(input_dims)
+                converted_transforms.append(transform_tc)
+            elif isinstance(transform, v2.RandomCrop):
+                if dimension_order != "NCHW":
+                    raise ValueError(
+                        "TorchVision v2 RandomCrop is only supported for NCHW "
+                        "dimension order. Please use the TorchCodec RandomCrop "
+                        "transform instead."
+                    )
```
NicolasHug: Is this validation actually needed? I understand TV only accepts NCHW, but I don't understand yet how that restricts what we should allow. Isn't the actual filtergraph transformation always applied on the HWC AVframe anyway?

I'm happy to be more conservative, just checking my understanding. If this is more of a design decision than a technical limitation, maybe a short comment would help.

src/torchcodec/decoders/_video_decoder.py
NicolasHug: I was confused as well, I didn't realize the returned `input_dims` would be used by the next transform in the `for` loop.  A comment would definitely help!

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -66,28 +72,167 @@ class Resize(DecoderTransform):
...
+            )
+        if width is None:
+            raise ValueError(
+                "Video metadata has no width. RandomCrop can only be used when input frame dimensions are known."
+            )
+
+        # Note that TorchVision v2 transforms only accept NCHW tensors.
+        params = tv_random_crop.make_params(
+            torch.empty(size=(3, height, width), dtype=torch.uint8)
+        )
```
NicolasHug: That's OK for now, but creating a dummy input tensor is not ideal. we could consider exposing a private torchvision API that simply takes the input shape as parameter, rather than the input itself. Ultimately its shape is all that's needed. Maybe leave a TODO?

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -66,28 +72,167 @@ class Resize(DecoderTransform):
...
+
+    Args:
+        size: (sequence of int): Desired output size. Must be a sequence of
+            the form (height, width).
+    """
+
+    size: Sequence[int]
+    _top: Optional[int] = None
+    _left: Optional[int] = None
+    _input_dims: Optional[Tuple[int, int]] = None
```
NicolasHug: crazy thought, but this isn't thread-safe, is it?

If I have two threads, each thread has its own `VideoDecoder(..., transforms=[crop])` where `crop` is the *same* `RandomCrop` instance, and each video has different dimensions... bad things happen?

We probably don't care... right now? I wonder how many of the TV  transforms are also not thread-safe. This is almost never a problem in practice I assume, because users create new transforms instance all the time instead of re-using the same object.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -66,28 +72,167 @@ class Resize(DecoderTransform):
...
+        if height is None:
+            raise ValueError(
+                "Video metadata has no height. RandomCrop can only be used when input frame dimensions are known."
+            )
+        if width is None:
+            raise ValueError(
+                "Video metadata has no width. RandomCrop can only be used when input frame dimensions are known."
+            )
+
+        self._input_dims = (height, width)
```
NicolasHug: I was surprised to find that we're setting something in what I thought was a simple "getter".

Generally I have a hard time reasoning about when each attribute is set. Part of that is that we have to support both native transforms and TV transforms, but part of that is also probably due to the `dataclass` design. I suspect addressing the refactor will help a bit?

src/torchcodec/transforms/_decoder_transforms.py
NicolasHug: I think an error is best, it has a documentation purpose for the code reader.

test/test_transform_ops.py
```diff
@@ -145,6 +145,131 @@ def test_resize_fails(self):
...
+        height_scaling_factor,
+        width_scaling_factor,
+        video,
+    ):
+        height = int(video.get_height() * height_scaling_factor)
+        width = int(video.get_width() * width_scaling_factor)
+
+        # We want both kinds of RandomCrop objects to get arrive at the same
+        # locations to crop, so we need to make sure they get the same random
+        # seed.
```
NicolasHug: Can I be paranoid and suggest that we test other seeds e.g.?

```py
@pytest.mark.parametrize("seed", (0, 1234))
```

test/test_transform_ops.py
```diff
@@ -145,6 +145,131 @@ def test_resize_fails(self):
...
+            int(num_frames * 0.3),
+            int(num_frames * 0.4),
+            int(num_frames * 0.5),
+            int(num_frames * 0.75),
+            int(num_frames * 0.90),
+            num_frames - 1,
+        ]:
+            frame_random_crop = decoder_random_crop[frame_index]
+            frame_random_crop_tv = decoder_random_crop_tv[frame_index]
+            assert_frames_equal(frame_random_crop, frame_random_crop_tv)
```
NicolasHug, line 195: I forgot to mention it earlier but almost everywhere in this file we might want to use `torch.testing.assert_close` instead of our own `assert_frames_equal`, because `assert_frames_equal` is less strict on other platforms like GPU (not supported yet by transforms) or Mac, Windows, and potentially soon aarch64.

But since we want bitwise equality against TV, we should probable use `torch.testing.assert_close`. It ensures strict bitwise equality on uint8.

(this can be a TODO)

src/torchcodec/transforms/_decoder_transforms.py
NicolasHug: The point above about thread-safety made me think about: what happens if I call 2 `VideoDecoder` on 2 different videos with the same transform instance?

TV transforms are stateless and handle that correctly: we can use the same instance on different videos with different resolutions. But `TC`'s `RandomCrop`  seem to be stateful, and applying it to the first video affects its behavior on the second video.

```py
    @pytest.mark.parametrize("t", ("TC", "TV"))
    @pytest.mark.parametrize("seed", range(50))
    def test_random_crop_same_instance(self, t, seed):
        torch.manual_seed(seed)

        if t == "TC":
            t = torchcodec.transforms.RandomCrop(size=(100, 100))
        else:
            t = v2.RandomCrop(size=(100, 100))

        assert VideoDecoder(TEST_SRC_2_720P.path, transforms=[t])[0].shape == (3, 100, 100)
        assert VideoDecoder(NASA_VIDEO.path, transforms=[t])[0].shape == (3, 100, 100)
```

Gives

```
FAILED test/test_transform_ops.py::TestPublicVideoDecoderTransformOps::test_random_crop_same_instance[49-TC] - RuntimeError: Crop x position out of bounds

==================================================================== 47 failed, 53 passed, 49 deselected in 2.99s ====================================================================
```

All failed are from TC. TC succeeds 3 times, by chance.

I think what's happening is that with the TC transform, there's no random sampling for the second video, so it tries to crop the second smaller video at the previous larger video's crop dimensions?



Removing this condition fixes it (but I haven't checked that it doesn't lead to more issues)

```
if self._top is None or self._left is None:  # remove this
```

src/torchcodec/transforms/_decoder_transforms.py
scotts: @NicolasHug, yup agreed. However: we may not need to ask the TorchVision transforms for their parameters at all, if we just always rely on the logic we have in TorchCodec. I think that's a lot simpler, and probably what we should do. I'm going to make that change and put the PR up for review again.

src/torchcodec/transforms/_decoder_transforms.py
scotts: Yeah, this will definitely get into problems if folks reuse their objects. , which, again, I think we can fix by just not calling the TorchVision transforms.

src/torchcodec/transforms/_decoder_transforms.py
scotts: @NicolasHug, yes, the refactor should help, as well as not calling the TorchVision objects. We'll still do this business with the dimensions, but we can name it better.

src/torchcodec/decoders/_video_decoder.py
scotts: @NicolasHug, yes, this validation _was_ needed. We were calling `make_params()` on the TorchVision object, and it would fail with the wrong dimensions. We no longer do that, so now it's no longer needed.

Note that I deleted my previous comment because I mis-remembered my own code. Being away for a week does that. :)

src/torchcodec/transforms/_decoder_transforms.py
scotts: Resolving because in the new code, the TorchCodec transforms are stateless, except for testing. I'll point this out in a separate thread.

src/torchcodec/transforms/_decoder_transforms.py
scotts: The business with the dimensions is better: no more internal dimension state.

test/test_transform_ops.py
```diff
@@ -145,6 +145,118 @@ def test_resize_fails(self):
...
+            assert frame_random_crop_tv.shape == expected_shape
+
+            frame_full = decoder_full[frame_index]
+            frame_tv = v2.functional.crop(
+                frame_full,
+                top=tc_random_crop._top,
+                left=tc_random_crop._left,
+                height=tc_random_crop.size[0],
+                width=tc_random_crop.size[1],
+            )
```
scotts: [Accessing RandomCrop's internal state]

We want to know what the random values that `RandomCrop` used so we can do the exact same logic with the TorchVision functional, but the actual values are stored deep in the C++ layer in `SingleStreamDecoder`'s copy of the transform objects. Just storing the values in the `DecoderTransform` object seems simpler with less consequences.

I'm happy to consider cleaner solutions here, but I can't think of one. Because we only access `_top` and `_left` during testing, the TorchCodec version of `RandomCrop` is still stateless in the ways that matter, same as the TorchVision transform.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -65,29 +71,144 @@ class Resize(DecoderTransform):
...
+            the form (height, width).
+    """
+
+    size: Sequence[int]
+
+    # Note that these values are never read by this object or the decoder. We
+    # record them for testing purposes only.
+    _top: Optional[int] = None
+    _left: Optional[int] = None
+
```
scotts: For where these values are used, see my comment with the text [Accessing RandomCrop's internal state].

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -65,29 +73,160 @@ class Resize(DecoderTransform):
...
+                "Video metadata has no height. "
+                "RandomCrop can only be used when input frame dimensions are known."
+            )
+        if width is None:
+            raise ValueError(
+                "Video metadata has no width. "
+                "RandomCrop can only be used when input frame dimensions are known."
+            )
+
+        return (self.size[0], self.size[1])
```
NicolasHug, line 201: There are a few things I do not completely understand with `_calculate_output_dims` above. They may be resolved when this becomes an actual class instead of a dataclass, but I think it's worth clarifying now:

- we have `assert len(self.size) == 2` while in `_make_transform_spec()`, which is called later, we have a more user-facing assertion. This seems duplicated, and maybe the nice user-facing assertion should be here in `_calculate_output_dims` ?
- The validation of `input_dims`:
   -  is duplicated here and in `_make_transform_spec`
   - doesn't seem to be needed here, since we're not using `input_dims` to compute anything - only for validation (but that validation can and does happen later in `_make_transform_spec`

The last bullet point makes me wonder: do we actually need to pass `input_dims` to all the `_calculate_output_dims()` methods? None of the existing transforms see to rely on it to calculate their output dimensions: for both resize and crop, the output dimensions are known statically at construction time. Maybe there's an opportunity to significantly simplify this PR?

scotts: 1. Assertions on `self.size` are a placeholder as there will eventually be only one in the constructor once we move away from being a dataclass.
2. Agreed on all of that, I've greatly simplified things and added a test which ensures we do the right thing with a simple pipeline of operators.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -448,76 +451,101 @@ def _get_and_validate_stream_metadata(
...
+    #
+    # The consequence of the above facts is that we need to statically track
+    # frame dimensions in the pipeline while we pre-process it. The input
+    # frame's dimensions to A, our first transform, is always what we know from
+    # our metadata. For each transform, we always calculate its output
+    # dimensions from its input dimensions. We store these with the converted
+    # transform, to be all used together when we generate the specs.
+    converted_transforms: list[
+        Tuple[DecoderTransform, Tuple[Optional[int], Optional[int]]]
+    ] = []
```
NicolasHug, line 522: Nit: I think it'd help to indicate that `Tuple[Optional[int], Optional[int]]` is for Height and Width.

NicolasHug, line 522: Not a nit this time :D

They can't / shouldn't be Optional at this stage, can they? I think we can and should be able to make all dims non-Optional? We should probably make `_get_output_dims()` output type non-Optional too?


EDIT: I understand now this is due to:

```py
    # Transforms that change the dimensions of their input frame return a value.
    # Transforms that don't return None; they can rely on this default
    # implementation.
    def _get_output_dims(self) -> Optional[Tuple[Optional[int], Optional[int]]]:
        return None
```

But we don't have such transform currently. We will eventually have such transform (e.g. flip). But I think a better option will be then to just pass the input dims to `_get_output_dims()`, or better yet to the constructor?

So I still think we can simplify and remove the `Optional` type in this PR?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -448,76 +451,101 @@ def _get_and_validate_stream_metadata(
...
-                   only mentions its base, nn.Module. We don't want to take a
-                   hard dependency on TorchVision.
-
-    Returns:
-        String of transforms in the format the core API expects: transform
-        specifications separate by semicolons.
-    """
-    if transforms is None:
-        return ""
+        curr_input_dims = output_dims if output_dims is not None else curr_input_dims
```
NicolasHug, line 545: This is related to my comment above, I don't think `output_dims` being `None` is valid? This could be:

```
curr_input_dims = output_dims
```

The tests pass. The type-checker fails, but I think that's more because the types above, and those for `_get_output_dims()` don't need to be `Optional`?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -448,76 +451,101 @@ def _get_and_validate_stream_metadata(
...
 
-    transforms = _convert_to_decoder_transforms(transforms)
-    return ";".join([t._make_transform_spec() for t in transforms])
+    return ";".join([t._make_transform_spec(dims) for t, dims in converted_transforms])
```
NicolasHug, line 547: This entire loop above, can I please suggest we refactor it a bit to avoid the code duplication in all of the individual transforms `if` blocks? I would find it a lot easier to understand, as it's clearer that we are applying the *exact same logic* to each transform. It's currently not obvious. Something like this:

```py
    curr_input_dims = input_dims
    for transform in transforms:
        if not isinstance(transform, DecoderTransform):
            if not tv_available:
                raise ValueError(
                    f"The supplied transform, {transform}, is not a TorchCodec "
                    " DecoderTransform. TorchCodec also accepts TorchVision "
                    "v2 transforms, but TorchVision is not installed."
                )
            elif isinstance(transform, v2.Resize):
                transform = Resize._from_torchvision(transform)
            elif isinstance(transform, v2.RandomCrop):
                transform = RandomCrop._from_torchvision(transform)
            else:
                raise ValueError(
                    f"Unsupported transform: {transform}. Transforms must be "
                    "either a TorchCodec DecoderTransform or a TorchVision "
                    "v2 transform."
                )

        output_dims = transform._get_output_dims()
        converted_transforms.append((transform, curr_input_dims))
        curr_input_dims = output_dims

    return ";".join([t._make_transform_spec(dims) for t, dims in converted_transforms])
```

It re-defines the `transform` object - that's OK, that's Python :D

test/test_transform_ops.py
NicolasHug: We have two alternative options:

1. Create a seeded v2.RandomCrop() object and apply it to the `full_frame`, in each iteration of the loop. This makes sure the RNG is consistently the same for all frames.

```py
        for frame_index in ...:
           # ...
            torch.manual_seed(seed)
            frame_tv = v2.RandomCrop(size=(height, width))(decoder_full[frame_index])
            assert_frames_equal(frame_random_crop, frame_tv)
```

2. Sa

```py

        torch.manual_seed(seed)
        tv_crop = v2.RandomCrop(size=(height, width))
        params = tv_crop.make_params(torch.empty(3, video.get_height(), video.get_width()))

        for frame_index in ...:
        # ...
            frame_full = decoder_full[frame_index]
            frame_tv = v2.functional.crop(
                frame_full,
                top=params["top"],
                left=params["left"],
                height=params["height"],
                width=params["width"],
            )
            assert_frames_equal(frame_random_crop, frame_tv)
```

I far prefer the first one as it's simpler. The second one is kinda replicating the internal implementation of a TV transform.


-----

Both suggestions above are stricter tests than the current one. The current one is merely testing that TV's `crop` will crop at the same place as TC's crop, when provided the same crop parameters (`_top` and `_left`).
But because `_top_` and `_left` come from TC, this test isn't validating that the RNG of TC is the same as the RNG of TV! Both options above ensure that.

test/test_transform_ops.py
```diff
@@ -145,6 +145,163 @@ def test_resize_fails(self):
...
+        # Create a spec which causes us to calculate the random crop location.
+        _ = random_crop._make_transform_spec((1000, 1000))
+        first_top = random_crop._top
+        first_left = random_crop._left
+
+        # Create a spec again, which should calculate a different random crop
+        # location.
+        _ = random_crop._make_transform_spec((1000, 1000))
+        assert first_top != random_crop._top
+        assert first_left != random_crop._left
```
NicolasHug: I think we'll be able to remove `_top` and `_left`, so we may need to re-implement this test in another way.

Separately, I think we will also want to make sure that changing the input dims changes the output behavior. The current test passes `(1000, 1000)` twice.

src/torchcodec/decoders/_video_decoder.py
scotts: No, this isn't a type-checker thing; `output_dims` being `None` is valid and expected. See how `_get_output_dims()` is implemented on the base class now. This matches the behavior of the complementary C++ code, for what it's worth.

src/torchcodec/decoders/_video_decoder.py
scotts: > But I think a better option will be then to just pass the input dims to _get_output_dims(), or better yet to the constructor?

We had a discussion yesterday about how it's best to remove `input_dims` from `_get_output_dims()`. :) I think we can do the constructor, but that means I'll have to pull the changes from #1081 into this PR. It's actually not that much code change, so that's okay.

scotts: @NicolasHug, what do you think about defining dataclasses:
```python
@dataclass
# Explain why they are optional
class MetadataDimension:
    height: Optional[int]
    width: Optional[int]

@dataclass
class Dimensions:
    height: int
    width: int
```
Rather than using tuples everywhere.

NicolasHug: My honest opinion is that they're both overkill. I think we can get-by with non-optional tuples everywhere.

The risk with that is that we try to frame every single use of `H` and `W` with the `Dimensions` dataclass, including in places where they don't need to be. Like in public interfaces.

NicolasHug: Yeah, sorry - to be clear I'm not suggesting we go back to adding `input_dims` to `_get_output_dims()` in this PR. We might have to do it later when we implement more transforms, but at that time it'll be easier to figure out what the most sensible design is.

What I am trying to apply here YAGNI. In this PR, we don't seem to need to pass `input_dims` as input to `_get_output_dims()`, and we don't seem to need for `height` and `width` to be Optional either.

scotts: Scratch the above: we can't put `input_dims` in the constructor because users construct these objects. So then we're back to figuring out which we think is the least-ugly:

1. Passing `input_dims` to `_get_output_dims()`, which will be unused by transforms that themselves determine the output dimensions (Resize, Crop, etc).
2. A default implementation of `_get_output_dims()` which returns `None` when it makes no changes to the input frame dimensions; that's currently implemented.

I have a slight preference for 2, as that's how things work in the C++ code, and it does reduce the amount of passing around variables that don't get used.

scotts: From an offline discussion: one of my goals here is to establish the machinery so that subsequent transforms don't require refactors. I believe (but may be wrong) that these interfaces allow new transforms without changes to this code in `VideoDecoder` or changes to the `DecoderTransform` base class. Yes, we don't need these interfaces for this PR, but I'm confident we'll need it soon.

src/torchcodec/transforms/_decoder_transforms.py
```diff
@@ -37,9 +38,48 @@ class DecoderTransform(ABC):
...
+                that don't will ignore it. The individual values in the tuple are
+                optional because the original values come from file metadata which
+                may be missing. We maintain the optionality throughout the APIs so
+                that we can decide as late as possible that it's necessary for the
+                values to exist. That is, if the values are missing from the
+                metadata and we have transforms which ignore the input dimensions,
+                we want that to still work.
+
+                Note: This method is the moral equivalent of TorchVision's
+                `Transformer.make_params()`.
```
NicolasHug: Nit, transformers are a different thing


```suggestion
                `Transform.make_params()`.
```

PR #1074, opened 2025-11-26
Benchmark encoding against ffmpeg cli 

Dan-Flores:
This PR creates a benchmark to compare VideoEncoder against FFmpeg CLI. These tools aren't one-to-one, so some assumptions are made:

For VideoEncoder, we use this simple workflow:
```
encoder = VideoEncoder(frames=frames, frame_rate=30)
encoder.to_file(dest=output_path, codec="h264_nvenc", extra_options={"qp": 1})
```

For FFmpeg CLI, we count the time used to write frames from a tensor to a file if the flag is used: `--write_frames`
```
if write_frames:
	raw_frames = frames.permute(0, 2, 3, 1).contiguous()[:num_frames]
    	with open(raw_path, "wb") as f:
        	f.write(raw_frames.cpu().numpy().tobytes())
        
ffmpeg_cmd = [...]
subprocess.run(ffmpeg_cmd, check=True, capture_output=True)

```

### Result Summary:
* `VideoEncoder` shows better performance on GPU + CPU.
	* When the time required to write frames to bytes is added, FFmpeg CLI is much slower.
* On GPU, `VideoEncoder` shows a significant speed improvement, up to 3.5x faster than FFmpeg CLI for decoding 30 frames, without adding the time required to write frames to bytes.
	* NVENC utilization is higher for `VideoEncoder`, while Median GPU memory used values are the same.
* On CPU, FFmpeg CLI  has a slight edge without adding the time required to write frames to bytes. Otherwise, `VideoEncoder` is significantly faster.
  * I suspect there are optimizations we could make in `VideoEncoder::encode` to close the gap, but lets land this benchmark as is.
<details>
All benchmarks are run using a 1280x720 video:
Command to generate video: `ffmpeg -f lavfi -i testsrc2=duration=600:size=1280x720:rate=30 -c:v libx264 -pix_fmt yuv420p test/resources/testsrc2_10min.mp4`

#### $ `python benchmarks/encoders/benchmark_encoders.py`
```
Benchmarking 390 frames from nasa_13013.mp4 over 30 runs:
Decoded 390 frames of size 270x480

VideoEncoder on GPU   med = 119.26 ms, max = 122.06 ms, fps = 3270.1
GPU memory used:      med = 1231.0 MB, max = 1231.0 MB
NVENC utilization:    med = 30.0%,     max = 38.0%

FFmpeg CLI on GPU     med = 1174.55 ms, max = 1524.59 ms, fps = 332.0
GPU memory used:      med = 1231.0 MB, max = 1231.0 MB
NVENC utilization:    med = 15.0%,     max = 22.0%

VideoEncoder on CPU   med = 408.43 ms, max = 454.66 ms, fps = 954.9

FFmpeg CLI on CPU     med = 1184.47 ms, max = 1219.28 ms, fps = 329.3
```
#### $ `python benchmarks/encoders/benchmark_encoders.py --skip-write-frames`
```
Benchmarking 390 frames from nasa_13013.mp4 over 30 runs:
Decoded 390 frames of size 270x480

VideoEncoder on GPU   med = 120.21 ms, max = 122.40 ms, fps = 3244.4
GPU memory used:      med = 1231.0 MB, max = 1231.0 MB
NVENC utilization:    med = 26.0%,     max = 39.0%

FFmpeg CLI on GPU     med = 419.66 ms, max = 1189.17 ms, fps = 929.3
GPU memory used:      med = 1231.0 MB, max = 1231.0 MB
NVENC utilization:    med = 18.0%,     max = 23.0%

VideoEncoder on CPU   med = 408.86 ms, max = 449.01 ms, fps = 953.9

FFmpeg CLI on CPU     med = 383.65 ms, max = 410.91 ms, fps = 1016.5
```
</details>

### DIFF
--- a/benchmarks/encoders/benchmark_encoders.py
+++ b/benchmarks/encoders/benchmark_encoders.py
@@ -161,7 +161,8 @@ def main():
     temp_dir = Path(tempfile.mkdtemp())
     raw_frames_path = temp_dir / "input_frames.raw"
 
-    # By default, frames will be written inside the benchmark function
+    # If skip_write_frames is True, we will not benchmark the time it takes to write the frames.
+    # Here, we still write the frames for FFmpeg to use!
     if args.skip_write_frames:
         write_raw_frames(frames, str(raw_frames_path))
 


benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+        print(
+            f"NVENC utilization:    median = {util_median:.1f}%, max = {util_max:.1f}%"
+        )
+
+
+def encode_torchcodec(frames, output_path, device="cpu"):
+    encoder = VideoEncoder(frames=frames, frame_rate=30)
+    if device == "cuda":
+        encoder.to_file(dest=output_path, codec="h264_nvenc", extra_options={"qp": 1})
```
mollyxu: Are we currently using `qp=1` for torchcodec encoder vs `qp=0` for ffmpeg cli? (line 155)

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        "--write-frames",
+        action="store_true",
+        help="Include raw frame writing time in FFmpeg CLI benchmarks for fairer comparison with tensor-based workflows",
+    )
+    args = parser.parse_args()
+
+    print(
+        f"Benchmarking up to {args.max_frames} frames from {Path(args.path).name} over {args.average_over} runs:"
+    )
+    cuda_available = torch.cuda.is_available()
```
mollyxu, line 149: should we also skip in the case where `nvenc` is not supported on the gpu (ie. a100)? Or it shouldn't really matter since this is just for benchmarking

mollyxu:
Great work on the benchmarks @Dan-Flores! I liked the detailed analysis of the results. I left two clarifying questions.

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: Yes, and we should not be, thanks for catching this!

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: For this benchmark, I believe its acceptable to only check for `cuda` availability. 
I'm not sure how we could generically check for `nvenc` support, besides trying to use it in FFmpeg and catching the error, but that is outside of the scope of this benchmark.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        "--max-frames",
+        type=int,
+        default=10,
+        help="Maximum number of frames to decode for benchmarking",
+    )
+    parser.add_argument(
+        "--write-frames",
+        action="store_true",
+        help="Include raw frame writing time in FFmpeg CLI benchmarks for fairer comparison with tensor-based workflows",
+    )
```
NicolasHug, line 144: I think the default behavior of the benchmark script should be the "fair" comparison with FFmpeg, that is, benchmarking the time it takes to write raw frame bytes. Because that's what users would have to do in the wild.
Benchmarking without accounting for the time it takes to write the frames is still valuable for us, so have an idea of the efficiency of the core decoding loop, but that can be left as the non-default behavior.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        samples = []
+        for line in stdout.strip().split("\n"):
+            if line.strip():
+                res = [float(x.strip()) for x in line.split(",")]
+                samples.append({"utilization": res[0], "memory_used": res[1]})
+
+        self.metrics = {
+            "utilization": [s["utilization"] for s in samples],
+            "memory_used": [s["memory_used"] for s in samples],
+        }
```
NicolasHug: On `NVENCMonitor` above, I think we might want to use `pynvml` instead, as done e.g. in P1984513849.

The main reason is that `NVENCMonitor` is sampling utilization value every 50ms, which isn't exactly in sync with the number of iterations in the loop. That is, the returned `nvenc_tensor` doesn't contain the same amount of values the `times` tensor, and so their reported values aren't averaged over the same amount of experiments either.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        "s": 1e-9,
+    }[unit]
+    unit_times = times * mul
+    std = unit_times.std().item()
+    med = unit_times.median().item()
+    mean = unit_times.mean().item()
+    min = unit_times.min().item()
+    max = unit_times.max().item()
+    print(
+        f"\n{prefix}   {med = :.2f}, {mean = :.2f} +- {std:.2f}, {min = :.2f}, {max = :.2f} - in {unit}, fps = {fps:.1f}"
```
NicolasHug: Nit, no strong opinion but I find that reporting the `mean` isn't super useful when the std is small enough. Same for min and max (one is enough). It makes the logs slightly easier to read.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+    )
+
+    if nvenc_metrics is not None:
+        mem_used_max = nvenc_metrics["memory_used"].max().item()
+        mem_used_median = nvenc_metrics["memory_used"].median().item()
+        util_max = nvenc_metrics["utilization"].max().item()
+        # For median utilization, only consider non-zero samples as NVENC is often idle
+        util_nonzero = nvenc_metrics["utilization"][nvenc_metrics["utilization"] > 0]
+        util_median = util_nonzero.median().item() if len(util_nonzero) > 0 else 0.0
```
NicolasHug: I think we'll want the complete view an not exclude the zero values. That the utilization is sometimes zero is actually relevant information!

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+
+    print(
+        f"Benchmarking up to {args.max_frames} frames from {Path(args.path).name} over {args.average_over} runs:"
+    )
+    cuda_available = torch.cuda.is_available()
+    if not cuda_available:
+        print("CUDA not available. GPU benchmarks will be skipped.")
+
+    decoder = VideoDecoder(str(args.path))
+    valid_max_frames = min(args.max_frames, len(decoder))
```
NicolasHug: Do we need this? I think we should be able to remove the corresponding parameter in `write_raw_frames` and just use `len(frames)`?

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+    encoder = VideoEncoder(frames=frames, frame_rate=30)
+    if device == "cuda":
+        encoder.to_file(dest=output_path, codec="h264_nvenc", extra_options={"qp": 0})
+    else:
+        encoder.to_file(dest=output_path, codec="libx264", crf=0)
+
+
+def write_raw_frames(frames, num_frames, raw_path):
+    # Convert NCHW to NHWC for raw video format
+    raw_frames = frames.permute(0, 2, 3, 1).contiguous()[:num_frames]
```
NicolasHug: The call to `contiguous()` should not be needed.

Neither should `[:num_frames]`, but I might be missing something

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+    with open(raw_path, "wb") as f:
+        f.write(raw_frames.cpu().numpy().tobytes())
+
+
+def write_and_encode_ffmpeg_cli(
```
NicolasHug: Nit: Since there is a `write_frames` parameter (whose default should be True), this function doesn't always writes. So it can be named `encode_ffmpeg_cli()` instead.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        "ffmpeg",
+        "-y",
+        "-f",
+        "rawvideo",
+        "-pix_fmt",
+        "rgb24",
+        "-s",
+        f"{frames.shape[3]}x{frames.shape[2]}",
+        "-r",
+        "30",  # frame_rate is 30
```
NicolasHug: define FRAME_RATE as a global var and use it here and in `encode_torchcodec`

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+    parser.add_argument(
+        "--average-over",
+        type=int,
+        default=30,
+        help="Number of runs to average over",
+    )
+    parser.add_argument(
+        "--max-frames",
+        type=int,
+        default=10,
```
NicolasHug: I think we'll want the default to be None and just encode the entire video. Having this parameter is useful for quick benchmarking but I suspect results will mostly be useful when benchmarking across longer sequences of frames, so not having a  default is probably best.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+        write_frames=args.write_frames,
+        average_over=args.average_over,
+    )
+    prefix = "FFmpeg CLI on CPU  "
+    report_stats(times, frames.shape[0], prefix=prefix)
+
+    try:
+        shutil.rmtree(temp_dir, ignore_errors=True)
+    except Exception:
+        pass
```
NicolasHug: We shouldn't need to try/except this, can you help me understand why it's needed?

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+            output_path=str(gpu_output),
+            device="cuda",
+            gpu_monitoring=True,
+            average_over=args.average_over,
+        )
+        report_stats(
+            times, frames.shape[0], nvenc_metrics, prefix="VideoEncoder on GPU"
+        )
+    else:
+        print("Skipping VideoEncoder GPU benchmark (CUDA not available)")
```
NicolasHug: Nit, something similar is already printed above. We only need one

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,275 @@
...
+            output_path=str(ffmpeg_gpu_output),
+            device="cuda",
+            gpu_monitoring=True,
+            write_frames=args.write_frames,
+            average_over=args.average_over,
+        )
+        prefix = "FFmpeg CLI on GPU  "
+        report_stats(times, frames.shape[0], nvenc_metrics, prefix=prefix)
+    else:
+        print("Skipping FFmpeg CLI GPU benchmark (CUDA not available)")
```
NicolasHug: Same here. We should only need a single `if cuda_available:` block and benchmark both torchcodec and the ffmpeg cli in there.

NicolasHug (commented):
Thanks @Dan-Flores , this looks good!

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: Thanks for the example, I'll update to use `pynvml`.

I see how arbitrarily selecting 50ms will not get the same amount of values as the `times` tensor, but I don't completely understand how `pynvml.nvmlDeviceGetDecoderUtilization` manages it. It seems like it is always sampling the device for usage, then when called returns a single median/max/average over an automatically determined sampling period when called?

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: The reason I skipped any zero value was that for shorter benchmarks with the previous NVENC Monitor approach, the NVENC was not active most of the time, so the median was often zero. It seems the `pynvml` library is smarter about sampling frequency, so this is not a problem anymore!

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: Good catch, I copied the `contiguous()` call from a test and is not needed.

I see your point about `[:num_frames]` and `valid_max_frames`, we can just decode the requested the number of frames once, then stop passing around this parameter.

benchmarks/encoders/benchmark_encoders.py
Dan-Flores: It is not with `ignore_errors=True`, I'll remove this.

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,226 @@
...
+    print(
+        f"Decoded {frames.shape[0]} frames of size {frames.shape[2]}x{frames.shape[3]}"
+    )
+
+    temp_dir = Path(tempfile.mkdtemp())
+    raw_frames_path = temp_dir / "input_frames.raw"
+
+    # By default, frames will be written inside the benchmark function
+    if args.skip_write_frames:
+        write_raw_frames(frames, str(raw_frames_path))
```
NicolasHug: Lol this makes sense but it's a bit surprising to read "if skip write frames, then write frames". Here's a suggestion below.


```suggestion
    # If skip_write_frames is True, then we don't benchmark the time it takes to write the frames.
    # But we still need to write them for FFmpeg to find them!
    if args.skip_write_frames:
        write_raw_frames(frames, str(raw_frames_path))
```

benchmarks/encoders/benchmark_encoders.py
```diff
@@ -0,0 +1,226 @@
...
+            mem_info = pynvml.nvmlDeviceGetMemoryInfo(handle)
+            mem_used = mem_info.used / (1_000_000)  # Convert bytes to MB
+            utilizations.append(util)
+            memory_usage.append(mem_used)
+
+    times_tensor = torch.tensor(times).float()
+    return times_tensor, {
+        "utilization": torch.tensor(utilizations).float() if gpu_monitoring else None,
+        "memory_used": torch.tensor(memory_usage).float() if gpu_monitoring else None,
+    }
```
NicolasHug, line 48: Just replying to https://github.com/meta-pytorch/torchcodec/pull/1074#discussion_r2669968606 here for visibility:

> It seems like it is always sampling the device for usage, then when called returns a single median/max/average over an automatically determined sampling period when called?

Yes, this is also my understanding. And I think this is also what was happening with your  previous implementation with `nvidia-smi`!

I think there are two main variables:

- The "query frequency", i.e. the frequency at which we call nvml/nvidia-smi. In your previous implementation, this was every 50ms. In the current implementation, it's every time we enter the `for _ in range(average_over)` loop.
- The "sampling period" over which nvml/nvidia-smi average and report their results. It's a *different* variable from the query frequency! And, as far as I can tell, we *do not* have control over this one. I can't find a relevant parameter for it, and claude says it's really determined by the underlying driver (which I'm inclined to believe)

PR #1076, opened 2025-11-28
Rename `getKeyFrameIndexForPts` into `getKeyFrameIdentifier`

NicolasHug:
See comment for an explanation why


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1092,13 +1092,6 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
   // Returns true if we can avoid seeking in the AVFormatContext based on
   // heuristics that rely on the target cursor_ and the last decoded frame.
   // Seeking is expensive, so we try to avoid it when possible.
-  // Note that this function itself isn't always that cheap to call: in
-  // particular the calls to getKeyFrameIndexForPts below in approximate mode
-  // are sometimes slow.
-  // TODO we should understand why (is it because it reads the file?) and
-  // potentially optimize it. E.g. we may not want to ever seek, or even *check*
-  // if we need to seek in some cases, like if we're going to decode 80% of the
-  // frames anyway.
```
NicolasHug, line 1101: Drive-by, my previous comment above is incorrect: `getKeyFrameIndexForPts()` in approximate mode is not slow. But it's not accurate (it doesn't return what we want it to return), so it's still the cause for the slowness. See https://github.com/meta-pytorch/torchcodec/issues/1077 for more details.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1365,7 +1358,19 @@ torch::Tensor SingleStreamDecoder::maybePermuteHWC2CHW(
...
+  // index that is returned depends on various factors:
+  // - If seek_mode is exact, we return the index of the key frame in the
+  //   scanned key-frame vector (streamInfo.keyFrames). So the returned value is
+  //   in [0, num_key_frames).
+  // - If seek_mode is approximate, we use av_index_search_timestamp() which
+  //   may return a value in [0, num_key_frames) like for mkv, but also a value
+  //   in [0, num_frames) like for mp4. It really depends on the container.
+  //
+  //  The range of the "identifier" doesn't matter that much, for now we only
+  //  use it to uniquely identify a key frame in canWeAvoidSeeking().
```
NicolasHug, line 1373: I can confirm the above on some test videos with 2 key frames. `mkv` would return 0 and 1 like in exact mode, while `mp4` returned 0 and 250!

PR #1079, opened 2025-11-28
Add sitemap config

NicolasHug:
We have very bad SEO on Google. This is an attempt at addressing that, following @svekars recomendations from https://fburl.com/workplace/ydutm35c


NicolasHug:
Docs CI jobs are happy, merging

PR #1081, opened 2025-12-01
Refactor decoder transforms

scotts:
My original thinking of the `DecoderTransform` object is that they would just be a holder of values with no methods. They're evolving to actually have methods, so let's just make them normal objects.


docs/source/api_ref_transforms.rst
```diff
@@ -14,4 +14,5 @@ For a tutorial, see: TODO_DECODER_TRANSFORMS_TUTORIAL.
     :template: dataclass.rst
 
     DecoderTransform
+    RandomCrop
```
scotts, line 17: Drive-by: I forgot to update this in previous PR.

NicolasHug (approved):

PR #1083, opened 2025-12-02
Add CPU tests for aarch64 GPU wheels

NicolasHug:
Yes you read that right: CPU tests for GPU wheels. We can build the ARM GPU wheels but we can't test them (on GPU).

### DIFF
--- a/.github/workflows/cpp_tests.yaml
+++ b/.github/workflows/cpp_tests.yaml
@@ -1,7 +1,7 @@
 name: CPP tests
 
 on:
-  # push:
+  push:
     branches: [ main ]
   pull_request:
 
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -3,7 +3,7 @@ name: Docs
 on:
   push:
     branches: [ main ]
-  # pull_request:
+  pull_request:
 
 permissions:
   id-token: write
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -3,7 +3,7 @@ name: Lint
 on:
   push:
     branches: [ main ]
-  # pull_request:
+  pull_request:
 
 concurrency:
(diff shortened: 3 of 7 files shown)


.github/workflows/cpp_tests.yaml
```diff
@@ -1,7 +1,7 @@
 name: CPP tests
 
 on:
-  push:
+  # push:
```
NicolasHug: I'll remove those before merging. That's just to avoid triggering these other jobs as I was testing the new ones.

.github/workflows/linux_cuda_aarch64_wheel.yaml
```diff
@@ -56,3 +56,80 @@ jobs:
       architecture: aarch64
       build-platform: "python-build-package"
       build-command: "BUILD_AGAINST_ALL_FFMPEG_FROM_S3=1 ENABLE_CUDA=1 python -m build --wheel -vvv --no-isolation"
+
+  install-and-test:
+    runs-on: linux.arm64.2xlarge
+    container:
+      image: pytorch/manylinuxaarch64-builder:cuda12.6
+    env:
+      cuda_version_without_periods: "126"
```
Dan-Flores, line 65: This cuda version is specified outside of the `matrix`, as done in `linux_cuda_wheel`. Is that intentional? 
https://github.com/meta-pytorch/torchcodec/blob/957a150ee2d924a4a1a8346a554fa74e95fa1f53/.github/workflows/linux_cuda_wheel.yaml#L63-L72

NicolasHug: Good catch - yes that's intentional. In this new job we only test on a CUDA 12.6 machine, not a any other CUDA versions (see `image: pytorch/manylinuxaarch64-builder:cuda12.6`) above.

We don't need to test more CUDA versions because we can't run CUDA tests on ARM anyway (that's a `test-infra` limitation, from what I understand). And we still have to use a CUDA docker image because the wheels were built with CUDA support.

PR #1085, opened 2025-12-02
Remove transforms from public decoder API

Dan-Flores:
This PR removes transforms from the public `_video_decoder.py` API. To be certain, the `torchcodec.transforms` imports are removed.
* `test/test_transform_ops.py` is deleted
* The docs reference to `transforms` and its uses are deleted

The implementations in  https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/transforms/_decoder_transforms.py and corresponding `__init__.py` are not deleted, as they are private classes, and cannot be used in `_video_decoder.py` after these changes.

### DIFF
--- a/src/torchcodec/__init__.py
+++ b/src/torchcodec/__init__.py
@@ -9,7 +9,7 @@ from pathlib import Path
 # Note: usort wants to put Frame and FrameBatch after decoders and samplers,
 # but that results in circular import.
 from ._frame import AudioSamples, Frame, FrameBatch  # usort:skip # noqa
-from . import decoders, encoders, samplers, transforms  # noqa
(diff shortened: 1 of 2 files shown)


scotts:
I think we should still remove this entire file: https://github.com/meta-pytorch/torchcodec/blob/f2c8c69cd6cbc7eb29145cf32a26909ce45c0bd6/src/torchcodec/transforms/__init__.py#L7 If we don't, users will still be able to see `torchcodec.transforms` and the two classes it exposes. And then I think we need to remove `transforms` from the top-level init: https://github.com/meta-pytorch/torchcodec/blob/f2c8c69cd6cbc7eb29145cf32a26909ce45c0bd6/src/torchcodec/__init__.py#L12

PR #1093, opened 2025-12-03
Expose cpu_fallback in VideoDecoder API

mollyxu:
## Expose cpu_fallback in VideoDecoder API

Expose the `cpu_fallback` property in the public API to allow users to check whether the video decoder fell back to CPU decoding. #943 

The `FallbackInfo` class provides a clean interface for users to check decoder fallback status:
- `bool(fallback_info)` returns True if any fallback occurred
- `str(fallback_info)` provides a human-readable explanation of why fallback happened
- Tracks two types of fallback: NVcuvid not found and unsupported video formats

The `cpu_fallback` property on `VideoDecoder` returns a `FallbackInfo` instance that can be queried after decoding at least one frame. This helps users understand why GPU decoding may not be used and debug performance issues.

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -270,10 +270,13 @@ class VideoDecoder:
 
                 if "CPU fallback" in backend_details:
                     self._cpu_fallback._is_fallback = True
-                    if "NVCUVID not available" in backend_details:
-                        self._cpu_fallback._nvcuvid_unavailable = True
-                    elif self._cpu_fallback._backend == "Beta CUDA":
-                        self._cpu_fallback._video_not_supported = True
+                    if self._cpu_fallback._backend == "Beta CUDA":
+                        # Only the beta interface can provide details.
+                        # if it's not that nvcuvid is missing, it must be video-specific
+                        if "NVCUVID not available" in backend_details:
+                            self._cpu_fallback._nvcuvid_unavailable = True
+                        else:
+                            self._cpu_fallback._video_not_supported = True
 
         return self._cpu_fallback
 
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1766,7 +1766,6 @@ class TestVideoDecoder:
         if "beta" in device:
             # For beta interface, status is known immediately
             assert decoder.cpu_fallback.status_known
-            decoder.get_frame_at(0)
             assert decoder.cpu_fallback
             # Beta interface provides the specific reason for fallback
             assert "Video not supported" in str(decoder.cpu_fallback)


src/torchcodec/decoders/_video_decoder.py
```diff
@@ -180,13 +222,42 @@ def __init__(
             custom_frame_mappings=custom_frame_mappings_data,
         )
 
+        self._fallback_info = FallbackInfo()
+        self._has_decoded_frame = False
```
mollyxu: Instead of tracking whether a frame has been decoded, we could also just call `_update_cpu_fallback` inside every method that would decode a frame. The problem with that approach would be that we would be returning non tensors in the compiled methods. As such, we would have to do something like inside 
```
if torch.compiler.is_compiling():
    return
```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -19,7 +20,48 @@
     create_decoder,
     ERROR_REPORTING_INSTRUCTIONS,
 )
-from torchcodec.transforms import DecoderTransform, RandomCrop, Resize
+from torchcodec.transforms import DecoderTransform, Resize
+
+
+@dataclass
+class FallbackInfo:
```
NicolasHug: We could keep this class private for now, but I was thinking, maybe it could be public and named `CpuFallbackStatus`? I'm hoping that it might resolve both @scotts preference and mine:

The `CpuFallbackStatus` class will be public and visible from the docs, and it will be clear from the docstring that the `dec.cpu_fallback` attribute is an instance of this class. The attribute name `cpu_fallback` is fairly generic and allows us to extend the functionality in the future, while the class name itself communicates that it it's not just a simple `bool`, or a simple string.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -19,7 +20,48 @@
...
+        - Use ``bool(fallback_info)`` to check if any fallback occurred
+
+    Attributes:
+        status_known (bool): Whether the fallback status has been determined.
+    """
+
+    def __init__(self):
+        self.status_known = False
+        self.__nvcuvid_unavailable = False
+        self.__video_not_supported = False
```
NicolasHug: Using double underscores works, it's the more "hardcore" version to indicate that something is private. ([ref](https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name)). We don't use double underscores in TC, we've mostly just used single underscore up to now. For consistency, I think it's best to stick to our current practice of using single undercores.

EDIT: see my other comment below, the name mangling is a bit surprising, so let's definitely use single underscores

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -19,7 +20,48 @@
...
+            reasons.append("NVcuvid unavailable")
+        if self.__video_not_supported:
+            reasons.append("Video not supported")
+
+        if reasons:
+            return "Fallback status: Falling back due to: " + ", ".join(reasons)
+        return "Fallback status: No fallback required"
 
 
 class VideoDecoder:
```
NicolasHug, line 79: We'll need to update the docstring of the VideoDecder and add the new field to the `Attributes:` section

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -183,13 +222,42 @@ def __init__(
...
+
     def __len__(self) -> int:
         return self._num_frames
 
+    @property
+    def cpu_fallback(self) -> FallbackInfo:
+        # We can only determine whether fallback to CPU is happening when this
+        # property is accessed and requires that at least one frame has been decoded.
+        self._update_cpu_fallback()
+        return self._fallback_info
```
NicolasHug: 2 nits:
- No need to define a separate `_update_cpu_fallback()` function, it can just be inlined here
- The convention (I *think* it's a convention??) is to use the same name for the `@property` and for the underlying cached object. That is, I think `self._fallback_info` should just be `self._cpu_fallback`. It makes it more obvious that it relates to the `@cpu_fallback` property.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -183,13 +222,42 @@ def __init__(
...
+        has been decoded.
+        """
+        if not self._fallback_info.status_known and self._has_decoded_frame:
+            backend_details = core._get_backend_details(self._decoder)
+
+            self._fallback_info.status_known = True
+
+            if "CPU fallback" in backend_details:
+                if "NVCUVID not available" in backend_details:
+                    self._fallback_info._FallbackInfo__nvcuvid_unavailable = True
```
NicolasHug: Ah, so this `_FallbackInfo__nvcuvid_unavailable` is the name-mangling consequence of using double leading underscore. Let's definitely use single underscores :)

src/torchcodec/decoders/_video_decoder.py
NicolasHug: I think we should implement this `_has_decoded_frame` logic in C++ for 2 reasons:

- it's simpler and can be localized to a single place (see suggested implementation below)
- it's only relevant for the default interface, not for the beta interface.

So maybe all we need is to update the default interface's returned string, and let it indicate whether its status is known or unknown yet. Here:

https://github.com/meta-pytorch/torchcodec/blob/38fa96c6226ce502c0f86fffa664835059d35ce9/src/torchcodec/_core/CudaDeviceInterface.cpp#L357-L363


To know whether a frame has been decoded yet, I think we can simply set a boolean field to true when 

https://github.com/meta-pytorch/torchcodec/blob/38fa96c6226ce502c0f86fffa664835059d35ce9/src/torchcodec/_core/CudaDeviceInterface.cpp#L238

is called. This would be a new private boolean attribute on the `CudaDeviceInterface` class.

test/test_decoders.py
```diff
@@ -1737,6 +1737,92 @@ def test_set_cuda_backend(self):
...
+
+    @needs_cuda
+    def test_cpu_fallback_h265_video_beta_cuda(self):
+        """Test that H265 video triggers CPU fallback on Beta CUDA interface."""
+        with set_cuda_backend("beta"):
+            decoder = VideoDecoder(H265_VIDEO.path, device="cuda")
+
+        _ = decoder.get_frame_at(0)
+
+        assert decoder.cpu_fallback.status_known
```
NicolasHug: For the beta interface, we should be able to assert that `status_known` is true *before* we even decode any frame. I think you might need some slight modification to the implementation above in order to achieve that.

test/test_decoders.py
```diff
@@ -1737,6 +1737,92 @@ def test_set_cuda_backend(self):
...
+        _ = decoder.get_frame_at(0)
+        assert decoder.cpu_fallback.status_known
+        status_after_get_frame = str(decoder.cpu_fallback)
+
+        _ = decoder.get_frames_in_range(1, 3)
+        assert str(decoder.cpu_fallback) == status_after_get_frame
+
+        _ = decoder.get_frame_played_at(0.5)
+        assert str(decoder.cpu_fallback) == status_after_get_frame
+
```
NicolasHug: We have a bunch of existing calls to `_core._get_backend_details()` in this file, mostly in https://github.com/meta-pytorch/torchcodec/blob/38fa96c6226ce502c0f86fffa664835059d35ce9/test/test_decoders.py#L1698-L1738

It'd be great to update these calls to now rely on your much better API :)

This is only a suggestion, it can be done in a separate PR, or left-over as an easy TODO.

NicolasHug (commented):
Thanks for working on this @mollyxu ! Left a few comments and suggestions regarding the implementation, let me know if I can clarify anything.

test/test_decoders.py
mollyxu: In `test_set_cuda_backend` the calls to `_code._get_backend_details()` are checking the exact backend that is being used. My current API does not support this but I could add a `_backend` field in `CpuFallbackStatus` like the following
```    
def __init__(self):
        self.status_known = False
        self._nvcuvid_unavailable = False
        self._video_not_supported = False
        self._backend = ""
```
I tried implementing this in the updated PR. Let me know if this is what you had in mind

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -23,6 +24,51 @@
...
+    """Information about CPU fallback status.
+
+    This class tracks whether the decoder fell back to CPU decoding.
+
+    Usage:
+        - Use ``str(cpu_fallback_status)`` or ``print(cpu_fallback_status)`` to see the cpu fallback status
+        - Use ``bool(cpu_fallback_status)`` to check if any fallback occurred
+
+    Attributes:
+        status_known (bool): Whether the fallback status has been determined.
```
NicolasHug: I think it's OK to expose publicly. Let's just document that:

- for the Beta CUDA backend this is always known
- for the ffmpeg one, it's known after decoding the first frame.

We can link to this concept of CUDA backend by linking to https://meta-pytorch.org/torchcodec/stable/generated/torchcodec.decoders.set_cuda_backend.html#torchcodec.decoders.set_cuda_backend

test/test_decoders.py
```diff
@@ -1672,22 +1672,27 @@ def test_beta_cuda_interface_cpu_fallback(self):
...
-        )
+
+        # Before accessing any frames, status should be unknown
+        assert not ref_dec.cpu_fallback.status_known
+
+        ref_frame = ref_dec.get_frame_at(0)
+
+        assert "FFmpeg CUDA" in str(ref_dec.cpu_fallback)
+        assert ref_dec.cpu_fallback.status_known
+        assert bool(ref_dec.cpu_fallback)
```
NicolasHug: Here and everywhere else, we don't need to explicitly call `bool()`. The Pythonic way is to rely on the fact that `bool()` is essentially called automatically in contexts where it's needed (it's not necessarily how it's implemented in CPython but the idea is the same). This includes:

- `assert cond`
- `if cond`
- etc.

For example, to check whether a list is empty, the Pythonic way is to just check `if l: ...`. We don't do `if bool(l):`

test/test_decoders.py
```diff
@@ -1715,7 +1720,8 @@ def test_set_cuda_backend(self):
         # Check that the default is the ffmpeg backend
         assert _get_cuda_backend() == "ffmpeg"
         dec = VideoDecoder(H265_VIDEO.path, device="cuda")
-        assert _core._get_backend_details(dec._decoder).startswith("FFmpeg CUDA")
+        _ = dec.get_frame_at(0)
+        assert "FFmpeg CUDA" in str(dec.cpu_fallback)
```
NicolasHug: 2 comments:

- Syntax: we don't need to use `_ = ...`. Just call a raw `dec.get_frame_at(0)`
- We shouldn't need to decode a frame here, because we should already be able to tell that it's the FFmpeg backend. We don't know whether there's a fallback, but that's not needed info here.

test/test_decoders.py
NicolasHug: Yes this is great!

test/test_decoders.py
```diff
@@ -1737,6 +1743,69 @@ def test_set_cuda_backend(self):
                 with set_cuda_backend(backend):
                     VideoDecoder(H265_VIDEO.path, device=f"cuda:{bad_device_number}")
 
+    def test_cpu_fallback_no_fallback_on_cpu_device(self):
+        """Test that CPU device doesn't trigger fallback (it's not a fallback scenario)."""
+        decoder = VideoDecoder(NASA_VIDEO.path, device="cpu")
+
+        _ = decoder[0]
+
+        assert decoder.cpu_fallback.status_known
```
NicolasHug: In this scenario, we should be able to assert that the status is known before we decode any frame?

test/test_decoders.py
```diff
@@ -1737,6 +1743,69 @@ def test_set_cuda_backend(self):
...
+        # because its dimensions are too small
+        decoder = VideoDecoder(H265_VIDEO.path, device="cuda")
+
+        assert not decoder.cpu_fallback.status_known
+
+        _ = decoder.get_frame_at(0)
+
+        assert decoder.cpu_fallback.status_known
+        assert bool(decoder.cpu_fallback)
+        assert "Fallback status: Falling back due to:" in str(decoder.cpu_fallback)
```
NicolasHug: Let's assert that it's due to the video being not supported

test/test_decoders.py
NicolasHug: ping on this, we seem to not have tests for the beta interface now?

test/test_decoders.py
```diff
@@ -1737,6 +1743,69 @@ def test_set_cuda_backend(self):
...
+
+    @needs_cuda
+    def test_cpu_fallback_no_fallback_on_supported_video(self):
```
NicolasHug: let's parametrize this over both interfaces - the beta and ffpmeg ones

test/test_decoders.py
```diff
@@ -1737,6 +1743,69 @@ def test_set_cuda_backend(self):
...
+        """Test that supported videos don't trigger fallback on CUDA."""
+        decoder = VideoDecoder(NASA_VIDEO.path, device="cuda")
+
+        _ = decoder[0]
+
+        assert not bool(decoder.cpu_fallback)
+        assert "No fallback required" in str(decoder.cpu_fallback)
+
+    def test_cpu_fallback_status_cached(self):
```
NicolasHug: IIUC this test mostly tests that the output value doesn't change, not that it's cached. I think testing the cache behavior is potentially really difficult, and perhaps not needed after all. I'd suggest to remove it?

test/test_decoders.py
```diff
@@ -1737,6 +1743,69 @@ def test_set_cuda_backend(self):
...
+        first_status = str(decoder.cpu_fallback)
+        assert decoder.cpu_fallback.status_known
+
+        _ = decoder[1]
+        second_status = str(decoder.cpu_fallback)
+        assert decoder.cpu_fallback.status_known
+
+        assert first_status == second_status
+
+    def test_cpu_fallback_multiple_access_methods(self):
```
NicolasHug: I think this test is technically a subset of the previous one: if we 'cache' the `cpu_fallback` result, then a consequence is that calling different methods isn't going to change it.

Maybe want you wanted to check is that the status becomes "known" and that it works with multiple decoding method? If that's the case I think we'd need to re-create the VideoDecoder object in-between method calls. But TBH, I'm not sure it's a critical test to have, so I might suggest to remove it too.

src/torchcodec/decoders/_video_decoder.py
NicolasHug: Also, we'll probably want to document this class publicly in `docs/source/api_ref_decoders.rst`. We should indicate that users should never instantiate this class directly, and only accessed via the `VideoDecoder.cpu_fallback` attribute.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -184,9 +234,34 @@ def __init__(
...
 
+        self._cpu_fallback = CpuFallbackStatus()
+
     def __len__(self) -> int:
         return self._num_frames
 
+    @property
+    def cpu_fallback(self) -> CpuFallbackStatus:
+        # We can only determine whether fallback to CPU is happening when this
+        # property is accessed and requires that at least one frame has been decoded.
```
NicolasHug: I think this comment is slightly misleading because it's only really true for the ffmpeg interface. Can I suggest the following - and also please check me on my understanding here:

> We only query the CPU fallback info if status is unknown. That happens either when:
> - this @property has never been called before
> - no frame has been decoded yet on the FFmpeg interface.
> 
> Note that for the beta interface, we're able to know the fallback status right when the VideoDecoder
> is instantiated, but the `status_known` attribute is initialized to False.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -186,9 +240,41 @@ def __init__(
...
+            backend_details = core._get_backend_details(self._decoder)
+
+            if "status unknown" not in backend_details:
+                self._cpu_fallback.status_known = True
+
+                if "CPU fallback" in backend_details:
+                    if "NVCUVID not available" in backend_details:
+                        self._cpu_fallback._nvcuvid_unavailable = True
+                    else:
+                        self._cpu_fallback._video_not_supported = True
```
NicolasHug: So I think this logic is correct for the Beta interface i.e. if there's a fallabck and nvcuvid is available, then the reason for the fallback is video-specific.

But for the FFmpeg interface we actually don't know the reason, it could be either.

Sorry for not mentioning that earlier, I forgot that this was another way the interfaces differ!

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -25,6 +26,55 @@
...
+        """Returns a human-readable string representation of the cpu fallback status."""
+        if not self.status_known:
+            return f"[{self._backend}] Fallback status: Unknown"
+
+        reasons = []
+        if self._nvcuvid_unavailable:
+            reasons.append("NVcuvid unavailable")
+        if self._video_not_supported:
+            reasons.append("Video not supported")
+
```
NicolasHug: Following up on my other comment below, I think we'll want this to be something like that. Let's also make that part of the test and we should be good to go!

```suggestion
        elif self._video_not_supported:
            reasons.append("Video not supported")
        else:
            reasons.append("Unknown reason - try the Beta interface to know more!")


```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -25,6 +26,56 @@
...
+    This class tracks whether the decoder fell back to CPU decoding.
+    Users should not instantiate this class directly; instead, access it
+    via the :attr:`VideoDecoder.cpu_fallback` attribute.
+
+    Usage:
+
+    - Use ``str(cpu_fallback_status)`` or ``print(cpu_fallback_status)`` to see the cpu fallback status
+    - Use ``if cpu_fallback_status:`` to check if any fallback occurred
+    """
```
Dan-Flores, line 41: Nice work, this all renders correctly on the docs!

test/test_decoders.py
```diff
@@ -1737,6 +1743,53 @@ def test_set_cuda_backend(self):
...
+        decoder.get_frame_at(0)
+
+        assert decoder.cpu_fallback.status_known
+        assert decoder.cpu_fallback
+        if "beta" in device:
+            # Beta interface provides the specific reason for fallback
+            assert "Video not supported" in str(decoder.cpu_fallback)
+        else:
+            # FFmpeg interface doesn't know the specific reason
+            assert "Unknown reason" in str(decoder.cpu_fallback)
```
Dan-Flores: I believe this if / else condition can be merged with the one above on lines 1764-1767 to make this test more linear. 
We would check `decoder.cpu_fallback.status_known`, then assert the `str(decoder.cpu_fallback)` value directly after.

Dan-Flores: We could also extend the assert to check for the full string "Unknown reason - try the Beta interface to know more" to be certain it is used.

test/test_decoders.py
```diff
@@ -1737,6 +1743,53 @@ def test_set_cuda_backend(self):
...
+    def test_cpu_fallback_h265_video(self, device):
+        """Test that H265 video triggers CPU fallback on CUDA interfaces."""
+        # H265_VIDEO is known to trigger CPU fallback on CUDA
+        # because its dimensions are too small
+        decoder, _ = make_video_decoder(H265_VIDEO.path, device=device)
+
+        if "beta" in device:
+            # For beta interface, status is known immediately
+            assert decoder.cpu_fallback.status_known
+            decoder.get_frame_at(0)
```
NicolasHug: We shouldn't need to call that with the beta backend
```suggestion
```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -186,9 +241,42 @@ def __init__(
...
+
+            if "status unknown" not in backend_details:
+                self._cpu_fallback.status_known = True
+
+                if "CPU fallback" in backend_details:
+                    self._cpu_fallback._is_fallback = True
+                    if "NVCUVID not available" in backend_details:
+                        self._cpu_fallback._nvcuvid_unavailable = True
+                    elif self._cpu_fallback._backend == "Beta CUDA":
+                        self._cpu_fallback._video_not_supported = True
```
NicolasHug: The current logic is correct but it's not immediately obvious that only the Beta interface can provide details. I think we can make it more explicit with something like this. The logic should be the same.

```suggestion
                if "CPU fallback" in backend_details:
                    self._cpu_fallback._is_fallback = True
                    if self._cpu_fallback._backend == "Beta CUDA":
                        # Only the beta interface can provide details.
                        # if it's not that nvcuvid is missing, it must be video-specific
                        if "NVCUVID not available" in backend_details:
                            self._cpu_fallback._nvcuvid_unavailable = True
                        else: 
                            self._cpu_fallback._video_not_supported = True
```

NicolasHug (approved):
Thanks a lot for working on this @mollyxu ! Left 2 last minor comments below but LGTM

PR #1094, opened 2025-12-04
Implement CenterCrop

scotts:
Depends on #1081.

Implementation of `torchcodec.transforms.CenterCrop`.

The only potential surprise here is the change to the C++ `Crop` transform, where we are using it to implement both `CenterCrop` and `RandomCrop`. My rationale here is that the C++ objects have a 1:1 mapping to the FFmpeg filters, and the objects in `torchcodec.transforms.DecoderTransform` can map to either a TorchVision transform or directly to an FFmpeg filter. (We don't have any of the latter yet.) The code in `custom_ops.cpp` is the bridge between the two.

I'm open to instead making the C++ side not share crop implementations and define a C++ `CenterCrop` that takes no x and y coordinates.

### DIFF
--- a/docs/source/api_ref_transforms.rst
+++ b/docs/source/api_ref_transforms.rst
@@ -14,5 +14,6 @@ For a tutorial, see: TODO_DECODER_TRANSFORMS_TUTORIAL.
     :template: dataclass.rst
 
     DecoderTransform
+    CenterCrop
     RandomCrop
     Resize
--- a/src/torchcodec/_core/Transform.cpp
+++ b/src/torchcodec/_core/Transform.cpp
@@ -46,6 +46,8 @@ CropTransform::CropTransform(const FrameDims& dims, int x, int y)
 }
 
 std::string CropTransform::getFilterGraphCpu() const {
+  // For the FFmpeg filter crop, if the x and y coordinates are left
+  // unspecified, it defaults to a center crop.
   std::string coordinates = x_.has_value()
       ? (":" + std::to_string(x_.value()) + ":" + std::to_string(y_.value()))
       : "";
--- a/test/test_transform_ops.py
+++ b/test/test_transform_ops.py
@@ -154,7 +154,6 @@ class TestPublicVideoDecoderTransformOps:
                 transforms=[torchcodec.transforms.Resize(size=(100, 100, 100))],
             )
 
-<<<<<<< HEAD
     @pytest.mark.parametrize(
         "height_scaling_factor, width_scaling_factor",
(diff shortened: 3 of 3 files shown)


docs/source/api_ref_transforms.rst
```diff
@@ -14,4 +14,5 @@ For a tutorial, see: TODO_DECODER_TRANSFORMS_TUTORIAL.
     :template: dataclass.rst
 
     DecoderTransform
+    RandomCrop
```
NicolasHug, line 18: I think this is https://github.com/meta-pytorch/torchcodec/pull/1081 's driveby, we also need CenterCrop here now :)

src/torchcodec/_core/Transform.cpp
```diff
@@ -37,16 +37,20 @@ std::optional<FrameDims> ResizeTransform::getOutputFrameDims() const {
...
 }
 
 std::string CropTransform::getFilterGraphCpu() const {
+  std::string coordinates = x_.has_value()
+      ? (":" + std::to_string(x_.value()) + ":" + std::to_string(y_.value()))
+      : "";
   return "crop=" + std::to_string(outputDims_.width) + ":" +
-      std::to_string(outputDims_.height) + ":" + std::to_string(x_) + ":" +
-      std::to_string(y_) + ":exact=1";
+      std::to_string(outputDims_.height) + coordinates + ":exact=1";
```
NicolasHug, line 55: Just to check my understanding, it's a **filtergraph built-in behavior** that passing `"crop=w:h"` leads to a center crop?

src/torchcodec/_core/Transform.h
```diff
@@ -61,14 +61,17 @@ class CropTransform : public Transform {
  public:
   CropTransform(const FrameDims& dims, int x, int y);
 
+  // Becomes a center crop if x and y are not specified.
+  CropTransform(const FrameDims& dims);
```
NicolasHug, line 65: To follow-up on your main comment about using `CropTransform` for both RandomCrop and CenterCrop, I think what you have here, and the rationale behind it, make sense.

src/torchcodec/_core/Transform.cpp
scotts: Yes, exactly. I can update the comment to make that clear. From [the docs](https://ffmpeg.org/ffmpeg-filters.html#crop):

> **x**
> The horizontal position, in the input video, of the left edge of the output video. It defaults to (`in_w-out_w)/2`. This expression is evaluated per-frame.
>
> **y**
> The vertical position, in the input video, of the top edge of the output video. It defaults to `(in_h-out_h)/2`. This expression is evaluated per-frame.

PR #1100, opened 2025-12-04
Update Python type annotations to modern syntax and types

### LINKED ISSUE #1033: Use modern Python type annotation syntax
We still use the old syntax and types. For example, we say something like:
```python
def foo(x: Optional[List[Union[int, float]]]):
```
While the newer way would be:
```python
def foo(x: list[int | float] | None):
```
Using `list` versus `List` is supported since Python 3.9, and the better optional and union syntax since 3.10. Our minimum has been 3.10 since TorchCodec v0.8. See: https://docs.python.org/3/library/typing.html

scotts:
Fixes #1033.

I did not do this manually. :) Claude did the conversions, with a few rounds of prompting. I had to do some manual un-doing of changes to our tutorials, as those are formatted for the particular rendering we get.

Note that one actual change is that I upped our Python version in `pyproject.toml`. Because we never bumped that up from 3.8, Claude initially put `from __future__ import annotations` all over the place.

### DIFF
--- a/src/torchcodec/_core/ops.py
+++ b/src/torchcodec/_core/ops.py
@@ -164,8 +164,8 @@ def create_from_file_like(
     assert _pybind_ops is not None
     return _create_from_file_like(
         _pybind_ops.create_file_like_context(
-            file_like, False
-        ),  # False means not for writing
+            file_like, False  # False means not for writing
+        ),
         seek_mode,
     )
 


scotts:
@NicolasHug, yup, agreed, I created #1102 to track the CI change.

PR #1105, opened 2025-12-05
Skip av1_nvenc test if in github ci

Dan-Flores:
Linux cuda machines on Github CI do not have support for the `av1_nvenc` codec.

This PR adds a marker to skip a test if running in github CI, similar to the [IN_OSS_CI mark in torchvision](https://github.com/pytorch/vision/blob/6b56de1cc83386025f2bad87abf608077d1853f7/test/common_utils.py#L28).

The marker checks that `GITHUB_ACTIONS` is true, which is [always set in a Github Action](https://docs.github.com/en/actions/reference/workflows-and-actions/variables).

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1146,6 +1146,8 @@ class TestVideoDecoder:
             key_frame_indices, h265_reference_key_frame_indices, atol=0, rtol=0
         )
 
+    # TODO investigate why this is failing from the nightlies of Dec 09 2025.
+    @pytest.mark.skip(reason="TODO investigate")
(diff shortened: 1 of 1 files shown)


test/test_encoders.py
```diff
@@ -17,6 +17,7 @@
     assert_tensor_close_on_at_least,
     get_ffmpeg_major_version,
     get_ffmpeg_minor_version,
+    IN_GIT_CI,
```
NicolasHug: I think we'll want to name it "GITHUB" CI rather than "GIT" CI

Issue #1014, opened 2025-10-31
python windows 3.8 +


deepblue96:
### 📚 The doc issue

please add this to documentation took me 2h to find it...
Python ≥3.8 on Windows does not use PATH for dependent DLLs. You must add the FFmpeg DLL folder to the current process with os.add_dll_directory(...) before importing torchcodec. Adding PATH via os.system("set PATH=...") does not affect the running process. Order is also tricky if you add multiple directories. ([Python documentation](https://docs.python.org/3/whatsnew/3.8.html))

test file:
import torch, os
os.add_dll_directory("C:\\ffmpeg\\bin")
import torchcodec, platform
print(f"Torch: {torch.__version__}")
print(f"TorchCodec: {torchcodec.__version__}")
print(f"Python: {platform.python_version()}")

NicolasHug:
Hi @deepblue96 , sorry, I'm not sure I completely understand what is needed, or why.

TorchCodec never supported Python < 3.8 so I don't think this is directly related to 3.8+.
As far as I know we haven't had issues related to PATH on Windows, so I suspect this is something specific to your own env?

deepblue96:
i'm not sure on 2 different system had to do the same procedure...
environment:
windows 11 10.0.26200 Build 26200
python 3.12.6
steps to reproduce:
downloaded and installed ffmpeg in standard folder c:\ffmpeg
added to PATH enviroment variables
restarted pc
created venv
installed pytorch 2.9.0 cu126
installed torchcodec
error ffmpeg not found when running any torchcodec related operation

litagin02:
I want to share what I’ve found and confirmed about torchcodec and ffmpeg in Windows.

On Windows, simply installing ffmpeg, adding it to PATH, and installing torchcodec is not enough. We need either:
- Use conda for everything as written in the README, or
- Install the **shared build** of ffmpeg (e.g., in `C:\ffmpeg`), and before every `import torchcodec`, call `os.add_dll_directory("C:\ffmpeg\bin")` where some `.dll` files are located.

I think this behavior should be documented. At first, I assumed that since I already had ffmpeg installed, added to PATH, and with a compatible version, I would be able to use it without conda. But I still hit an error on `import torchcodec`, and it took me a while to figure out what was really going on.

NicolasHug:
This should hopefully be fixed from `torchcodec 0.9.1` (compatible with torch 2.9). I'll close this issue, feel free to open a new one if the problem persists.

PR #1109, opened 2025-12-09
Fix load_torchcodec_shared_libraries on Windows

### LINKED ISSUE #1108: Cannot find FFmpeg DLLs by default
### 🐛 Describe the bug

Continuing from #912 as the underlying issue has not been fixed.
The issue is that torchcodec is not able to find FFmpeg DLLs, even if the FFmpeg bin folder is added to PATH.

Sample of the error:
```
[start of libtorchcodec loading traceback]
FFmpeg version 7: Could not find module 'C:\Python313\site-packages\torchcodec\libtorchcodec_core7.dll' (or one of its dependencies). Try using the full path with constructor syntax.
FFmpeg version 6: Could not find module 'C:\Python313\site-packages\torchcodec\libtorchcodec_core6.dll' (or one of its dependencies). Try using the full path with constructor syntax.
FFmpeg version 5: Could not find module 'C:\Python313\site-packages\torchcodec\libtorchcodec_core5.dll' (or one of its dependencies). Try using the full path with constructor syntax.
FFmpeg version 4: Could not find module 'C:\Python313\site-packages\torchcodec\libtorchcodec_core4.dll' (or one of its dependencies). Try using the full path with constructor syntax.
[end of libtorchcodec loading traceback].
```

Currently, one fix, as recommended in this [French comment](https://github.com/meta-pytorch/torchcodec/issues/912#issuecomment-3450048176), is to directly copy the DLLs into the `torchcodec` site-packages folder.

### Versions

Using torch 2.8+cu128, torchcodec 0.7, and Python 3.13.

aboood40091:
Add support for locating FFmpeg DLLs on Windows.

Fixes #1108, #1014, and several other issues.

### DIFF
--- a/src/torchcodec/_core/ops.py
+++ b/src/torchcodec/_core/ops.py
@@ -89,12 +89,10 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
 
 
 if sys.platform == "win32" and hasattr(os, "add_dll_directory"):
-    # We need to locate the directory containing FFmpeg DLLs
-    # and temporarily add it to the DLL search path.
-    # This seems to be needed on some users machine, but not on our CI. We don't
-    # know why.
-    ffmpeg_path = shutil.which("ffmpeg")
-    if ffmpeg_path:
+    # On windows we try to locate the FFmpeg DLLs and temporarily add them to
+    # the DLL search path. This seems to be needed on some users machine, but
+    # not on our CI. We don't know why.
+    if ffmpeg_path := shutil.which("ffmpeg"):
 
         def cm():  # noqa: F811
             ffmpeg_dir = Path(ffmpeg_path).parent


src/torchcodec/_core/ops.py
```diff
@@ -70,7 +74,8 @@ def load_torchcodec_shared_libraries():
     raise RuntimeError(
         f"""Could not load libtorchcodec. Likely causes:
           1. FFmpeg is not properly installed in your environment. We support
-             versions 4, 5, 6, 7, and 8.
+             versions 4, 5, 6, 7, and 8. On Windows, ensure you've installed
+             the "full-shared" version.
```
NicolasHug: I did a quick search and I wasn't able to find an authoritative source on what a full-shared version is. I assume it's an version that ships dll instead of statically linking all libraries.

My concern is that this error message may not be directly actionable for users, unless we add more details on how they can install such a "full-shared" version. These details aren't really in scope here, so maybe we can leave this error message as it was?

aboood40091: Correct. "full-shared" is the version that ships DLLs instead of statically linking them. "full-shared" is the term used in the download page.

Screenshot_20251209-145038_Brave
Screenshot_20251209-145314_Brave

Personally, I did not know the shared version is the one that has all DLLs until I started digging enough, so I figured this information would be helpful. 

If you still would like me to reverse the change to the message, please let me know.

aboood40091:
> Thanks @aboood40091
> 
> ...
> 
> What I do not understand though: why are users experiencing these issues, when our Windows CI jobs are just fine? There must be a fundamental env difference that I'm currently missing?

Personally, I don't know. It actually should *not* work for you under normal circumstances according to the Python documentation due to a breaking change in Python 3.8. Refer to:
https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

The use of `add_dll_directory` is *required* now as PATH will no longer be searched for DLLs. Perhaps whatever environment you have somehow overrides this, or it's already called `add_dll_directory` somewhere behind the scenes.

aboood40091:
@NicolasHug in your latest changes, `cm` is undefined if ffmpeg path is null. Is this intentional?

NicolasHug:
:open_mouth: thanks for catching this @aboood40091 . Let me submit a follow-up. We're planning to publish a bugfix release later today with your windows fix.

zwukong:
still error . I even copy all the ffmpeg 8.0.1 bin/dlls to the site-packages\torchcodec dir. 
 
 torch 2.9.1cu130.  torchcodec-0.10.0
```
FFmpeg version 8:
Traceback (most recent call last):
  File "M:\ComfyUI\312_cu128\python_embeded\Lib\site-packages\torch\_ops.py", line 1488, in load_library
    ctypes.CDLL(path)
  File "ctypes\__init__.py", line 379, in __init__
FileNotFoundError: Could not find module 'M:\ComfyUI\312_cu128\python_embeded\Lib\site-packages\torchcodec\libtorchcodec_core8.dll' (or one of its dependencies). Try using the full path with constructor syntax.

The above exception was the direct cause of the following exception:
```

PR #1121, opened 2025-12-10
Add pytest skip to test_contiguity

Dan-Flores:
The `VideoEncoder::test_contiguity` test is flaky, and fails most consistently on FFmpeg4 + CUDA. Lets disable it for now to reduce the noise on CI.


NicolasHug (approved):
Thanks @Dan-Flores

PR #1123, opened 2025-12-11
Transforms tutorial

scotts:
First draft of the transform tutorial. Things to consider:

1. The order in which things are presented is, to me, a natural teaching order. But the actual thing we're demonstrating comes a quarter of the way down the page! Does that seem okay when read, or should we try to find a way to pull the `transforms` usage in `VideoDecoder` up higher?
2. I just copied the guarantees that are part of the `DecoderTransform` docstring because I feel that information is critical, and I didn't see a point in trying to rephrase it.
3. The second "Note," after those guarantees, has a possibility of being confusing. I think we have to say something on this, but we've made this difficult to talk about because we accept TorchVision transform objects. So in writing, it's hard to distinguish between *accepting* TorchVision transform objects and *applying* TorchVision transform objects without getting super wordy. Let me know if you find it potentially confusing.
4. I am underwhelmed by the lack of demo for memory efficiency, but I don't have a way around it. And I think it needs to be said.
5. The runtime guidance is subtle. Too subtle?

Once we align on 4 and 5, we should also update the performance tutorial. I think that should be a separate PR.


examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+print(f"Penguin video metadata: {VideoDecoder(penguin_video_path).metadata}")
+
+# %%
+# As shown above, the video is 37 seconds long and has a height of 2160 pixels
+# and a width of 3840 pixels.
+#
+# .. note::
+#
+#     The colloquial way to report the dimensions of this video would be as
+#     3840x2160; that is, (`width`, `height`). In the PyTorch ecosystem, image
```
NicolasHug, line 93: Just noting that the single ticks render as italics (which is fine!)

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+#
+# The main motivation for decoder transforms is *memory efficiency*,
+# particularly when applying transforms that reduce the size of a frame, such
+# as resize and crop. Because the transforms are applied during decoding, the
+# full frame is never returned to the Python layer. As a result, there is
+# significantly less pressure on the Python gargabe collector.
+#
+# In `benchmarks <https://github.com/meta-pytorch/torchcodec/blob/f6a816190cbcac417338c29d5e6fac99311d054f/benchmarks/decoders/benchmark_transforms.py>`_
+# reducing frames from (1080, 1920) down to (135, 240), we have observed a
+# reduction in peak resident set size from 4.3 GB to 0.4 MB.
```
NicolasHug: just to confirm: it's 0.4 **MB** ? Not 0.4 GB?

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+#   1. **Normalization**: Videos can have many different lengths, resolutions,
+#      and frame rates. Normalizing all videos to the same characteristics
+#      leads to better model performance.
+#   2. **Data reduction**: Training on higher resolution frames may lead to better
+#      model performance, but it will be more expensive both at training and
+#      inference time. As a consequence, many video pre-processing pipelines reduce
+#      frame dimensions through resizing and cropping.
+#   3. **Variety**: Applying random transforms (flips, crops, perspective shifts)
+#      to the same frames during training can improve model performance.
+#
```
NicolasHug: On the above:

"normalization" is very overloaded but in the context of transforms, it usually refers to the [Normalize](https://docs.pytorch.org/vision/main/generated/torchvision.transforms.Normalize.html) transform, which is basically just used to "standardize" (yet another overloaded term) the pixel values. It's not about frame resolution or rates. Also, the fact that we resize frames to the same dimensions isn't truely about model performance: it's usually a **technical constraint** that models expect a unique input dimensions (usually related to the last fully-connected layers)

"variety" is usually called "data augmentation" which clashes with the "data reduction" term above.

I can suggest alternative names for these, but I think we can make our lives easier and just remove most of this section. The users coming into this tutorial should already know why they want to apply transforms, and why they're useful. I think we can accept that as a premise?

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+# Below is a simple example of applying the
+# :class:`~torchvision.transforms.v2.Resize` transform to a single frame:
```
NicolasHug:
```suggestion
# Below is a simple example of applying TorchVision's
# :class:`~torchvision.transforms.v2.Resize` transform to a single frame **after** the decoder returns it:
```

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+frame = full_decoder[5]
+resized_after = v2.Resize(size=(480, 640))(frame)
+
+plot(resized_after, title="Resized to 480x640 after decoding")
+
+# %%
+# In the example above, ``full_decoder`` returns a video frame that has the
+# dimensions (2160, 3840) which is then resized down to (480, 640). But with the
+# ``transforms`` parameter of :class:`~torchcodec.decoders.VideoDecoder` we can
+# specify for the resize to  happen during decoding:
```
NicolasHug: To emphasize the difference with my other suggestion above:

```suggestion
# specify for the resize to  happen **during** decoding!
```

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+# %%
+# TorchCodec's relationship to TorchVision transforms
+# -----------------------------------------------------
+# Notably, in our examples we are passing in TorchVision
+# :class:`~torchvision.transforms.v2.Transform` objects as our transforms. We
+# would have gotten equivalent behavior if we had passed in the
+# :class:`~torchcodec.transforms.Resize` object that is a part of TorchCodec.
+# :class:`~torchcodec.decoders.VideoDecoder` accepts both objects as a matter of
+# convenience and to clarify the relationship between the transforms that TorchCodec
+# applies and the transforms that TorchVision offers.
```
NicolasHug, line 146: I don't have a specific suggestion on the above but I think we should try to make it more obvious that torchvision is purely an optional dependency and that users will find equivalent torchcodec-native transforms.

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+#
+# Importantly, the two frames are not identical, even though we can see they
+# *look* very similar:
+
+abs_diff = (resized_after.float() - resized_during.float()).abs()
+(abs_diff == 0).all()
+
+# %%
+# But they're close enough that models won't be able to tell a difference:
+(abs_diff <= 1).float().mean() >= 0.998
```
NicolasHug: Nit, this could be an `assert`, which makes the output look slightly better IMO (because shorter).

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+
+# %%
+# While :class:`~torchcodec.decoders.VideoDecoder` accepts TorchVision transforms as
+# *specifications*, it is not actually using the TorchVision implementation of these
+# transforms. Instead, it is mapping them to equivalent
+# `FFmpeg filters <https://ffmpeg.org/ffmpeg-filters.html>`_. That is,
+# :class:`torchvision.transforms.v2.Resize` is mapped to
+# `scale <https://ffmpeg.org/ffmpeg-filters.html#scale-1>`_ and
+# :class:`torchvision.transforms.v2.CenterCrop` is mapped to
+# `crop <https://ffmpeg.org/ffmpeg-filters.html#crop>`_.
```
NicolasHug, line 166: To emphasize that both TV and TC transforms are accepted, and that they use the same underlying implem:

```suggestion
# `FFmpeg filters <https://ffmpeg.org/ffmpeg-filters.html>`_. That is,
# :class:`torchvision.transforms.v2.Resize`  and :class:`torchcodec.transforms.Resize`  are mapped to
# `scale <https://ffmpeg.org/ffmpeg-filters.html#scale-1>`_ and
# both :class:`torchvision.transforms.v2.CenterCrop` and :class:`torchcodec.transforms.CenterCrop` are mapped to
# `crop <https://ffmpeg.org/ffmpeg-filters.html#crop>`_.
```

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+# the originally decoded frame. The output of that transform becomes the input
+# to the next transform in the list, and so on.
+#
+# A simple example:
+
+crop_resize_decoder = VideoDecoder(
+    penguin_video_path,
+    transforms = [
+        v2.Resize(size=(480, 640)),
+        v2.CenterCrop(size=(315, 220))
```
NicolasHug: It usually makes more sense to first crop and then resize, because resize will then work on a smaller surface.

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+
+# %%
+# Performance: memory efficiency and speed
+# ----------------------------------------
+#
+# The main motivation for decoder transforms is *memory efficiency*,
+# particularly when applying transforms that reduce the size of a frame, such
+# as resize and crop. Because the transforms are applied during decoding, the
+# full frame is never returned to the Python layer. As a result, there is
+# significantly less pressure on the Python gargabe collector.
```
NicolasHug: I think there's another core reason why that's more memory efficient: the decompressed RGB frame is never materialized in its original resolution.

Without decoder-native transform we have:

YUV compressed frame in original res -> RGB decompressed frame in original res -> RGB decompressed frame in final (smaller) res

WIth the decoder-native transform we have:

YUV compressed frame in original res -> RGB decompressed frame in final (smaller) res

i.e. we can skip the "RGB decompressed frame in original res" materialization, which is the most memory-expensive bit.

The garbage collector being less pressure is a consequence of that.

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+    )
+    frames = samplers.clips_at_regular_indices(
+        decoder,
+        num_clips=1,
+        num_frames_per_clip=200
+    )
+    transformed_frames = []
+    for frame in frames.data[0]:
+        frame = v2.Resize(size=(480, 640))(frame)
+        frame = v2.CenterCrop(size=(315, 220))(frame)
```
NicolasHug: The recommended way to use the tv transforms is to instanciate the transform objects once, and they re-use them. You could also use the functional instead. Here, I'd just suggested to define

```py
transforms = v2.Compose([v2.Resize(..), ...])
# and then just call
frame transforms(frame)
```


Also, it's best to transform the whole batch at once rather than transforming individual frames.

examples/decoding/transforms.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+print(f"decoder transforms:    {bench(sample_decoder_transforms, num_threads=0)}")
+print(f"torchvision transform: {bench(sample_torchvision_transforms, num_threads=0)}")
+
+# %%
+# The reason is that FFmpeg is applying the decoder transforms in parallel.
+# However, if the number of threads is 1 (as is the default), then there often is no
+# runtime benefit to using decoder transforms. Using the TorchVision transforms may
+# even be faster!
```
NicolasHug, line 318: When explicitly setting ffmpeg_num_threads to 1 I think it makes sense to also use `torch.set_num_threads(1)`. Otherwise we're comparing a single-threaded implem (ffmpeg) against a potentially multi-threaded one (TV).

examples/decoding/transforms.py
scotts: Yup, it's intentional. I'm considering those as more "math mode" than "code mode".

examples/decoding/transforms.py
scotts: Good catch - it's 0.4 GB. I had thought the default number in top for RSS was bytes, but it's kb.

examples/decoding/transforms.py
scotts: Sure, that simplifies things.

examples/decoding/transforms.py
scotts: I'm not using the functionals to keep the relationship between the TC and TV side clear. Good call on batch and keeping one object.

examples/decoding/transforms.py
scotts: That's not entirely accurate - we definitely never get the "RGB decompressed frame in original res" in the Python layer, but it exists in FFmpeg. This is because we ensure that the FFmpeg filters get applied in the output color space. So without decoder transforms we have (parenthesis to indicate where it happens, TC or TV):
```
YUV compressed, original res (TC) -> 
RGB decompressed , original res (TC) -> 
RGB decompressed, smaller res (TV)
```
With decoder transforms it's:
```
YUV compressed, original res (TC) -> 
RGB decompressed, original res (TC) -> 
RGB decompressed, smaller res (TC)
```
So we really do go through the same steps in decoder transforms. That middle step - getting the RGB image in the original resolution - is because of this line: https://github.com/meta-pytorch/torchcodec/blob/ee8ce04431ba295a6e4c98935eb6ab38dfe7bd47/src/torchcodec/_core/CpuDeviceInterface.cpp#L84

Eliminating the explicit `"format=rgb24"` does improve performance a lot, but at the cost of similarity with using TorchVision transforms on full frames.

Since the filtergraph inputs and outputs are known statically, I suspect they're able to optimize things and reuse memory. That is, it's possible for them to allocate exactly the memory they need for each step and reuse it every time. But I don't *know* that's the case. I'll try to say something about all this.

examples/decoding/transforms.py
scotts: Indeed it does, and curiously, it actually makes decoder transforms faster than the TorchVision version now (at least on my dev machine).

Results with the old way:
```
0:
decoder transforms:    times_med = 1474.17ms +- 79.85
torchvision transform: times_med = 4683.55ms +- 28.71

1:
decoder transforms:    times_med = 18486.50ms +- 165.66
torchvision transform: times_med = 16066.02ms +- 164.19
```
Results with the new way:
```
0:
decoder transforms:    times_med = 1352.46ms +- 34.86
torchvision transform: times_med = 4077.44ms +- 45.63

1:
decoder transforms:    times_med = 14771.99ms +- 148.83
torchvision transform: times_med = 16112.88ms +- 62.15
```

examples/decoding/transforms.py
scotts: Tried to address this by just explicitly stating it with bold text.

Dan-Flores:
Let's add a grid card item so this tutorial appears on the "Home" tab / `index` page:
https://github.com/meta-pytorch/torchcodec/blob/6300361dfdb16f37f7c5c2756beaf22fdf242f4e/docs/source/index.rst?plain=1#L87-L93

examples/decoding/transforms.py
```diff
@@ -0,0 +1,319 @@
...
+"""
+=======================================================
+Decoder Transforms: Applying transforms during decoding
+=======================================================
+
+In this example, we will demonstrate how to use the ``transforms`` parameter of
+the :class:`~torchcodec.decoders.VideoDecoder` class. This parameter allows us
+to specify a list of :class:`torchcodec.transforms.DecoderTransform` or
+:class:`torchvision.transforms.v2.Transform` objects. These objects serve as
+transform specificiations that the :class:`~torchcodec.decoders.VideoDecoder`
```
Dan-Flores: nit: specifications

examples/decoding/transforms.py
```diff
@@ -0,0 +1,319 @@
...
+will apply during the decoding process.
+"""
+
+# %%
+# First, a bit of boilerplate and definitions that we will use later:
```
Dan-Flores: Regarding point 1 in the PR description about the demonstration starting a quarter down the page - we have a pattern of having a link to skip past the boiler plate section, that might help this gap feel smaller.

https://github.com/meta-pytorch/torchcodec/blob/6300361dfdb16f37f7c5c2756beaf22fdf242f4e/examples/decoding/sampling.py#L17-L20

examples/decoding/transforms.py
scotts: I should probably start asking Claude to do spell check on the comments. 🤔

examples/decoding/transforms.py
```diff
@@ -0,0 +1,324 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates.
+# All rights reserved.
+#
+# This source code is licensed under the BSD-style license found in the
+# LICENSE file in the root directory of this source tree.
+
+"""
```
mollyxu, line 7: Adding meta directive at the top could help with discoverability

For example
```
.. meta::
   :description: Learn how to apply resize and crop transforms during video decoding with TorchCodec's VideoDecoder for improved memory efficiency and performance.
```

examples/decoding/transforms.py
```diff
@@ -0,0 +1,324 @@
...
+# The reason is that FFmpeg is applying the decoder transforms in parallel.
+# However, if the number of threads is 1 (as is the default), then there is often
+# less benefit to using decoder transforms. Using the TorchVision transforms may
+# even be faster!
+
+print(f"decoder transforms:    {bench(sample_decoder_transforms, num_threads=1)}")
+print(f"torchvision transform: {bench(sample_torchvision_transforms, num_threads=1)}")
+
+# %%
+# In brief, our performance guidance is:
```
mollyxu, line 324: Would it be worth mentioning decoder native transforms in the performance tips docs?

scotts: @mollyxu, yes, absolutely. I'd like to do that in a follow-up PR.

examples/decoding/transforms.py
```diff
@@ -0,0 +1,324 @@
...
+#
+#    1. If you are applying a transform pipeline that signficantly reduces
```
mollyxu, line 326: Would there be a case in which users need to perform some transforms that are not currently supported natively by torchcodec in addition to those that are supported natively by torchcodec? In that case would the advice be to stick with the torch vision transforms in order to not mix the transforms?

scotts: Actually, no, I think it would be fine to use a mix of transforms in that case. The mix we want to avoid is, say, using a decoder transform resize during training, but TorchVision's resize during inference. Let me try to clarify that point in the note above.

PR #1124, opened 2025-12-11
Find default hardware encoder in VideoEncoder

Dan-Flores:
To improve ease of use of the `VideoEncoder` on GPU, this PR adds logic to find and use a hardware enabled `codec` if one is available.

A boolean arg is added to [findCodec](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/CudaDeviceInterface.cpp#L339) to flip the function to search for encoders. This means it has the following behaviors:
- For video container formats where the default `codec` has a hardware equivalent (ex. `libx264`, `h264_nvenc`), the function will correctly return the hardware encoder.
- For video container formats where the default `codec` does NOT have a hardware equivalent(ex. the `mpeg` codec used for`avi`), no hardware encoder will be used by default.

### DIFF
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -1336,7 +1336,6 @@ class TestVideoEncoder:
 
     @needs_ffmpeg_cli
     @pytest.mark.needs_cuda
-    # @pytest.mark.parametrize("format", ("mov", "mp4", "mkv"))
     @pytest.mark.parametrize("method", ("to_file", "to_tensor", "to_file_like"))
     # TODO-VideoEncoder: Enable additional pixel formats ("yuv420p", "yuv444p")
     @pytest.mark.parametrize(
@@ -1390,8 +1389,8 @@ class TestVideoEncoder:
         ffmpeg_cmd.extend(["-qp", str(qp)])
         ffmpeg_cmd.extend([ffmpeg_encoded_path])
         subprocess.run(ffmpeg_cmd, check=True, capture_output=True)
-        encoder = VideoEncoder(frames=source_frames, frame_rate=frame_rate)
 
+        encoder = VideoEncoder(frames=source_frames, frame_rate=frame_rate)
         encoder_extra_options = {"qp": qp}
         if method == "to_file":
             encoder_output_path = str(tmp_path / f"nvenc_output.{format}")


test/test_encoders.py
```diff
@@ -1337,28 +1335,13 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
-            ("avi", "h264_nvenc"),
-            pytest.param(
-                ("mkv", "av1_nvenc"),
-                marks=pytest.mark.skipif(
-                    IN_GITHUB_CI, reason="av1_nvenc is not supported on CI"
-                ),
-            ),
-        ],
-    )
+    @pytest.mark.parametrize("format", ("mov", "mp4", "mkv"))
```
Dan-Flores: These formats all default to `h264_nvenc`, so `IN_GITHUB_CI` is no longer necessary to avoid `av1_nvenc`.

test/test_encoders.py
```diff
@@ -1337,28 +1335,13 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
 
     @needs_ffmpeg_cli
     @pytest.mark.needs_cuda
-    # TODO-VideoEncoder: Auto-select codec for GPU encoding
-    @pytest.mark.parametrize(
-        "format_codec",
-        [
-            ("mov", "h264_nvenc"),
-            ("mp4", "hevc_nvenc"),
-            ("avi", "h264_nvenc"),
```
Dan-Flores, line 1346: For the container format `avi`, FFmpeg defaults to using the `mpeg` codec, which does not have a hardware equivalent. Users encoding into `avi` will need to specify a hardware enabled codec.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -755,6 +755,15 @@ void VideoEncoder::initializeEncoder(
...
+    // If frames are on a CUDA device, try to substitute the default codec
+    // with its hardware equivalent
+    if (frames_.device().is_cuda() && deviceInterface_) {
+      auto hwCodec = deviceInterface_->findHardwareEncoder(
+          avFormatContext_->oformat->video_codec);
+      if (hwCodec.has_value()) {
+        avCodec = hwCodec.value();
+      }
+    }
     TORCH_CHECK(avCodec != nullptr, "Video codec not found");
```
NicolasHug: Not from this PR but we might as well address it here: 

We have this `TORCH_CHECK` just above, and we also have a similar one in the first `if` branch above

```C
    TORCH_CHECK(
        avCodec != nullptr,
        "Video codec ",
        codec,
        " not found. To see available codecs, run: ffmpeg -encoders");
 
```

Let's just have a single TORCH_CHECK **after** this whole `if/else` block? The error message mentioning `ffmpeg -encoders` is good.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -488,4 +488,25 @@ void CudaDeviceInterface::setupHardwareFrameContextForEncoding(
...
+    for (int j = 0; (config = avcodec_get_hw_config(codec, j)) != nullptr;
+         ++j) {
+      if (config->device_type == AV_HWDEVICE_TYPE_CUDA) {
+        return codec;
+      }
+    }
+  }
+
+  return std::nullopt;
+}
```
NicolasHug: This is 99% the same code as the existing `findCodec()`. Let's just extend `findCodec` and add an optional `bool decoder` parameter who's default is true.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -755,6 +755,15 @@ void VideoEncoder::initializeEncoder(
         avFormatContext_->oformat != nullptr,
         "Output format is null, unable to find default codec.");
     avCodec = avcodec_find_encoder(avFormatContext_->oformat->video_codec);
```
NicolasHug: I think we should logically call `avcodec_find_encoder` *after* we have tried `findHardwareEncoder`. The reason is: for whatever reason, `avcodec_find_encoder` could return an error (like if there's no CPU codec for a given configuration) but `findHardwareEncoder` might be able to find the right one. So I thikn we should logically do:

```
if onCUDA:
  codec = findHardwareEncoder()  # first try NVENC specific stuff
if codec is still not found:
   codec = avcodec_find_encoder()
```

src/torchcodec/_core/Encoder.cpp
```diff
@@ -755,6 +755,15 @@ void VideoEncoder::initializeEncoder(
...
+    // If frames are on a CUDA device, try to substitute the default codec
+    // with its hardware equivalent
+    if (frames_.device().is_cuda() && deviceInterface_) {
```
NicolasHug: This `frames_.device().is_cuda() && deviceInterface_` is a bit imprecise because when frames are on CUDA, we should always have defined a device interface. If it's not defined, then it's a bug. This should be:

```py
if frames on cuda:
   assert device_interface exists
   # continue here
```

test/test_encoders.py
NicolasHug: By removing codec we have lost important testing coverage:

-  manually passing a codec that is not `None`
- passing a non-default codec for that given container format 
- testing on av1

Let's put all these back. To test the new behavior that this PR is enabling, all we have to do is to specify a `codec=None` case. Something like this (with comments to help understanding the intent):

```py
    @pytest.mark.parametrize(
        ("format", "codec"),
        [
            ("mov", None), # should default to h264_nvenc
            ("mov", "h264_nvenc"),
            ("mp4", "hevc_nvenc"),  # non-default
            ("avi", "h264_nvenc"),
            pytest.param(
                ("mkv", "av1_nvenc"),
                marks=pytest.mark.skipif(
                    IN_GITHUB_CI, reason="av1_nvenc is not supported on CI"
                ),
            ),
        ],
    )
```


Note also that I separated `"format"` and `"codec"` into their own parameter input - that's the more conventional way of passing couples of parameters with pytest. You'll have to update the function's signature to accept both format and codec separately.

Dan-Flores: Good point, I've added back these tests so we are explicitly testing `hevc_nvenc` and `av1_nvenc`.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Thanks for the suggestion, I've updated the code to conditionally call `avcodec_find_encoder` iff there is no hardware encoder for the default codec.

test/test_encoders.py
```diff
@@ -1337,28 +1336,29 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
 
     @needs_ffmpeg_cli
     @pytest.mark.needs_cuda
-    # TODO-VideoEncoder: Auto-select codec for GPU encoding
+    # @pytest.mark.parametrize("format", ("mov", "mp4", "mkv"))
```
NicolasHug: can remove

test/test_encoders.py
```diff
@@ -1337,28 +1336,29 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
     )
-    @pytest.mark.parametrize("method", ("to_file", "to_tensor", "to_file_like"))
-    # TODO-VideoEncoder: Enable additional pixel formats ("yuv420p", "yuv444p")
-    def test_nvenc_against_ffmpeg_cli(self, tmp_path, format_codec, method):
+    def test_nvenc_against_ffmpeg_cli(self, tmp_path, method, format, codec):
         # Encode with FFmpeg CLI using nvenc codecs
-        format, codec = format_codec
         device = "cuda"
-        qp = 1  # Lossless (qp=0) is not supported on av1_nvenc, so we use 1
+        qp = 1  # Use near lossless encoding to reduce noise and support av1_nvenc
```
NicolasHug, line 1360: QQ would we be able to assert closer expected results (stricter tolerance) if we set qp = 0 for the other codecs? If we can, then it might be worth it

test/test_encoders.py
```diff
@@ -1381,21 +1381,18 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, format_codec, method):
...
         ]
+        # CLI requires explicit codec for nvenc
+        ffmpeg_cmd.extend(["-c:v", codec if codec is not None else "h264_nvenc"])
+        # VideoEncoder will select an NVENC encoder by default since the frames are on GPU.
 
         ffmpeg_cmd.extend(["-pix_fmt", "nv12"])  # Output format is always NV12
-        if codec == "av1_nvenc":
-            ffmpeg_cmd.extend(["-rc", "constqp"])  # Set rate control mode for AV1
-        ffmpeg_cmd.extend(["-qp", str(qp)])  # Use lossless qp for other codecs
+        ffmpeg_cmd.extend(["-qp", str(qp)])
```
NicolasHug, line 1389: There are two `if codec == "av1_nvenc":` that were removed, above and below.  Just checking if that was by design, and why?

test/test_encoders.py
Dan-Flores: No, at `qp = 0` or `qp = 1` the highest assertion we can make is 96% of frames match.

test/test_encoders.py
Dan-Flores: yes this is intentional, I realized `constqp` is the default rate control mode when `qp` is set for the nvenc codecs, so its not necessary to set it here.

PR #1125, opened 2025-12-11
Add color handling to VideoEncoder GPU

Dan-Flores:
This PR resolves this TODO:
`// TODO-VideoEncoder: Enable configuration of color properties, similar to FFmpeg.`

Support is added for the following parameters:
* Color spaces: `BT.601`, `BT.709`, `BT.2020`
* Color ranges: `tv` (limited), `pc` (full)

As a result, 6 `ColorConversionMatrices` are stored and utilized for NPP color conversion functions. To my understanding, these are the most commonly used or newer color spaces parameters ([docs for AVColorSpace](https://www.ffmpeg.org/doxygen/3.1/pixfmt_8h.html#aff71a069509a1ad3ff54d53a1c894c85))

The testing caveats:
* FFmpeg 4 and 6 fails frame equal assertions in `test_nvenc_against_ffmpeg_cli`. It only passes when using non-default color range and colorspace (specifically, not limited range and BT601). Since these tests pass on FFmpeg 7 and 8, I suspect there is some issue in FFmpeg 4 and 6. I've opened #1140 to track this issue.
* The `av1_nvenc` test is disabled on FFmpeg4, as the codec is not implemented.

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -542,7 +542,7 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
 
   NppiSize oSizeROI = {width, height};
   NppStatus status;
-  // Convert to NV12, as CUDA_PIXEL_FORMAT is always NV12 currently
+  // Convert to NV12, as CUDA_ENCODING_PIXEL_FORMAT is always NV12 currently
   status = nppiRGBToNV12_8u_ColorTwist32f_C3P2R_Ctx(
       static_cast<const Npp8u*>(hwcFrame.data_ptr()),
       hwcFrame.stride(0) * hwcFrame.element_size(),
@@ -555,7 +555,7 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
   TORCH_CHECK(
       status == NPP_SUCCESS,
       "Failed to convert RGB to ",
-      av_get_pix_fmt_name(CUDA_PIXEL_FORMAT),
+      av_get_pix_fmt_name(DeviceInterface::CUDA_ENCODING_PIXEL_FORMAT),
       ": NPP error code ",
       status);
 
@@ -578,7 +578,7 @@ void CudaDeviceInterface::setupHardwareFrameContextForEncoding(
       hwFramesCtxRef != nullptr,
       "Failed to allocate hardware frames context for codec");
 
-  codecContext->sw_pix_fmt = CUDA_PIXEL_FORMAT;
+  codecContext->sw_pix_fmt = DeviceInterface::CUDA_ENCODING_PIXEL_FORMAT;
   // Always set pixel format to support CUDA encoding.
   codecContext->pix_fmt = AV_PIX_FMT_CUDA;
 
--- a/src/torchcodec/_core/CudaDeviceInterface.h
+++ b/src/torchcodec/_core/CudaDeviceInterface.h
@@ -43,7 +43,6 @@ class CudaDeviceInterface : public DeviceInterface {
 
   std::string getDetails() override;
 
-  constexpr static AVPixelFormat CUDA_PIXEL_FORMAT = AV_PIX_FMT_NV12;
   UniqueAVFrame convertCUDATensorToAVFrameForEncoding(
       const torch::Tensor& tensor,
       int frameIndex,
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -795,7 +795,7 @@ void VideoEncoder::initializeEncoder(
   } else {
     if (frames_.device().is_cuda()) {
       // Default to nv12 pixel format when encoding on GPU.
-      outPixelFormat_ = AV_PIX_FMT_NV12;
+      outPixelFormat_ = DeviceInterface::CUDA_ENCODING_PIXEL_FORMAT;
     } else {
       const AVPixelFormat* formats = getSupportedPixelFormats(*avCodec);
       // Use first listed pixel format as default (often yuv420p).
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -1345,9 +1345,15 @@ class TestVideoEncoder:
             pytest.param(
                 "mkv",
                 "av1_nvenc",
-                marks=pytest.mark.skipif(
-                    IN_GITHUB_CI, reason="av1_nvenc is not supported on CI"
-                ),
+                marks=[
+                    pytest.mark.skipif(
+                        IN_GITHUB_CI, reason="av1_nvenc is not supported on CI"
+                    ),
+                    pytest.mark.skipif(
+                        get_ffmpeg_major_version() == 4,
+                        reason="av1_nvenc is not supported on FFmpeg 4",
+                    ),
+                ],
             ),
         ],
     )
@@ -1370,8 +1376,6 @@ class TestVideoEncoder:
             pytest.skip(
                 "Non-default color space and range have lower accuracy on FFmpeg 4 and 6"
             )
-        if ffmpeg_version == 4 and codec == "av1_nvenc":
-            pytest.skip("av1_nvenc is not supported on FFmpeg 4")
         # Encode with FFmpeg CLI using nvenc codecs
         device = "cuda"
         qp = 1  # Use near lossless encoding to reduce noise and support av1_nvenc
@@ -1478,14 +1482,12 @@ class TestVideoEncoder:
                 # when color_range='tv' and color_space=None on FFmpeg 7/8.
                 # Since this failure is rare, I suspect its a bug related to these
                 # older container formats on newer FFmpeg versions.
-                if (
+                if not (
                     ffmpeg_version in (7, 8)
                     and color_range == "tv"
                     and color_space is None
                     and format in ("mov", "avi")
                 ):
-                    pass
-                else:
                     assert color_range == encoder_metadata["color_range"]
             if color_space is not None:
                 assert color_space == encoder_metadata["color_space"]
--- a/src/torchcodec/_core/DeviceInterface.h
+++ b/src/torchcodec/_core/DeviceInterface.h
@@ -139,6 +139,9 @@ class DeviceInterface {
     return "";
   }
 
+  // Pixel format used for encoding on CUDA devices
+  static constexpr AVPixelFormat CUDA_ENCODING_PIXEL_FORMAT = AV_PIX_FMT_NV12;
+
   // Function used for video encoding, only implemented in CudaDeviceInterface.
   // It is here to isolate CUDA dependencies from CPU builds
   // TODO Video-Encoder: Reconsider using video encoding functions in device


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -373,22 +373,72 @@ std::string CudaDeviceInterface::getDetails() {
...
-// NPP ColorTwist function used below expects the limited range
-// color conversion matrix, and this matches FFmpeg's default behavior.
-const Npp32f defaultLimitedRangeRgbToNv12[3][4] = {
-    // Y = 16 + 0.859 * (0.299*R + 0.587*G + 0.114*B)
-    {0.257f, 0.504f, 0.098f, 16.0f},
-    // U = -0.148*R - 0.291*G + 0.439*B + 128 (BT.601 coefficients)
-    {-0.148f, -0.291f, 0.439f, 128.0f},
-    // V = 0.439*R - 0.368*G - 0.071*B + 128 (BT.601 coefficients)
-    {0.439f, -0.368f, -0.071f, 128.0f}};
+// RGB to YUV conversion matrices to use in NPP color conversion functions
```
NicolasHug, line 441: Can you share how these were derived? What were the original values that were used as reference?

test/test_encoders.py
```diff
@@ -1426,13 +1446,19 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, format_codec, method):
...
-            # pix_fmt nv12 is stored as yuv420p in metadata
-            assert encoder_metadata["pix_fmt"] == "yuv420p"
-            assert ffmpeg_metadata["pix_fmt"] == "yuv420p"
+            metadata_fields = ["color_range", "color_space"]
+            ffmpeg_metadata = self._get_video_metadata(
+                ffmpeg_encoded_path, metadata_fields
+            )
+            encoder_metadata = self._get_video_metadata(encoder_output, metadata_fields)
+            assert encoder_metadata["color_range"] == ffmpeg_metadata["color_range"]
+            assert encoder_metadata["color_space"] == ffmpeg_metadata["color_space"]
```
NicolasHug, line 1477: We'll want to be stricter here:


```suggestion
            assert encoder_metadata["color_range"] == ffmpeg_metadata["color_range"] == color_range
            assert encoder_metadata["color_space"] == ffmpeg_metadata["color_space"] == color_space
```

test/test_encoders.py
```diff
@@ -1426,13 +1446,19 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, format_codec, method):
...
+            assert_tensor_close_on_at_least(
+                ff_frame, enc_frame, percentage=percentage, atol=2
+            )
 
         if method == "to_file":
-            ffmpeg_metadata = self._get_video_metadata(ffmpeg_encoded_path, ["pix_fmt"])
-            encoder_metadata = self._get_video_metadata(encoder_output, ["pix_fmt"])
-            # pix_fmt nv12 is stored as yuv420p in metadata
-            assert encoder_metadata["pix_fmt"] == "yuv420p"
-            assert ffmpeg_metadata["pix_fmt"] == "yuv420p"
```
NicolasHug, line 1431: Looks like we're not assert pix_fmt anymore, which makes it hard to verify that the changes in this PR are correct. IIRC, passing NV12 actually resulted in a `yuv420p` format at the end. We should try to undertand why that was the case. It may not add a lot of value to support both nv12 and yuv420p as parameter values if they're both the same thing (and if they both end-up being yuv420 anyway).

Dan-Flores: I suspect the format changes occur based on codec implementation. By adding back this assertion, I observed a deprecated pixel format `yuvj420p` is set when `pc` (full) color range is used by `h264_nvenc` and `hevc_nvenc`, but not `av1_nvenc`. 
I can incorporate pixel formats into my benchmarking PR, to see if there is some optimization to using `nv12`.

src/torchcodec/_core/CudaDeviceInterface.cpp
Dan-Flores: These follow the pattern described in the note in `CudaCommon`, I can add a comment referencing that note here  https://github.com/meta-pytorch/torchcodec/blob/ee8ce04431ba295a6e4c98935eb6ab38dfe7bd47/src/torchcodec/_core/CUDACommon.cpp#L43-L44

NicolasHug: The note is about YUV -> RGB so it's not 100% targeted to what the matrices are doing. But yes, add a ref to that note, it's still useful.

You asked me offline whether we should update the note to explain limited range: yes, we should :)
There's a TODO in the note for that, but I never had the chance to do it - and frankly I forgot the underlying logic lol. If you'd like to give it a go, please do it - in a follow-up PR.

test/test_encoders.py
NicolasHug: The new assertions you added are good. But I personally still do not understand why passing NV12 actually ends up being reported yuv420.
Is it actually still NV12, and it's just FFmpeg that can't tell the difference? Or is it indeed yuv420?

Until we get a good understanding on that, I think we should refrain from allowing `pixel_format` with CUDA encoding. We have a surprising behavior that we cannot explain right now: passing NV12 leads to yuv420. If we're surprised, our users will be surprised too, and we won't have a good explanation to give them.  It is safer to simply not expose this functionality just yet, and let them rely on the default behavior.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -790,23 +790,30 @@ void VideoEncoder::initializeEncoder(
   outHeight_ = inHeight_;
 
   if (videoStreamOptions.pixelFormat.has_value()) {
+    // TODO-VideoEncoder: Enable pixel formats to be set by user
+    // and handled with the appropriate NPP function on GPU.
```
Dan-Flores, line 786: I moved this TODO from `setupHardwareFrameContextForEncoding` to here in `initializeEncoder` to centralize `pixel_format` handling. 

The behavior is unchanged: If `pixel_format` argument is used while frames are on GPU, an error is raised.
The default usage of `nv12` is moved into initializeEncoder` as well.

mollyxu: are we raising an error for `pixel_format` on gpu because of what nicolas mentioned below? that passing NV12 leads to yuv420?

Dan-Flores: Essentially yes. Because we do not understand the codec's behavior yet, we do not want the user to set or expect a pixel format.

test/test_encoders.py
Dan-Flores: I added these stricter checks below, with the caveat that when the color params are `None`, the encodec metadata may set a value other than `None`.

test/test_encoders.py
```diff
@@ -1351,7 +1351,24 @@ def test_extra_options_utilized(self, tmp_path, profile, colorspace, color_range
...
+        # TODO-VideoEncoder: Investigate why FFmpeg 4 and 6 fail with non-default color space and range.
+        # See https://github.com/meta-pytorch/torchcodec/issues/1140
+        if ffmpeg_version in (4, 6) and not (
+            color_space == "bt470bg" and color_range == "tv"
+        ):
+            pytest.skip(
+                "Non-default color space and range have lower accuracy on FFmpeg 4 and 6"
+            )
+        if ffmpeg_version == 4 and codec == "av1_nvenc":
+            pytest.skip("av1_nvenc is not supported on FFmpeg 4")
```
NicolasHug: For this one, we might be able to use the `skipif` mark as part of the parametrization, which is our new preferred strategy to keep internal CI healthy.

But, I am curious: is this actually related to this PR? If `av1_nvenc` really isn't supported on FFmpeg, then how can our existing test (on `main`) pass?

test/test_encoders.py
```diff
@@ -1379,16 +1396,23 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, method, format, codec):
             temp_raw_path,
         ]
         # CLI requires explicit codec for nvenc
+        # VideoEncoder will default to h264_nvenc since the frames are on GPU.
         ffmpeg_cmd.extend(["-c:v", codec if codec is not None else "h264_nvenc"])
-        # VideoEncoder will select an NVENC encoder by default since the frames are on GPU.
-
         ffmpeg_cmd.extend(["-pix_fmt", "nv12"])  # Output format is always NV12
-        ffmpeg_cmd.extend(["-qp", str(qp)])
+        ffmpeg_cmd.extend(["-qp", str(qp)])  # Use lossless qp for other codecs
```
NicolasHug: I'm confused by this comment:

- It is described as "lossless" while also described as "near lossless" on line 1374
- What are the "other codecs" mentioned?

test/test_encoders.py
```diff
@@ -1419,13 +1443,37 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, method, format, codec):
...
 
         assert ffmpeg_frames.shape[0] == encoder_frames.shape[0]
+        # The combination of full range + bt709 results in worse accuracy
+        percentage = 91 if color_range == "full" and color_space == "bt709" else 96
         for ff_frame, enc_frame in zip(ffmpeg_frames, encoder_frames):
             assert psnr(ff_frame, enc_frame) > 25
-            assert_tensor_close_on_at_least(ff_frame, enc_frame, percentage=96, atol=2)
+            assert_tensor_close_on_at_least(
+                ff_frame, enc_frame, percentage=percentage, atol=2
+            )
```
NicolasHug: Out of curiosity can you try to validate an upper bound on the absolute difference? E.g. something like

```
assert_close(..., rtol=0, atol=5)
```

would be pretty good. Might have to be higher than 5. If it's something like 50 then it's not worth it, as it's not informative. But if we can get a low upper-bound, that gives confidence that we're not too far off.

test/test_encoders.py
```diff
@@ -1419,13 +1443,37 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, method, format, codec):
...
+                encoder_metadata["pix_fmt"]
+                == ffmpeg_metadata["pix_fmt"]
+                == expected_pix_fmt
+            )
+            assert encoder_metadata["color_range"] == ffmpeg_metadata["color_range"]
+            assert encoder_metadata["color_space"] == ffmpeg_metadata["color_space"]
+            # Default values vary by codec, so we only assert when
+            # color_range and color_space are not None.
+            if color_range is not None:
+                color_range = ffmpeg_metadata["color_range"]
```
NicolasHug: I think you meant an `assert` here.

Also, while this is functionally equivalent (because of the `assert encoder_metadata[...] == ffmpeg_metadata[...]` call above), what we *logically* want to assert here is that the `encode_metadata` respected the input `color_range` and the input `color_space`. So I'd suggest to write this as `assert encode_metadata[...] == color_range.

Same below

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -383,22 +383,76 @@ std::string CudaDeviceInterface::getDetails() {
...
+  static constexpr Npp32f BT2020_LIMITED[3][4] = {
+      {0.2256f, 0.5823f, 0.0509f, 16.0f},
+      {-0.1227f, -0.3166f, 0.4392f, 128.0f},
+      {0.4392f, -0.4039f, -0.0353f, 128.0f}};
+
+  static constexpr Npp32f BT2020_FULL[3][4] = {
+      {0.2627f, 0.6780f, 0.0593f, 0.0f},
+      {-0.139630f, -0.360370f, 0.5000f, 128.0f},
+      {0.5000f, -0.459786f, -0.040214f, 128.0f}};
+};
```
NicolasHug, line 472: For the above, did you get the values by running a Python snippet like the one from 

https://github.com/meta-pytorch/torchcodec/blob/f91d24aaccb6cc0c87f5eeef1a00262e023f9b20/src/torchcodec/_core/CUDACommon.cpp#L90-L100

, or was this mostly from claude code?

I think it'd help to validate that we can get those values from a python snippet (that we understand) that takes the raw Ffmpeg coefficient as input and that gives us those matrices as output - similarly to what was done in the note.

The other thing I'd like to understand is how were the values for limited range derived. I remember [doing it for one matrix](https://github.com/meta-pytorch/torchcodec/blob/f91d24aaccb6cc0c87f5eeef1a00262e023f9b20/src/torchcodec/_core/CUDACommon.cpp#L216-L224), but I didn't write down the logic for it and completely forgot. Since we now are checking-in such matrices, it's important that we undertand and document how they were generated.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -844,7 +851,7 @@ void VideoEncoder::initializeEncoder(
   if (frames_.device().is_cuda() && deviceInterface_) {
     deviceInterface_->registerHardwareDeviceWithCodec(avCodecContext_.get());
     deviceInterface_->setupHardwareFrameContextForEncoding(
-        avCodecContext_.get());
+        avCodecContext_.get(), outPixelFormat_);
```
NicolasHug: So, we have this invariant throughout the GPU encoder that the output pixel format is always `AV_PIX_FMT_NV12`. it's good practice to enforce this invariant in the APIs themselves. Here for example, there is no need for `deviceInterface_->setupHardwareFrameContextForEncoding()` to take the pixel format as a parameter, because it's an invariant that it should always be `AV_PIX_FMT_NV12`.

The same is true for all the other APIs of the device interface. There are different ways to enforce this invariant in the API, one simple one is to make `outputPixelFormatForEncoding_` a public constexpr member.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -437,34 +491,44 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
...
+          getConversionMatrix(codecContext),
+          *nppCtx_);
+      break;
+    default:
+      TORCH_CHECK(
+          false,
+          "GPU encoding expected to encode into nv12 pixel format, but got ",
+          av_get_pix_fmt_name(targetPixelFormat),
+          ". This should not happen, please report this to the TorchCodec repo");
+  }
```
NicolasHug: Above: this is related to my other comment below about enforcing `targetPixelFormat` to be `AV_PIX_FMT_NV12` through the APIs. It's great that you are raising on this, and indeed this should never happen and this is a torchcodec bug. But we can avoid risking the bug entirely by simply just not making `targetPixelFormat` a parameter! That's the value of "enforcing the invariant through the API".

src/torchcodec/_core/CudaDeviceInterface.cpp
Dan-Flores: A little of both, I had claude code generate a script to create the matrices using the python snippet from the note: P2110894531

I've published #1154 to document the limited range methodology.

test/test_encoders.py
Dan-Flores: The lowest bound I tested that passed was `atol=40`, which is pretty high. The current test tells us that 96% of the frames are within 2, which seems to be fairly high accuracy

test/test_encoders.py
Dan-Flores: This comment is an artifact of deciding what `qp` value to use.  I'll remove it as it is irrelevant now.

test/test_encoders.py
Dan-Flores: These skips are examples of covariate parametrization, where we do not know both values in either parametrization line. I considered using the new pattern, but I believe it is impossible in these cases.

> is this actually related to this PR? 

It is not related to this PR, I likely did not locally test on FFmpeg 4 previously, because my installation of FFmpeg 4 does not support `av1_nvenc`. 

<details>

```

(ffmpeg4_cuda12_6) dev@gpu-dev-d63de2d5:~/torchcodec$ ffmpeg -encoders | grep nvenc    
ffmpeg version 4.4.2 Copyright (c) 2000-2021 the FFmpeg developers
  built with gcc 10.4.0 (conda-forge gcc 10.4.0-16)
  configuration: --prefix=/home/dev/.conda/envs/ffmpeg4_cuda12_6 --cc=/home/conda/feedstock_root/build_artifacts/ffmpeg_1660334214659/_build_env/bin/x86_64-conda-linux-gnu-cc --cxx=/home/conda/feedstock_root/build_artifacts/ffmpeg_1660334214659/_build_env/bin/x86_64-conda-linux-gnu-c++ --nm=/home/conda/feedstock_root/build_artifacts/ffmpeg_1660334214659/_build_env/bin/x86_64-conda-linux-gnu-nm --ar=/home/conda/feedstock_root/build_artifacts/ffmpeg_1660334214659/_build_env/bin/x86_64-conda-linux-gnu-ar --disable-doc --disable-openssl --enable-avresample --enable-demuxer=dash --enable-hardcoded-tables --enable-libfreetype --enable-libfontconfig --enable-libopenh264 --enable-gnutls --enable-libmp3lame --enable-libvpx --enable-pthreads --enable-vaapi --enable-gpl --enable-libx264 --enable-libx265 --enable-libaom --enable-libsvtav1 --enable-libxml2 --enable-pic --enable-shared --disable-static --enable-version3 --enable-zlib --pkg-config=/home/conda/feedstock_root/build_artifacts/ffmpeg_1660334214659/_build_env/bin/pkg-config
  libavutil      56. 70.100 / 56. 70.100
  libavcodec     58.134.100 / 58.134.100
  libavformat    58. 76.100 / 58. 76.100
  libavdevice    58. 13.100 / 58. 13.100
  libavfilter     7.110.100 /  7.110.100
  libavresample   4.  0.  0 /  4.  0.  0
  libswscale      5.  9.100 /  5.  9.100
  libswresample   3.  9.100 /  3.  9.100
  libpostproc    55.  9.100 / 55.  9.100
 V....D h264_nvenc           NVIDIA NVENC H.264 encoder (codec h264)
 V..... nvenc                NVIDIA NVENC H.264 encoder (codec h264)
 V..... nvenc_h264           NVIDIA NVENC H.264 encoder (codec h264)
 V..... nvenc_hevc           NVIDIA NVENC hevc encoder (codec hevc)
 V....D hevc_nvenc           NVIDIA NVENC hevc encoder (codec hevc)
```
</details>

> how can our existing test (on main) pass?

They do not!
The `av1_nvenc` test is skipped if `IN_GITHUB_CI`, so it would not have appeared there. Since this test is also marked as `@needs_ffmpeg_cli`, its not run internally either.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: I was aiming to follow up quickly on adding support for other pixel formats, but I see how this parameter is confusing at the moment. I will remove it from the APIs until it is needed.

test/test_encoders.py
NicolasHug: Got it thank you - I understand how that's relevant for this PR now.

Agreed the  first skip (the one with `color_space == "bt470bg" and color_range == "tv"`) is a covariate parametrization, but I think this one just above is not: the only thing that is parametrized over is `codec`, so I think we could use something like `pytest.param("av1_nvenc", marks=pytest.mark.skipif(get_ffmpeg_major_version() == 4)`.

That's not very critical in any case, feel free to ignore.

test/test_encoders.py
```diff
@@ -1424,8 +1451,41 @@ def test_nvenc_against_ffmpeg_cli(self, tmp_path, method, format, codec):
...
+            assert encoder_metadata["color_range"] == ffmpeg_metadata["color_range"]
+            assert encoder_metadata["color_space"] == ffmpeg_metadata["color_space"]
+            # Default values vary by codec, so we only assert when
+            # color_range and color_space are not None.
+            if color_range is not None:
+                # FFmpeg and torchcodec encode color_range as 'unknown' for mov and avi
+                # when color_range='tv' and color_space=None on FFmpeg 7/8.
+                # Since this failure is rare, I suspect its a bug related to these
+                # older container formats on newer FFmpeg versions.
+                if (
```
NicolasHug: use `if not ...` rather than `if: pass; else: ....`

src/torchcodec/_core/CudaDeviceInterface.h
```diff
@@ -43,6 +43,7 @@ class CudaDeviceInterface : public DeviceInterface {
 
   std::string getDetails() override;
 
+  constexpr static AVPixelFormat CUDA_PIXEL_FORMAT = AV_PIX_FMT_NV12;
```
NicolasHug: Nit, name this `CUDA_ENCODING_PIXEL_FORMAT` (or any variant that is explicitly about encoding), because this file and class contain a lot of stuff about decoding. We don't want to imply that this is a "decoding pixel format".

src/torchcodec/_core/Encoder.cpp
```diff
@@ -782,23 +782,30 @@ void VideoEncoder::initializeEncoder(
...
-    // Use first listed pixel format as default (often yuv420p).
-    // This is similar to FFmpeg's logic:
-    // https://www.ffmpeg.org/doxygen/4.0/decode_8c_source.html#l01087
-    // If pixel formats are undefined for some reason, try yuv420p
-    outPixelFormat_ = (formats && formats[0] != AV_PIX_FMT_NONE)
-        ? formats[0]
-        : AV_PIX_FMT_YUV420P;
+    if (frames_.device().is_cuda()) {
+      // Default to nv12 pixel format when encoding on GPU.
+      outPixelFormat_ = AV_PIX_FMT_NV12;
```
NicolasHug: To avoid duplication of hard-coded constants, we can use the interface's `CUDA_PIXEL_FORMAT` field here (if it's public, which I think it is?)

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -383,16 +383,120 @@ std::string CudaDeviceInterface::getDetails() {
...
+// To perform color conversion in NPP, we are required to provide these color
+// conversion matrices to ColorTwist functions, for example,
+// `nppiRGBToNV12_8u_ColorTwist32f_C3P2R_Ctx`.
+// https://docs.nvidia.com/cuda/npp/image_color_conversion.html
+//
+// These offsets are added in the 4th column of each conversion matrix below.
+// - In limited range, Y is offset by 16 to add the lower margin.
+// - In both color ranges, U,V are offset by 128 to be centered around 0.
+//
+// RGB to YUV conversion matrices to use in NPP color conversion functions
```
NicolasHug, line 441: Thanks for the note, it will be super useful if we ever need to go back to this in the future.

NicolasHug (approved):
Thanks @Dan-Flores !

PR #1129, opened 2025-12-12
Refactor CI jobs

NicolasHug:
This PR refactors most of our CI job. More refac will probably come later to finalize things.

It's easier to review these few commits individually:

- [ed5f8c5](https://github.com/meta-pytorch/torchcodec/pull/1129/commits/ed5f8c55fe51787b3df242d6a224d7dc8a456109) puts our "install pytorch and maybe torchvision" logic in a separate script
- [80aea14](https://github.com/meta-pytorch/torchcodec/pull/1129/commits/80aea1431b4e9176af62ef96e64a07a8613a8a24) installs test-dependencies in a separate script
- [25e0b77](https://github.com/meta-pytorch/torchcodec/pull/1129/commits/25e0b771961ecec1a1c9bab82a77c843c8d3f255) puts the  ffmpeg installation as well as the torchcodec wheel install in separate scripts
- [ab32c0c](https://github.com/meta-pytorch/torchcodec/pull/1129/commits/ab32c0cfff430b8570453e525a1e162b24614114) removes `docs.yaml` to add the doc job as a dependency of the cuda job. This is more efficient: the doc job was building a wheel on its own. But that wheel was  already created by the cuda job. The doc job now just depends on the wheel-building part of the cuda job. They still both run in parallel.

The rest of the commits are just attempts at fixing stuff (reordering some steps), which I think you can safely ignore

### DIFF
--- a/.github/workflows/linux_cuda_wheel.yaml
+++ b/.github/workflows/linux_cuda_wheel.yaml
@@ -139,7 +139,7 @@ jobs:
       CUDA_VERSION: '12.6'
       FFMPEG_VERSION: '7'
     container:
-      image: "pytorch/manylinux2_28-builder:cuda${{ env.CUDA_VERSION }}"
+      image: "pytorch/manylinux2_28-builder:cuda12.6"  # must be same as env!!
       options: "--gpus all -e NVIDIA_DRIVER_CAPABILITIES=video,compute,utility"
     needs: build
     steps:


.github/workflows/windows_wheel.yaml
```diff
@@ -74,12 +74,14 @@ jobs:
...
-          path: pytorch/torchcodec/dist/
+      - name: Check out repo
+        uses: actions/checkout@v3
+
+      - name: Remove src/ folder
+        run: bash packaging/remove_src.sh
+
       - name: Setup conda env
-        uses: conda-incubator/setup-miniconda@v2
+        uses: conda-incubator/setup-miniconda@v3
```
Dan-Flores, line 84: I'm curious what effect upgrading this version had? Perhaps we should update the remaining uses of miniconda@v2 in `reference_resources.yaml` and `lint.yaml`.

.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -109,49 +112,90 @@ jobs:
...
         run: |
-          wheel_path=`find pytorch/torchcodec/dist -type f -name "*cu${{ env.cuda_version_without_periods }}-cp${{ env.python_version_without_periods }}*.whl"`
-          echo Installing $wheel_path
-          ${CONDA_RUN} python -m pip install $wheel_path -vvv
+          ${CONDA_RUN} time python benchmarks/decoders/gpu_benchmark.py --devices=cuda:0,cpu --resize_devices=none
+
+  build-docs:
+    runs-on: linux.4xlarge.nvidia.gpu
+    env:
+      PYTHON_VERSION: '3.10'
```
Dan-Flores, line 138: Lets keep the comment clarifying why python 3.10 is used here:
https://github.com/meta-pytorch/torchcodec/blob/1e171fa71dc56dd1b0b8a1a4dfa589685de7ae13/.github/workflows/docs.yaml#L54-L57

.github/workflows/cpp_tests.yaml
```diff
@@ -37,17 +37,7 @@ jobs:
       - name: Update pip
```
mollyxu, line 37: Would this be something that we could move to the `install_pytorch.sh` script or would it be better to keep this separate?

.github/workflows/linux_cuda_wheel.yaml
Dan-Flores: I see this comment still exists above on lines 64-68, resolved.

.github/workflows/cpp_tests.yaml
Dan-Flores: We could move `Update pip` into `install_pytorch.sh`, but I would say it makes sense to keep here, as it is a high level workflow step (like `Setup conda env` or `Install PyTorch`), and the `install_pytorch.sh` script should be focused on one dependency.

.github/workflows/windows_wheel.yaml
NicolasHug: The old versions typically stop working at some point, because they get deprecated, or use dependencies that don't exist anymore. It's usually best to use the latest version that works. Yes you're right, it'd be good to align everywhere on the same version!

Issue #1053, opened 2025-11-14
Add UniqueAVDictionary class and utilize it in VideoEncoder


Dan-Flores:
In #1050, VideoEncoder is updated to use raw `AVDictionary` pointers (`avFormatOptions_`, `avCodecOptions`) and manual calls to `av_dict_free()` to clean up those objects.

```
void sortCodecOptions(
    const std::map<std::string, std::string>& extraOptions,
    AVDictionary** codecDict,
    AVDictionary** formatDict)
. . .
av_dict_free(&avCodecOptions);
```
 
We should create a `UniqueAVDictionary` wrapper class (similar to existing `UniqueAVFrame`) that automatically manages the AVDictionary objects using the RAII pattern, then refactor VideoEncoder to use UniqueAVDictionary and remove manual cleanup calls.

PR #1136, opened 2025-12-16
Add UniqueAVDictionary class

Dan-Flores:
This PR resolves issue #1053 by implementing `UniqueAVDictionary`, similar to a `unique_ptr`, with some workarounds.
* Primarily, `getAddress()` is implemented to return a double pointer (`AVDictionary**`) which is needed in certain FFmpeg functions. 
  * For our use case, `unique_ptr` is limiting. 
  * [FFmpeg functions modify the underlying pointer's location](https://github.com/FFmpeg/FFmpeg/blob/master/libavutil/dict.c#L121-L122). If we tried to use the temporary pointer copy returned by `unique_ptr`'s `get()` function, we would not be able to update the stored `AVDictionary`. 
  
To follow the rule of 5, the copy and move operators are explicitly deleted.
* The copy operators are not permitted in our `unique_ptr` implementation.
* The move operators are not needed at this time, so I thought it would be simplest to delete it until we need it. As a result, the move constructor operator of `VideoEncoder` is now deleted.

### DIFF
--- a/src/torchcodec/_core/FFMPEGCommon.h
+++ b/src/torchcodec/_core/FFMPEGCommon.h
@@ -103,8 +103,8 @@ using UniqueAVBufferSrcParameters = std::unique_ptr<
     AVBufferSrcParameters,
     Deleterv<AVBufferSrcParameters, void, av_freep>>;
 
-// Wrapper class for AVDictionary to support FFmpeg's double-pointer functions
-// such as av_dict_set.
+// Wrapper class for AVDictionary, similar to unique_ptr, to support FFmpeg's
+// functions that require a double-pointer to AVDictionary, such as av_dict_set.
 // https://ffmpeg.org/doxygen/trunk/group__lavu__dict.html#ga8d9c2de72b310cef8e6a28c9cd3acbbe
 class UniqueAVDictionary {
  private:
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -103,6 +103,39 @@ using UniqueAVBufferSrcParameters = std::unique_ptr<
...
 
+// Wrapper class for AVDictionary, similar to unique_ptr, to support FFmpeg's
+// functions that require a double-pointer to AVDictionary, such as av_dict_set.
+// https://ffmpeg.org/doxygen/trunk/group__lavu__dict.html#ga8d9c2de72b310cef8e6a28c9cd3acbbe
+class UniqueAVDictionary {
+ private:
+  AVDictionary* dict_ = nullptr;
+
+ public:
+  UniqueAVDictionary() = default;
```
mollyxu, line 114: Is there a reason why we have to explicitly define the default constructor? Is it better practice in terms of clarity?

mollyxu:
Thank you for the PR @Dan-Flores ! I found it to be really easy to read because of the linked context and informative comments.

src/torchcodec/_core/FFMPEGCommon.h
Dan-Flores: From [cppreference](https://en.cppreference.com/w/cpp/language/default_constructor.html), "If there is no user-declared constructor ... for a class type, the compiler will implicitly declare a default constructor" 

Since we explicitly declared the copy and move constructors (with `= delete`) below, we are required to set the constructor to `default`, as it will not be generated automatically.

PR #1046, opened 2025-11-13
Improve load_torchcodec_shared_libraries error reporting

JAEarly:
Added when investigated issues related to https://github.com/meta-pytorch/torchcodec/issues/1023. 
Doesn't fix the underlying issue, but improves error reporting for users.

PR #1138, opened 2025-12-17
Enhance error handling in load_torchcodec_shared_libraries by capturing full exception tracebacks for better debugging.

suo:
The existing message will hide whether, e.g. there is a symbol mismatch which is super annoying.

### DIFF
--- a/src/torchcodec/_core/ops.py
+++ b/src/torchcodec/_core/ops.py
@@ -10,6 +10,7 @@ import json
 import os
 import shutil
 import sys
+import traceback
 import warnings
 from contextlib import nullcontext
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


scotts:
Hi, @suo! The challenge with this approach is that users will hit a giant wall of text where some of the errors are actually expected. That is, if the user has FFmpeg 6 installed, this will show all of the errors when attempting to load the libraries for FFmpeg 4, 5, 6 7 and 8. The user only cares about the errors for FFmpeg 6, but we don't know that; this code is how we figure out which version of FFmpeg will work.

With that said, #1046 proposed basically the same fix. The wall of text is a lot, but folks seem to want the information. The only other way to get this info is to make this hack yourself, so I'm inclined to say let's do this until we get logging in. *However*, we should also probably update the error message to give some actual guidance on understanding what's going on. @Dan-Flores, @mollyxu, thoughts?

Dan-Flores:
I agree @scotts, lets merge this PR to enable better debugging.

mollyxu:
Yep I agree @scotts @Dan-Flores! I can work on a follow-up PR that gives users a bit more context on the wall of text

suo:
thanks all!

PR #1141, opened 2025-12-17
Give context on exception tracebacks

mollyxu:
Give users a bit more context on the wall of text that they would see when they encounter an exception.

### DIFF
--- a/src/torchcodec/_core/ops.py
+++ b/src/torchcodec/_core/ops.py
@@ -77,10 +77,10 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
         f"""Could not load libtorchcodec. Likely causes:
           1. FFmpeg is not properly installed in your environment. We support
              versions 4, 5, 6, 7, and 8, and we attempt to load libtorchcodec
-             for each of those versions. Errors for
-             versions not installed on your system are expected; only the error
-             for your installed FFmpeg version is relevant. On Windows, ensure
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/ops.py
```diff
@@ -83,6 +83,11 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
              table:
              https://github.com/pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec.
           3. Another runtime dependency; see exceptions below.
+
+        Note: We attempt to load libtorchcodec for FFmpeg versions 8 down to 5.
```
scotts: Let's incorporate this in point 1 above, since it's related. That would also make it so that we don't need to restate the supported versions.

src/torchcodec/_core/ops.py
```diff
@@ -76,13 +76,17 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
     raise RuntimeError(
         f"""Could not load libtorchcodec. Likely causes:
           1. FFmpeg is not properly installed in your environment. We support
-             versions 4, 5, 6, 7, and 8. On Windows, ensure you've installed
-             the "full-shared" version which ships DLLs.
+             versions 4, 5, 6, 7, and 8, and we attempt to load libtorchcodec
+             for each of those versions. Errors for
```
scotts: Let's fix the formatting so that it looks word-wrapped.

PR #1142, opened 2025-12-17
Remove remaining tests from collection

mollyxu:
## Remove remaining skipped tests from collection in internal environment

Prevent internal CI failures from skipped tests by excluding skipped tests from collection in internal environment.

TorchCodec tests were flaky / broken internal because:
- Tests were using runtime `pytest.skip()` calls, so they were collected but then skipped during execution

Follow up to #1068

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1191,13 +1191,6 @@ class TestVideoDecoder:
             assert_frames_equal(ref_frame3, frames[1].data)
             assert_frames_equal(ref_frame5, frames[2].data)
 
-    # Non-regression test for
-    # https://github.com/pytorch/torchcodec/issues/677 and
-    # https://github.com/pytorch/torchcodec/issues/676.
-    # More accurately, this is a non-regression test for videos which do
-    # *not* specify pts values (all pts values are N/A and set to
-    # INT64_MIN), but specify *dts* value - which we fallback to.
-    #
     # The test video we have is from
     # https://huggingface.co/datasets/raushan-testing-hf/videos-test/blob/main/sample_video_2.avi
     # We can't check it into the repo due to potential licensing issues, so
@@ -1208,6 +1201,12 @@ class TestVideoDecoder:
     @pytest.mark.parametrize("seek_mode", ("exact", "approximate"))
     @pytest.mark.skip(reason="TODO: Need video with no pts values.")
     def test_pts_to_dts_fallback(self, seek_mode):
+        # Non-regression test for
+        # https://github.com/pytorch/torchcodec/issues/677 and
+        # https://github.com/pytorch/torchcodec/issues/676.
+        # More accurately, this is a non-regression test for videos which do
+        # *not* specify pts values (all pts values are N/A and set to
+        # INT64_MIN), but specify *dts* value - which we fallback to.
         path = "/home/nicolashug/Downloads/sample_video_2.avi"
         decoder = VideoDecoder(path, seek_mode=seek_mode)
         metadata = decoder.metadata


test/test_decoders.py
```diff
@@ -1295,16 +1301,6 @@ def test_10bit_videos(self, device, asset):
         # This just validates that we can decode 10-bit videos.
         # TODO validate against the ref that the decoded frames are correct
 
-        if device == "cuda:beta" and asset is H264_10BITS:
```
mollyxu, line 1298: no longer applicable because the BETA interface fallsback to the CPU now

test/test_encoders.py
```diff
@@ -34,6 +34,33 @@
...
 
+VIDEO_ENCODE_PARAMS = (
+    {"pixel_format": "yuv444p", "crf": 0, "preset": None},
+    {"pixel_format": "yuv420p", "crf": 30, "preset": None},
+    {"pixel_format": "yuv420p", "crf": None, "preset": "ultrafast"},
+    {"pixel_format": "yuv420p", "crf": None, "preset": None},
+)
+
+
+def _get_format_encode_params(formats, encode_params=VIDEO_ENCODE_PARAMS):
```
mollyxu: to skip tests that depend on both format and encoding parameters 
(ie. avi / flv with yuv444p)

test/test_decoders.py
```diff
@@ -1629,17 +1636,20 @@ def test_beta_cuda_interface_get_frames_played_at(self, asset, seek_mode):
...
             BT709_FULL_RANGE,
             TEST_SRC_2_720P_H265,
-            AV1_VIDEO,
+            pytest.param(
+                AV1_VIDEO,
+                marks=pytest.mark.skipif(
+                    in_fbcode(),
+                    reason="AV1 CUDA not supported internally",
+                ),
+            ),
```
Dan-Flores, line 1632: Nice job on these! It is easier to understand where the test is skipped now that the skip occurs in the parametrization.

test/test_encoders.py
Dan-Flores: Ideally, we would handle all parametrization for a test in one place.

By adding a helper function, it is harder to add new test cases and to understand what is being tested. If we cannot fit the logic into a `pytest.param`, it might be best to let these test cases appear skipped on CI.

This is one of the downsides to having a test with many parameters -  we cannot easily isolate these parameter combinations that fail.

test/utils.py
```diff
@@ -45,11 +45,29 @@ def needs_ffmpeg_cli(test_item):
 _CUDA_BETA_DEVICE_STR = "cuda:beta"
 
 
-def all_supported_devices():
+def all_supported_devices(*, cuda_marks: list[pytest.MarkDecorator] | None = None):
```
Dan-Flores: I see how this covers the test skip in `test_get_frame_at_av1`, its a tricky case.

Since there's only one case where we need the `cuda_marks` argument, I'm not sure its worth exposing at the moment. While other usages of `all_supported_devices` are not affected by the new arg, its not obvious that this function should be used to add marks to the `cuda` device params. 

In this case, it might be bes to keep the `pytest.skip` in the `test_get_frame_at_av1`

test/conftest.py
```diff
@@ -38,8 +38,13 @@ def pytest_collection_modifyitems(items):
...
 
         if in_fbcode():
-            # fbcode doesn't like skipping tests, so instead we  just don't collect the test
+            # fbcode doesn't like skipping tests, so instead we just don't collect the test
             # so that they don't even "exist", hence the continue statements.
+
+            # Current we skip all marked tests internally, whether they are marked
+            # with needs_ffmpeg_cli, skip, or skipif. Future skipped tests
+            # should follow the pattern of needs_cuda and needs_ffmpeg_cli if we are
+            # skipping a test because of a dependency.
```
NicolasHug: I have a bit of trouble understanding this comment, can you clarify what we mean by "follow the pattern of needs_cuda and needs_ffmpeg_cli if we are skipping a test because of a dependency" ?

I think that part of the reason I'm confused is that this mentioned `needs_cuda`, which isn't used within this `if in_fbcode()` block - maybe this comment was meant to be somewhere else?

test/test_decoders.py
```diff
@@ -1191,6 +1197,7 @@ def get_some_frames(decoder):
             assert_frames_equal(ref_frame3, frames[1].data)
             assert_frames_equal(ref_frame5, frames[2].data)
 
+    @pytest.mark.skip(reason="TODO: Need video with no pts values.")
```
NicolasHug: Make sure to also move the related comment block here as well. Starting with "The test video we have ...".

test/conftest.py
mollyxu: Thanks for the review! The original comment was a bit confusing so I clarified it and moved it to a better section. The intention was to encourage people to follow the decorator pattern for `needs_cuda` and `needs_ffmpeg_cli` rather than individual `skip`when we are conditionally skipping tests due to dependencies.

mollyxu:
Thanks @Dan-Flores @NicolasHug for the review!

PR #1144, opened 2025-12-20
Refactor logic to close AVIOContext

Dan-Flores:
This PR consolidates the logic used to flush and close `AVIOContext`, and should fix the flaky test bug described in https://github.com/meta-pytorch/torchcodec/issues/724.

The issue is related to the member variables `avFormatContext_` and `avioContextHolder_` pointing to the same `AVIOContext` object, and the logic of `close_avio`. 
https://github.com/meta-pytorch/torchcodec/blob/50686bb306dae586efa53533e43dba0c82aaedf9/src/torchcodec/_core/Encoder.cpp#L185
In `close_avio`, if `avioContextHolder_` is not defined (if using `to_file`), the `AVIOContext` pointer is correctly set to `nullptr`, preventing double frees in the member variables destructors. **Otherwise, it will be destroyed in the member variable destructors.** In those destructors, both will attempt to free the same address, as it is never set to a `nullptr`: https://github.com/meta-pytorch/torchcodec/blob/50686bb306dae586efa53533e43dba0c82aaedf9/src/torchcodec/_core/AVIOContextHolder.cpp#L50-L54
https://github.com/meta-pytorch/torchcodec/blob/50686bb306dae586efa53533e43dba0c82aaedf9/src/torchcodec/_core/FFMPEGCommon.h#L68-L70
The updated `closeAVIOContext` function avoids this by 
* Returning early if `avFormatContext_` or `avFormatContext_->pb` are set to `nullptr` (so it is safe to call multiple times).
* Calls `avio_close` on the `AVIOContext` pointer if `avioContextHolder_` is not defined (when using `to_file`). Otherwise, the reference in `avFormatContext_` will be set to `nullptr`.
  * During `avioContextHolder`'s destructor,  the `AVIOContext` will be freed (only once!).

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -107,23 +107,23 @@ AVSampleFormat findBestOutputSampleFormat(const AVCodec& avCodec) {
 }
 
 void closeAVIOContext(
-    AVFormatContext* formatContext,
+    AVFormatContext* avFormatContext,
     AVIOContextHolder* avioContextHolder) {
-  if (!formatContext || !formatContext->pb) {
+  if (!avFormatContext || !avFormatContext->pb) {
     return;
   }
 
-  if (formatContext->pb->error == 0) {
-    avio_flush(formatContext->pb);
+  if (avFormatContext->pb->error == 0) {
+    avio_flush(avFormatContext->pb);
   }
 
-  if (avioContextHolder == nullptr) {
-    if (formatContext->pb->error == 0) {
-      avio_close(formatContext->pb);
+  if (!avioContextHolder) {
+    if (avFormatContext->pb->error == 0) {
+      avio_close(avFormatContext->pb);
     }
   }
 
-  formatContext->pb = nullptr;
+  avFormatContext->pb = nullptr;
 }
 
 } // namespace


src/torchcodec/_core/Encoder.cpp
```diff
@@ -106,26 +106,30 @@ AVSampleFormat findBestOutputSampleFormat(const AVCodec& avCodec) {
...
-      }
-      // avoids closing again in destructor, which would segfault.
-      avFormatContext_->pb = nullptr;
+  if (!avioContextHolder) {
+    if (avFormatContext->pb->error == 0) {
+      avio_close(avFormatContext->pb);
     }
   }
+
+  avFormatContext->pb = nullptr;
```
NicolasHug, line 126: Can you check my understanding that the main logic difference is that we now set `avFormatContext->pb = nullptr;` **after** the `if (!avioContextHolder) { ...}` block, and not inside it?

NicolasHug (approved):
Thanks @Dan-Flores , let's try this. The refactor is useful regardless of the logic change!

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Yes exactly, after and outside of the condition.

Issue #1133, opened 2025-12-15
FPS: enforcing an output frame rate


scotts:
### 🚀 The feature

The [FFmpeg fps filter](https://ffmpeg.org/ffmpeg-filters.html#fps-1) is often used during decoding to control the output frame rate. From the docs:

> Convert the video to specified constant frame rate by duplicating or dropping frames as necessary.

At first, it seems obvious to support this functionality in TorchCodec as a [`DecoderTransfomr`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/transforms/_decoder_transforms.py#L16). However, fps is not 1:1. That is, one input frame does not always produce one output frame. Transforms like resize and crop *are* 1:1; one input frame always produces one output frame. Transforms that are not 1:1 do not fit into our transforms framework as they break a lot of other assumptions in `VideoDecoder`.

Instead, the more natural way to support this functionality is as a parameter to our time-based batch API:

```
def get_frames_played_in_range(
    self,
    start_seconds: float,
    stop_seconds: float,
    fps: int | None = None,
) -> FrameBatch:
```
The way to read this API if fps is:

* `None`, default: returns decoded frames between `start_seconds` and `stop_seconds` at the input frame rate. This is the existing behavior.
* `int`: returns decoded frames between `start_seconds` and `stop_seconds` where the output frame rate matches the value of `fps`.

The real logic for implementing this feature will probably need to live in [`SingleStreamDecoder::getFramesPlayedInRange`](https://github.com/meta-pytorch/torchcodec/blob/1e171fa71dc56dd1b0b8a1a4dfa589685de7ae13/src/torchcodec/_core/SingleStreamDecoder.cpp#L845). It will also need an `fps` parameter. While decoding frames, we will need to drop or duplicate frames to match the requested frame rate.

Regarding testing, we should aim for an exact match for what the `ffmpeg` CLI produces. For testing, I think we won't need to make any changes to our testing code as we do support a [`filters`](https://github.com/meta-pytorch/torchcodec/blob/1e171fa71dc56dd1b0b8a1a4dfa589685de7ae13/test/generate_reference_resources.py#L41) parameter when generating reference frames and when [retrieving them](https://github.com/meta-pytorch/torchcodec/blob/1e171fa71dc56dd1b0b8a1a4dfa589685de7ae13/test/utils.py#L410).

### Motivation, pitch

[...]

Issue #428, opened 2024-12-09
Stricter GPU frame comparison


NicolasHug:
Placeholder, will fill later.

NicolasHug:
Oops, no idea what I had in mind with that. 🤷‍♂️

PR #1148, opened 2025-12-30
Add FPS to `get_frames_played_in_range()`

mollyxu:
## Add FPS to `get_frames_played_in_range()`

Context: #1133 

This PR adds support for specifying a target frame rate when calling `get_frames_played_in_range()`. When `target_fps` is provided, the method resamples the video to output frames at evenly-spaced timestamps matching the target FPS.

```
def get_frames_played_in_range(
    self,
    start_seconds: float,
    stop_seconds: float,
    fps: float | None = None,
) -> FrameBatch:
```
The way to read this API if fps is:

* `None`, default: returns decoded frames between `start_seconds` and `stop_seconds` at the input frame rate. This is the existing behavior.
* `float`: returns decoded frames between `start_seconds` and `stop_seconds` where the output frame rate matches the value of `fps`.

### Implementation
1. **Frame count calculation:** `num_output_frames = round((stop - start) * fps)`
2. **Timestamp generation:** For each output frame `i`, compute `target_pts = start + i / fps`
3. **Source frame lookup:** Find the source frame displayed at `target_pts` using lower-bound search
4. **Decode optimization:** When consecutive output frames map to the same source frame (common when upsampling), the already-decoded tensor is copied rather than re-decoded

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -1137,10 +1137,11 @@ class TestVideoDecoder:
 
         # All duplicated frames should have the same content as frame 0
         frame0_data = decoder.get_frame_at(0).data
-        for i in range(len(frames_high_fps)):
-            torch.testing.assert_close(
-                frames_high_fps.data[i], frame0_data, atol=0, rtol=0
-            )
+        if not (device == "cuda" and get_ffmpeg_major_version() == 4):
+            for i in range(len(frames_high_fps)):
+                torch.testing.assert_close(
+                    frames_high_fps.data[i], frame0_data, atol=0, rtol=0
+                )
 
         # Test that fps=None returns the original behavior (same as not passing fps)
         frames_no_fps = decoder.get_frames_played_in_range(start_seconds, stop_seconds)
@@ -1148,9 +1149,10 @@ class TestVideoDecoder:
             start_seconds, stop_seconds, fps=None
         )
         assert len(frames_no_fps) == len(frames_none_fps)
-        torch.testing.assert_close(
-            frames_no_fps.data, frames_none_fps.data, atol=0, rtol=0
-        )
+        if not (device == "cuda" and get_ffmpeg_major_version() == 4):
+            torch.testing.assert_close(
+                frames_no_fps.data, frames_none_fps.data, atol=0, rtol=0
+            )
 
     @pytest.mark.parametrize("device", all_supported_devices())
     @pytest.mark.parametrize("seek_mode", ("exact", "approximate"))


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,116 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
             std::to_string(maxSeconds.value()) + ").");
   }
 
+  // Resample frames to match the target frame rate
```
mollyxu, line 910: We could add an early break if requested `fps` is the same as the current fps. Not sure if it is a necessary use case

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -844,7 +844,8 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedAt(
 
 FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     double startSeconds,
-    double stopSeconds) {
+    double stopSeconds,
+    std::optional<double> fps) {
```
scotts, line 848: I have a slight preference for `SingleStreamDecoder` to not take optionals, and for the bridge code in `custom_ops.cpp` to take care of optional logic. But, there are other APIs in `SingleStreamDecoder` which do take optionals, I'll invite @Dan-Flores and @NicolasHug to make a call here.

src/torchcodec/_core/SingleStreamDecoder.cpp
scotts: Let's handle that with a follow-up. Because we'll need to do equality on a double, there may be some subtle things to handle.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,112 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
             std::to_string(maxSeconds.value()) + ").");
   }
 
+  // Resample frames to match the target frame rate
+  if (fps.has_value()) {
+    TORCH_CHECK(
+        fps.value() > 0,
+        "fps must be positive, got " + std::to_string(fps.value()));
+
+    double fpsVal = static_cast<double>(fps.value());
```
scotts: Nit: no need to do a cast, as `fps.value()` returns a `double`.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,112 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+    double frameDuration = 1.0 / fpsVal;
+
+    // Calculate the exact number of output frames for the half-open range
+    // [startSeconds, stopSeconds). FFmpeg's fps filter uses rounding to
+    // determine the frame count, which we replicate here for consistency.
+    // Examples:
+    // - 35.23 fps, 1.0s: round(35.23) = 35 frames
+    // - 35.87 fps, 1.0s: round(35.87) = 36 frames
```
scotts: Can you put a link to the resource you learned the exact techniques that the fps filter is using? I assume it's the source code, in which case a stable link to that code would be great. We want to make it easy for others to make changes to this code, and they may need to look at that as well.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,112 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+    int64_t numOutputFrames = static_cast<int64_t>(std::round(product));
+
+    if (numOutputFrames <= 0) {
+      FrameBatchOutput frameBatchOutput(
+          0,
+          resizedOutputDims_.value_or(metadataDims_),
+          videoStreamOptions.device);
+      frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+      return frameBatchOutput;
+    }
```
scotts: Is it ever valid for `numOutputFrames < 0`? That is, we ensure that `startSeconds <= stopSeconds` at the very top, and we also ensure that `fpsVal > 0`. So I think that if `numOutputFrames < 0`, that is an error.

And in the case that `numOutputFrames == 0`, I don't think we actually need to special case it. I think the remaining logic should be robust to that case.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,112 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+      int64_t outputIdx =
+          static_cast<int64_t>(std::round((framePts - startSeconds) * fpsVal));
+
+      if (outputIdx >= numOutputFrames) {
+        break;
+      }
+      if (outputIdx >= 0) {
+        sourceFrameIndices[outputIdx] = j;
+      }
+    }
```
scotts: This mapping logic will need to be seek mode aware. In approximate mode, `streamInfo.allFrames` won't be available and we'll need to do math to calculate the value. Check out `secondsToIndexLowerBound()` for inspiration.

We'll also need to test exact and approximate seek modes. :)

test/test_decoders.py
```diff
@@ -1098,6 +1101,127 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
         with pytest.raises(ValueError, match="Invalid stop seconds"):
             frame = decoder.get_frames_played_in_range(0, 23)  # noqa
 
+    @pytest.mark.parametrize("device", all_supported_devices())
```
scotts, line 1100: Also parameterize on seek mode - I think we can skip custom and just do `"exact"` and `"approximate"`. Based on the code above, I think `"approximate"` will currently fail.

test/test_decoders.py
```diff
@@ -1098,6 +1101,127 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        start_seconds = decoder.get_frame_at(0).pts_seconds
+        stop_seconds = start_seconds + 1.0
+
+        with pytest.raises(RuntimeError, match="fps must be positive"):
+            decoder.get_frames_played_in_range(start_seconds, stop_seconds, fps=0)
+
+        with pytest.raises(RuntimeError, match="fps must be positive"):
+            decoder.get_frames_played_in_range(start_seconds, stop_seconds, fps=-10)
+
+    @needs_ffmpeg_cli
```
scotts: Let's also parametrize this test on seek mode.

test/test_decoders.py
```diff
@@ -1098,6 +1101,127 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
             frame = decoder.get_frames_played_in_range(0, 23)  # noqa
 
+    @pytest.mark.parametrize("device", all_supported_devices())
+    def test_get_frames_played_in_range_with_fps(self, device):
+        decoder, _ = make_video_decoder(
+            NASA_VIDEO.path, device=device, seek_mode="exact"
+        )
+
+        source_fps = decoder.metadata.average_fps
+        duration_seconds = 1.0
```
scotts, line 1108: We'll need some tests that extract a longer duration. In particular, we'll want to make sure we have tests that decode the entire video. I suspect we'll hit some edge-cases in our dropping and duping logic when we decode all frames.

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: It seems natural to me to have an optional here. Letting `custom_ops.cpp` be the bridge, I think, would mean that we'd have to create more private method to split the implementation (for e.g. the common checks at the beginning, and the common logic at the end).

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -452,32 +452,31 @@ def get_frames_played_at(self, seconds: torch.Tensor | list[float]) -> FrameBatc
         )
 
     def get_frames_played_in_range(
-        self, start_seconds: float, stop_seconds: float
+        self, start_seconds: float, stop_seconds: float, fps: float | None = None
```
NicolasHug, line 454: If we expose the fps functionality in this method, then for convenience I think we should make sure to also expose a `get_all_frames(fps)` method, so that users don't have to manually specify both `start` and `stop`. (Exposing this `get_all_frames()` method is something we should do regardless).

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,125 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+    int64_t numOutputFrames = static_cast<int64_t>(std::round(product));
+
+    // Generate target timestamps using index-based calculation to avoid
+    // floating-point accumulation errors
+    std::vector<double> targetTimestamps;
+    targetTimestamps.reserve(numOutputFrames);
+    for (int64_t i = 0; i < numOutputFrames; ++i) {
+      double targetPts = startSeconds + i * frameDuration;
+      targetTimestamps.push_back(targetPts);
+    }
```
NicolasHug: This logic looks a lot like what we already have in the [time-based samplers](https://meta-pytorch.org/torchcodec/stable/api_ref_samplers.html) with the `seconds_between_frames` parameter:

https://github.com/meta-pytorch/torchcodec/blob/50686bb306dae586efa53533e43dba0c82aaedf9/src/torchcodec/samplers/_time_based.py#L94-L101


I could wrong, but I suspect there is some overlap between the new `fps` parameter and the existing `seconds_between_frames` parameter, in terms of functionality. It's been a while since I implemented the latter, so I don't fully remember the details and how exactly it behaves (does it also drop and duplicate frames in the same way? I'm not sure).

Generally, a good design principle is that [*There should be one-- and preferably only one --obvious way to do it.*](https://peps.python.org/pep-0020/)

I think it might be worth resolving this ambiguity - @mollyxu, can I ask you to do a bit of investigation and figure out the main behavior differences between `fps` and `seconds_between_frames`? Once we know how they differ in terms of behavior, it'll be easier to figure out a correct, unambiguous and holistic API for fps output.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,125 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+
+    // TODO: add an early break if requested fps is the same as the current fps
+
+    double fpsVal = fps.value();
+    double frameDuration = 1.0 / fpsVal;
+
+    // Calculate the exact number of output frames for the half-open range
+    // [startSeconds, stopSeconds). FFmpeg's fps filter drops the frames until
+    // the next frame would overshoot.
+    // https://github.com/FFmpeg/FFmpeg/blob/n7.1/libavfilter/vf_fps.c#L290-L291
```
scotts: We should make this https://github.com/FFmpeg/FFmpeg/blob/b08d7969c550a804a59511c7b83f2dd8cc0499b8/libavfilter/vf_fps.c#L290-L291 instead. That is, we want a permalink which includes the commit hash in the URL. That will never change, even as the code is changed on main. As the code on main changes, lines 290 and 291 will eventually be different code.

(On GitHub, to get a permalink, click the three dots next to a line number, and it brings up a dialog box you can use to copy the permalink.)

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        frame1_pts = decoder.get_frame_at(1).pts_seconds
+        stop_seconds = start_seconds + duration_seconds
+
+        # Test downsampling: request lower fps than source
+        fps_low = 5
+        frames_low_fps = decoder.get_frames_played_in_range(
+            start_seconds, stop_seconds, fps=fps_low
+        )
+        expected_frames_low = round(duration_seconds * fps_low)
+        assert len(frames_low_fps) == expected_frames_low
```
NicolasHug, line 1119: This is a great test overall. I think we could assert here that `frames_low_fps[0]` is frame 0, and that `frames_low_fps[1]` is *not* frame 1?

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        )
+        # All frames should be duplicates of frame 0 since we're within frame 0's display time
+        frame_duration = frame1_pts - start_seconds
+        expected_frames_high = round(frame_duration * fps_high)
+        assert len(frames_high_fps) == expected_frames_high
+
+        # All duplicated frames should have the same content as frame 0
+        frame0_data = decoder.get_frame_at(0).data
+        for i in range(len(frames_high_fps)):
+            assert_frames_equal(frames_high_fps.data[i], frame0_data)
```
NicolasHug: Here and everywhere else I think we'll want to use `torch.testing.assert_close(..., atol=0)`, like you did in other places?

`assert_frames_equal` has looser tolerances because they're needed when testing against the FFmpeg  reference on GPU or other platforms like Macos. But here, since we're comparing output coming from TorchCodec itself, we should be able to enforce bitwise equality.

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        with pytest.raises(RuntimeError, match="fps must be positive"):
+            decoder.get_frames_played_in_range(start_seconds, stop_seconds, fps=0)
+
+        with pytest.raises(RuntimeError, match="fps must be positive"):
+            decoder.get_frames_played_in_range(start_seconds, stop_seconds, fps=-10)
+
+    @pytest.mark.parametrize("fps", [5.0, 24.0, 30.1, 60.0])
+    def test_get_frames_played_in_range_fps_matches_torchvision(self, fps):
+        """Test that TorchCodec's fps output matches torchvision's resampling logic."""
+        video_path = str(NASA_VIDEO.path)
```
NicolasHug: Nit: no need to define the video_path variable, it's only used in a single place. We could just pass `VideoDecoder(NASA_VIDEO.path)`, which is what we usually do.

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        )
+
+        # Compute expected indices using torchvision's resampling logic:
+        # https://github.com/pytorch/vision/blob/1e53952f57462e4c28103835cf1f9e504dbea84b/torchvision/datasets/video_utils.py#L278
+        # For each output frame i, select source frame at index floor(i * step)
+        # where step = original_fps / target_fps
+        original_fps = decoder.metadata.average_fps
+        step = original_fps / fps
+        if step.is_integer():
+            expected_indices = slice(None, None, int(step))
```
NicolasHug: This step was meant to be a perf optimization. To simplify our mental model around the reference implem, I think we can remove it?

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+            fps=fps,
+        )
+
+        # Get all source frames in the range
+        all_source_frames = decoder.get_frames_played_in_range(
+            start_seconds=start_seconds,
+            stop_seconds=stop_seconds,
+        )
+
+        # Compute expected indices using torchvision's resampling logic:
```
NicolasHug, line 1199: The ref implem is re-defined in two tests: best to extract it out as helper method and re-use it!

EDIT: actually, I might suggest below that we just merge both tests, in which case this could be in-lined within the test

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+                .floor()
+                .to(torch.int64)
+            )
+        expected_frames = all_source_frames.data[expected_indices]
+
+        # Verify frame counts match
+        assert len(tc_frames_batch) == len(expected_frames), (
+            f"Frame count mismatch: TorchCodec={len(tc_frames_batch)}, "
+            f"expected={len(expected_frames)}"
+        )
```
NicolasHug: We usually don't add extra error messages on tests, pytest is very good at providing a good error message already.

Note also that this assertion is somewhat redundant with the `torch.testing.assert_close` call below (which would fail if lengths were different), so we could just remove it!

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        torch.testing.assert_close(
+            tc_frames_batch.data,
+            expected_frames,
+            rtol=0,
+            atol=0,
+            msg="Frames differ between TorchCodec fps and torchvision resampling",
+        )
+
+    @pytest.mark.parametrize("fps", [5.0, 15.0, 29.97, 60.0])
+    def test_get_frames_played_in_range_full_video_fps_matches_torchvision(self, fps):
```
NicolasHug: Both tests look very similar, I think we should be able to test both logics over a single parametrized test? IIUC, the main difference is just start_seconds and stop_seconds?

test/test_decoders.py
```diff
@@ -1098,6 +1097,203 @@ def test_get_frames_by_pts_in_range_fails(self, device, seek_mode):
...
+        all_frames = decoder.get_all_frames()
+
+        assert len(all_frames) == len(decoder)
+
+        frames_in_range = decoder.get_frames_played_in_range(
+            start_seconds=decoder.metadata.begin_stream_seconds,
+            stop_seconds=decoder.metadata.end_stream_seconds,
+        )
+        assert len(all_frames) == len(frames_in_range)
+        assert_frames_equal(all_frames.data, frames_in_range.data)
```
NicolasHug: following up on my other comment, here and below should also be `torch.testing.assert_close`

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,50 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+  // Resample frames to match the target frame rate
+  if (fps.has_value()) {
+    TORCH_CHECK(
+        fps.value() > 0,
+        "fps must be positive, got " + std::to_string(fps.value()));
+
+    // TODO: add an early break if requested fps is the same as the current fps
+
+    double fpsVal = fps.value();
+    double frameDuration = 1.0 / fpsVal;
```
NicolasHug: Nit, for consistency with the other variable names (like stopSeconds)

```suggestion
    double frameDurationSeconds = 1.0 / fpsVal;
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,50 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+        numOutputFrames,
+        resizedOutputDims_.value_or(metadataDims_),
+        videoStreamOptions.device);
+
+    // Decode frames, reusing already-decoded frames for duplicates
+    int64_t lastDecodedSourceIndex = -1;
+    torch::Tensor lastDecodedData;
+
+    for (int64_t i = 0; i < numOutputFrames; ++i) {
+      double targetPts = startSeconds + i * frameDuration;
```
NicolasHug:
```suggestion
      double targetPtsSeconds = startSeconds + i * frameDuration;
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,50 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+      int64_t sourceIdx = secondsToIndexLowerBound(targetPts);
+
+      if (sourceIdx == lastDecodedSourceIndex && lastDecodedSourceIndex >= 0) {
+        frameBatchOutput.data[i].copy_(lastDecodedData);
+      } else {
+        FrameOutput frameOutput =
+            getFrameAtIndexInternal(sourceIdx, frameBatchOutput.data[i]);
+        lastDecodedData = frameBatchOutput.data[i];
```
NicolasHug: This should be a cheap shallow copy (which doesn't copy the underlying pixel data), so that's probably fine. In other places for similar, we used an index in the output batch instead:

https://github.com/meta-pytorch/torchcodec/blob/48844c8cc2bfec40100c395b4e79fa80af0ad915/src/torchcodec/_core/SingleStreamDecoder.cpp#L698-L699

I think what you have is fine, but it might be even simpler by just doing 

```
frameBatchOutput.data[i].copy_(frameBatchOutput.data[i - 1]);
```

(if my logic is correct, which it might not!)

test/test_decoders.py
mollyxu: It seems like for [FFmpeg 4](https://github.com/meta-pytorch/torchcodec/actions/runs/21219748285/job/61054376581?pr=1148), the output coming from TorchCodec itself does not satisfy bitwise equality. I saw that in the codebase there is an example of using `assert_frames_equal` for FFmpeg 4 and I was wondering if this is something we should do here too. The comment points to an issue #428 that has not been closed yet.https://github.com/meta-pytorch/torchcodec/blob/e330e6b006aa07ed85c0f233be68caf16bab0eab/test/test_decoders.py#L380-L383

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -452,32 +451,31 @@ def get_frames_played_at(self, seconds: torch.Tensor | list[float]) -> FrameBatc
...
     ) -> FrameBatch:
         """Returns multiple frames in the given range.
 
-        Frames are in the half open range [start_seconds, stop_seconds). Each
-        returned frame's :term:`pts`, in seconds, is inside of the half open
-        range.
```
NicolasHug, line 461: Just checking, was this removed on purpose?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -906,6 +907,47 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
...
+        lastDecodedSourceIndex = sourceIdx;
+      }
+
+      frameBatchOutput.ptsSeconds[i] = targetPtsSeconds;
+      frameBatchOutput.durationSeconds[i] = frameDurationSeconds;
+    }
+
+    frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+    return frameBatchOutput;
+  }
```
NicolasHug: As a general rule (which we break often, and sometimes for good reasons), we should try to avoid `return` statements in the middle of a function. It can make it harder to understand from a quick look what  the output of the function can be. And here, because we have the following structure

```py

if fps.has_value():
  # branch A with return
# branch B with return
```

it's not immediately obvious that branch A and branch B are *completely* disjoint logically: that they're disjoint relies on the fact that there's a `return` statement in branch A, and that's not always trivially clear or expected.

So I'd suggest to follow the following structure:

```py
if fps.has_value():
    # branch A with return
else:
    # branch B with return
```

It's logically the same thing, but the `else` block makes it clearer that the 2 branches are completely separate. From there we have 2 options, and I don't have a strong preference for either:

- inline the code of branch A and branch B (as done here)
- put branch A in a private method (e.g. `getFramesPlayedInRangeWithFps()`) and branch B in another private method (e.g. `getFramesPlayedInRangeWithoutFps()`.

test/test_decoders.py
NicolasHug: Oh, oops, sorry for leaving https://github.com/meta-pytorch/torchcodec/issues/428  blank! I probably investigated it a bit, but now I don't remember what it was about.

It's a shame we can't get bit-wise equality with FFmpeg 4, but I'd still suggest to use `assert_close` here instead of `assert_frames_equal`. Looking at the [logs](https://github.com/meta-pytorch/torchcodec/actions/runs/21219748285/job/61054376581?pr=1148), the differences we get happen on very few pixels (up to 1.5%), and it also only happens on FFmpeg 4 **with** the FFmpeg CUDA interface. Both CPU and the Beta interface seem to be fine. So I think we're safe to consider that an FFmpeg 4 bug, and just not test for strict equality in such case. We *could* use`assert_frames_equal` exclusively for FFmpeg4 + CUDA FFmpeg interface if we wanted to be super thorough, but I don't think it's strictly necessary.

src/torchcodec/_core/SingleStreamDecoder.cpp
mollyxu: Thanks for pointing this out! Here's the summary from our team discussions:

Upon investigation, `fps` and `seconds_between_frames` are essentially inverses of each other (`fps = 1 / seconds_between_frames`). However, the frame selection logic differs between FFmpeg's fps filter and our samplers:

* FFmpeg fps filter: Determines for each output index which input index should map to it using an algorithm that involves [rescaling and rounding](https://superuser.com/questions/843292/ffmpeg-how-does-ffmpeg-decide-which-frames-to-pick-when-fps-value-is-specified/843363#843363)
* Samplers: Determine which input frame to duplicate based on input presentation timestamps

Decision:
We'll keep the two APIs separate. This makes it clearer for users rather than having users reason about the relationship between the two parameters.
For the frame selection logic in `get_frames_played_in_range(fps=...)`, we'll use the same logic as the samplers for consistency with other libraries like TorchVision.

mollyxu:
Thank you everyone for your feedback and reviews!

PR #1154, opened 2026-01-07
Extend color range note to explain limited range

Dan-Flores:
This PR extends the note on color range to mention how limited range matrices are generated. 
This method is used to generate the matrices in #1125.

### DIFF
--- a/src/torchcodec/_core/CUDACommon.cpp
+++ b/src/torchcodec/_core/CUDACommon.cpp
@@ -180,8 +180,8 @@ void initializeCudaContextWithPytorch(const torch::Device& device) {
 //   the decoder.
 // - But *internally*, `nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx` needs U and V to
 //   be centered around 0, i.e. in [-128, 127]. So we need to apply a -128
-//   offset to U and V. Y doesn't need to be offset. The offset can be applied
-//   by adding a 4th column to the matrix.
+//   offset to U and V. Y needs an offset of -16, only when using limited range.
+//   The offsets can be applied by adding a 4th column to the matrix.
 //
 //
 // So our conversion matrix becomes the following, with new offset column:
@@ -189,7 +189,7 @@ void initializeCudaContextWithPytorch(const torch::Device& device) {
 //         [ 1.0000e+00, -1.8732e-01, -4.6812e-01,     -128]
 //         [ 1.0000e+00,  1.8556e+00,  4.6231e-09 ,    -128]])
 //
-// And that's what we need to pass for BT701, full range.
+// And that's what we need to pass for BT709, full range.
 /* clang-format on */
 
 // BT.709 full range color conversion matrix for YUV to RGB conversion.


src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -110,6 +109,50 @@ void initializeCudaContextWithPytorch(const torch::Device& device) {
...
+// ])
+//
+// print("RGB->YUV matrix (Limited Range BT.709):")
+// print(rgb_to_yuv_limited)
+// ```
+//
+// This yields:
+// tensor([[ 0.1826,  0.6142,  0.0620],
+//         [-0.1006, -0.3386,  0.4392],
+//         [ 0.4392, -0.3989, -0.0403]])
```
NicolasHug, line 151: Ideally, could we find an authoritative source from which we can validate that these values are correct? Something similar to what we have for the full range:

https://github.com/meta-pytorch/torchcodec/blob/f91d24aaccb6cc0c87f5eeef1a00262e023f9b20/src/torchcodec/_core/CUDACommon.cpp#L111

It doesn't have to be for 709 specifically, but validating it for one limited range config would be great.

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -110,6 +109,50 @@ void initializeCudaContextWithPytorch(const torch::Device& device) {
...
+//
+// Next, lets consider encoding RGB -> YUV in the limited color range.
+// Y is in the range [16-235] and U,V are in [16-240].
+// The reduced range provides margins for errors during processing.
+// https://en.wikipedia.org/wiki/YCbCr#Y%E2%80%B2PbPr_to_Y%E2%80%B2CbCr
+//
+// To encode RGB -> YUV in limited range, we start with the full range conversion
+// matrix derived above, then scale it to compress into the limited ranges:
+// - RGB [0,255] -> Y [16,235]: compress by (235-16)/255 ≈ 219/255
+// - RGB [0,255] -> U,V [16,240]: compress by (240-16)/255 ≈ 224/255
```
NicolasHug, line 121: Is this meant to be "≈" or should it be "=" ?

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -110,6 +109,50 @@ void initializeCudaContextWithPytorch(const torch::Device& device) {
 //
 // Which matches https://en.wikipedia.org/wiki/YCbCr#ITU-R_BT.709_conversion
 //
+//
+// Next, lets consider encoding RGB -> YUV in the limited color range.
+// Y is in the range [16-235] and U,V are in [16-240].
```
NicolasHug, line 114: Reading this might be slightly confusing because the first part of the note above started with

> Conventionally Y is in [0, 1] range, and U and V are in [-0.5, 0.5]

And that's also how the matrices below are defined. I think we can get around this by saying something like:

> In limited range, the [0, 255] range is mapped into [16-235] for Y, and into  [16-240] for U and V.

NicolasHug (commented):
Thanks for the PR @Dan-Flores . It looks good, I left a few comments and questions.

One general comment: The new section is slightly at odds with the existing note structure. The existing note is mostly about *decoding* (not encoding), and largely about YUV -> RGB conversion. The last section for example mainly refers to `nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx`, not to `nppiRGBToNV12_8u_ColorTwist32f_C3P2R_Ctx`, and the intro specifically mentions decoding. The new section doesn't flow very well within this structure (and switches from a decoding context to an encoding context without notice).

Re-writing the entire note to make it holistically consistent would be a fair amount of work, and it's probably not worth it. I'd suggest to insead write the new section you added into a separate place. For example, in the CudaDeviceInterface code, where the new matrices are. Maybe this can just be included as part of #1154 ?

src/torchcodec/_core/CUDACommon.cpp
Dan-Flores: I'll try to find a source that gets our exact values. There is one section on that wiki page that gets close, but results in different values due to an optimization: https://en.wikipedia.org/wiki/YCbCr#Approximate_8-bit_matrices_for_BT.601

Dan-Flores:
Closing this PR and moving the note into https://github.com/meta-pytorch/torchcodec/pull/1125.

PR #1155, opened 2026-01-07
Fix skipif marker handling

mollyxu:
Fixes a bug where tests with `@pytest.mark.skipif(cond, ...)` were being incorrectly excluded in internal CI even when `cond` was `False`.

The issue is that pytest markers are always present regardless of whether the condition is true or false.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -38,8 +38,9 @@ def pytest_collection_modifyitems(items):
 
         # For skipif, the marker is always present regardless of whether the
         # condition is True or False, so we must check the actual condition.
-    skipif_condition_is_true = any(skipif_marker.args[0] for skipif_marker in item.iter_markers("skipif"))
-
+        skipif_condition_is_true = any(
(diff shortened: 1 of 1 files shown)


test/conftest.py
```diff
@@ -35,12 +35,20 @@ def pytest_collection_modifyitems(items):
...
-        has_skipif_marker = item.get_closest_marker("skipif") is not None
+
+        # For skipif, the marker is always present regardless of whether the
+        # condition is True or False, so we must check the actual condition.
+        skipif_condition_is_true = False
+        for skipif_marker in item.iter_markers("skipif"):
+            condition = skipif_marker.args[0]
+            if condition:
+                skipif_condition_is_true = True
+                break
```
NicolasHug: I think we can simplify this a bit with something of the sort


```suggestion
    skipif_condition_is_true = any(skipif_marker.args[0] for skipif_marker in item.iter_markers("skipif"))

```

PR #1160, opened 2026-01-09
Modified copyright, privacy, terms in docs

mollyxu:
Modified copyright, privacy, terms in homepage to comply with Meta Open Source policies


docs/source/_templates/layout.html
```diff
@@ -15,7 +15,22 @@
     var collapsedSections = ['Introduction', 'Getting Started', 'Tutorials']
 </script> -->
 
-<script script type="text/javascript">
+<script type="text/javascript">
     var collapsedSections = []
 </script>
+
+<script type="text/javascript">
+    // Override footer links to Meta's Terms and Privacy Policy
```
mollyxu, line 23: Currently, we use pytorch-sphinx-theme, which have these links to privacy and terms in raw html. We can't directly override it in our layout.html. As such, we would have to use javascript to replace those links.

We could also ask pytorch-sphinx-theme if they could wrap these links in an overridable block but Im not sure if it makes sense for them to support that.

docs/source/conf.py
```diff
@@ -150,7 +150,7 @@ def __call__(self, filename):
 master_doc = "index"
 
 # General information about the project.
-copyright = "2023-present, TorchCodec Contributors"
+copyright = "©2023 Meta Platforms, Inc"
```
mollyxu: Meta Open Source policy has a script regex that requires the copyright symbol to appear after "Copyright." This will result in a double copyright symbol effectively and render as `© Copyright ©2023 Meta Platforms, Inc.`The official PyTorch docs also have a double copyright symbol

Alternatively, we could replace the footer's copyright section using JavaScript

Dan-Flores: It seems on the [official pytorch docs](https://docs.pytorch.org/docs/stable/index.html), the first copyright symbol precedes "PyTorch", but ours only precedes the word "Copyright"? 
'© PyTorch. Copyright © The Linux Foundation®'

Either way, it's probably fine to have two copyright symbols.

Torchaudio keeps the year current, lets reuse this code: https://github.com/pytorch/audio/blob/e1232690308a6b5297fcd06e925899a9b64f7280/docs/source/conf.py#L174

docs/source/_templates/layout.html
Dan-Flores: I think the search and replace approach makes sense, but is somewhat fragile. If the `pytorch_sphinx_theme` footer updates, we will no longer update the link to have the correct policy. 

The fix should probably be upstream, where this link is [currently set](https://github.com/pytorch/pytorch_sphinx_theme/blob/2d7873510947895ec27f3ca829430df3bd6a474c/pytorch_sphinx_theme/layout.html#L540-L546).

NicolasHug:
Closing as I think we resolved this by just migrating to new pytorch sphinx theme

PR #1161, opened 2026-01-10
Add TORCHCODEC_FFMPEG_DIR to specify FFmpeg dynamic libs search path.

bcw222:
Maybe sometimes FFmpeg's DLLs is not in `PATH` so we need to add them manually.


src/torchcodec/_core/ops.py
```diff
@@ -103,6 +103,14 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
         def expose_ffmpeg_dlls():  # noqa: F811
             ffmpeg_dir = Path(ffmpeg_path).parent
             return os.add_dll_directory(str(ffmpeg_dir))  # that's the actual CM
+        
+        if chocolateyInstall := os.environ.get("ChocolateyInstall"):
+            # Chocolatey creates shim so we need to get the actual path to the FFmpeg executable.
+            chocolateyInstallPath = Path(chocolateyInstall)
+            if Path(ffmpeg_path).parent.parent == chocolateyInstallPath and (chocolateyFFmpegSharedPaths := list(chocolateyInstallPath.rglob("lib/ffmpeg-shared/tools/ffmpeg-*-shared/bin"))):
```
Dan-Flores: To clarify my understanding, this line checks that the `ffmpeg_path` found by `shutil` is installed via Chocolatey but is the [shimmed executable](https://docs.chocolatey.org/en-us/features/shim/#usage), so the second half of the line finds the actual FFmpeg executable?

Could you share how you installed `ffmpeg`? I suspect the path searched here might be specific to how `ffmpeg` was installed (ex. `choco install ffmpeg` vs `ffmpeg-full`).

Dan-Flores:
Thanks for the PR @bcw222!

Let's try to implement a more general solution. Perhaps we could expose a global variable that we pass to `os.add_dll_directory` if defined. That way, chocolatey users can still pass in the directory of the actual FFmpeg executable, without hardcoding any paths in. 

Would that work for your workflow?

src/torchcodec/_core/ops.py
bcw222: Only `ffmpeg-shared` contains the required DLLs. So I was actually verifying whether it was installed via Chocolatey and this fix is only applied in that case.

bcw222:
Hmm, that sounds good, but it would result in a feature that should work out-of-the-box (on devices with the required FFmpeg already installed) still requiring additional manual steps.

In a narrow sense, my fix already avoids hardcoding paths as much as possible by reading the `ChocolateyInstall` environment variable. Broadly speaking, due to Chocolatey's peculiarities, we inevitably have to hardcode certain things.

Using an environment variable is certainly a good idea, but this pull request is just a quick fix I made after encountering the issue while using `torchcodec`. I have not looked closely into the project's environment variable naming conventions, so I did not add that feature to avoid inconsistencies in the code style.

I would be happy to refine this further if I have the time, or if someone else is willing to help.

Dan-Flores:
Got it, I understand your concern that an environment variable would require an extra manual step. From our perspective, there are too many package managers to implement specific solutions for each. Our thinking is that an environment variable will provide a general solution that helps simplify your use case and others.

Roughly, the implementation would be:
* First check for `TORCHCODEC_FFMPEG_DIR` environment variable
* If not set, fallback to the existing `shutil` logic
* Call `add_dll_directory()` if an FFmpeg directory is found

Would you be interested in contributing this change?

bcw222:
This way?

src/torchcodec/_core/ops.py
```diff
@@ -97,11 +97,23 @@ def load_torchcodec_shared_libraries() -> tuple[int, str]:
...
+                ffmpeg_dir = Path(ffmpeg_path).parent
+                return os.add_dll_directory(str(ffmpeg_dir))  # that's the actual CM
+
+            if chocolateyInstall := os.environ.get("ChocolateyInstall"):
+                # Chocolatey creates shim so we need to get the actual path to the FFmpeg executable.
+                chocolateyInstallPath = Path(chocolateyInstall)
+                if Path(ffmpeg_path).parent.parent == chocolateyInstallPath and (chocolateyFFmpegSharedPaths := list(chocolateyInstallPath.rglob("lib/ffmpeg-shared/tools/ffmpeg-*-shared/bin"))):
+                    chocolateyFFmpegSharedPath = chocolateyFFmpegSharedPaths[0]
+                    def expose_ffmpeg_dlls():  # noqa: F811
+                        return os.add_dll_directory(str(chocolateyFFmpegSharedPath))
```
Dan-Flores: With `TORCHCODEC_FFMPEG_DIR` environment variable, we can remove the package manager specific logic and instead have users set `TORCHCODEC_FFMPEG_DIR` to the directory containing their `ffmpeg` DLLs.

Dan-Flores (commented):
Thanks for updating this @bcw222! I left a comment on avoiding package manager specific logic.

bcw222:
Reasonable. But maybe we need to document this.

README.md
```diff
@@ -211,6 +211,10 @@ format you want. Refer to Nvidia's GPU support matrix for more details
    conda install -c conda-forge "torchcodec=*=*cuda*"
    ```
 
+### Specify FFmpeg Path Manually
+
+If torchcodec cannot detect the FFmpeg installation correctly, you can set the `TORCHCODEC_FFMPEG_DIR` environment variable to the directory containing the FFmpeg DLLs.
```
Dan-Flores:
```suggestion
If torchcodec cannot detect the FFmpeg installation correctly, you can set the `TORCHCODEC_FFMPEG_DIR` environment variable to the directory containing the FFmpeg shared libraries.
```

src/torchcodec/_core/ops.py
```diff
@@ -26,12 +26,15 @@
...
     # the DLL search path. This seems to be needed on some users machine, but
     # not on our CI. We don't know why.
-    if ffmpeg_path := shutil.which("ffmpeg"):
-
-        def expose_ffmpeg_dlls():  # noqa: F811
-            ffmpeg_dir = Path(ffmpeg_path).parent.absolute()
-            return os.add_dll_directory(str(ffmpeg_dir))  # that's the actual CM
+    if ffmpeg_dir := os.getenv("TORCHCODEC_FFMPEG_DIR"):
+        def expose_ffmpeg_dlls():
+            return os.add_dll_directory(str(ffmpeg_dir))
```
NicolasHug: This is in a windows-specific path, I think we want to support `TORCHCODEC_FFMPEG_DIR` across platforms.

bcw222: Does this issue really exist on other platforms?

Dan-Flores: Yes, definitely @bcw222! Failure to locate FFmpeg libs is a somewhat common issue. Hopefully, `TORCHCODEC_FFMPEG_DIR` will be a generic solution in many of these cases.

bcw222: OK I've changed this.

bcw222:
This way?

Dan-Flores:
Thanks for the update @bcw222! I'll push a quick change to load the FFmpeg libraries in macOS and linux, since `add_dll_directory()` is only defined for Windows.

src/torchcodec/_core/ops.py
```diff
@@ -22,13 +24,30 @@
...
+        def expose_ffmpeg_dlls():  # type: ignore[no-redef]  # noqa: F811
+            return os.add_dll_directory(str(ffmpeg_dir))
+
+    else:
+        ffmpeg_library_glob = "*.dylib" if sys.platform == "darwin" else "*.so*"
+        ffmpeg_library_paths = glob.glob(os.path.join(ffmpeg_dir, ffmpeg_library_glob))
+        if not ffmpeg_library_paths:
+            raise RuntimeError(
+                "TORCHCODEC_FFMPEG_DIR is set, but no FFmpeg shared libraries "
+                f"were found in {repr(ffmpeg_dir)}."
```
NicolasHug, line 39: `repr` seems overkill

src/torchcodec/_core/ops.py
```diff
@@ -22,13 +24,30 @@
...
+            )
+        for ffmpeg_library_path in ffmpeg_library_paths:
+            ctypes.CDLL(ffmpeg_library_path)
```
NicolasHug, line 42: This is not ideal: this is loading every single .so file that is present in the `TORCHCODEC_FFMPEG_DIR`. 

If, for example, the user specifies `TORCHCODEC_FFMPEG_DIR=$CONDA_PREFIX/lib/` then we're loading hundreds and hundreds of .so files that we'll never use.

We shouldn't load the .so files ourselves, we should continue leaving it up to `load_torchcodec_shared_libraries()` load all the transitive .so files it needs. What we want is to tell `load_torchcodec_shared_libraries()` where to find some extra .so files. I.e. I think we just want to update `LD_LIBRARY_PATH` on linux and its equivalent on MacOS.

Something along these lines:

```py
    custom_path = os.environ.get('TORCHCODEC_FFMPEG_LIBRARIES')
    if custom_path:
        # Prepend so it takes priority, just like LD_LIBRARY_PATH
        ld_path = os.environ.get('LD_LIBRARY_PATH', '')
        os.environ['LD_LIBRARY_PATH'] = f"{custom_path}:{ld_path}" if ld_path else custom_path
# now call load_torchcodec_shared_libraries() as usual
```

bcw222: That seems reasonable on Linux. However, as far as I know, on macOS, `DYLD_LIBRARY_PATH` is ignored when using the system Python because of System Integrity Protection (SIP). Should we simply document this limitation and recommend that users use an external Python installation, or should we keep the logic to load the shared libraries manually with `ctypes`?

Dan-Flores: Thanks for pointing this out @bcw222! Unfortunately this behavior on Linux and MacOS means we need to know the names of all the required .so files to avoid loading many unnecessary .so files (as @NicolasHug points out), and we load the .so files before `load_torchcodec_shared_libraries()`. 

These introduce complexity and potential for unexpected side effects. Based on our discussion, what we actually want is for `TORCHCODEC_FFMPEG_DIR` to act as an alias for `LD_LIBRARY_PATH`. 

So reconsidering this approach, we should be able to have FFmpeg libs be correctly detected by setting `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH` / `PATH`.

We can update this PR to document this in `README.md`:

```bash
# Conda on Linux
LD_LIBRARY_PATH="$CONDA_PREFIX/lib:$LD_LIBRARY_PATH"

# Conda on macOS
DYLD_LIBRARY_PATH="$CONDA_PREFIX/lib:$DYLD_LIBRARY_PATH"
```

```sh
# Conda on Windows Powershell
$env:PATH = "$env:CONDA_PREFIX\Library\bin;" + "$env:PATH"
```

bcw222: Wait, are you sure? On Linux it should be `LD_LIBRARY_PATH` while on Windows no environment variable equivalent exists. That's why we need `TORCHCODEC_FFMPEG_DIR` on Windows.

Dan-Flores: My apologies, those examples were very incorrect as I sent my message while drafting, I'll fix my previous comment.

My understanding is that the `PATH` variable is the Windows equivalent to make DLLs discoverable, and the syntax to use it would be very similar to `LD_LIBRARY_PATH` or `DYLD_LIBRARY_PATH`.

NicolasHug: > On Linux it should be LD_LIBRARY_PATH while on Windows no environment variable equivalent exists. That's why we need TORCHCODEC_FFMPEG_DIR on Windows.

@bcw222  our understanding is that on Windows, the `LD_LIBRARY_PATH` equivalent is actually the `PATH` env var. Is that not the case?

bcw222: Indeed, but globally modifying `PATH` would clutter the environment (essentially losing the advantages of using shims like chocolatey provides; for example, when using Tab auto-completion; although Windows is already quite messy in this regard); while manually passing it every time is cumbersome. Especially considering the differences across platforms, I think using a dedicated environment variable (`TORCHCODEC_FFMPEG_DIR`) to configure it would be better.
However, given that this is essentially just an alias for existing environment variables (`PATH`/`LD_LIBRARY_PATH`/`DYLD_LIBRARY_PATH`), perhaps documenting these considerations for users would be sufficient.
So what solution do you all think is more appropriate?

Dan-Flores: Given that the OS differences require us to:
1. Know the names of all .so files we want to load on Linux/MacOS (we could grep on `libav*`, `libsw*`, but thats not guaranteed to always work).
2. Load the library using `ctypes` on Linux/MacOS, which diverges from the Windows logic and introduces library loading logic, which adds complexity and potentially side effects.

Lets stick to the standard environmental variables, and add a section documenting how to use them! It can be similar to my edited comment above and list the various `PATH` variables. 

> manually passing it every time is cumbersome

Depending on your set up, could you call `os.add_dll_directory()` before `import torchcodec`?

bcw222: You're right. So we have two solutions: with the original logic untouched, on Windows we could either modify `PATH` or directly call `os.add_dll_directory()` before `import torchcodec`.

Issue #1145, opened 2025-12-22
Segmentation Fault on 16 Channel WAV


SirNate0:
### 🐛 Describe the bug

Loading a 16 channel wav file causes a segmentation fault. (Converting to an ogg file with ffmpeg and then loading works, for whatever reason).
```sh
#!/bin/sh
ffmpeg -f lavfi -i "sine=frequency=10000:duration=5" sine.wav
ffmpeg -i sine.wav -i sine.wav -i sine.wav -i sine.wav \
        -i sine.wav -i sine.wav -i sine.wav -i sine.wav \
        -i sine.wav -i sine.wav -i sine.wav -i sine.wav \
        -i sine.wav -i sine.wav -i sine.wav -i sine.wav \
        -filter_complex "[0:a][1:a][2:a][3:a][4:a][5:a][6:a][7:a][8:a][9:a][10:a][11:a][12:a][13:a][14:a][15:a]amerge=inputs=16[a]" \
        -map "[a]" -c:a pcm_s16le 16ch.wav
ffprobe 16ch.wav 
# ... Stream #0:0: Audio: pcm_s16le ([1][0][0][0] / 0x0001), 44100 Hz, 16 channels, s16, 11289 kb/s
python3 -c "from torchcodec.decoders import AudioDecoder; dec=AudioDecoder('16ch.wav'); dec.get_all_samples()" 
# Segmentation fault (core dumped)
ffmpeg -i 16ch.wav 16ch.ogg
ffprobe 16ch.ogg
# ... Stream #0:0: Audio: vorbis, 44100 Hz, 16 channels, fltp, 1280 kb/s ...
python3 -c "from torchcodec.decoders import AudioDecoder; dec=AudioDecoder('16ch.ogg'); print(dec.get_all_samples())" 
# AudioSamples:
#  data (shape): torch.Size([16, 220736])
#  pts_seconds: 0.0
#  duration_seconds: 5.005351473922903
#  sample_rate: 44100


```

### Versions

```
PyTorch version: 2.9.1+cu128
Is debug build: False
CUDA used to build PyTorch: 12.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.3 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: Could not collect
CMake version: version 3.22.1
Libc version: glibc-2.35

[...]

PR #1166, opened 2026-01-14
Enable decoding audio with >8 channels

Dan-Flores:
This PR resolves the segfault in https://github.com/meta-pytorch/torchcodec/issues/1145 and enables 	`convertAudioAVFrameSamples` to handle >8 channels.

#### The Fix
The fix is to replace [AVFrame->data](https://www.ffmpeg.org/doxygen/trunk/structAVFrame.html#a1d0f65014a8d1bf78cec8cbed2304992), which only holds up to 8 channel pointers as defined by `AV_NUM_DATA_POINTERS`, with [AVFrame->extended_data](https://www.ffmpeg.org/doxygen/trunk/structAVFrame.html#afca04d808393822625e09b5ba91c6756), which holds all the channels that may not fit within AVFrame->data. 

Since the original reproducing example used S16 format (`-c:a pcm_s16le`), and `outputFormat` is always set to `AV_SAMPLE_FMT_FLTP` within `convertAudioAVFrameToFrameOutput`, this bug was reached within `convertAudioAVFrameSamples`.

#### The Longterm Fix
To ensure we are able to decode into any valid number of channels, we only assert that `audioStreamOptions.numChannels > 0`, and remove the assertion that `audioStreamOptions.numChannels <= AV_NUM_DATA_POINTERS`.

This way,  higher valid channel counts (16, 24) can still be used. In the case of invalid channel counts, we error when initializing `SwrContext`.

#### Testing
* I added a new `TestAudio` called `SINE_16_CHANNELS`, and added it to `test_get_all_samples` and `test_metadata`. 
* `test_num_channels` is updated to use higher valid `num_channels=16, 24`. 
* `test_num_channels_errors` is updated to use higher invalid values `num_channels=15, 23`.


test/utils.py
```diff
@@ -834,3 +834,20 @@ def sample_format(self) -> str:
         )
     },
 )
+
+# 16-channel audio for testing multi-channel (>8) support. Generated with:
+# ffmpeg -f lavfi -i "sine=f=440:r=16000:d=1" -filter_complex "[0]asplit=16[s0][s1][s2][s3][s4][s5][s6][s7][s8][s9][s10][s11][s12][s13][s14][s15];[s0][s1][s2][s3][s4][s5][s6][s7][s8][s9][s10][s11][s12][s13][s14][s15]amerge=inputs=16" -c:a pcm_s16le test/resources/16ch.wav
```
Dan-Flores: This is a complex looking FFmpeg line. It uses filtergraph to split and label the existing stream 0 into 16 channels, then merges them into a 16 channel audio file.

test/utils.py
```diff
@@ -834,3 +834,20 @@ def sample_format(self) -> str:
         )
     },
 )
+
+# 16-channel audio for testing support for >8 channels. Generated with:
+# ffmpeg -i test/resources/sine_mono_s32.wav -t 1 -filter_complex "[0]asplit=16[s0][s1][s2][s3][s4][s5][s6][s7][s8][s9][s10][s11][s12][s13][s14][s15];[s0][s1][s2][s3][s4][s5][s6][s7][s8][s9][s10][s11][s12][s13][s14][s15]amerge=inputs=16" -c:a pcm_s16le test/resources/16ch.wav
+SINE_16_CHANNELS = TestAudio(
+    filename="16ch.wav",
```
mollyxu: nit: since all the videos follow the format the format of `file_name.format`, perhaps this video should be `sine_16ch.wav` rather than `16ch.wav`

test/test_decoders.py
```diff
@@ -1848,7 +1849,7 @@ def test_get_all_samples_with_range(self, asset, stop_seconds):
         assert samples.sample_rate == asset.sample_rate
         assert samples.pts_seconds == asset.get_frame_info(idx=0).pts_seconds
 
-    @pytest.mark.parametrize("asset", (NASA_AUDIO, NASA_AUDIO_MP3))
+    @pytest.mark.parametrize("asset", (NASA_AUDIO, NASA_AUDIO_MP3, SINE_16_CHANNELS))
```
mollyxu: Although the original segmentation fault was only on `get_all_samples`, would we want to add this audio file to other tests too?

mollyxu:
@Dan-Flores Thanks for looking into this! I left two small comments but LGTM!

test/test_decoders.py
```diff
@@ -2117,8 +2120,7 @@ def test_samples_duration(self, asset, sample_rate):
     # that the extra tensor allocation that happens within
     # maybeFlushSwrBuffers() is correct.
     @pytest.mark.parametrize("sample_rate", (None, 16_000))
-    # FFmpeg can handle up to AV_NUM_DATA_POINTERS=8 channels
-    @pytest.mark.parametrize("num_channels", (1, 2, 8, None))
+    @pytest.mark.parametrize("num_channels", (1, 2, 8, 16, None))
```
NicolasHug: Passing 17 channels now leads to this error message:

```
E       RuntimeError: Couldn't initialize SwrContext: Invalid argument. If the error says 'Invalid argument', it's likely that you are using a buggy FFmpeg version. FFmpeg4 is known to fail here in some valid scenarios. Try to upgrade FFmpeg?
```

which isn't really accurate anymore. Let's have a `TORCH_CHECK` for when `num_channels > 16` (ideally this would be an FFmpeg constant like `AV_NUM_DATA_POINTERS`, but I don't know if it exists - please check!) and we should have a small test against this.

test/test_decoders.py
```diff
@@ -2132,12 +2134,5 @@ def test_num_channels(self, asset, sample_rate, num_channels):
...
-        with pytest.raises(
-            RuntimeError, match="num_channels must be > 0 and <= AV_NUM_DATA_POINTERS"
-        ):
+        with pytest.raises(RuntimeError, match="num_channels must be > 0"):
             AudioDecoder(asset.path, num_channels=0)
-        with pytest.raises(
-            RuntimeError, match="num_channels must be > 0 and <= AV_NUM_DATA_POINTERS"
-        ):
-            # FFmpeg can handle up to AV_NUM_DATA_POINTERS=8 channels
-            AudioDecoder(asset.path, num_channels=9)
```
NicolasHug, line 2143: Following up on my comment above, we should keep this test and have num_channels=17 here instead.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -512,12 +512,16 @@ UniqueAVFrame convertAudioAVFrameSamples(
...
+  // for all channels, while data only contains AV_NUM_DATA_POINTERS (8).
+  // https://ffmpeg.org/doxygen/trunk/structAVFrame.html#afca04d808393822625e09b5ba91c6756
   auto numConvertedSamples = swr_convert(
       swrContext.get(),
-      convertedAVFrame->data,
+      convertedAVFrame->extended_data,
       convertedAVFrame->nb_samples,
       static_cast<const uint8_t**>(
-          const_cast<const uint8_t**>(srcAVFrame->data)),
+          const_cast<const uint8_t**>(srcAVFrame->extended_data)),
```
NicolasHug, line 524: We should note that this `convertAudioAVFrameSamples` function is called from different places: in the decoder, but also in the encoder!

In the encoder, we use and validate `data` instead of `extended_data`. I *think* we're fine, because `extended_data` and `data` should be the same thing when num_channels <= 8, which we validate against in the encoder (that's what Claude says, at least).

Still, to be safe, we should add a `TORCH_CHECK(data == extended_data)` in the encoder's code, right before we call `convertAudioAVFrameSamples`.

test/test_decoders.py
NicolasHug: That's a good thought - I think the current test coverage is fine. The other tests where we parametrize over other audio files aren't testing this critical path, so adding `SINE_16_CHANNELS` would be a bit redundant.

test/test_decoders.py
Dan-Flores: Agreed, I updated this test to document a few other failing `num_channel` values.

test/test_decoders.py
Dan-Flores: Thank you for catching this, I am surprised it would error when initializing `SwrContext`. 
I observed this test fails with `num_channels = 15, 17, 18, 20, 22` but passes with `num_channels=24`.

I found that `getOutputChannelLayout()` calls `av_channel_layout_default()` to try to generate a default channel layout given a number of channels, and for some unusual numbers of channels it is unable to do so, which causes `swrContext` init to fail.

To generically catch this, I implemented `TORCH_CHECK`s  in `getOutputChannelLayout` to ensure that the channel order output is not `AV_CHANNEL_ORDER_UNSPEC` when the user requests a different number of channels from the original audio file.

test/test_decoders.py
```diff
@@ -2131,13 +2133,15 @@ def test_num_channels(self, asset, sample_rate, num_channels):
...
+    @pytest.mark.parametrize("num_channels", (0, 15, 23))
+    def test_num_channels_errors(self, asset, num_channels):
+        msg = (
+            "num_channels must be > 0"
+            if num_channels == 0
+            else f"Cannot convert audio to {num_channels} channels"
+        )
+        with pytest.raises(RuntimeError, match=msg):
+            decoder = AudioDecoder(asset.path, num_channels=num_channels)
+            # Call get_all_samples to trigger num_channels conversion.
```
NicolasHug: The parametrization is a bit overkill here. It also forces us to call both `AudioDecoder()` and `decoder.get_all_samples()` within the `with pytest.raises(RuntimeError, match=msg):`, when in reality we only need to call  `AudioDecoder()` to trigger the `num_channels == 0` exception.
I think it's best to remove the parametrization and to de-factorize the 2 tests cases. To test both 15 and 23, we can just use a for loop!

src/torchcodec/_core/Encoder.cpp
```diff
@@ -361,6 +361,9 @@ UniqueAVFrame AudioEncoder::maybeConvertAVFrame(const UniqueAVFrame& avFrame) {
         avFrame,
         outNumChannels_));
   }
+  TORCH_CHECK(
```
NicolasHug, line 367: Thanks for adding this. Since it won't be obvious why this TORCH_CHECK exists, we can add a small comment, e.g.

> convertAudioAVFrameSamples relies on extended_data, so we check it's the same as data. This should always be the case since we validated earlier that we have less than AV_NUM_DATA_POINTERS channels.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -288,6 +288,16 @@ AVChannelLayout getOutputChannelLayout(
...
+    // AV_CHANNEL_ORDER_UNSPEC indicates FFmpeg did not find a default layout.
+    // Since the channel number is changing, SwrContext will need a valid layout
+    // to do mixing. We will fail to initialize SwrContext in this case, so we
+    // error here instead.
+    TORCH_CHECK(
+        outLayout.order != AV_CHANNEL_ORDER_UNSPEC,
+        "Cannot convert audio to ",
+        outNumChannels,
+        " channels. FFmpeg did not find a default channel layout for this "
+        "channel count.");
```
NicolasHug: I think this check should be somewhere else (if we really want to have it - which I'm not sure about):  this is not called only when channels are changing. This is also called transitively within `convertAudioAVFrameSamples`, which may or may not involve `num_channels` conversion:

 https://github.com/meta-pytorch/torchcodec/blob/5d5f46462088cefd6d6be0f8afcffdd1b9daa07d/src/torchcodec/_core/FFMPEGCommon.cpp#L507

For these cases, it's not clear that `AV_CHANNEL_ORDER_UNSPEC` actually leads to a problem.

I think it'd make more sense to have this check here:

https://github.com/meta-pytorch/torchcodec/blob/5d5f46462088cefd6d6be0f8afcffdd1b9daa07d/src/torchcodec/_core/FFMPEGCommon.cpp#L369-L371


----

But, are we 100% sure this check is correct? We have established that passing `num_channels = 15, 17, etc.` leads to having `AV_CHANNEL_ORDER_UNSPEC`, but what we don't know is whether swresample is always going to fail when `AV_CHANNEL_ORDER_UNSPEC`? In other words, are we sure adding this `TORCH_CHECK` isn't going to lead to errors that do not need to be errors?

Dan-Flores: Thats a good point, one point of uncertainty is how different audio codecs will handle `AV_CHANNEL_ORDER_UNSPEC`. 

I'll remove the `TORCH_CHECK` related to `AV_CHANNEL_ORDER_UNSPEC`, but leave the updated `TORCH_CHECK` In `SingleStreamDecoder.cpp` to only check `audioStreamOptions.numChannels > 0`, without the assertion that `audioStreamOptions.numChannels <= AV_NUM_DATA_POINTERS`. 

This way we can still test higher valid channel counts (16, 24) and test that we error on higher invalid channel counts (15, 23).

PR #1169, opened 2026-01-14
Add CUDA encoding section to VideoEncoder tutorial

Dan-Flores:
This section covers the basics of CUDA Encoding:
* It is enabled by passing the `frames` tensor on a CUDA device
* VideoEncoder will try to select a valid CUDA enabled `codec` by default
* `nv12` is currently the only supported pixel format

### DIFF
--- a/src/torchcodec/encoders/_video_encoder.py
+++ b/src/torchcodec/encoders/_video_encoder.py
@@ -8,14 +8,15 @@ from torchcodec import _core
 
 
 class VideoEncoder:
-    """A video encoder.
+    """A video encoder on CPU or CUDA..
 
     Args:
         frames (``torch.Tensor``): The frames to encode. This must be a 4D
             tensor of shape ``(N, C, H, W)`` where N is the number of frames,
             C is 3 channels (RGB), H is height, and W is width.
             Values must be uint8 in the range ``[0, 255]``.
-            The device of the frames tensor will be used for encoding.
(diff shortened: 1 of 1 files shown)


examples/encoding/video_encoding.py
```diff
@@ -71,6 +71,28 @@ def play_video(encoded_bytes):
...
+#     On GPU, the pixel format is always set to ``nv12`` (which does equivalent chroma subsampling
+#     to ``yuv420p``). The ``pixel_format`` parameter is not supported for GPU encoding.
+#
+# .. code-block:: python
+#
+#     gpu_frames = frames.to("cuda")  # Move frames to GPU
+#     gpu_encoder = VideoEncoder(frames=gpu_frames, frame_rate=frame_rate)
+#
+# That's it! The rest of the encoding process is the same as on CPU.
+
```
NicolasHug, line 95: This looks good, we should make sure to update the docstring of the encoder as well:

- the one-liner description (current "A video encoder") could make it clear that it supports CUDA, e.g. it could be "A video encoder on CPU or CUDA"
- The `frames` attribute should indicate these could be a CUDA tensor
- the pixel_format parameter should clarify that it should be left to None if frames are on CUDA
- we should clarify that the `to_tensor()` method always returns a CPU tensor, regardless of frames.device

NicolasHug (approved):

Issue #1140, opened 2025-12-17
Enable color parameters in NVENC test on FFmpeg 4 and 6


Dan-Flores:
Investigate why FFmpeg 4 and 6 fail with non-default color space and range.
The `test_nvenc_against_ffmpeg_cli` test is currently skipped on FFmpeg 4 and FFmpeg 6 when `colorspace` and `color_range` are set to values besides the default `BT.601` and `tv` . 

The code below demonstrates that the color parameters are set on the codec context, and the encoded video has the correct metadata. The error may be caused by other steps in the test.
<details>

```
import subprocess
import torch
import json
from torchcodec.encoders import VideoEncoder

# Create test frames and encode with specific color parameters
frames = torch.randint(0, 256, (5, 3, 256, 256), dtype=torch.uint8).to("cuda")
encoder = VideoEncoder(frames=frames, frame_rate=30)

encoder.to_file(
    dest="/tmp/test_video.mp4",
    codec="h264_nvenc",
    extra_options={
        "colorspace": "bt709",
        "color_range": "pc",
    }
)

result = subprocess.run(
        [
          "ffprobe",
          "-v",
          "error",
          "-select_streams",
          "v:0",
          "-show_entries",
          "stream=color_space,color_range,pix_fmt",
          "-of",
          "json",
          "/tmp/test_video.mp4"
        ],
        capture_output = True,
        text = True,
        check = True)

metadata = json.loads(result.stdout)["streams"][0]
print(json.dumps(metadata, indent=2))
```
</details>

PR #1172, opened 2026-01-16
Prioritize remaining TODO-VideoEncoder

Dan-Flores:
This PR updates the remaining `TODO-VideoEncoder` with a priority label to start a discussion on which are high priority.
Some `TODO`s are duplicates or not relevant anymore, so they are dropped.

P1: 
(somewhat important features)
* Add and utilize output `frame_rate` option
* Allow `height` and `width` to be set

P2:
(features unlikely to be frequently used)
* Enable tensors in NHWC format
* Enable `pixel_format` to be set by user on GPU

(bug fix)
* Investigate why FFmpeg 4 and 6 fail with non-default color space and range.

(refactors)
* Unify `AVFrame` creation with CPU method
* Consider using pytorch to allocate CUDA memory for efficiency

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -517,7 +517,7 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
   int height = static_cast<int>(tensor.size(1));
   int width = static_cast<int>(tensor.size(2));
 
-  // TODO-VideoEncoder: (P2) Unify AVFrame creation with CPU method
+  // TODO-VideoEncoder: (P1) Unify AVFrame creation with CPU method
   avFrame->format = AV_PIX_FMT_CUDA;
   avFrame->height = height;
   avFrame->width = width;
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -770,14 +770,14 @@ void VideoEncoder::initializeEncoder(
   avCodecContext_.reset(avCodecContext);
 
   // Store dimension order and input pixel format
-  // TODO-VideoEncoder: (P2) Enable tensors in NHWC format
+  // TODO-VideoEncoder: (P2) Enable tensors in NHWC shape
   auto sizes = frames_.sizes();
   inPixelFormat_ = AV_PIX_FMT_GBRP;
   inHeight_ = static_cast<int>(sizes[2]);
   inWidth_ = static_cast<int>(sizes[3]);
 
   // Use specified dimensions or input dimensions
-  // TODO-VideoEncoder: (P1) Allow height and width to be set
+  // TODO-VideoEncoder: (P2) Allow height and width to be set
   outWidth_ = inWidth_;
   outHeight_ = inHeight_;
 


test/test_encoders.py
```diff
@@ -1389,7 +1388,7 @@ def test_nvenc_against_ffmpeg_cli(
         self, tmp_path, method, format, codec, color_space, color_range
     ):
         ffmpeg_version = get_ffmpeg_major_version()
-        # TODO-VideoEncoder: Investigate why FFmpeg 4 and 6 fail with non-default color space and range.
+        # TODO-VideoEncoder: (P2) Investigate why FFmpeg 4 and 6 fail with non-default color space and range.
```
Dan-Flores, line 1391: This is potentially not a real bug in the encoder. I opened https://github.com/meta-pytorch/torchcodec/issues/1140 to document my findings. 

I suspect that the color space and color range change values are not being respected during decoding, not by the encoder. However, since we need to decode the frames to compare them, I have not been able to verify this.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -770,19 +770,19 @@ void VideoEncoder::initializeEncoder(
...
 
   // Use specified dimensions or input dimensions
-  // TODO-VideoEncoder: Allow height and width to be set
+  // TODO-VideoEncoder: (P1) Allow height and width to be set
   outWidth_ = inWidth_;
   outHeight_ = inHeight_;
 
   if (videoStreamOptions.pixelFormat.has_value()) {
-    // TODO-VideoEncoder: Enable pixel formats to be set by user
+    // TODO-VideoEncoder: (P2) Enable pixel formats to be set by user
```
Dan-Flores: For users encoding on GPU, performance is likely a priority, so I suspect lossless encoding (`yuv444p`) will not be needed.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -770,19 +770,19 @@ void VideoEncoder::initializeEncoder(
...
-  // TODO-VideoEncoder: Remove assumption that tensor in NCHW format
+  // TODO-VideoEncoder: (P2) Enable tensors in NHWC format
   auto sizes = frames_.sizes();
   inPixelFormat_ = AV_PIX_FMT_GBRP;
   inHeight_ = static_cast<int>(sizes[2]);
   inWidth_ = static_cast<int>(sizes[3]);
 
   // Use specified dimensions or input dimensions
-  // TODO-VideoEncoder: Allow height and width to be set
+  // TODO-VideoEncoder: (P1) Allow height and width to be set
```
Dan-Flores: This seems like a core feature we could expose, if resizing isn't more commonly done during other steps in a workflow (ex. decoding with transforms). Please let me know if this is the correct priority.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -770,19 +770,19 @@ void VideoEncoder::initializeEncoder(
   avCodecContext_.reset(avCodecContext);
 
   // Store dimension order and input pixel format
-  // TODO-VideoEncoder: Remove assumption that tensor in NCHW format
+  // TODO-VideoEncoder: (P2) Enable tensors in NHWC format
```
NicolasHug: I think we mean shape instead of format?

src/torchcodec/_core/Encoder.cpp
NicolasHug: IMHO this is a P2 until  we get explicit feature request for it. No strong opinion.

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -517,14 +517,14 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
   int height = static_cast<int>(tensor.size(1));
   int width = static_cast<int>(tensor.size(2));
 
-  // TODO-VideoEncoder: Unify AVFrame creation with CPU version of this method
+  // TODO-VideoEncoder: (P2) Unify AVFrame creation with CPU method
```
NicolasHug: Loosely held opinion: could be a P1, not because it's more urgent than the rest, but mostly because it's a reasonably easy TODO and it's good practice not to live with pending refacs for too long.

NicolasHug (approved):
Thanks @Dan-Flores ! I shared my thoughts below but no strong opinion, this PR LGTM as-is

src/torchcodec/_core/CudaDeviceInterface.cpp
Dan-Flores: I think this will be a part of a larger DeviceInterface refactor, so I'm not sure how easy it will be to do, but I'll set it to P1 since it is a pending refactor.

Issue #1084, opened 2025-12-02
Apply or expose rotation


johan-sightic:
### 🚀 The feature

I have videos collected with an iPhone, some of which are rotated +/- 90 degrees. This shows up in the output of `ffprobe`:
```
$ ffprobe video.mp4 
...
  Stream #0:0[0x1](und): Video: hevc (Main) (hvc1 / 0x31637668), yuvj420p(pc, smpte170m/bt709/bt709), 1440x1080, 80078 kb/s, 59.91 fps, 59.94 tbr, 600 tbn (default)
    Metadata:
      creation_time   : 2023-09-01T15:16:03.000000Z
      handler_name    : Core Media Video
      vendor_id       : [0][0][0][0]
    Side data:
      displaymatrix: rotation of -90.00 degrees
```
[PyAV](https://github.com/PyAV-Org/PyAV) exposes this on each frames as `frame.rotation`.

Could `torchcodec` expose this too in the metadata or even better apply the rotation when decoding?

### Motivation, pitch

Currently I need to get the rotation in another way and transpose the return tensor. It would be nice to not have to deal with that

NicolasHug:
Thanks for the feature request @johan-sightic . I think we'll want to support that, and natively return the frames in the orientation that respects the rotation metadata. That's potentially a BC-breaking change, but it probably qualifies as a bug-fix?

johan-sightic:
That would be great! Yeah could be classified as a bug

NicolasHug:
This was addressed in https://github.com/meta-pytorch/torchcodec/pull/1173 and was released with 0.11, thanks again for the feature request @johan-sightic

PR #1173, opened 2026-01-16
Rotation metadata support

mollyxu:
## Add rotation metadata support

Context: #1084

Videos recorded on mobile devices often store rotation metadata indicating how the video should be displayed. TorchCodec now extracts this metadata and applies the rotation during decoding so frames are returned in the correct orientation.

This PR adds support for extracting and applying rotation metadata from video streams. The rotation angle (in degrees counter clockwise) is extracted from the display matrix side data and:
1. Exposed via `VideoStreamMetadata.rotation`
2. Automatically applied during decoding — frames are rotated so they display correctly

When there are user transforms, the rotation is applied before the transforms as a part of filtergraph. In the future, we should benchmark this against alternatives and see if there should be optimizations.

```python

decoder = VideoDecoder("rotated_video.mp4")
print(decoder.metadata.rotation)  # e.g., 90
```

### The `rotation` property

- **`None`**: The video has no rotation metadata (most videos)
- **`float`**: The rotation angle in degrees (counter-clockwise rounded to the nearest 90 degrees) needed for correct display. Common values are 0, 90, 180, -90, or -180.

### Implementation

**Metadata extraction** (`FFMPEGCommon.cpp`):
- Added `getRotationFromStream()` which extracts the rotation from `AV_PKT_DATA_DISPLAYMATRIX` side data
- Handles FFmpeg version differences:
  - FFmpeg ≥ 6.1: Uses `av_packet_side_data_get()` with `codecpar->coded_side_data`
  - FFmpeg 5/6: Uses `av_stream_get_side_data()` with deprecation warning suppression for FFmpeg 6
  - FFmpeg 4: Uses `av_stream_get_side_data()` with `int*` size parameter
- Uses `av_display_rotation_get()` to compute the angle from the display matrix

**Metadata propagation**:
- `StreamMetadata` struct now includes an `std::optional<double> rotation` field (`Metadata.h`)
- `SingleStreamDecoder` populates this during stream scanning
- Python `VideoStreamMetadata` dataclass exposes `rotation: float | None` (`_metadata.py`)
- JSON metadata serialization includes the rotation value (`custom_ops.cpp`)

### Test coverage
- Added `nasa_13013_rotated.mp4` test resource with 90° rotation metadata
-  generated video using 
```
ffmpeg -y -display_rotation 90 -i test/resources/nasa_13013.mp4 \
  -c copy \
  test/resources/nasa_13013_rotated.mp4
```
- `test_rotation_metadata` (in `test_metadata.py`): Verifies rotation is correctly extracted for rotated videos and is `None` for non-rotated videos
- `test_rotation_applied_to_frames` (in `test_decoders.py`): Performs pixel-exact comparison with ffmpeg CLI output
- `test_rotation_with_transform` (in `test_decoders.py`): Verifies that user-specified transforms (Resize, CenterCrop, RandomCrop) work correctly with rotated videos, ensuring output dimensions match the requested (H, W) regardless of rotation metadata
- `test_rotation_with_transform_pipeline` (in `test_decoders.py`): Verifies that chained transforms work correctly with rotated videos, with each transform operating in post-rotation coordinate space

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -718,9 +718,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
   const auto& streamInfo = streamInfos_[activeStreamIndex_];
   const auto& videoStreamOptions = streamInfo.videoStreamOptions;
   FrameBatchOutput frameBatchOutput(
-      frameIndices.numel(),
-      resizedOutputDims_.value_or(preRotationDims_),
-      videoStreamOptions.device);
+      frameIndices.numel(), getOutputDims(), videoStreamOptions.device);
 
   auto previousIndexInVideo = -1;
   for (int64_t f = 0; f < frameIndices.numel(); ++f) {
@@ -777,9 +775,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesInRange(
   int64_t numOutputFrames = std::ceil((stop - start) / double(step));
   const auto& videoStreamOptions = streamInfo.videoStreamOptions;
   FrameBatchOutput frameBatchOutput(
-      numOutputFrames,
-      resizedOutputDims_.value_or(preRotationDims_),
-      videoStreamOptions.device);
+      numOutputFrames, getOutputDims(), videoStreamOptions.device);
 
   for (int64_t i = start, f = 0; i < stop; i += step, ++f) {
     FrameOutput frameOutput =
@@ -912,9 +908,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
   // below. Hence, we need this special case below.
   if (startSeconds == stopSeconds) {
     FrameBatchOutput frameBatchOutput(
-        0,
-        resizedOutputDims_.value_or(preRotationDims_),
-        videoStreamOptions.device);
+        0, getOutputDims(), videoStreamOptions.device);
     frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
     return frameBatchOutput;
   }
@@ -957,9 +951,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     int64_t numOutputFrames = static_cast<int64_t>(std::round(product));
 
     FrameBatchOutput frameBatchOutput(
-        numOutputFrames,
-        resizedOutputDims_.value_or(preRotationDims_),
-        videoStreamOptions.device);
+        numOutputFrames, getOutputDims(), videoStreamOptions.device);
 
     // Decode frames, reusing already-decoded frames for duplicates
     int64_t lastDecodedSourceIndex = -1;
@@ -1000,9 +992,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     int64_t numFrames = stopFrameIndex - startFrameIndex;
 
     FrameBatchOutput frameBatchOutput(
-        numFrames,
-        resizedOutputDims_.value_or(preRotationDims_),
-        videoStreamOptions.device);
+        numFrames, getOutputDims(), videoStreamOptions.device);
     for (int64_t i = startFrameIndex, f = 0; i < stopFrameIndex; ++i, ++f) {
       FrameOutput frameOutput =
           getFrameAtIndexInternal(i, frameBatchOutput.data[f]);
@@ -1550,6 +1540,20 @@ int64_t SingleStreamDecoder::getPts(int64_t frameIndex) {
     default:
       TORCH_CHECK(false, "Unknown SeekMode");
   }
+}
+
+FrameDims SingleStreamDecoder::getOutputDims() const {
+  const auto& streamMetadata =
+      containerMetadata_.allStreamMetadata[activeStreamIndex_];
+  Rotation rotation = rotationFromDegrees(streamMetadata.rotation);
+  // If there is a rotation, then resizedOutputDims_ is necessarily non-null
+  // (the rotation transform would have set it).
+  if (rotation != Rotation::NONE) {
+    TORCH_CHECK(
+        resizedOutputDims_.has_value(),
+        "Internal error: rotation is applied but resizedOutputDims_ is not set");
+  }
+  return resizedOutputDims_.value_or(preRotationDims_);
 }
 
 // --------------------------------------------------------------------------
--- a/src/torchcodec/_core/SingleStreamDecoder.h
+++ b/src/torchcodec/_core/SingleStreamDecoder.h
@@ -296,6 +296,16 @@ class SingleStreamDecoder {
 
   int64_t getPts(int64_t frameIndex);
 
+  // Returns the output frame dimensions for video frames.
+  // If resizedOutputDims_ is set (via resize, crop, or rotation transforms),
+  // returns that. Otherwise, returns preRotationDims_.
+  //
+  // Note: if resizedOutputDims_ is null, there is no rotation (the
+  // rotation transform would have set it), so preRotationDims_ ==
+  // postRotationDims_. This makes it safe to use preRotationDims_ as the
+  // fallback.
+  FrameDims getOutputDims() const;
+
   // --------------------------------------------------------------------------
   // STREAM AND METADATA APIS
   // --------------------------------------------------------------------------


src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -612,6 +612,66 @@ int64_t computeSafeDuration(
...
+  if (sideData != nullptr && sideData->size >= 9 * sizeof(int32_t)) {
+    displayMatrix = reinterpret_cast<const int32_t*>(sideData->data);
+  }
+#else
+  // FFmpeg < 7: Use av_stream_get_side_data.
+  // This function was deprecated in FFmpeg 6.1, but the replacement
+  // (codecpar->coded_side_data) wasn't available until FFmpeg 7.
+  // Suppress the deprecation warning only for FFmpeg 6.
+#if LIBAVFORMAT_VERSION_MAJOR == 60
+#pragma GCC diagnostic push
```
mollyxu: I'm not sure if there is a better way to handle the case of FFmpeg 6.1 deprecating `av_stream_get_side_data` while the replacement `codecpar->coded_side_data` not being available until FFmpeg 7

src/torchcodec/_core/FFMPEGCommon.h
```diff
@@ -280,6 +281,11 @@ int64_t computeSafeDuration(
     const AVRational& frameRate,
     const AVRational& timeBase);
 
+// Extracts the rotation angle in degrees from the stream's display matrix
+// side data. The display matrix is used to specify how the video should be
+// rotated for correct display.
+std::optional<double> getRotationFromStream(const AVStream* avStream);
```
mollyxu, line 287: Not sure if there is a huge difference between using `optional` rather than `0` to represent no rotation.

Would it be useful to distinguish the case of a video with no rotation metadata and a video that is known to not been rotated before (ie. rotation of 0 degrees)?

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -612,6 +612,69 @@ int64_t computeSafeDuration(
...
+      avStream->codecpar->coded_side_data,
+      avStream->codecpar->nb_coded_side_data,
+      AV_PKT_DATA_DISPLAYMATRIX);
+  if (sideData != nullptr && sideData->size >= 9 * sizeof(int32_t)) {
+    displayMatrix = reinterpret_cast<const int32_t*>(sideData->data);
+  }
+#elif LIBAVFORMAT_VERSION_MAJOR >= 60 // FFmpeg 6.0
+  // FFmpeg 6.0: Use av_stream_get_side_data (deprecated but still available)
+  // Suppress deprecation warning for this specific call
+#pragma GCC diagnostic push
```
mollyxu, line 642: I'm not sure if there is a better way to handle the case of FFmpeg 6 deprecating `av_stream_get_side_data` while the replacement `codecpar->coded_side_data` not being available until FFmpeg 6.1

test/test_decoders.py
```diff
@@ -1752,6 +1752,67 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+        import subprocess
+
+        stream_index = 0
+        decoder = VideoDecoder(
+            NASA_VIDEO_ROTATED.path,
+            stream_index=stream_index,
+            dimension_order=dimension_order,
+        )
+
+        frame = decoder[0]
```
Dan-Flores: Lets modify this test to check accuracy for multiple frames.

Also, the `nasa_13013` video contains multiple video streams, so we could parametrize over `stream_index` as well.

test/test_decoders.py
```diff
@@ -1752,6 +1752,67 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+
+        # Get frame dimensions based on dimension order
+        if dimension_order == "NCHW":
+            # Frame is (C, H, W)
+            frame_h, frame_w = frame.shape[1], frame.shape[2]
+        else:
+            # Frame is (H, W, C)
+            frame_h, frame_w = frame.shape[0], frame.shape[1]
+
+        # Pixel-exact comparison with ffmpeg CLI (autorotate is on by default)
```
Dan-Flores: Does `autorotate` refer to an ffmpeg CLI option? It's not used in the command below, so this comment is a bit confusing.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -253,6 +261,28 @@ def __init__(
...
+        Returns:
+            The rotated frame tensor.
+        """
+        if self._rotation_k == 0:
+            return frame
+
+        if self._dimension_order == "NCHW":
+            dims = (-2, -1)
+        else:
+            dims = (-3, -2)
```
Dan-Flores: Is there a reason we use negative indices here? It works, but I had to think about it for a second.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -241,6 +240,15 @@ def __init__(
             custom_frame_mappings=custom_frame_mappings_data,
         )
 
+        # Precompute rotation parameters for torch.rot90
+        # k is the number of 90-degree counter-clockwise rotations
+        rotation = self.metadata.rotation
+        if rotation is not None and rotation != 0:
```
Dan-Flores: Checking `rotation != 0` shouldn't matter here, we will set `self._rotation_k` to 0 either way.

src/torchcodec/_core/Metadata.h
```diff
@@ -56,6 +56,11 @@ struct StreamMetadata {
   std::optional<int> width;
   std::optional<int> height;
   std::optional<AVRational> sampleAspectRatio;
+  // Rotation angle in degrees counter-clockwise (0, 90, 180, -90, -180).
+  // Extracted from the display matrix side data. This indicates how the video
+  // should be rotated for correct display. Common for videos recorded on mobile
+  // devices.
+  std::optional<double> rotation;
```
Dan-Flores, line 61: The information from this comment is also in `_metadata.py`, it's probably not necessary here.

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -612,6 +612,69 @@ int64_t computeSafeDuration(
...
+// The coded_side_data API was added in FFmpeg 6.1 (libavcodec 60.31.x)
+// FFmpeg >= 6.1: Use codecpar->coded_side_data
+#if LIBAVCODEC_VERSION_INT >= AV_VERSION_INT(60, 31, 100)
+  const AVPacketSideData* sideData = av_packet_side_data_get(
+      avStream->codecpar->coded_side_data,
+      avStream->codecpar->nb_coded_side_data,
+      AV_PKT_DATA_DISPLAYMATRIX);
+  if (sideData != nullptr && sideData->size >= 9 * sizeof(int32_t)) {
+    displayMatrix = reinterpret_cast<const int32_t*>(sideData->data);
+  }
```
Dan-Flores, line 638: Why do we verify the size of `sideData` here to be `>= 9 * sizeof(int32_t)`?  I found [one example in FFmpeg](https://ffmpeg.org/doxygen/7.0/matroskaenc_8c_source.html#l01417) where there is only a nullptr check, but its possible I am missing something.

test/resources/nasa_13013_rotated.mp4
Dan-Flores, line 1: Since we test that `rotation` is present in the metadata, it makes sense to check in a video.

`test_rotation_applied_to_frames` only needs a few frames, perhaps we can reduce the size of the video file and associated `all_frames_info` JSONs?

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -284,7 +314,7 @@ def _getitem_int(self, key: int) -> Tensor:
         assert isinstance(key, int)
 
         frame_data, *_ = core.get_frame_at_index(self._decoder, frame_index=key)
-        return frame_data
+        return self._apply_rotation(frame_data)
```
Dan-Flores: A side effect of applying the rotation at the python level is that we have to update each frame retrieval API.

I wonder if this should be handled in C++, similar to `maybePermuteHWC2CHW`. We would still need to add it to every `SingleStreamDecoder::getFrame ...` method, but it keeps the complexity in the C++ implementation.  
Let me know your thoughts @mollyxu @NicolasHug

src/torchcodec/_core/FFMPEGCommon.h
Dan-Flores: I'm not sure if this distinction is important to users or FFmpeg, but `optional` seems like a reasonable type to use here. Correct me if I'm wrong, the only difference is that we need to check `rotation is not None`?

src/torchcodec/decoders/_video_decoder.py
mollyxu: Thanks for the suggestion! I think it would make the public API more clean to do something similar to `maybePermuteHWC2CHW`

I was also considering applying the ffmpeg transpose filter as an option to consolidate the calls in one place. The con of that would be the pixels being moved instead of it just being a view operation

src/torchcodec/_core/FFMPEGCommon.cpp
mollyxu: Thanks for the suggestion! I think I was being overly cautious to check that the display matrix has 9 elements.

src/torchcodec/decoders/_video_decoder.py
mollyxu: I was using negative indices to account for batch vs unbatched call but it might not be applicable after moving the logic to C++

test/test_decoders.py
NicolasHug: Agreed on Daniel's suggestions. Kinda related but: do we actually need to check against FFmpeg? I'm hoping we can get away with just comparing the output of `NASA_VIDEO` and `NASA_VIDEO_ROTATED` (where we manually apply the 90 rotation to either, in order to compare them)?
To be clear I do agree that comparing against FFmpeg is the "golden standard" test... but I think we can get away with the simpler approach described above.

On parametrizating over streams: if we're going to check-in both json files for both stream 0 and stream 3, I agree, we might as well parametrize over both. But do we need to? Would a single stream be enough here?

test/test_decoders.py
```diff
@@ -1752,6 +1752,67 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
         assert "No fallback required" in str(decoder.cpu_fallback)
 
+    @needs_ffmpeg_cli
+    @pytest.mark.parametrize("dimension_order", ["NCHW", "NHWC"])
+    def test_rotation_applied_to_frames(self, dimension_order, tmp_path):
+        """Test that rotation is correctly applied to decoded frames.
+
+        Verifies pixel-exact match with ffmpeg CLI.
+        """
+        import subprocess
```
NicolasHug: If we keep the test against FFmpeg, this should be a top-level import.

src/torchcodec/_core/_metadata.py
```diff
@@ -102,6 +101,13 @@ class VideoStreamMetadata(StreamMetadata):
     (SAR --- not to be confused with Storage Aspect Ratio, also SAR),
     is the ratio between the width and height of each pixel
     (``fractions.Fraction`` or None)."""
+    rotation: float | None
+    """Rotation angle in degrees (counter-clockwise) from the display matrix
+    metadata. This indicates how the video should be rotated for correct
+    display. Common values are 0, 90, 180, or -90, -180 degrees. Videos recorded on
+    mobile devices often have rotation metadata. TorchCodec automatically
+    applies this rotation during decoding, so the returned frames are
+    in the correct orientation (float or None)."""
```
NicolasHug: Great comment! We should indicate that we're rounding to the closest multiple of 90.

Also: The height and width that we report, are they *pre-rotation* or *post-rotation*? I think we'll want them to be post-rotation, since this is what we'll return from our decoding methods? No matter what we choose, we should make sure to document it!

src/torchcodec/decoders/_video_decoder.py
NicolasHug: That's a good call-out, I agree that we should try to implement this logic in C++. The main reason is that, while we don't want to have a public C++ decoding API, we should still aim to have a fully-featured C++ decoder. Ideally, we want the python layer to be as slim as possible (with no logic). We can't always do that, but that's what we should aim for.
It makes a lot of sense to me to treat this rotation logic in the same way we treat `maybePermuteHWC2CHW`.

On the implementation, fully agreed that we should prefer using `torch::rot90` (which should still exist in C++) so that we can return a view instead of incurring a copy through filtergraph.

src/torchcodec/_core/FFMPEGCommon.cpp
NicolasHug: This is awful 😄 . But I think you're doing the right thing!

I'd just add a comment at the top of the function, with a high-level explanation of what's going on. Something like this (please check me on the versions and the API names):

> `av_stream_get_side_data`  was deprecated in version FFmpeg 6.0, but its replacement (`av_packet_side_data_get` + `codecpar->coded_side_data`) is only available from version 6.1. We need  some `#pragma` magic to silence the deprecation warning which our compile chain would otherwise treat as an error.


 (CC @scotts, I'm sure you will enjoy this).

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -952,6 +952,9 @@ std::string get_stream_json_metadata(
     map["sampleAspectRatioDen"] =
         std::to_string((*streamMetadata.sampleAspectRatio).den);
   }
+  if (streamMetadata.rotation.has_value()) {
+    map["rotation"] = std::to_string(*streamMetadata.rotation);
+  }
```
NicolasHug, line 959: We might need to use the same as https://github.com/meta-pytorch/torchcodec/pull/1200 instead of to_string. Let's monitor which PR gets merged first and edit the other one.
CC @Dan-Flores as you might need to edit yours

src/torchcodec/_core/_metadata.py
```diff
@@ -102,6 +101,24 @@ class VideoStreamMetadata(StreamMetadata):
...
+
+    .. note::
+
+        The :attr:`~torchcodec.decoders.VideoStreamMetadata.width` and
+        :attr:`~torchcodec.decoders.VideoStreamMetadata.height` attributes report
+        the **post-rotation** dimensions, i.e., the dimensions of frames as they
+        will be returned by TorchCodec's decoding methods. For videos with 90
+        or 270 degree rotation, this means width and height are swapped
+        compared to the raw encoded dimensions in the container.
+    """
```
Dan-Flores, line 119: This docstring is great, and renders correctly in the docs! It is a bit verbose, perhaps we could reduce it to something like this:
```
    """Rotation angle in degrees (counter-clockwise) from the display matrix
    metadata. This indicates how the video should be rotated for correct
    display. TorchCodec automatically applies this rotation during decoding 
    (rounded to the nearest multiple of 90 degrees), so the returned frames are 
    in the correct orientation (float or None).

    .. note::

        The :attr:`~torchcodec.decoders.VideoStreamMetadata.width` and
        :attr:`~torchcodec.decoders.VideoStreamMetadata.height` attributes report
        the **post-rotation** dimensions, i.e., the dimensions of frames as they
        will be returned by TorchCodec's decoding methods. For videos with 90
        or 270 degree rotation, this means width and height are swapped
        compared to the raw encoded dimensions in the container.
    """
```

src/torchcodec/_core/_metadata.py
```diff
@@ -102,6 +101,22 @@ class VideoStreamMetadata(StreamMetadata):
     (SAR --- not to be confused with Storage Aspect Ratio, also SAR),
     is the ratio between the width and height of each pixel
     (``fractions.Fraction`` or None)."""
+    rotation: float | None
```
NicolasHug, line 104: It's worth documenting the range of the values. My understanding is that FFmpeg outputs values in [-180, 180]. What about us? We mention a possible value of 270 degrees below, so I just want to make sure.

src/torchcodec/_core/_metadata.py
```diff
@@ -102,6 +101,22 @@ class VideoStreamMetadata(StreamMetadata):
...
+    """Rotation angle in degrees (counter-clockwise) from the display matrix
+    metadata. This indicates how the video should be rotated for correct
```
NicolasHug: This reads as the *non-rounded* rotation angle. Am I reading this correctly?

src/torchcodec/_core/Metadata.h
```diff
@@ -56,6 +56,7 @@ struct StreamMetadata {
   std::optional<int> width;
   std::optional<int> height;
   std::optional<AVRational> sampleAspectRatio;
+  std::optional<double> rotation;
```
NicolasHug, line 61: Please indicate the range here, I think it's [-180, 180]?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -32,6 +32,17 @@ int64_t getPtsOrDts(const UniqueAVFrame& avFrame) {
   return avFrame->pts == INT64_MIN ? avFrame->pkt_dts : avFrame->pts;
 }
 
+// Computes the rotation parameter k for torch::rot90 from a rotation angle in
+// degrees. k is the number of 90-degree counter-clockwise rotations (0-3).
+// Returns 0 for no rotation, 1 for 90°, 2 for 180°, 3 for 270°.
```
NicolasHug: I may very well be wrong but I'm confused by reading 270 here. My undertanding is that the input rotationDegrees is in [-180, 180]

It'd help to clarify the expected ranges everywhere. If it really is ` [-180, 180]`, can you please double-check that the implementation below is correct?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -153,8 +164,22 @@ void SingleStreamDecoder::initializeDecoder() {
       if (fps > 0) {
         streamMetadata.averageFpsFromHeader = fps;
       }
-      streamMetadata.width = avStream->codecpar->width;
-      streamMetadata.height = avStream->codecpar->height;
+      streamMetadata.rotation = getRotationFromStream(avStream);
+
+      // Report post-rotation dimensions: swap width/height for 90 or 270
```
NicolasHug: same here and every where else 270 is mentioned :)

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1398,6 +1435,25 @@ torch::Tensor SingleStreamDecoder::maybePermuteHWC2CHW(
...
+    return torch::rot90(hwcTensor, streamInfo.rotationK, {0, 1});
+  } else if (numDimensions == 4) {
+    // Batch of frames: NHWC format, rotate on (H, W) which are dims (1, 2)
+    return torch::rot90(hwcTensor, streamInfo.rotationK, {1, 2});
+  } else {
+    TORCH_CHECK(
+        false, "Expected tensor with 3 or 4 dimensions, got ", numDimensions);
+  }
+}
+
```
NicolasHug: There are a few places where we call `maybePermuteHWC2CHW` but we don't call `applyRotation`. Just checking: is that intentional?

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -241,6 +241,11 @@ class SingleStreamDecoder {
 
     VideoStreamOptions videoStreamOptions;
     AudioStreamOptions audioStreamOptions;
+
```
NicolasHug: Somewhere else in this file I think we'll want to add a comment to
```
FrameDims metadataDims_;
```

are these the dimensions before, or after rotation? My understanding is that they're post rotation. But... I'm confused. If they're post-rotation, then surely the way we allocate an output tensor here, is incorrect

https://github.com/meta-pytorch/torchcodec/blob/a9e1ae3440d1a3941be049a5c3d3c14c102e4b59/src/torchcodec/_core/SingleStreamDecoder.cpp#L740-L743

----

I can confirm that `        decoder_rotated.get_frames_in_range(0, 1)` in the test leads to:

```
E       RuntimeError: Expected pre-allocated tensor of shape 270x480x3, got [480, 270, 3]
```

I suspect other batch APIs which pre-allocate the output tensor might have the same issue.

src/torchcodec/_core/_metadata.py
mollyxu: Thanks for pointing this out! Our rotation metadata values should be in [-180, 180] as well. I used 270 in comments that were related to `torch::rot90` because the internal k parameter was the number of 90 degree rotations counter clockwise. It's definitely more confusing for users to see 270 rather than -90.

src/torchcodec/_core/_metadata.py
mollyxu: Would it be better to move `rounded to the nearest multiple of 90 degrees` from line 108 to line 105 so it's something like `Rotation angle in degrees (counter-clockwise rounded to the nearest multiple of 90 degrees)` for more clarity?

NicolasHug: yes, I think that would be very clear that it's the rounded angle!

src/torchcodec/_core/SingleStreamDecoder.cpp
mollyxu: Thanks for pointing this out! I missed these instances of `maybePermuteHWC2CHW` after merging the FPS changes from main

src/torchcodec/_core/SingleStreamDecoder.cpp
mollyxu: I have updated the comment to clarify that our rotation metadata values should be in [-180, 180]. The implementation below should be correct since -90 would be `- 1 % 4` which evaluates to `-1` in C++. Then the correction `k += 4` should result in k = 3, (3x90 degrees rotation counter clockwise), which is 90 degrees clockwise.

test/test_metadata.py
```diff
@@ -150,6 +154,18 @@ def test_get_metadata_audio_file(metadata_getter):
...
+def test_rotation_metadata():
+    """Test that rotation metadata is correctly extracted for rotated video."""
+    # NASA_VIDEO_ROTATED has 90-degree rotation metadata
+    decoder = VideoDecoder(NASA_VIDEO_ROTATED.path)
+    assert decoder.metadata.rotation is not None
+    assert decoder.metadata.rotation == 90
+
+    # NASA_VIDEO has no rotation metadata
+    decoder = VideoDecoder(NASA_VIDEO.path)
+    assert decoder.metadata.rotation is None
```
NicolasHug, line 161: Let's also assert something like this, to check that the height and width are reported post-rotation.

```py
assert (decoder_rotated.metadata.height, decoder_rotated.metadata.width) == (decoder.metadata.width, decoder.metadata.height)

```

src/torchcodec/_core/Metadata.h
```diff
@@ -53,9 +53,15 @@ struct StreamMetadata {
   std::optional<int64_t> numFramesFromContent;
 
   // Video-only fields
+  // Post-rotation dimensions
   std::optional<int> width;
   std::optional<int> height;
```
NicolasHug: Let's make it obvious from the name

```suggestion
  // Post-rotation dimensions
std::optional<int> postRotationWidth;
std::optional<int> postRotationHeight;
```

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -373,7 +380,13 @@ class SingleStreamDecoder {
   // 3. metdataDims_; the dimensions we learned from the metadata.
   std::vector<std::unique_ptr<Transform>> transforms_;
   std::optional<FrameDims> resizedOutputDims_;
+  // Post-rotation dimensions from metadata. Used for user-facing metadata and
+  // final output shape validation.
   FrameDims metadataDims_;
```
NicolasHug: Intuitively for me the "metadata dims" would be the pre-rotation dims, since this is what `avStream->codecpar->width` reports. So I'd suggest to rename this into `preRotationDims_`.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -684,7 +729,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
   const auto& videoStreamOptions = streamInfo.videoStreamOptions;
   FrameBatchOutput frameBatchOutput(
       frameIndices.numel(),
-      resizedOutputDims_.value_or(metadataDims_),
+      resizedOutputDims_.value_or(preRotationDims_),
```
NicolasHug: I think there's an issue here: if `resizedOutputDims_` is not null then we allocate an output tensor of the corresonding size.. and then rotate it. Which means that if the user expected (300, 500), they'll get (500, 300)?

See e.g. below. I'm not sure how to address that yet 🤔 

```py
from torchcodec.transforms import Resize


desired_H, desired_W = 300, 500
dec = VideoDecoder("test/resources/nasa_13013_rotated.mp4", transforms=[Resize((desired_H, desired_W))])

assert dec[0].shape == (3, desired_H, desired_W)  # fails, output is (3, desired_W, desired_H)
```

mollyxu: Good catch! We should rotate the frame first then apply the transforms to be consistent with the expected behavior. 

Sticking with the current approach of using `torch::rot90()` would mean we need to map the coordinates for the transforms, increasing code complexity. We could use the ffmpeg `transpose` filter before the other transforms instead of `torch::rot90()`. Do you think this would be a good tradeoff between performance and code complexity?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -543,24 +578,68 @@ void SingleStreamDecoder::addVideoStream(
...
+    for (const auto& transform : transforms_) {
+      auto dims = transform->getOutputFrameDims();
+      if (dims.has_value()) {
+        FrameDims swapped(dims->width, dims->height);
+        if (transform->isResize()) {
+          preRotationTransforms.push_back(
+              std::make_unique<ResizeTransform>(swapped));
+        } else {
+          preRotationTransforms.push_back(
+              std::make_unique<CropTransform>(swapped));
```
mollyxu: This would not preserve the behavior for RandomCrop since we would be losing `x_` and `y_`

We can't access `x_` and `y_` because they are immutable private fields on the CropTransform object. We could add getters such that we could pass them swapped.

NicolasHug: Thanks for raising this - I think it means we'd be implicitly converting `RandomCrop` into `CenterCrop`? That's not desirable.

Maybe we should expose a new method on transforms e.g. `transform.transpose()` which would do that work internally? Then maybe we'd only need one single loop (the original one), which would look something like this:

```C
for (auto& transform : transforms) {
  if (k == 1 || k == 3){
    transform = transform.transpose()
  }
  // rest is the same
}
```

Hopefully `resizedOutputDims_` would be natively handled?

I'm not sure I'm being super clear, happy to sync to chat more.

-----

On a separate note, I'm wondering: does this logic still work if we have more than one transform potentially changing the dimensions? E.g. we should make sure that passing e.g. `[Resize(...), Resize(...), Crop(...)]` still works as expected.

src/torchcodec/_core/Transform.cpp
```diff
@@ -125,4 +125,53 @@ void CropTransform::validate(const FrameDims& inputDims) const {
...
+std::string RotationTransform::getFilterGraphCpu() const {
+  switch (rotation_) {
+    case Rotation::None:
+      return "";
+    case Rotation::Ccw90:
+      return "transpose=cclock";
+    case Rotation::Cw90:
+      return "transpose=clock";
+    case Rotation::Rotate180:
+      return "hflip,vflip";
```
mollyxu, line 174: From my understanding, `hflip,vflip` could be faster than `cclock,cclock` for 180 degrees

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -540,19 +558,37 @@ void SingleStreamDecoder::addVideoStream(
         activeStreamIndex_, customFrameMappings.value());
   }
 
-  metadataDims_ =
-      FrameDims(streamMetadata.height.value(), streamMetadata.width.value());
-  FrameDims currInputDims = metadataDims_;
+  // Store pre-rotation dimensions (raw encoded dimensions from FFmpeg)
+  preRotationDims_ = FrameDims(
+      streamMetadata.preRotationHeight.value(),
+      streamMetadata.preRotationWidth.value());
```
NicolasHug: Nit, it might best to set `preRotationDims_` as soon as `streamMetadata.preRotationHeight` and `streamMetadata.preRotationWidth` are available. That's just a little bit easier to follow.

NicolasHug: Going back to this, it seems that `preRotationWidth` and `preRotationHeight` are only ever used here, to define `preRotationDims_`:

```bash
~/dev/torchcodec (rotation*) » git grep -n -C 1 -e preRotationHeight -e preRotationWidth
src/torchcodec/_core/Metadata.h-59-  // Pre-rotation dimensions
src/torchcodec/_core/Metadata.h:60:  std::optional<int> preRotationWidth;
src/torchcodec/_core/Metadata.h:61:  std::optional<int> preRotationHeight;
src/torchcodec/_core/Metadata.h-62-  std::optional<AVRational> sampleAspectRatio;
--
src/torchcodec/_core/SingleStreamDecoder.cpp-155-      // Store pre-rotation (raw encoded) dimensions
src/torchcodec/_core/SingleStreamDecoder.cpp:156:      streamMetadata.preRotationWidth = avStream->codecpar->width;
src/torchcodec/_core/SingleStreamDecoder.cpp:157:      streamMetadata.preRotationHeight = avStream->codecpar->height;
src/torchcodec/_core/SingleStreamDecoder.cpp-158-
--
src/torchcodec/_core/SingleStreamDecoder.cpp-562-  preRotationDims_ = FrameDims(
src/torchcodec/_core/SingleStreamDecoder.cpp:563:      streamMetadata.preRotationHeight.value(),
src/torchcodec/_core/SingleStreamDecoder.cpp:564:      streamMetadata.preRotationWidth.value());
src/torchcodec/_core/SingleStreamDecoder.cpp-565-
```

so if we move the definition of `preRotationDims_` earlier as suggested, we'll be able to remove those fields!

src/torchcodec/_core/Metadata.h
```diff
@@ -53,9 +53,15 @@ struct StreamMetadata {
...
 
   // Video-only fields
-  std::optional<int> width;
-  std::optional<int> height;
+  // Post-rotation dimensions
+  std::optional<int> postRotationWidth;
+  std::optional<int> postRotationHeight;
+  // Pre-rotation dimensions
+  std::optional<int> preRotationWidth;
+  std::optional<int> preRotationHeight;
```
NicolasHug: See my other comments below, I think we should be able to get rid of both preRotationWidth and preRotationHeight

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -540,19 +558,37 @@ void SingleStreamDecoder::addVideoStream(
...
+
+  FrameDims currInputDims = preRotationDims_;
+
+  // If there's rotation, prepend a RotationTransform to handle it in the
+  // filter graph. This way user transforms (resize, crop) operate in
+  // post-rotation coordinate space, preserving x/y coordinates for crops.
```
NicolasHug, line 565: Can I also suggest to add something along these lines (feel free to edit):

```
It is critical to apply the rotation *before* applying any user-supplied transforms.
By design, we want 
A: VideoDecoder(..., transforms=tv_transforms)[i]  # decoder-native transforms
to be equivalent to
B: tv_transforms(VideoDecoder(...)[i])  # transforms applied after decoding
In B, the rotation is clearly applied before the transforms.
```

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -540,19 +558,37 @@ void SingleStreamDecoder::addVideoStream(
...
+  // TODO: benchmark the performance of doing this additional filtergraph
+  // transform
+  if (streamMetadata.rotation.has_value()) {
+    Rotation rotation = rotationFromDegrees(*streamMetadata.rotation);
```
NicolasHug: Nit, if `rotationFromDegrees` were to accept a std::optional and return `Rotation::None` upon `nullopt`, we could avoid the first `if (streamMetadata.rotation.has_value()) `.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -540,19 +558,37 @@ void SingleStreamDecoder::addVideoStream(
...
+    transform->validate(currInputDims);
     if (transform->getOutputFrameDims().has_value()) {
       resizedOutputDims_ = transform->getOutputFrameDims().value();
+      currInputDims = resizedOutputDims_.value();
     }
-    transform->validate(currInputDims);
-    currInputDims = resizedOutputDims_.value_or(metadataDims_);
-
-    // Note that we are claiming ownership of the transform objects passed in to
-    // us.
```
NicolasHug, line 555: I think we'll want to preserve this comment?

src/torchcodec/_core/Transform.cpp
NicolasHug: Would you have any ref to share for that? A quick benchmark would be helpful too (claude can probably come up with something quick with `time` and the `ffmpeg` CLI)

src/torchcodec/_core/Transform.h
```diff
@@ -83,4 +83,33 @@ class CropTransform : public Transform {
...
+enum class Rotation {
+  None, // 0°
+  Ccw90, // 90° counter-clockwise
+  Cw90, // 90° clockwise (or -90°)
+  Rotate180 // 180° (or -180°)
+};
+
+// Converts rotation degrees from video metadata to Rotation enum.
+// Rounds to nearest multiple of 90 degrees before converting.
+// Returns Rotation::None for unsupported rotation values (e.g., 270°).
```
NicolasHug: I think this will be less surprising to read if we clarify that the input is expected in [-180, 180].
BTW instead of returning None we might as well `TORCH_CHECK(false)`? I think that'd qualify as a bug if we ever had something outside of that range, and we'd rather know

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -680,7 +716,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
   const auto& videoStreamOptions = streamInfo.videoStreamOptions;
   FrameBatchOutput frameBatchOutput(
       frameIndices.numel(),
-      resizedOutputDims_.value_or(metadataDims_),
+      resizedOutputDims_.value_or(preRotationDims_),
```
NicolasHug: I got quite confused reading this, I thought it should be `postRotationDims_`.
But IIUC this is equivalent, because *if* there is a rotation, then `resizedOutputDims_` is necessarily non-null. Which means that if `resizedOutputDims_` is null, then there *cannot* be a rotation, which means that `preRotationDims_` == `postRotationDims_`.

🤔 Not sure how to make that obvious. Any thoughts?

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -362,18 +362,20 @@ class SingleStreamDecoder {
...
+  //    always takes priority.
+  // 2. The dimensions of the actual decoded AVFrame. This can change
   //    per-decoded frame, and is unknown in SingleStreamDecoder. Only the
   //    DeviceInterface learns it immediately after decoding a raw frame but
-  //    before the color transformation.
-  // 3. metdataDims_; the dimensions we learned from the metadata.
+  //    before the color conversion.
+  // 3. preRotationDims_; the raw encoded dimensions from FFmpeg metadata
+  //    (before any rotation is applied). Used as fallback for tensor
+  //    allocation when resizedOutputDims_ is not set.
```
NicolasHug: We might want to clarify here that when that's the case, no rotation is needed, so `preRotationDims_` is the right thing

src/torchcodec/_core/Transform.h
```diff
@@ -83,4 +83,33 @@ class CropTransform : public Transform {
   std::optional<int> y_;
 };
 
+// Rotation values for RotationTransform.
+// These correspond to video metadata rotation angles.
+enum class Rotation {
+  None, // 0°
+  Ccw90, // 90° counter-clockwise
+  Cw90, // 90° clockwise (or -90°)
+  Rotate180 // 180° (or -180°)
```
NicolasHug: Nit, traditionally, enum value are written in `ALL_CAPS`. I find it slightly easier to parse - but maybe that's just my own habit.

src/torchcodec/_core/Transform.h
```diff
@@ -83,4 +83,33 @@ class CropTransform : public Transform {
...
+};
+
+// Converts rotation degrees from video metadata to Rotation enum.
+// Rounds to nearest multiple of 90 degrees before converting.
+// Returns Rotation::None for unsupported rotation values (e.g., 270°).
+Rotation rotationFromDegrees(double degrees);
+
+// Applies rotation using FFmpeg's transpose/flip filters.
```
NicolasHug: let's just clarify this just applies rotation as multiples of 90 degrees. TorchVision has a RandomRotate transform that can do arbitrary angle rotation, we want to make sure it's obvious that it's not the same thing.

test/test_decoders.py
```diff
@@ -1910,6 +1911,204 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+            assert frames.data.shape == (2, 3, desired_H, desired_W)
+
+    def test_rotation_with_transform_pipeline(self):
+        """Test that a pipeline of multiple transforms works correctly with rotated videos.
+
+        This test verifies that chaining multiple transforms (e.g., Resize -> Resize -> Crop)
+        works as expected when the video has rotation metadata. Each transform should
+        operate on the output of the previous transform in post-rotation coordinate space.
+        """
+        from torchcodec.transforms import CenterCrop, Resize
```
NicolasHug: Any reason to import those here? (same for other tests)

test/test_decoders.py
```diff
@@ -1910,6 +1911,204 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+        from torchcodec.transforms import RandomCrop
+
+        # Test with various crop dimensions
+        test_cases = [
+            (100, 150),  # landscape crop
+            (150, 100),  # portrait crop
+            (100, 100),  # square crop
+        ]
+
+        for desired_H, desired_W in test_cases:
```
NicolasHug: Use a pytest parametrization instead of a for loop. 

for loops are OK when there is a lot of costly initialization work to be done before the for loop (which would be unnecessarily duplicated when parametrizing over). But here, that's not the case.

test/test_decoders.py
```diff
@@ -1910,6 +1911,204 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+            # Use fixed seed for reproducibility
+            torch.manual_seed(42)
```
NicolasHug: We should actually not need to set a seed, this test isn't testing anything random: the output should always be of the expected shape, regardless of the seed.

test/test_decoders.py
```diff
@@ -1910,6 +1911,204 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+            decoder = VideoDecoder(
+                NASA_VIDEO_ROTATED.path,
+                transforms=[RandomCrop((desired_H, desired_W))],
+            )
+            frame = decoder[0]
+
+            # Frame shape should be (C, H, W) in NCHW format (default)
```
NicolasHug: I'd suggest to remove this comment, and most others. They don't add any value over the actual code - claude has a tendency to be too verbose.

test/test_decoders.py
```diff
@@ -1910,6 +1911,204 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
...
+        (H, W) regardless of the video's rotation metadata.
+        """
+        from torchcodec.transforms import RandomCrop
+
+        # Test with various crop dimensions
+        test_cases = [
+            (100, 150),  # landscape crop
+            (150, 100),  # portrait crop
+            (100, 100),  # square crop
+        ]
```
NicolasHug: Let's remove these comments, they don't add anything

test/test_decoders.py
NicolasHug: Speaking of parametrization, from a quick glance these 3 tests `test_rotation_with_resize_transform`, `test_rotation_with_center_crop_transform`, and `test_rotation_with_random_crop_transform` are 90% the same. We could consider factorizing them into a single test that parametrizes over the transform itself? There would be 2 levels of parametrization, something like this:


```py
@pytest.mark.parametrize("desired_H, desired_W", ((100, 150), (...))
@pytest.mark.parametrize("TransformClass", (Resize, CenterCrop, RandomCrop))
def the_test...
```

src/torchcodec/_core/SingleStreamDecoder.cpp
mollyxu: We could either
1. change `preRotationDims_` back to `metadataDims_`
2. add a helper function `getOutputDims()` in `SingleStreamDecoder.h` and a comment explaining why it would be `resizedOutputDims_.value_or(preRotationDims_)`
3. add comments at all call sites as is (this might be not as clean as there are a few call sites)

src/torchcodec/_core/Transform.cpp
mollyxu: Here's what I got from running a quick benchmark

```
Video: /home/mollyxu/torchcodec/benchmarks/decoders/../../test/resources/nasa_13013.mp4
Resolution: 320x180
Frames: 13.000000
Duration: 325s

Running benchmark (10 iterations each)...

Results:
--------------------------------------------------
  hflip,vflip                                153.77 ms (baseline)
  cclock,cclock                              220.49 ms (1.43x)
  clock,clock                                195.48 ms (1.27x)
```

Issue #353, opened 2024-11-08
Test performance of numFfmpegThreads with CUDA


ahmadsharif1:
I don't know how these two interact. From some experiments I ran earlier this year, CUDA performance was best with numFfmpegThreads=1. But we should do these experiments again and that's the case, we should always set 1 thread when using CUDA decoding.

NicolasHug:
We should also update the docstring of `num_ffmpeg_threads` to reflect our findings, as originally attempted in https://github.com/pytorch/torchcodec/pull/348

oliverdain:
@NicolasHug I was looking for exactly this information and found this thread. I also checked #348 and it seems like the thinking is that the number of ffmpeg threads is releavant on GPU but it's not clear if it helps or not. Do you have some information you could post here?

NicolasHug:
@oliverdain I looked at it again and ran benchmarks. I can confirm this has no effect on GPU decoding. I opened https://github.com/meta-pytorch/torchcodec/pull/1174/ to document this.

PR #1174, opened 2026-01-20
Clarify that `num_ffmpeg_threads` is only for CPU

NicolasHug:
Closes https://github.com/meta-pytorch/torchcodec/issues/353

The `num_ffmpeg_threads` parameter of `VideoDecoder` only affects CPU decoding. It doesn't affect GPU decoding.

To verify this I looked for uses of  `thread_count` in the FFmpeg code-base: it's only used for decoding, never for demuxing, and it's never used in the nvdec codec:

```bash
~/dev/FFmpeg (master) » git grep thread_count libavformat/ | wc   # not used for demuxing
      0       0       0
```

```bash
~/dev/FFmpeg (master) » git grep -C 1 thread_count libavcodec/*nvdec*  # not used in nvdec codec. Weirdly there's one hit for logging.
libavcodec/nvdec.c-            av_log(avctx, AV_LOG_WARNING, "Try lowering the amount of threads. Using %d right now.\n",
libavcodec/nvdec.c:                   avctx->thread_count);
libavcodec/nvdec.c-        }


```

And to be extra sure, I ran benchmark showing no difference between calls:

```py


import torch


    for _ in range(warmup):
        f(*args, **kwargs)

    times = []
    for _ in range(num_exp):
        start = perf_counter_ns()
        f(*args, **kwargs)
        end = perf_counter_ns()
        times.append(end - start)

    mul = {
        "ns": 1,
        "µs": 1e-3,
        "ms": 1e-6,
        "s": 1e-9,
    }[unit]
    times = times * mul
    std = times.std().item()
    med = times.median().item()
    return med


def f(num_ffmpeg_threads):
    decoder = VideoDecoder("test/resources/testsrc2.mp4", device="cuda", num_ffmpeg_threads=num_ffmpeg_threads)
    decoder[:]

report_stats(bench(f, warmup=1, num_exp=20, num_ffmpeg_threads=0))
report_stats(bench(f, warmup=1, num_exp=20, num_ffmpeg_threads=1))
report_stats(bench(f, warmup=1, num_exp=20, num_ffmpeg_threads=4))
report_stats(bench(f, warmup=1, num_exp=20, num_ffmpeg_threads=16))
```

```
~/dev/torchcodec-cuda (main*) » python bench_ffmpeg_threads.py                   nicolashug@nicolashug-fedora-PW0H326Y
med = 274.31ms +- 3.57
med = 281.61ms +- 5.24
med = 275.53ms +- 4.08
med = 280.13ms +- 5.53
```

### DIFF
--- a/benchmarks/encoders/benchmark_encoders.py
+++ b/benchmarks/encoders/benchmark_encoders.py
@@ -71,7 +71,7 @@ def report_stats(times, num_frames, nvenc_metrics=None, prefix="", unit="ms"):
             f"GPU memory used:      med = {mem_used_median:.1f} MB, max = {mem_used_max:.1f} MB"
         )
         print(
-            f"NVENC utilization:    med = {nvenc_metrics["utilization"].median():.1f}%,     max = {util_max:.1f}%"
+            f"NVENC utilization:    med = {nvenc_metrics['utilization'].median():.1f}%,     max = {util_max:.1f}%"
         )
 
 
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -104,7 +104,8 @@ class VideoDecoder:
                 cheap no-copy operation that allows these frames to be
                 transformed using the `torchvision transforms
                 <https://pytorch.org/vision/stable/transforms.html>`_.
-        num_ffmpeg_threads (int, optional): The number of threads to use for decoding.
+        num_ffmpeg_threads (int, optional): The number of threads to use for CPU decoding.
+            This has no effect when using GPU decoding.
             Use 1 for single-threaded decoding which may be best if you are running multiple
             instances of ``VideoDecoder`` in parallel. Use a higher number for multi-threaded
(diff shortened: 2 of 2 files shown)


benchmarks/encoders/benchmark_encoders.py
```diff
@@ -71,7 +71,7 @@ def report_stats(times, num_frames, nvenc_metrics=None, prefix="", unit="ms"):
             f"GPU memory used:      med = {mem_used_median:.1f} MB, max = {mem_used_max:.1f} MB"
         )
         print(
-            f"NVENC utilization:    med = {nvenc_metrics["utilization"].median():.1f}%,     max = {util_max:.1f}%"
+            f"NVENC utilization:    med = {nvenc_metrics['utilization'].median():.1f}%,     max = {util_max:.1f}%"
```
NicolasHug, line 74: Drive-by, `flake8` was failing on this locally. The syntax was using double quotes within a double-quoted `f-string`. This is valid syntax only from Python 3.12 (https://peps.python.org/pep-0701/).
Our linter didn't catch it because it's running on 3.12.

PR #1175, opened 2026-01-20
Add cmake option to disable setting rpath for homebrew ffmpeg

Dan-Flores:


src/torchcodec/_core/CMakeLists.txt
```diff
@@ -8,6 +8,8 @@ find_package(pybind11 REQUIRED)
 find_package(Torch REQUIRED)
 find_package(Python3 ${PYTHON_VERSION} EXACT COMPONENTS Development)
 
+option(TORCHCODEC_DISABLE_HOMEBREW_RPATH "Do not add Homebrew FFmpeg library path to rpath on macOS" OFF)
```
NicolasHug: Nit: use the same pattern as below for `TORCHCODEC_DISABLE_COMPILE_WARNING_AS_ERROR` so that we don't have to define the default value in  two separate places?

PR #1176, opened 2026-01-21
Add NVDEC reconfiguration capability to improve NVDEC cache hit rate and reduce initialization overhead.

IteratorandIterator:
Added NVDEC decoder reconfiguration capability on top of the NVDEC cache, which improves cache hit rate. Re-creating a completely new decoder takes approximately 50–70 milliseconds to initialize, whereas reconfiguring an existing decoder reduces initialization time to about 5–6 milliseconds—a 10× reduction in initialization latency (tested on an H100 GPU).

### DIFF
--- a/examples/decoding/performance_tips.py
+++ b/examples/decoding/performance_tips.py
@@ -52,7 +52,6 @@ to increase performance.
 # - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_at` for timestamps
 # - :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_in_range` for time ranges
 #
-# %%
 # **When to use:**
 #
 # - Decoding multiple frames
@@ -136,9 +135,8 @@ to increase performance.
 # (NVDEC) on supported hardware. This keeps decoded tensors in GPU memory,
 # avoiding expensive CPU-GPU transfers for downstream GPU operations.
 #
-# %%
-# **Recommended:  use the Beta Interface!!**
-# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# **Recommended: use the Beta Interface!!**
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 #
 # We recommend you use the new "beta" CUDA interface which is significantly faster than the previous one, and supports the same features:
 #
@@ -147,7 +145,6 @@ to increase performance.
 #     with set_cuda_backend("beta"):
 #         decoder = VideoDecoder("file.mp4", device="cuda")
 #
-# %%
 # **When to use:**
 #
 # - Decoding large resolution videos
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -517,14 +517,14 @@ UniqueAVFrame CudaDeviceInterface::convertCUDATensorToAVFrameForEncoding(
   int height = static_cast<int>(tensor.size(1));
   int width = static_cast<int>(tensor.size(2));
 
-  // TODO-VideoEncoder: Unify AVFrame creation with CPU version of this method
+  // TODO-VideoEncoder: (P1) Unify AVFrame creation with CPU method
   avFrame->format = AV_PIX_FMT_CUDA;
   avFrame->height = height;
   avFrame->width = width;
   avFrame->pts = frameIndex;
 
   // FFmpeg's av_hwframe_get_buffer is used to allocate memory on CUDA device.
-  // TODO-VideoEncoder: Consider using pytorch to allocate CUDA memory for
+  // TODO-VideoEncoder: (P2) Consider using pytorch to allocate CUDA memory for
   // efficiency
   int ret =
       av_hwframe_get_buffer(codecContext->hw_frames_ctx, avFrame.get(), 0);
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -770,19 +770,19 @@ void VideoEncoder::initializeEncoder(
   avCodecContext_.reset(avCodecContext);
 
   // Store dimension order and input pixel format
-  // TODO-VideoEncoder: Remove assumption that tensor in NCHW format
+  // TODO-VideoEncoder: (P2) Enable tensors in NHWC shape
   auto sizes = frames_.sizes();
   inPixelFormat_ = AV_PIX_FMT_GBRP;
   inHeight_ = static_cast<int>(sizes[2]);
   inWidth_ = static_cast<int>(sizes[3]);
 
   // Use specified dimensions or input dimensions
-  // TODO-VideoEncoder: Allow height and width to be set
+  // TODO-VideoEncoder: (P2) Allow height and width to be set
   outWidth_ = inWidth_;
   outHeight_ = inHeight_;
 
   if (videoStreamOptions.pixelFormat.has_value()) {
-    // TODO-VideoEncoder: Enable pixel formats to be set by user
+    // TODO-VideoEncoder: (P2) Enable pixel formats to be set by user on GPU
     // and handled with the appropriate NPP function on GPU.
     if (frames_.device().is_cuda()) {
       TORCH_CHECK(
@@ -813,7 +813,7 @@ void VideoEncoder::initializeEncoder(
   avCodecContext_->width = outWidth_;
   avCodecContext_->height = outHeight_;
   avCodecContext_->pix_fmt = outPixelFormat_;
-  // TODO-VideoEncoder: Add and utilize output frame_rate option
+  // TODO-VideoEncoder: (P1) Add and utilize output frame_rate option
   avCodecContext_->framerate = av_d2q(inFrameRate_, INT_MAX);
   avCodecContext_->time_base = av_inv_q(avCodecContext_->framerate);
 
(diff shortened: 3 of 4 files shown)


src/torchcodec/_core/NVDECCache.cpp
```diff
@@ -25,36 +25,133 @@ NVDECCache& NVDECCache::getCache(const torch::Device& device) {
...
+        it2->second.erase(bg);
+        return std::make_tuple(cache_type, std::move(decoder), context.second.decoderID);
+      }
+    }
+
+    for (auto bg = it2->second.begin(), ed = it2->second.end(); bg != ed; ++bg) {
+      const auto& context = *bg;
+      if (
+        context.second.ulMaxWidth >= videoFormat->coded_width &&
+        context.second.ulMaxHeight >= videoFormat->coded_height
```
NicolasHug: Quick question: does there need to be any condition on `numDecodeSurfaces` for the decoder to be eligible for re-configuration?

IteratorandIterator: The numDecodeSurfaces value returned by each reconfiguration should be less than the ulMaxNumDecodeSurfaces specified when creating the video parser. Since the current code fixes ulMaxNumDecodeSurfaces to 8 when creating the video parser, if the tests pass under this setting, it indicates that for most videos the required ulNumDecodeSurfaces is less than or equal to 8. Therefore, I did not add an explicit check here, but it is indeed necessary and important to enforce this constraint.

IteratorandIterator: I will add this constraint soon~

IteratorandIterator:
> Thanks a lot for the great PR @IteratorandIterator ! This is definitely something we'll want to include if we can verify the performance improvements.
> 
> Just checking my understanding first: before, we would never reconfigure a decoder, so to have a cache hit we expected all of codecType, width , height, chromaFormat, bitDepthLumaMinus8, and numDecodeSurfaces to match exactly. Now, we still require all of these to match to re-use the decoder, but we allow width , height, and numDecodeSurfaces to differ as long as width and height are smaller than the cache entry - in which case we can reconfigure.
> 
> Is that understanding correct?
> 
> Regarding the performance improvement, would you be able to share a quick reproducible benchmark showing the performance gain? I assume we should observe the most gain when decoding multiple videos sequentially with decreasing height and width?
> 
> Let me know if that's something you have the bandwidth for, and thanks a lot for the PR!

Hi, @NicolasHug !

When creating a decoder for the first time, it is usually necessary to specify ulMaxHeight and ulMaxWidth. During subsequent reconfiguration, it is only required that the new bitstream’s coded_height and coded_width are less than or equal to the ulMaxHeight and ulMaxWidth specified at creation time. In other words, If relatively large ulMaxHeight and ulMaxWidth values (for example, 4K) are specified manually when creating the decoder, it will result in a higher cache hit rate. In this PR, the reason for setting these values to the resolution of the first video that uses the decoder is to remain consistent with the existing code logic, and I believe this approach is also reasonable.

Regarding ulNumDecodeSurfaces, based on my reading of the official NVIDIA documentation as well as my own experimental results, the value set when creating the decoder usually does not affect the outcome. This is because the pfnSequenceCallback function returns the actual ulNumDecodeSurfaces required by the current bitstream and updates the value dynamically. However, the ulNumDecodeSurfaces value returned during reconfiguration must not exceed the ulMaxNumDecodeSurfaces specified when creating the video parser; otherwise, the decoding result will be incorrect.

Oh— I just realized that I did not add a constraint to ensure that the current bitstream’s ulNumDecodeSurfaces is less than or equal to ulMaxNumDecodeSurfaces. That said, the current code fixes ulMaxNumDecodeSurfaces to 8, so the test videos likely all use a ulNumDecodeSurfaces value that is less than or equal to 8 by default.

I would be very happy to provide a reproducible benchmark. I also noticed that some test results did not pass, and I am not sure whether this is due to a logic issue in the code I provided or some edge cases that I did not consider. If you could let me know how to inspect the source code of the tests being run, I would be glad to fix these bugs and help merge the reconfiguration functionality into torchcodec.

NicolasHug:
Thanks for the details!

Regarding the test failures, if you click on one of those links

image

you should then have access to the logs like this one: https://github.com/meta-pytorch/torchcodec/actions/runs/21201605282/job/61015703549?pr=1176

The logs aren't always easy to read because stdout and stderr are interleaved. Here, I think the relevant errors are:

```
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp: In member function ‘int facebook::torchcodec::BetaCudaDeviceInterface::reconfigureNVDECDecoder(CUVIDEOFORMAT*)’:
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::ulHeight’ [-Werror=missing-field-initializers]
  392 |   CUVIDRECONFIGUREDECODERINFO info = { 0 };
      |                                          ^
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::ulTargetWidth’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::ulTargetHeight’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::ulNumDecodeSurfaces’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::reserved1’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::display_area’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::target_rect’ [-Werror=missing-field-initializers]
/__w/torchcodec/torchcodec/meta-pytorch/torchcodec/src/torchcodec/_core/BetaCudaDeviceInterface.cpp:392:42: error: missing initializer for member ‘_CUVIDRECONFIGUREDECODERINFO::reserved2’ [-Werror=missing-field-initializers]
cc1plus: all warnings being treated as errors
```

IteratorandIterator:
Thanks a lot! 

I've already fixed the bug. If this code passes testing, I'll start preparing the reproducible benchmark.

IteratorandIterator:
Sorry @NicolasHug , I misexplained the part about ulNumDecodeSurfaces yesterday. It seems that ulMaxNumDecodeSurfaces, specified when creating the parser, is actually the one that can be dynamically changed, and the number of surfaces used during reconfiguration should be less than the ulNumDecodeSurfaces set when creating the decoder. If this passes testing, the conclusion should be correct.

IteratorandIterator:
Hi @NicolasHug , I found all errors stem from the issue below. My added code shouldn't introduce new dependencies—it compiles and runs fine on my local Linux system. Why is this happening🙏?
image

NicolasHug:
The errors were not related to your code, I think the CI was triggered at a very specific (unlucky) time where the wheels were built with yesterday's torch nightly while the tests were against today's torch nightlies, which caused a conflict. Things should be fine now!

IteratorandIterator:
Thanks! The two errors do not appear to be related to my code. Based on this, can we conclude that the tests have been successfully passed?

image

image

NicolasHug:
> Based on this, can we conclude that the tests have been successfully passed?

Yes :) 

I'm happy to go ahead, pending benchmark results. I haven't reviewed in details yet but I suspect I'll be making some minor changes to the PR: for the sake of time, if that's OK with you, I'll be directly editing the PR?

> When creating a decoder for the first time, it is usually necessary to specify ulMaxHeight and ulMaxWidth. During subsequent reconfiguration, it is only required that the new bitstream’s coded_height and coded_width are less than or equal to the ulMaxHeight and ulMaxWidth specified at creation time. In other words, If relatively large ulMaxHeight and ulMaxWidth values (for example, 4K) are specified manually when creating the decoder, it will result in a higher cache hit rate. In this PR, the reason for setting these values to the resolution of the first video that uses the decoder is to remain consistent with the existing code logic, and I believe this approach is also reasonable

Thanks for the details. Based on this, do you think it'd make sense to eventually implement a LRU cache eviction mechanism?

IteratorandIterator:
> for the sake of time, if that's OK with you, I'll be directly editing the PR?

I’m okay with that, please edit the PR as needed. And I’m happy to help if anything comes up.

IteratorandIterator:
> do you think it'd make sense to eventually implement a LRU cache eviction mechanism?

Yes, I think it would make sense. I also think it's worth considering assigning each decoder an eviction priority—for instance, among decoders with the same codecType, evict those with smaller ulMaxHeight, ulMaxWidth, or ulMaxNumDecodeSurfaces first?

Issue #1103, opened 2025-12-05
Reference resource generation CI jobs are failing


scotts:
See: https://github.com/meta-pytorch/torchcodec/actions/runs/19951109109/job/57245904008?pr=1100

It happens every time, so I think it's not a transient failure. And no other CI jobs encounter it. We may have missed some test-infra related change we did to other CI jobs.

PR #1187, opened 2026-01-23
CI jobs cleanups

NicolasHug:
Some minor CI cleanups and alignment:

- Align on using `miniforge` instead of `miniconda`. For a reason I ignore the `.github/workflows/reference_resources.yaml` job fails when using `miniforge`, so it stays on `miniconda`.
- Update `conda-incubator/setup-miniconda@v2` to v3
- Use the same `pytest` invocation in all CI jobs.

I might follow-up with crazier refacs - not sure yet.

The resource-generation job is failing, but it's *very* unlikely for these changes to be the source of the failure. They are failing already on `main` (https://github.com/meta-pytorch/torchcodec/pull/1196) and they've been failing for weeks (https://github.com/meta-pytorch/torchcodec/issues/1103). I'm attempting to fix them separately in https://github.com/meta-pytorch/torchcodec/pull/1197, so I kindly ask that we ignore these failures on this PR.

### DIFF
--- a/.github/workflows/reference_resources.yaml
+++ b/.github/workflows/reference_resources.yaml
@@ -55,10 +55,9 @@ jobs:
         python-version: ['3.10']
         ffmpeg-version-for-tests: ['4.4.2', '5.1.2', '6.1.1', '7.0.1']
     steps:
-      - uses: actions/download-artifact@v4
-        with:
-          name: meta-pytorch_torchcodec__${{ matrix.python-version }}_cpu_x86_64
-          path: pytorch/torchcodec/dist/
+      - name: Check out repo
+        uses: actions/checkout@v3
+
       - name: Setup conda env
         uses: conda-incubator/setup-miniconda@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/windows_wheel.yaml
```diff
@@ -113,5 +113,4 @@ jobs:
         run: bash packaging/install_test_dependencies.sh
 
       - name: Run Python tests
-        run: |
-          pytest test -vvv
+        run: pytest --override-ini="addopts=-v" test
```
NicolasHug, line 116: Before, we were using

```
pytest test -vvv
```

which would use the default pytest options we have defined in our `pyproject.toml`:

https://github.com/meta-pytorch/torchcodec/blob/e2ae6940764d0258e283977d19f5634df9052ac2/pyproject.toml#L53-L55

In effect, Windows CI jobs would be skipping the "slow" tests!


We now use `pytest --override-ini="addopts=-v" test` for Windows, just like for all other platforms (Linux etc.), which overrides the pytest options and disable the "slow skip logic".

test/test_encoders.py
```diff
@@ -936,7 +936,7 @@ def encode_to_tensor(frames):
                     pytest.mark.slow,
                     pytest.mark.skipif(
                         get_ffmpeg_major_version() == 4
-                        or (IS_WINDOWS and get_ffmpeg_major_version() in (6, 7)),
+                        or (IS_WINDOWS and get_ffmpeg_major_version() >= 6),
```
NicolasHug, line 939: We used to skip slow tests on Windows, now we don't, which is why the CI is exposing that this slow test is actually *also* failing on FFmpeg 8.

PR #1189, opened 2026-01-23
Ci scripts crazy refac

NicolasHug:


NicolasHug:
I'll drop this, happy to reconsider if we want to factorize it further

PR #1199, opened 2026-01-26
Update performance tips tutorials with decoder native transforms

mollyxu:
Add guidance on using decoder native transforms in performance tips tutorial

### DIFF
--- a/examples/decoding/performance_tips.py
+++ b/examples/decoding/performance_tips.py
@@ -197,22 +197,21 @@ to increase performance.
 # This can lead to significant memory savings, especially when decoding
 # high-resolution videos that will be resized to smaller dimensions.
 #
-# :class:`~torchcodec.decoders.VideoDecoder` accepts both TorchVision
-# :class:`~torchvision.transforms.v2.Transform` objects and TorchCodec
-# :class:`~torchcodec.transforms.DecoderTransform` objects as transform
-# specifications. TorchVision transforms are accepted as a convenience;
-# TorchVision is **not required** to use decoder transforms.
+# :class:`~torchcodec.decoders.VideoDecoder` accepts both TorchCodec
+# :class:`~torchcodec.transforms.DecoderTransform` objects and TorchVision
+# :class:`~torchvision.transforms.v2.Transform` objects as transform
+# specifications. TorchVision is **not required** to use decoder transforms.
 #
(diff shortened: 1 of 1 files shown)


examples/decoding/performance_tips.py
```diff
@@ -187,14 +188,60 @@
...
+#
+# - If you are applying a transform pipeline that significantly reduces the
+#   dimensions of your input frames and memory efficiency matters.
+# - If you are using multiple FFmpeg threads, decoder transforms may be faster.
+#   Experiment with your setup to verify.
+#
+# **Caveat:**
+#
+# - If you are using a single FFmpeg thread, decoder transforms may be slower.
+#   Experiment with your setup to verify.
```
Dan-Flores: nit: This caveat is implicit in the bullet above. Maybe we can link to the full benchmark explanation in the transform tutorial here instead?

examples/decoding/performance_tips.py
```diff
@@ -187,14 +188,60 @@
...
+#
+# .. code-block:: python
+#
+#     from torchcodec.decoders import VideoDecoder
+#     from torchvision.transforms import v2
+#
+#     decoder = VideoDecoder(
+#         "file.mp4",
+#         transforms=[v2.Resize(size=(480, 640))]
+#     )
```
Dan-Flores, line 215: Maybe lets add an example using the `DecoderTransform` class here as well, since we mention TorchVision is not required.

Though I noticed the current transforms tutorial does not have any examples of `DecoderTransform`, so I am not certain on this point.

NicolasHug: I think it makes sense to use the `DecoderTransform` here. I think we'd rather have one single example instead of 2: the point of this tutorial is to be an entry point to other tutorials, so ideally we want to keep it short and to-the-point. The example with the TV transforms can be left-out for the more complete dedicated tutorial.

Related: I just opened https://github.com/meta-pytorch/torchcodec/pull/1204 to illustrate the use of TC transforms in the dedicated tutorial.

examples/decoding/performance_tips.py
mollyxu: Thanks for the feedback! I think since the note below links to the full benchmarks I'll just remove the caveat.

Issue #1159, opened 2026-01-08
std::to_string using system locale resulting in JSONDecodeError


spezialspezial:
### 🐛 Describe the bug

```
Traceback (most recent call last):
  File "/+DEV/conda/envs/vidvid/lib/python3.10/site-packages/torchcodec/decoders/_video_decoder.py", line 143, in __init__
    ) = _get_and_validate_stream_metadata(
  File "/+DEV/conda/envs/vidvid/lib/python3.10/site-packages/torchcodec/decoders/_video_decoder.py", line 397, in _get_and_validate_stream_metadata
    container_metadata = core.get_container_metadata(decoder)
  File "/+DEV/conda/envs/vidvid/lib/python3.10/site-packages/torchcodec/_core/_metadata.py", line 200, in get_container_metadata
    container_dict = json.loads(jmd)
  File "/+DEV/conda/envs/vidvid/lib/python3.10/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/+DEV/conda/envs/vidvid/lib/python3.10/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/+DEV/conda/envs/vidvid/lib/python3.10/json/decoder.py", line 353, in raw_decode
    obj, end = self.scan_once(s, idx)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 4 column 20 (char 75)
```

```
dbg::get_container_metadata:: {
"bestAudioStreamIndex": 1,
"bestVideoStreamIndex": 0,
"bitRate": 3645156,000000,
"durationSecondsFromHeader": 8627,104000,
"numStreams": 3
} <class 'str'>
```

std::to_string in [get_container_json_metadata](https://github.com/meta-pytorch/torchcodec/blob/f91d24aaccb6cc0c87f5eeef1a00262e023f9b20/src/torchcodec/_core/custom_ops.cpp#L866)
seems to use the current system locale's LC_NUMERIC setting resulting in a JSONDecodeError for some configurations (where comma is the decimale separator for example). 

maybe consider (locally) using something like:
```
#include <clocale>
std::setlocale(LC_NUMERIC, "C")
```

### Versions

Linux, torchcodec 0.9.1

Dan-Flores:
Thanks for the issue @spezialspezial!

I was able to reproduce this, I'll take a look.
```
python -c "
  import locale
[...]

PR #1200, opened 2026-01-26
Avoid JSONDecodeError by using classic C locale

Dan-Flores:
This PR implements `doubleToString` to always use "." as the decimal separator so that the metadata JSON can be decoded, which should resolve https://github.com/meta-pytorch/torchcodec/issues/1159..
 
For the double metadata fields defined in [Metadata.h](https://github.com/meta-pytorch/torchcodec/blob/e93fc3143238ce83507cbe4c7612fe10d3d21329/src/torchcodec/_core/Metadata.h#L4), `std::to_string` is replaced by `doubleToString`.

SIdenote: The transient [JSON Decode error](https://github.com/meta-pytorch/torchcodec/actions/runs/21380808895/job/61548988932) appeared on this PR, so we can be certain its unrelated.

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -175,6 +175,7 @@ std::map<std::string, std::string> unflattenExtraOptions(
 // Convert double to string using the classic "C" locale.
 // This prevents JSON parsing errors in locales that use ',' as decimal
 // separator.
+// See https://github.com/meta-pytorch/torchcodec/issues/1159
 std::string doubleToString(double value) {
   std::ostringstream oss;
   oss.imbue(std::locale::classic());
--- a/test/test_metadata.py
+++ b/test/test_metadata.py
@@ -5,7 +5,6 @@
 # LICENSE file in the root directory of this source tree.
 
 import functools
-import locale
 from fractions import Fraction
 
 import pytest
@@ -21,19 +20,6 @@ from torchcodec._core import (
 from torchcodec.decoders import AudioDecoder, VideoDecoder
 
 from .utils import get_ffmpeg_major_version, NASA_AUDIO_MP3, NASA_VIDEO
-
-
-def _is_locale_available(locale_name):
-    # Check if a locale is available by trying to set it.
-    # The original locale is always restored.
-    saved = locale.getlocale(locale.LC_NUMERIC)
-    try:
-        locale.setlocale(locale.LC_NUMERIC, locale_name)
-        return True
-    except locale.Error:
-        return False
-    finally:
-        locale.setlocale(locale.LC_NUMERIC, saved)
 
 
 # TODO: Expected values in these tests should be based on the assets's
@@ -210,21 +196,3 @@ def test_repr():
   sample_format: fltp
 """
     )
-
-
-@pytest.mark.skipif(
-    not _is_locale_available("fr_FR.UTF-8"),
-    reason="Locale that uses comma as decimal separator not available",
-)
-def test_metadata_with_comma_decimal_locale():
-    # Test that metadata parsing works when comma is used as decimal separator.
-    # Regression test for https://github.com/meta-pytorch/torchcodec/issues/1159
-    saved_locale = locale.getlocale(locale.LC_NUMERIC)
-    try:
-        locale.setlocale(locale.LC_NUMERIC, "fr_FR.UTF-8")
-        decoder = VideoDecoder(NASA_VIDEO.path)
-        metadata = decoder.metadata
-        assert metadata.average_fps == pytest.approx(29.97, abs=0.001)
-        assert metadata.duration_seconds == pytest.approx(13.013, abs=0.001)
-    finally:
-        locale.setlocale(locale.LC_NUMERIC, saved_locale)


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -170,6 +172,17 @@ std::map<std::string, std::string> unflattenExtraOptions(
...
+// Convert double to string using the classic "C" locale.
+// This prevents JSON parsing errors in locales that use ',' as decimal
+// separator.
+std::string doubleToString(double value) {
+  std::ostringstream oss;
+  oss.imbue(std::locale::classic());
+  // Set precision to match std::to_string
+  oss << std::fixed << std::setprecision(6) << value;
+  return oss.str();
+}
```
Dan-Flores, line 185: I considered using [std::to_chars](https://en.cppreference.com/w/cpp/utility/to_chars.html) here, but I found it less intuitive than the std::ostringstream approach.

test/test_metadata.py
```diff
@@ -196,3 +210,21 @@ def test_repr():
   sample_format: fltp
 """
     )
+
+
+@pytest.mark.skipif(
+    not _is_locale_available("fr_FR.UTF-8"),
+    reason="Locale that uses comma as decimal separator not available",
+)
```
Dan-Flores: To follow the `pytest.mark.skipif` pattern, I had to define a function to check if switching locale's is available. I defined `_is_locale_available` in this file since it probably will not be used elsewhere.

src/torchcodec/_core/custom_ops.cpp
NicolasHug: Wow.

For my own ref: https://claude.ai/share/064f2168-e417-4cdf-bcd4-0c42be71d072

Curious about the precision though. Is that possibly related to the internal bug we mentioned [before](https://fburl.com/diff/eev31408)? CC @mollyxu as I think you're working on that as a BE item

test/test_metadata.py
```diff
@@ -22,6 +23,19 @@
...
+    # Check if a locale is available by trying to set it.
+    # The original locale is always restored.
+    saved = locale.getlocale(locale.LC_NUMERIC)
+    try:
+        locale.setlocale(locale.LC_NUMERIC, locale_name)
+        return True
+    except locale.Error:
+        return False
+    finally:
+        locale.setlocale(locale.LC_NUMERIC, saved)
```
NicolasHug: If we're going to keep this, move it right next to where it's used

test/test_metadata.py
```diff
@@ -196,3 +210,21 @@ def test_repr():
...
+
+@pytest.mark.skipif(
+    not _is_locale_available("fr_FR.UTF-8"),
+    reason="Locale that uses comma as decimal separator not available",
+)
+def test_metadata_with_comma_decimal_locale():
+    # Test that metadata parsing works when comma is used as decimal separator.
+    # Regression test for https://github.com/meta-pytorch/torchcodec/issues/1159
+    saved_locale = locale.getlocale(locale.LC_NUMERIC)
+    try:
```
NicolasHug: If we keep this, let's define a context manager instead of the try/finally. Hopefully, we can use that context manager in `_is_locale_available` as well.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -170,6 +172,17 @@ std::map<std::string, std::string> unflattenExtraOptions(
   return optionsMap;
 }
 
+// Convert double to string using the classic "C" locale.
+// This prevents JSON parsing errors in locales that use ',' as decimal
+// separator.
```
NicolasHug, line 177: Maybe mention https://github.com/meta-pytorch/torchcodec/issues/1159 in the comment (you also did it on the test, which is great!)

test/test_metadata.py
NicolasHug: Potentially controversial suggestion: let's not have a test for this.

This test is even more complex than the thing it's testing, and the probability of a regression is very, very low since you put the right comments in place. In other words, I feel like this test is more likely to fail because of its own complexity (false positive) rather than for a true-positive (the underlying fix being actually broken). 

Happy to be outvoted though :)

NicolasHug: Hm, I just checked a random CI job and this test is [being skipped](https://github.com/meta-pytorch/torchcodec/actions/runs/21380808896/job/61548101396?pr=1200):

```
test/test_metadata.py::test_metadata_with_comma_decimal_locale SKIPPED   [ 86%]
```

Makes me err more on the "let's remove this" side. If it's not being run on the CI, it's mainly dead code and we liekly won't be catching regressions.

Dan-Flores: Removing the test sounds fine to me, I agree the test is possibly more complicated than the fix. 

Also, we are still relying on the metadata tests (such as `test_repr`) to ensure that the output values are correct.

src/torchcodec/_core/custom_ops.cpp
Dan-Flores: Since I am removing the test, I'll mention the issue here.

src/torchcodec/_core/custom_ops.cpp
mollyxu: I think the precision would be related to the BE item that I am working on. We would have to update this callsite with `fmt:to_string` depending on the order in which they are merged.

Dan-Flores:
Closing as https://github.com/meta-pytorch/torchcodec/pull/1203 will resolve the issue.

PR #1203, opened 2026-01-27
Switch to fmt::to_string for precision

mollyxu:
Preserve precision for double to string conversions by using `fmt::to_string()` instead of `string::to_string()`

Note: some tests are edited to match the behavior of `fmt::to_string()` instead of `string::to_string()`

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -52,7 +52,7 @@ function(make_torchcodec_sublibrary
         ${Python3_INCLUDE_DIRS}
     )
     # Use fmt library in header-only mode (bundled with PyTorch): the fmt
-    # library is bundled with PyTorch and the headers are available in 
+    # library is bundled with PyTorch and the headers are available in
     # TORCH_INSTALL_PREFIX/include/, which we add above. To be compatible with
     # torch, we want to rely on the "header-only" mode of these headers and hit
     # all the `#ifdef FMT_HEADER_ONLY` paths, so we define FMT_HEADER_ONLY here.```


src/torchcodec/_core/CMakeLists.txt
```diff
@@ -2,12 +2,21 @@ cmake_minimum_required(VERSION 3.18)
...
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON) # Needed for linking fmt into shared libraries
 
 set(PYBIND11_FINDPYTHON ON)
 find_package(pybind11 REQUIRED)
 find_package(Torch REQUIRED)
 find_package(Python3 ${PYTHON_VERSION} EXACT COMPONENTS Development)
 
+# Fetch fmt library for high-precision string formatting
+include(FetchContent)
```
mollyxu: Would `FetchContent` + setting `CMAKE_POSITION_INDEPENDENT_CODE` be the best approach for linking fmt? An alternative is we could just use `fmt::fmt-header-only`

Dan-Flores:
`fmt::to_string` is locale-independent, so this PR resolves the locale bug without requiring us to set precision manually. 
We can close https://github.com/meta-pytorch/torchcodec/pull/1200 in favor of this.

But to make the reproducing example pass, `"bit_rate"` needs to be set with `std::to_string` as it is a `double`. I believe all the other `double` metadata fields are handled:
https://github.com/meta-pytorch/torchcodec/blob/85d1dad2c387f4d844ac8a61c2e4bce2b4ecdb64/src/torchcodec/_core/custom_ops.cpp#L815
https://github.com/meta-pytorch/torchcodec/blob/85d1dad2c387f4d844ac8a61c2e4bce2b4ecdb64/src/torchcodec/_core/custom_ops.cpp#L876
https://github.com/meta-pytorch/torchcodec/blob/85d1dad2c387f4d844ac8a61c2e4bce2b4ecdb64/src/torchcodec/_core/custom_ops.cpp#L918
Please double check me on this!
```
python -c "
  import locale
  locale.setlocale(locale.LC_NUMERIC, 'de_DE.UTF-8')
  decoder = VideoDecoder('test/resources/nasa_13013.mp4')
  "
```

src/torchcodec/_core/CMakeLists.txt
Dan-Flores: It looks like [pytorch/pytorch](https://github.com/pytorch/pytorch/blob/main/cmake/Dependencies.cmake#L1601) uses `fmt-header-only`. 

Another concern is that we are hardcoding this URL and version of the `fmt` library, so we would have to handle updates to this third party library as [pytorch/pytorch](https://github.com/pytorch/pytorch/tree/main/third_party) does.

I'm not sure what the best way to handle this is, @NicolasHug any thoughts?

NicolasHug: Yeah we should be using whatever pytorch uses. I think we might be able to just rely on the headers that torch itself is shipping:

```
~/dev/torchcodec (rotation*) » find /home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/args.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/base.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/chrono.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/color.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/compile.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/core.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/format-inl.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/format.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/os.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/ostream.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/printf.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/ranges.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/std.h
/home/nicolashug/miniconda3/envs/codec/lib/python3.11/site-packages/torch/include/fmt/xchar.h
```

(sorry if this is a bit cryptic, I'll try to share more details tomorrow but that might unblock you for the time being)

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -49,6 +51,8 @@ function(make_torchcodec_sublibrary
         "${TORCH_INSTALL_PREFIX}/include"
         ${Python3_INCLUDE_DIRS}
     )
+    # Use fmt in header-only mode (bundled with PyTorch)
```
NicolasHug: Took me a bit to understand what this was doing and why we need it. Can I suggest something like this:


```suggestion
    # Use fmt library in header-only mode (bundled with PyTorch): the fmt
    # library is bundled with PyTorch and the headers are available in 
    # TORCH_INSTALL_PREFIX/include/, which we add above. To be compatible with
    # torch, we want to rely on the "header-only" mode of these headers and hit
    # all the `#ifdef FMT_HEADER_ONLY` paths, so we define FMT_HEADER_ONLY here.```

**Review by NicolasHug — APPROVED** [review]
Thanks a lot @mollyxu , and thanks @Dan-Flores for cross-checking with https://github.com/meta-pytorch/torchcodec/pull/1200 ! I agree we can merge this in favor of https://github.com/meta-pytorch/torchcodec/pull/1200 since it's quite a bit simpler.

LGTM but let's wait on @Dan-Flores to make a final pass.

-----

We have a fair amount of uses of `std::to_string` in the rest of the codebase. I think we should just use `fmt::to_string` everywhere? This could be a follow-up PR!

File: test/CMakeLists.txt [hunk 3]
```diff
@@ -45,6 +45,9 @@ target_include_directories(MetadataTest SYSTEM PRIVATE ${TORCH_INCLUDE_DIRS})
 target_include_directories(MetadataTest SYSTEM PRIVATE ${libav_include_dirs})
 target_include_directories(MetadataTest PRIVATE ../)
 
+# Use fmt in header-only mode (bundled with PyTorch)
+target_compile_definitions(MetadataTest PRIVATE FMT_HEADER_ONLY)
```
Dan-Flores: I don't think we need to add this header to `MetadataTest`. The only file we include include `fmt/format.h` in is `custom_ops.cpp`, which is covered by the changes to `_core/CMakeLists.txt`

PR #1204, opened 2026-01-28
Small update to transforms tutorial

NicolasHug:
Small update that moves one section higher-up, and uses the TC transforms (previously, only TV transforms were used). Now, the tutorial illustrates both uses.

### DIFF
--- a/examples/decoding/transforms.py
+++ b/examples/decoding/transforms.py
@@ -129,22 +129,6 @@ resized_during = resize_decoder[5]
 plot(resized_during, title="Resized to 480x640 during decoding")
 
 # %%
-# TorchCodec's relationship to TorchVision transforms
-# -----------------------------------------------------
-# Notably, in our examples we are passing in TorchVision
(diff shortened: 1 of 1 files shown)


examples/decoding/transforms.py
```diff
@@ -128,6 +128,18 @@ def plot(frames: torch.Tensor, title : str | None = None):
 
 plot(resized_during, title="Resized to 480x640 during decoding")
 
+# %%
```
NicolasHug, line 131: I just copy/pasted this section and moved it here. Previously, it was lower down, right after a section that was comparing TC transforms and TV transforms when used as decoder-native transforms. I thought it could potentially be confusing: this section started with "the two frames are not identical", but in TC, passing a TV transform or a TV transform provides the exact same result, since the same filtergraph implementation is used under the hood.

The fact that the 2 frames are not identical is because one is transformed within the decoder, while the other one is transformed after the decoder.

PR #1207, opened 2026-01-28
Add doc-preview step

Dan-Flores:
This PR enables doc previews in the pytorch-bot comments, largely inspired by the [doc-preview step from torchrec](https://github.com/meta-pytorch/torchrec/blob/fb70bb38f35fd7bc0c4d30bc84bb37db59ed825a/.github/workflows/docs.yml#L104).

The comment linked by pytorch-bot now shows a preview of the docs built in this PR!

### DIFF
--- a/.github/workflows/linux_cuda_wheel.yaml
+++ b/.github/workflows/linux_cuda_wheel.yaml
@@ -223,3 +223,57 @@ jobs:
           if-no-files-found: error
           path: docs
           s3-prefix: meta-pytorch/torchcodec/${{ github.event.pull_request.number }}
+
+  upload-docs:
+    # This job uploads built docs:
+    # - to the `main` folder of the gh-pages branch (https://meta-pytorch.org/torchcodec/main) on every commit to the `main` branch
+    # - to the (e.g.) `0.10` folder in the gh-pages branch whenever a corresponding tag is pushed, like `v0.10.0` (https://meta-pytorch.org/torchcodec/0.10).
+
+    needs: build-docs
+    if: github.repository == 'meta-pytorch/torchcodec' && github.event_name == 'push' &&
+        ((github.ref_type == 'branch' && github.ref_name == 'main') || github.ref_type == 'tag')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out gh-pages branch
+        uses: actions/checkout@v6
+        with:
+          ref: gh-pages
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: Built-Docs
+          path: docs-artifact/
+
+      - name: Update docs and push
+        run: |
+          set -euo pipefail
+
+          REF_TYPE=${{ github.ref_type }}
+          REF_NAME=${{ github.ref_name }}
+
+          if [[ "${REF_TYPE}" == branch ]]; then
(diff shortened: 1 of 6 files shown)


.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -199,3 +199,26 @@ jobs:
...
+    if: github.repository == 'meta-pytorch/torchcodec' && github.event_name == 'pull_request'
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: Built-Docs
+          path: docs
+
+      - name: Add no-index tag
+        run: |
+          find docs -name "*.html" -print0 | xargs -0 sed -i '/<head>/a \ \ <meta name="robots" content="noindex">';
```
Dan-Flores, line 216: Similar jobs add the `no-index` tag so that search engines do not index these ephemeral docs.

.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -199,3 +199,26 @@ jobs:
         with:
           name: Built-Docs
           path: docs/build/html/
+
+  doc-preview:
+    runs-on: [self-hosted, linux.2xlarge]
```
Dan-Flores, line 204: Similar jobs use `self-hosted` to force this job to run on pytorch's self-hosted runners, which enables us to upload to S3.

.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -199,3 +199,26 @@ jobs:
...
+        with:
+          name: Built-Docs
+          path: docs
+
+      - name: Add no-index tag
+        run: |
+          find docs -name "*.html" -print0 | xargs -0 sed -i '/<head>/a \ \ <meta name="robots" content="noindex">';
+
+      - name: Upload docs preview
+        uses: seemethere/upload-artifact-s3@v5
```
Dan-Flores, line 219: Rather than check out the test-infra repo to use the [`upload-artifact-s3` step](https://github.com/pytorch/test-infra/blob/6232f6188829512f26f48ab065ff7622f98a537d/.github/workflows/linux_job_v2.yml#L358), we just reuse this step from the torchrec workflow, which seems to accomplish the same thing.

.github/workflows/linux_cuda_wheel.yaml
mollyxu: Out of curiosity, how do we usually determine the best machine size for tasks?

mollyxu (approved):
Thanks for working on this! It's super useful to be able to see a preview of docs on our PRs. I left a question but this LGTM!

.github/workflows/linux_cuda_wheel.yaml
NicolasHug: Nit: add a comment in the file to explain exactly this :)

.github/workflows/linux_cuda_wheel.yaml
Dan-Flores: One way (certainly not the best) would be trial and error. If this machine OOMs, we would know to increase the size.
In this case, this is the machine used by the workflow I took inspiration from, so we can be somewhat confident it will work for our docs.

PR #1209, opened 2026-01-29
Fix frame indexing for videos with non-zero start time in approximate seek mode

eugenn:
## Problem Description

In approximate seek mode, the frame index calculations in `secondsToIndexLowerBound`, `secondsToIndexUpperBound`, and `getPts` assume that videos always start at time 0. This assumption is incorrect for videos that have a non-zero start time (e.g., trimmed videos, videos with start offset in the container header).

**Current behavior:**
- When converting seconds to frame index, the code directly multiplies `seconds * fps`, ignoring the video's actual start time
- When converting frame index to PTS, it doesn't account for the stream's beginning offset
- `getEndStreamSeconds()` incorrectly returns just the duration instead of `beginStreamSeconds + duration`

**Impact:**
- Frame indices are miscalculated for videos with non-zero start timestamps
- Seeking to a specific second or frame index returns wrong frames

**Fix:**
1. In `secondsToIndexLowerBound` and `secondsToIndexUpperBound`: subtract `beginStreamSeconds` from the requested seconds to get the relative time before calculating frame index
2. In `getPts`: add `beginStreamSeconds` when converting frame index back to PTS
3. In `getEndStreamSeconds`: properly calculate end time as `beginStreamSeconds + duration`
4. In `getBeginStreamSeconds`: add fallback to `beginStreamSecondsFromHeader` when content-derived value is not available

---

### 1. Create a test video with non-zero start time

```bash
ffmpeg -ss 10 -to 20 -copyts -avoid_negative_ts disabled -i https://download.samplelib.com/mp4/sample-30s.mp4 -map 0 -c copy -movflags +faststart test.mp4 -y
```

This command extracts a segment from 10s to 20s while preserving original timestamps.

### 2. Run the following Python code

```python
import torchcodec
print(torchcodec.__version__)

VIDEO_URL = "test.mp4"

decoder = VideoDecoder(VIDEO_URL, seek_mode="approximate")

t0 = decoder.metadata.begin_stream_seconds
print("begin_stream_seconds: ", t0)

frames = decoder.get_frames_played_in_range(5.0, 10.0)
print("\nRange 5-10s (played):")
print(f"  Num frames: {frames.pts_seconds.numel()}")
if frames.pts_seconds.numel() > 0:
    print(f"  First PTS: {frames.pts_seconds[0].item():.6f}")
    print(f"  Last PTS: {frames.pts_seconds[-1].item():.6f}")
```

### 3. Actual output (incorrect)

```
0.10.0
begin_stream_seconds:  0.0

Range 5-10s (played):
  Num frames: 150
  First PTS: 8.333008
  Last PTS: 13.299674
```

### 4. Expected behavior

The video's actual start time can be verified with ffmpeg:

```bash
ffmpeg -ss 5 -to 10 -copyts -avoid_negative_ts disabled -i test.mp4 -map 0 -c copy out.mp4
# Output: Duration: 00:00:11.80, start: 8.266009, bitrate: 5882 kb/s
```

Therefore, `decoder.metadata.begin_stream_seconds` should return `8.266009`, not `0.0`.

---

## Expected behavior after fix

```python
import torchcodec
print(torchcodec.__version__)

VIDEO_URL = "test.mp4"

decoder = VideoDecoder(VIDEO_URL, seek_mode="approximate")

t0 = decoder.metadata.begin_stream_seconds
print("begin_stream_seconds: ", t0)

frames = decoder.get_frames_played_in_range(t0 + 5.0, t0 + 10.0)
print("\nRange 5-10s (played):")
print(f"  Num frames: {frames.pts_seconds.numel()}")
if frames.pts_seconds.numel() > 0:
    print(f"  First PTS: {frames.pts_seconds[0].item():.6f}")
    print(f"  Last PTS: {frames.pts_seconds[-1].item():.6f}")
```

**Expected output:**

```
begin_stream_seconds:  8.333008

Range 5-10s (played):
  Num frames: 151
  First PTS: 13.333008
  Last PTS: 18.333008
```


src/torchcodec/_core/Metadata.cpp
```diff
@@ -69,7 +72,13 @@ std::optional<double> StreamMetadata::getEndStreamSeconds(
...
         return endStreamPtsSecondsFromContent.value();
       }
-      return getDurationSeconds(seekMode);
+      {
+        auto dur = getDurationSeconds(seekMode);
+        if (dur.has_value()) {
+          return getBeginStreamSeconds(seekMode) + dur.value();
+        }
+      }
+      return std::nullopt;
```
mollyxu: nit: for consistency, it would be great if you could move the braces to the case level
```
case SeekMode::approximate: {
  if (endStreamPtsSecondsFromContent.has_value()) {
    return endStreamPtsSecondsFromContent.value();
  }
  auto dur = getDurationSeconds(seekMode);
  if (dur.has_value()) {
    return getBeginStreamSeconds(seekMode) + dur.value();
  }
  return std::nullopt;
}
```

mollyxu (commented):
Hi @eugenn, thanks for submitting this fix! I had a small nit on the formatting but the changes LGTM. It would be great if you could also add the reproducing video to this PR as a non-regression test for future PRs. Thanks again for working on this!

src/torchcodec/_core/Metadata.cpp
```diff
@@ -50,6 +50,9 @@ double StreamMetadata::getBeginStreamSeconds(SeekMode seekMode) const {
       if (beginStreamPtsSecondsFromContent.has_value()) {
         return beginStreamPtsSecondsFromContent.value();
       }
```
NicolasHug, line 52: QQ @mollyxu on the lines above (not strictly related to this PR): do you remember why we added some logic with `beginStreamPtsSecondsFromContent` for `case SeekMode::approximate:`?
My understanding is that `beginStreamPtsSecondsFromContent` would always be nullopt in approximate mode, because we don't do any scan, but I remember there are quite a few subtleties when it comes to metadata. Maybe this is required?

src/torchcodec/_core/Metadata.cpp
eugenn: done

src/torchcodec/_core/Metadata.cpp
eugenn: @mollyxu of course. In step 1 above there is a link to the original video:
https://download.samplelib.com/mp4/sample-30s.mp4

Then, using the ffmpeg command, we generate a test video called test.mp4, which can be used to reproduce and detect the discrepancy between ffmpeg and torchcodec behavior.

mollyxu: @NicolasHug Thanks for bringing attention to this! I revisited the metadata refactor PR and saw that `beginStreamPtsSecondsFromContent` for approximate mode seems to be a mismapping of the previous metadata logic in python. 
```        
if self.begin_stream_seconds_from_content is None:
            return 0
else:
            return self.begin_stream_seconds_from_content
```

You are right in that `beginStreamPtsSecondsFromContent` would always be nullopt in approximate mode, because we don't do any scan. I believe `beginStreamPtsSecondsFromContent` in `getBeginStreamSeconds()` and `endStreamPtsSecondsFromContent`  in `getEndStreamSeconds()` should be removed in addition to the PR's changes.

I can push a change to this PR to resolve this discrepancy.

NicolasHug: Thanks for following-up @mollyxu , let's do that in a separate PR then, it's better to keep concerns separate

mollyxu:
@eugenn Thanks for addressing the comments so quickly! 

To clarify the comment on adding a video, I was wondering if you would be interested in adding a test in `test_decoders.py` that checks if frame retrieval methods return frames with expected PTS values, checking in a small video with a non-zero start time (similar to the ones we have in [test/utils.py ](https://github.com/meta-pytorch/torchcodec/blob/main/test/utils.py#L473)). No worries if not, we can handle it on our side.

eugenn:
> @eugenn Thanks for addressing the comments so quickly!
> 
> To clarify the comment on adding a video, I was wondering if you would be interested in adding a test in `test_decoders.py` that checks if frame retrieval methods return frames with expected PTS values, checking in a small video with a non-zero start time (similar to the ones we have in [test/utils.py ](https://github.com/meta-pytorch/torchcodec/blob/main/test/utils.py#L473)). No worries if not, we can handle it on our side.

Ok, I’ve added a test video fragment along with tests that check frame retrieval and PTS values. Hopefully everything works as expected now.

mollyxu:
@eugenn Thank you so much for adding the video and the test! I pushed a small commit to shorten the video. Thanks again for contributing!

Issue #100, opened 2024-07-22
More robust FFmpeg version getter


NicolasHug:
`_core.get_ffmpeg_library_versions()` returns a dict of strings.  As discussed in D58974580, versions are better exposed as namedtuples, like so:

```py
[nav] In [25]: _Version = namedtuple("_Version", ("major", "minor", "micro"))
[ins] In [26]: version = _Version(1, 2, 3)
[ins] In [27]: version
Out[27]: _Version(major=1, minor=2, micro=3)
[ins] In [28]: version < (1, 2, 4)
Out[28]: True
[ins] In [29]: version < (1, 2, 3, 4)
Out[29]: True
[ins] In [30]: version.major
Out[30]: 1
```

This is the same kind of struct that sys.version_info returns (we can't use their base class, but that doesn't matter).

-----

Also, we're relying on `av_version_info()` to get the major FFmpeg version, but that's [not recommended](https://ffmpeg.org/doxygen/4.2/group__lavu__ver.html#gabaea9f86706819534baceeac7ea5818a) :

> This string has no fixed format and can change any time. It should never be parsed by code.

NicolasHug:
For the FFmpeg major version, we should just use https://github.com/meta-pytorch/torchcodec/blob/ed44f0a53550d0cb71c81649f1990dbf1e6b4ad6/src/torchcodec/_core/ops.py#L109

PR #1211, opened 2026-01-29
Robust ffmpeg major version getter

Dan-Flores:
This PR updates `get_ffmpeg_major_version` to get the major version from `load_torchcodec_shared_libraries` to resolve https://github.com/meta-pytorch/torchcodec/issues/100.

This function is used many times throughout our testing ([code search](https://github.com/search?q=repo%3Ameta-pytorch%2Ftorchcodec%20get_ffmpeg_major_version&type=code)), so we can verify that the skips are still occuring by checking conditionally skipped tests. For example, this [encoder test](https://github.com/meta-pytorch/torchcodec/blob/736512985b430d88f388c0ea51d0d0e075d9e1ed/test/test_encoders.py#L228-L230) is still [skipped on FFmpeg 4](https://github.com/meta-pytorch/torchcodec/actions/runs/21492408138/job/61920113574).

Ideally, we would also migrate `get_ffmpeg_minor_version`, but since it only has 3 uses ([code search](https://github.com/search?q=repo%3Ameta-pytorch%2Ftorchcodec+get_ffmpeg_minor_version&type=code)) it can be lower priority.

### DIFF
--- a/test/test_encoders.py
+++ b/test/test_encoders.py
@@ -1060,7 +1060,6 @@ class TestVideoEncoder:
         pixel_format = encode_params["pixel_format"]
         crf = encode_params["crf"]
         preset = encode_params["preset"]
-        ffmpeg_version = ffmpeg_major_version
 
         if format in ("avi", "flv") and pixel_format == "yuv444p":
             pytest.skip(f"Default codec for {format} does not support {pixel_format}")
@@ -1148,7 +1147,7 @@ class TestVideoEncoder:
 
         # Only compare video metadata on ffmpeg versions >= 6, as older versions
         # are often missing metadata
-        if ffmpeg_version >= 6 and method == "to_file":
+        if ffmpeg_major_version >= 6 and method == "to_file":
             fields = [
                 "duration",
                 "duration_ts",
@@ -1387,10 +1386,9 @@ class TestVideoEncoder:
     def test_nvenc_against_ffmpeg_cli(
         self, tmp_path, method, format, codec, color_space, color_range
     ):
-        ffmpeg_version = ffmpeg_major_version
         # TODO-VideoEncoder: (P2) Investigate why FFmpeg 4 and 6 fail with non-default color space and range.
         # See https://github.com/meta-pytorch/torchcodec/issues/1140
-        if ffmpeg_version in (4, 6) and not (
+        if ffmpeg_major_version in (4, 6) and not (
             color_space == "bt470bg" and color_range == "tv"
         ):
             pytest.skip(
@@ -1503,7 +1501,7 @@ class TestVideoEncoder:
                 # Since this failure is rare, I suspect its a bug related to these
                 # older container formats on newer FFmpeg versions.
                 if not (
-                    ffmpeg_version in (7, 8)
+                    ffmpeg_major_version in (7, 8)
                     and color_range == "tv"
                     and color_space is None
                     and format in ("mov", "avi")


src/torchcodec/_core/ops.py
Dan-Flores, line 109: This is where `ffmpeg_major_version` is defined.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -1008,8 +1008,8 @@ std::string get_stream_json_metadata(
 // Returns version information about the various FFMPEG libraries that are
 // loaded in the program's address space.
 // TODO: ideally we'd have a more robust way of getting the ffmpeg version,
-// we're using av_version_info() which is not standardized and shouldn't be
-// parsed by code (which we do!). See
+// we're using av_version_info() to get the ffmpeg minor version, which is
+// not standardized and shouldn't be parsed by code (which we do!). See
```
mollyxu: To double check my understanding, we cannot get ffmpeg minor versions using the same method as this robust ffmpeg major version getter because the libraries are compiled against ffmpeg major versions only. As such, we would have to use `av_version_info()` to check what is installed on the users system at runtime

src/torchcodec/_core/ops.py
```diff
@@ -598,7 +598,13 @@ def scan_all_streams_to_update_metadata_abstract(decoder: torch.Tensor) -> None:
...
 def get_ffmpeg_library_versions():
     versions_json = _get_json_ffmpeg_library_versions()
-    return json.loads(versions_json)
+    versions = json.loads(versions_json)
+    # We reuse ffmpeg_major_version from load_torchcodec_shared_libraries() as
+    # it's more reliable than parsing ffmpeg_version from av_version_info().
+    # We still get the minor version from av_version_info() in _get_json_ffmpeg_library_versions().
+    # See https://github.com/pytorch/torchcodec/issues/100
+    versions["ffmpeg_major_version"] = ffmpeg_major_version
+    return versions
```
NicolasHug: I think this is a new key in the `versions` dict, previously the entries were like this:

```
In [2]: torchcodec._core.get_ffmpeg_library_versions()
Out[2]: 
{'libavfilter': [10, 4, 100],
 'libavutil': [59, 39, 100],
 'libavcodec': [61, 19, 100],
 'libavformat': [61, 7, 100],
 'ffmpeg_version': '7.1'}
```


I think we can do something more straightforward than adding a new key. I'd suggest not to add anything new to this dict and to *not* rely on the C++ `_get_json_ffmpeg_library_versions` to get the major version (see other comment below).

test/utils.py
```diff
@@ -91,24 +91,17 @@ def make_video_decoder(*args, **kwargs) -> tuple[VideoDecoder, str]:
     return dec, clean_device
 
 
-def _get_ffmpeg_version_string():
+def get_ffmpeg_major_version():
+    return get_ffmpeg_library_versions()["ffmpeg_major_version"]
```
NicolasHug: I think we can get rid of this `get_ffmpeg_major_version()` helper entirely now. We expose `torchcodec.ffmpeg_major_version`:

https://github.com/meta-pytorch/torchcodec/blob/a54cfaf5bbd1ab3436e8198fe26987a723e6415a/src/torchcodec/__init__.py#L27

so we can just use that in all the tests directly. This way, getting the major FFmpeg version never relies on `torchcodec._core.get_ffmpeg_library_versions()` and never calls the C++ `_get_json_ffmpeg_library_versions()`. Only getting the minor version requires going through C++.

I realize I didn't mention the existence of `torchcodec.ffmpeg_major_version` in https://github.com/meta-pytorch/torchcodec/issues/100, sorry for that!

src/torchcodec/_core/custom_ops.cpp
Dan-Flores: Yes, exactly!

test/utils.py
Dan-Flores: That's a lot cleaner, thanks, I'll update the PR.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -1009,8 +1009,8 @@ std::string get_stream_json_metadata(
...
```
NicolasHug: Just a nit, we might want to revert the comment to its original state because we also retrieve (e.g.) libavcodec versions ,as well as all other ffmpeg libraries here. It's not just the minor version. In fact even the major version number is available through this.

PR #1212, opened 2026-01-29
[wip] Create parallel EncodingDeviceInterface

Dan-Flores:


src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -19,8 +22,67 @@ PerGpuCache<NppStreamContext> g_cached_npp_ctxs(
     MAX_CUDA_GPUS,
     MAX_CONTEXTS_PER_GPU_IN_CACHE);
 
+PerGpuCache<AVBufferRef, Deleterp<AVBufferRef, void, av_buffer_unref>>
+    g_cached_hw_device_ctxs(MAX_CUDA_GPUS, MAX_CONTEXTS_PER_GPU_IN_CACHE);
+
```
Dan-Flores, line 27: This code is moved here from `CudaDeviceInterface.cpp` so that `EncodingCudaDeviceInterface.cpp` can reuse the context creation and caching.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -901,20 +892,14 @@ void VideoEncoder::encode() {
...
-      avFrame = std::move(cudaFrame);
-    } else {
-      avFrame = convertTensorToAVFrame(currFrame, i);
-    }
+    UniqueAVFrame avFrame = encodingInterface_->convertTensorToAVFrame(
+        currFrame, i, avCodecContext_.get());
+    TORCH_CHECK(
+        avFrame != nullptr,
+        "convertTensorToAVFrame failed for frame ",
+        i,
```
Dan-Flores, line 900: Encoding interface handles both CPU and CUDA, and removes repeated conditionals:
`if (frames_.device().is_cuda && deviceInterface_) { ... }`

src/torchcodec/_core/Encoder.cpp
```diff
@@ -927,72 +912,6 @@ void VideoEncoder::encode() {
       getFFMPEGErrorStringFromErrorCode(status));
 }
 
-UniqueAVFrame VideoEncoder::convertTensorToAVFrame(
```
Dan-Flores, line 930: CPU specific implementation is removed from Encoder.cpp.

src/torchcodec/_core/EncodingDeviceInterface.h
```diff
@@ -0,0 +1,95 @@
...
+struct EncodingDeviceInterfaceKey {
+  torch::DeviceType deviceType;
+
+  bool operator<(const EncodingDeviceInterfaceKey& other) const {
+    return deviceType < other.deviceType;
+  }
+
+  explicit EncodingDeviceInterfaceKey(torch::DeviceType type)
+      : deviceType(type) {}
+};
```
Dan-Flores, line 27: This is mostly copied over from DeviceInterface.h.

We could move these registration functions to a new header that is shared between `EncodingDeviceInterface` and `DeviceInterface`, but I'm not sure that's necessary. For now, I've added this parallel implementation using `Encoding...` types.

Alternatively, if we never expect to implement a `BetaCuda` variant for encoding, this can be simplified.

PR #1213, opened 2026-01-29
Combine encoding into DeviceInterface

Dan-Flores:
This PR merges encoding into the DeviceInterface pattern.
* The code within Encoder.cpp now always expects to have a `deviceInterface_` member variable defined, and uses it to call encoding functions.
* CUDA encoding functions were already present in CudaDeviceInterface.cpp
* CPU encoding functions are moved out of Encoder.cpp, into CpuDeviceInterface.cpp

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -766,17 +766,17 @@ void VideoEncoder::initializeEncoder(
   TORCH_CHECK(avCodecContext != nullptr, "Couldn't allocate codec context.");
   avCodecContext_.reset(avCodecContext);
 
-  // Store dimension order and input pixel format
+  // Store dimensions of input frames
   // TODO-VideoEncoder: (P2) Enable tensors in NHWC shape
   auto sizes = frames_.sizes();
-  inPixelFormat_ = AV_PIX_FMT_GBRP;
-  inHeight_ = static_cast<int>(sizes[2]);
-  inWidth_ = static_cast<int>(sizes[3]);
+  int inHeight = static_cast<int>(sizes[2]);
+  int inWidth = static_cast<int>(sizes[3]);
 
-  // Use specified dimensions or input dimensions
+  // Always use input dimensions as output dimensions
   // TODO-VideoEncoder: (P2) Allow height and width to be set
-  outWidth_ = inWidth_;
-  outHeight_ = inHeight_;
+  int outWidth = inWidth;
+  int outHeight = inHeight;
+  AVPixelFormat outPixelFormat = AV_PIX_FMT_NONE;
 
   if (videoStreamOptions.pixelFormat.has_value()) {
     // TODO-VideoEncoder: (P2) Enable pixel formats to be set by user on GPU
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -506,4 +506,78 @@ std::string CpuDeviceInterface::getDetails() {
   return std::string("CPU Device Interface.");
 }
 
+UniqueAVFrame CpuDeviceInterface::convertTensorToAVFrameForEncoding(
```
Dan-Flores, line 509: CPU specific encoding implementation is removed from Encoder.cpp and added here.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -901,20 +888,14 @@ void VideoEncoder::encode() {
...
-          "convertCUDATensorToAVFrameForEncoding failed for frame ",
-          i,
-          " on device: ",
-          frames_.device());
-      avFrame = std::move(cudaFrame);
-    } else {
-      avFrame = convertTensorToAVFrame(currFrame, i);
-    }
+    UniqueAVFrame avFrame = deviceInterface_->convertTensorToAVFrameForEncoding(
+        currFrame, i, avCodecContext_.get());
```
Dan-Flores, line 895: Device interface handles both CPU and CUDA, and removes repeated conditionals:
`if (frames_.device().is_cuda && deviceInterface_) { ... }`

src/torchcodec/_core/Encoder.h
```diff
@@ -160,16 +160,12 @@ class VideoEncoder {
...
-  UniqueAVFrame convertTensorToAVFrame(
-      const torch::Tensor& frame,
-      int frameIndex);
   void encodeFrame(AutoAVPacket& autoAVPacket, const UniqueAVFrame& avFrame);
   void flushBuffers();
 
   UniqueEncodingAVFormatContext avFormatContext_;
   UniqueAVCodecContext avCodecContext_;
   AVStream* avStream_ = nullptr;
-  UniqueSwsContext swsContext_;
```
Dan-Flores, line 172: This member variable is deleted in favor of the `swsContext_` in `CpuDeviceInterface.h`.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -142,11 +142,7 @@ class DeviceInterface {
   // Pixel format used for encoding on CUDA devices
   static constexpr AVPixelFormat CUDA_ENCODING_PIXEL_FORMAT = AV_PIX_FMT_NV12;
 
-  // Function used for video encoding, only implemented in CudaDeviceInterface.
-  // It is here to isolate CUDA dependencies from CPU builds
-  // TODO Video-Encoder: Reconsider using video encoding functions in device
-  // interface
-  virtual UniqueAVFrame convertCUDATensorToAVFrameForEncoding(
```
Dan-Flores, line 149: We replace `convertCUDATensor...` with `convertTensor...` to integrate into the device interface pattern.

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -156,9 +152,7 @@ class DeviceInterface {
   // Function used for video encoding, only implemented in CudaDeviceInterface.
   // It is here to isolate CUDA dependencies from CPU builds
   virtual void setupHardwareFrameContextForEncoding(
-      [[maybe_unused]] AVCodecContext* codecContext) {
-    TORCH_CHECK(false);
-  }
+      [[maybe_unused]] AVCodecContext* codecContext) {}
```
NicolasHug: I think we'll want to keep it with `TORCH_CHECK(false);`  instead of having a default void implementation? The reason is that there is still the beta cuda interface, for which this method is essentially invalid.

Dan-Flores: I agree with this, I think I was emulating the way `registerHardwareDeviceWithCodec` was called unconditionally as a [no-op](https://github.com/meta-pytorch/torchcodec/blob/377c6383d60660a2e917da2e3a6a5c3c39f11fdf/src/torchcodec/_core/DeviceInterface.h#L88-L89) in SingleStreamDecoder:
https://github.com/meta-pytorch/torchcodec/blob/377c6383d60660a2e917da2e3a6a5c3c39f11fdf/src/torchcodec/_core/SingleStreamDecoder.cpp#L477-L481

Perhaps `registerHardwareDeviceWithCodec` should also raise an error, since its unimplemented in the CPU and BetaCUDA interface.

src/torchcodec/_core/CpuDeviceInterface.cpp
NicolasHug: just checking: this is 100% copy/pasted, correct?

Dan-Flores: Yes, the main difference is that `Encoder.h`'s member variables are replaced by the local variables on lines 513-518.
This made me realize we can remove those member variables from `Encoder.h` entirely, since they are unused.

PR #1214, opened 2026-01-29
[wip] Move CUDA encoding into GpuEncoder

Dan-Flores:


src/torchcodec/_core/Encoder.h
```diff
@@ -3,10 +3,13 @@
 #include <map>
 #include <string>
 #include "AVIOContextHolder.h"
-#include "DeviceInterface.h"
 #include "FFMPEGCommon.h"
 #include "StreamOptions.h"
 
+#ifdef ENABLE_CUDA
+#include "GpuEncoder.h"
+#endif
```
Dan-Flores, line 11: As I was reviewing this, I recall this is why the initial `GpuEncoder` approach was abandoned. The `DeviceInterface` abstraction cleanly separates CUDA headers without using `#ifdef`

Issue #1135, opened 2025-12-15
Refactor sw_scale into new class


mollyxu:
Follow-up to todo mentioned in #1130 

We should probably refactor the sws logic into its own class, similar to what we did for filtergraph.

* Encapsulate the double swscale path (color conversion + resize)
* Handle context caching internally (like FilterGraph is created once and reused)
* Provide a clean convert() API similar to FilterGraph::convert()

Issue #925, opened 2025-10-03
Cleanup the naming of filtergraph and swscale constructs in `CpuDeviceInterface`


scotts:
From the code itself: https://github.com/meta-pytorch/torchcodec/blob/6d72f115c4fbbcb8ed62cac43f5f1cfe2bd5f692/src/torchcodec/_core/CpuDeviceInterface.h#L98-L105

Part of the problem here is that "context" is an overused term. We've used it to mean "the values for the current version of filtergraph or swscale that let us know if we can reuse it," but context is also commonly used in FFmpeg for the names of objects that control access to a thing.

Because "context" is so pervasive in FFmpeg type names, we should probably not use it our type names. This does mean we'll probably also want to rename [`FiltersContext`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/FilterGraph.h#L14) as well as [`SwsFrameContext`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/CpuDeviceInterface.h#L52)

mollyxu:
closing with #1215

NicolasHug:
@mollyxu any reason this was re-opened? Same with https://github.com/meta-pytorch/torchcodec/issues/1135

mollyxu:
@NicolasHug I accidentally closed it before the PR was merged. Closing now!

PR #1215, opened 2026-02-02
Refactor swscale into new class + rename filtergraph and swscale constructs

mollyxu:
Refactor `sw_scale` into new class similar to what we did for `filtergraph`. (#1135)
  - Encapsulate the double `swscale` path (color conversion + resize) in a new `SwScale` class with a clean `convert()` API
  - Merge `SwScaleContext` and `SwsFrameContext` into a single `SwsConfig` struct (colorspace is now part of the config rather than a separate parameter)
  - Add `encodingSwsContext_` for the encoding path, separating it from the decoding `swscale` which is now managed by `SwScale`

Rename `filtergraph` and `swscale` constructs in `CpuDeviceInterface` (#925)

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -4,7 +4,6 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include <c10/cuda/CUDAStream.h>
 #include <torch/types.h>
 #include <map>
 #include <mutex>
@@ -302,6 +301,8 @@ void BetaCudaDeviceInterface::initialize(
     const AVStream* avStream,
     const UniqueDecodingAVFormatContext& avFormatCtx,
     [[maybe_unused]] const SharedAVCodecContext& codecContext) {
+  STD_TORCH_CHECK(avStream != nullptr, "AVStream cannot be null");
+  rotation_ = rotationFromDegrees(getRotationFromStream(avStream));
   if (!nvcuvidAvailable_ || !nativeNVDECSupport(device_, codecContext)) {
     cpuFallback_ = createDeviceInterface(kStableCPU);
     STD_TORCH_CHECK(
@@ -315,7 +316,6 @@ void BetaCudaDeviceInterface::initialize(
     return;
   }
 
-  STD_TORCH_CHECK(avStream != nullptr, "AVStream cannot be null");
   timeBase_ = avStream->time_base;
   frameRateAvgFromFFmpeg_ = avStream->r_frame_rate;
 
@@ -583,8 +583,8 @@ int BetaCudaDeviceInterface::receiveFrame(UniqueAVFrame& avFrame) {
   // the NPP stream before any color conversion.
   // Re types: we get a cudaStream_t from PyTorch but it's interchangeable with
   // CUstream
-  procParams.output_stream = reinterpret_cast<CUstream>(
-      at::cuda::getCurrentCUDAStream(device_.index()).stream());
+  procParams.output_stream =
+      reinterpret_cast<CUstream>(getCurrentCudaStream(device_.index()));
 
   CUdeviceptr framePtr = 0;
   unsigned int pitch = 0;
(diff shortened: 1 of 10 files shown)


src/torchcodec/_core/SwScale.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+  // Create color conversion context (input format -> RGB24).
+  // When resizing is needed, color conversion outputs at input resolution.
+  // When no resize is needed, color conversion outputs at output resolution.
+  SwsFrameContext colorConversionFrameContext(
+      context_.inputWidth,
+      context_.inputHeight,
+      context_.inputFormat,
+      needsResize_ ? context_.inputWidth : context_.outputWidth,
+      needsResize_ ? context_.inputHeight : context_.outputHeight);
```
Dan-Flores: Since this is the color conversion step, I think we should always use the input dimensions instead of these ternary statements. 

I think this is always happening though:
* When `needsResize_` is true, we correctly use the input dimensions.
* When `needsResize_` is false, we use the output dimensions, which is also correct because both input and output dimensions are equal.

src/torchcodec/_core/SwScale.cpp
```diff
@@ -0,0 +1,136 @@
...
+        /*swsFlags=*/swsFlags_);
+  }
+}
+
+int SwScale::convert(
+    const UniqueAVFrame& avFrame,
+    torch::Tensor& outputTensor) {
+  // When no resize is needed, we do color conversion directly into the output
+  // tensor. When resize is needed, we first convert to an intermediate tensor
+  // at the input resolution, then resize into the output tensor.
```
Dan-Flores: The comments in `SwScale.h` are good, let's also retain the color space details from this comment in this documentation:
https://github.com/meta-pytorch/torchcodec/blob/377c6383d60660a2e917da2e3a6a5c3c39f11fdf/src/torchcodec/_core/CpuDeviceInterface.cpp#L287-L293

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -198,8 +198,23 @@ void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
...
+        avFrame->width,
+        avFrame->height,
+        avFrameFormat,
+        avFrame->colorspace,
+        outputDims.width,
+        outputDims.height);
+
+    if (!swScale_ || prevSwScaleContext_ != swScaleContext) {
+      swScale_ = std::make_unique<SwScale>(swScaleContext, swsFlags_);
+      prevSwScaleContext_ = swScaleContext;
```
NicolasHug: The `swScale_` field now stores a `SwScaleContext context_` field. My understanding is that it should actually correspond to `prevSwScaleContext_`, so maybe we can simplify this a bit by removing the `prevSwScaleContext_` field, and change the comparison to something like

```C++
 

 if (!swScale_ || swScale_->getContext() != swScaleContext) {                                                      
    swScale_ = std::make_unique<SwScale>(swScaleContext, swsFlags_);                                                   
  }                                                                                                                    
```

src/torchcodec/_core/CpuDeviceInterface.h
```diff
@@ -80,32 +76,26 @@ class CpuDeviceInterface : public DeviceInterface {
...
-  UniqueSwsContext swsContext_;
-  SwsFrameContext prevSwsFrameContext_;
-
-  // Cached swscale context for resizing in RGB24 space (used in double swscale
-  // path). Like the color conversion context above, we cache this to avoid
-  // recreating it for every frame.
-  UniqueSwsContext resizeSwsContext_;
-  SwsFrameContext prevResizeSwsFrameContext_;
+  std::unique_ptr<SwScale> swScale_;
+  SwScaleContext prevSwScaleContext_;
```
NicolasHug: See my suggestion above, I'm hoping we can remove this `SwScaleContext prevSwScaleContext_;` field.

src/torchcodec/_core/CpuDeviceInterface.h
```diff
@@ -80,32 +76,26 @@ class CpuDeviceInterface : public DeviceInterface {
...
+  // reuse them across frames. However, it is possible that subsequent frames
   // are different enough (change in dimensions) that we can't reuse the color
   // conversion object. We store the relevant frame context from the frame used
   // to create the object last time. We always compare the current frame's info
   // against the previous one to determine if we need to recreate the color
   // conversion object.
   //
   // TODO: The names of these fields is confusing, as the actual color
-  //       conversion object for Sws has "context" in the name,  and we use
+  //       conversion object for Sws has "context" in the name, and we use
```
NicolasHug: Things are getting quite difficult for me to understand now :D 

We have various types named `SwScale`, `SwScaleContext`, `SwsContext`, `UniqueSwsContext`, `SwsFrameContext`...

It might be worth addressing this TODO as part of this PR? I don't think we'll want to change  `SwsContext` and `UniqueSwsContext` because they're from FFmpeg.

I think `SwScale` is an OK name for our "all-encompassing swscale handler", but maybe `SwScaleContext` and `SwsFrameContext` should be name "config" instead of "context"?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -198,8 +198,23 @@ void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
     outputTensor = preAllocatedOutputTensor.value_or(
         allocateEmptyHWCTensor(outputDims, torch::kCPU));
 
-    int resultHeight =
-        convertAVFrameToTensorUsingSwScale(avFrame, outputTensor, outputDims);
+    enum AVPixelFormat avFrameFormat =
+        static_cast<enum AVPixelFormat>(avFrame->format);
+
+    SwScaleContext swScaleContext(
```
NicolasHug: `SwScaleContext` and the existing `SwsFrameContext` are essentially the same thing, it's just that `SwScaleContext` has the additional `AVColorSpace inputColorspace` in there. 

Q - Do we need to have both structs, or could we just use a single one and simplify a bit? Sorry for not being able to suggest an actual alternative at this time, I'm still trying to understand how everything fits together

mollyxu: I cleaned up the naming a bit and also think that they are essentially the same thing with `inputColorspace` being the only difference. I merged them into a single `SwsFrameConfig` that includes colorspace, and updated `createSwsContext` to pull colorspace from the config instead of taking it as a separate parameter.

src/torchcodec/_core/SwScale.cpp
```diff
@@ -0,0 +1,115 @@
...
+  int colorConvertedHeight = sws_scale(
+      colorConversionSwsContext_.get(),
+      avFrame->data,
+      avFrame->linesize,
+      0,
+      avFrame->height,
+      colorConvertedPointers,
+      colorConvertedLinesizes);
+
+  TORCH_CHECK(
```
NicolasHug:
```suggestion
  STD_TORCH_CHECK(
```

src/torchcodec/_core/SwScale.cpp
```diff
@@ -0,0 +1,115 @@
...
+  // When resizing is needed, we do sws_scale twice: first convert to RGB24 at
+  // original resolution, then resize in RGB24 space. This ensures transforms
+  // happen in the output color space (RGB24) rather than the input color space
+  // (YUV).
+  //
+  // When no resize is needed, we do color conversion directly into the output
+  // tensor.
+  torch::Tensor colorConvertedTensor = needsResize_
+      ? allocateEmptyHWCTensor(
+            FrameDims(config_.inputHeight, config_.inputWidth), torch::kCPU)
```
NicolasHug:
```suggestion
            FrameDims(config_.inputHeight, config_.inputWidth), kStableCPU)
```

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -748,17 +748,18 @@ UniqueAVFrame BetaCudaDeviceInterface::transferCpuFrameToGpuNV12(
       "Failed to allocate NV12 CPU frame buffer: ",
       getFFMPEGErrorStringFromErrorCode(ret));
 
-  SwsFrameContext swsFrameContext(
+  SwsFrameConfig swsFrameConfig(
       width,
       height,
       static_cast<AVPixelFormat>(cpuFrame->format),
+      cpuFrame->colorspace,
```
NicolasHug, line 755: This is great because I think it's also a bug fix. The creation of the `swsContext_` does depend on the color space, so if the color-space changes we should re-create the `swsContext_`. We weren't doing that in main previously!

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -681,39 +681,41 @@ std::optional<double> getRotationFromStream(const AVStream* avStream) {
   return rotation;
 }
 
-SwsFrameContext::SwsFrameContext(
+SwsFrameConfig::SwsFrameConfig(
```
NicolasHug: I think calling this "config" helps disambiguate. If I can be annoyingly picky once more, I think this could be `SwsConfig` instead of `SwsFrameConfig`. It makes it consistent with the `FitlersConfig` class, and it's also really clear that it's just a config for swscale itself.

NicolasHug (approved):
Thanks @mollyxu !

PR #1216, opened 2026-02-02
Find muxer options in VideoEncoder to test fragmented mp4

Dan-Flores:
This PR adds logic to `sortCodecOptions` to find and set muxer options (check available options with `ffmpeg -h muxer=mp4`)

With these changes, we can now test that fragment related `movflag` options are utilized correctly. 
But, without a 'streaming' video encoder, the benefits of fragmented mp4's are still not unlocked:
  * Fragmented MP4s are useful because they allow the video file to be consumed before it is closed. 
  * However, since our encoder expects all frames to be encoded at once, the fragments are not flushed to become available to consume until the end of the encoding process, so we cannot access the video before encoding is completed.


src/torchcodec/_core/Encoder.cpp
```diff
@@ -921,6 +940,13 @@ void VideoEncoder::encode() {
   flushBuffers();
 
   status = av_write_trailer(avFormatContext_.get());
+  // av_write_trailer returns mfra atom size (positive) for fragmented
+  // containers, which we'd misinterpret as an error. So we replace positive
+  // values with AVSUCCESS. See:
+  // https://github.com/FFmpeg/FFmpeg/blob/n8.0/libavformat/movenc.c#L8666
+  if (status > 0) {
+    status = AVSUCCESS;
+  }
```
Dan-Flores, line 943: Without this conversion, we interpret the positive returned integer as a random low level error unrelated to the encoding process (ex. `24` raises EMFILE: too many open files). It might just be a bug in the FFmpeg implementation that we can work around.

Since FFmpeg errors are negative, I think its fine to set the status to successful here.

mollyxu: Thanks for providing context! Would we want to also add a comment on how ffmpeg errors are negative and link to source code?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -631,24 +631,42 @@ void tryToValidateCodecOption(
...
+    // Check if option is muxer-specific option
+    // (Returned from `ffmpeg -h muxer=mp4`)
+    const AVOption* muxerOpt = nullptr;
+    if (muxerClass) {
+      muxerOpt = av_opt_find2(
+          &muxerClass,
+          key.c_str(),
+          nullptr,
+          0,
+          AV_OPT_SEARCH_FAKE_OBJ,
```
NicolasHug, line 665: Any reason not to use `AV_OPT_SEARCH_CHILDREN` here? It's used above for `fmtOpt`

test/test_encoders.py
```diff
@@ -1509,3 +1509,49 @@ def test_nvenc_against_ffmpeg_cli(
                     assert color_range == encoder_metadata["color_range"]
             if color_space is not None:
                 assert color_space == encoder_metadata["color_space"]
+
+    @pytest.mark.skipif(
+        ffmpeg_major_version == 4,
+        reason="On FFmpeg 4 we error on truncated packets",
```
NicolasHug: Where do we error? By "we" you mean TC, or FFmpeg (i.e. is this limitation enforced by us, or by FFmpeg)?

test/test_encoders.py
```diff
@@ -1509,3 +1509,49 @@ def test_nvenc_against_ffmpeg_cli(
...
+        truncated_size = int(len(full_content) * 0.5)
+        with open(encoded_path, "wb") as f:
+            f.write(full_content[:truncated_size])
+
+        # Decode the truncated file and verify first 10 frames match reference
+        truncated_decoder = VideoDecoder(encoded_path)
+        assert len(truncated_decoder) >= 10
+        for i in range(10):
+            truncated_frame = truncated_decoder.get_frame_at(i)
+            assert torch.equal(truncated_frame.data, reference_frames[i].data)
```
NicolasHug: Use `torch.testing.assert_equal(..., rtol=0, atol=0)`, it provides better error messages than `torch,equal`.

test/test_encoders.py
```diff
@@ -1509,3 +1509,49 @@ def test_nvenc_against_ffmpeg_cli(
...
+    @pytest.mark.parametrize("format", ["mp4", "mov"])
+    @pytest.mark.parametrize(
+        "extra_options",
+        [
+            # frag_keyframe with empty_moov (new fragment every keyframe)
+            {"movflags": "+frag_keyframe+empty_moov"},
+            # frag_duration creates fragments based on duration (in microseconds)
+            {"movflags": "+empty_moov", "frag_duration": "1000000"},
+        ],
+    )
```
NicolasHug, line 1526: This is a fun test!
Just to make sure, did you verify that it breaks if we don't pass the `extra_option` flags?

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Updated!

src/torchcodec/_core/Encoder.cpp
Dan-Flores: We only use `AV_OPT_SEARCH_CHILDREN` when an `AVClass` has children to iterate over. 
I checked that the `AVFormatContext` class defines `child_next` and `child_class_iterate` to iterate through child fields:
https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/options.c#L128-L137

While the muxer class for MOV does not define these functions:
https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/movenc.c#L132-L137

test/test_encoders.py
Dan-Flores: This is a `torchcodec` limitation, I was able to decode the truncated video with FFmpeg 4. I'll update the reason message.

The error we get is `Invalid data found when processing input`, which means we are hitting `AVERROR_INVALIDDATA` ([see errors](https://ffmpeg.org/doxygen/8.0/error_8h_source.html)), which we do not handle, we only check for `AVERROR_EOF`:
https://github.com/meta-pytorch/torchcodec/blob/6f10da06a32e44699329029b4fa41545a6880bc8/src/torchcodec/_core/SingleStreamDecoder.cpp#L265-L272

Since this error only occurs on FFmpeg 4, I think it should be ok to leave as is.

test/test_encoders.py
Dan-Flores: Yes, it fails to open the file when using `extra_options={}`:
`RuntimeError: Could not open input file: .../fragmented_output.mp4 Invalid data found when processing input`

Hopefully this fun test does not introduce flakiness!

PR #1221, opened 2026-02-04
[wip] Decode WAV with C++ backend

Dan-Flores:
To review:
* First, see the changes to the public `AudioDecoder` class.
  * This implementation changes the `_audio_decoder.py` Python file. To completely contain changes in C++ requires more complex changes, ex. implementing a `Decoder` class that `SingleStreamDecoder` and `WavDecoder` implement.
* Read `_is_uncompressed_wav`, and read a function it dispatches to, like `get_wav_metadata_from_file`. 
  * Skim through the input type handler classes, `WavFileReader` and `WavTensorReader`
* Read `WavDecoder::WavDecoder` init implementation, follow it to `parseHeader`.
* Wonder at `convertSamplesToFloat`
  * Using tensor operations here reduces the performance gains. This implementation does the conversion in a single pass.

### DIFF
--- a/src/torchcodec/_core/WavDecoder.cpp
+++ b/src/torchcodec/_core/WavDecoder.cpp
@@ -290,23 +290,18 @@ void WavDecoder::convertChunkToFloatDirect(
     effectiveFormat = header_.subFormat;
   }
 
+  std::vector<float> tempBuffer(numSamples * header_.numChannels);
+
   if (effectiveFormat == WAV_FORMAT_IEEE_FLOAT) {
     if (header_.bitsPerSample == 32) {
       const float* src = reinterpret_cast<const float*>(rawData);
-      // Convert from interleaved (s0c0, s0c1, s1c0, s1c1...) to (channels,
-      // samples)
-      for (int64_t s = 0; s < numSamples; ++s) {
-        for (int64_t c = 0; c < header_.numChannels; ++c) {
-          outputPtr[c * totalSamples + s] = src[s * header_.numChannels + c];
-        }
+      for (int64_t i = 0; i < numSamples * header_.numChannels; ++i) {
+        tempBuffer[i] = src[i];
       }
     } else if (header_.bitsPerSample == 64) {
       const double* src = reinterpret_cast<const double*>(rawData);
-      for (int64_t s = 0; s < numSamples; ++s) {
-        for (int64_t c = 0; c < header_.numChannels; ++c) {
-          outputPtr[c * totalSamples + s] =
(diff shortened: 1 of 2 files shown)


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -61,6 +63,26 @@ def __init__(
...
+                data=samples,
+                pts_seconds=0.0,
+                duration_seconds=metadata["durationSeconds"],
+                sample_rate=metadata["sampleRate"],
+            )
+            self.stream_index = 0
+            self._desired_sample_rate = metadata["sampleRate"]
+            self._decoder = None  # type: ignore[assignment]
+            self.metadata = AudioStreamMetadata.from_json(metadata)
+            return
```
Dan-Flores, line 85: The `AudioDecoder` exposes the input audio's metadata to the user. Using a C++ backend without FFmpeg, we pass over the information necessary to create `AudioStreamMetadata` from the C++ side. This implementation uses JSON, but its possible to pass each field individually and construct an `AudioStreamMetadata` object here as well.

src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -61,6 +63,26 @@ def __init__(
...
+            source, stream_index, sample_rate, num_channels
+        )
+        if metadata_json:
+            metadata = json.loads(metadata_json)
+            self._wav_samples = AudioSamples(
+                data=samples,
+                pts_seconds=0.0,
+                duration_seconds=metadata["durationSeconds"],
+                sample_rate=metadata["sampleRate"],
+            )
```
Dan-Flores, line 77: `self._wav_source` is only populated if WAV decoding was successful.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,412 @@
...
+  // Verify RIFF header
+  if (!checkFourCC(data, "RIFF")) {
+    throw std::runtime_error("Missing RIFF header");
+  }
+
+  // Verify WAVE format
+  if (!checkFourCC(data + 8, "WAVE")) {
+    throw std::runtime_error("Missing WAVE format identifier");
+  }
+
```
Dan-Flores, line 142: Once we find the `RIFF` and `WAVE` signatures, we can look for the `fmt` chunk which contains metadata, and the `data` chunk which contains the actual samples. We find the data chunk now to store `dataSize` and `dataOffset`, which will be needed later for decoding.

PR #1222, opened 2026-02-04
[wip] Decode WAV with Python backend

Dan-Flores:
To review:
* See the top level changes to `_audio_decoder.py`.
* Read the definition of `class WavDecoder`, and the associated `try_create`.
* Skim through `_parse_wav_chunks`, it looks complicated but just handles the [WAV file header](https://en.wikipedia.org/wiki/WAV#WAV_file_header).
* Skim through `_samples_from_bytes`.  This conversion is considerably easier to read than the C++ implementation.

### DIFF
--- a/src/torchcodec/decoders/_fast_wav.py
+++ b/src/torchcodec/decoders/_fast_wav.py
@@ -13,11 +13,13 @@ than going through FFmpeg's full decoding pipeline.
 """
 
 import io
+import os
 import struct
 from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
+import numpy as np
 import torch
 from torch import Tensor
 
@@ -187,8 +189,9 @@ def _samples_from_bytes(
             return data
 
         elif metadata.bits_per_sample == 24:
-            # s24: Use 24-bit helper + convert + deinterleave
-            int32_data = _convert_24bit_pcm(memoryview(audio_bytes))
+            # s24: Use 24-bit conversion + convert + deinterleave
+            np_data = np.frombuffer(audio_bytes, dtype=np.uint8)
+            int32_data = _convert_24bit(np_data)
             data = (
                 int32_data.to(torch.float32)
                 .view(num_samples, metadata.num_channels)
@@ -202,8 +205,6 @@ def _samples_from_bytes(
     else:
         raise ValueError(f"Unsupported audio format: {metadata.audio_format}")
 
-    return torch.empty((metadata.num_channels, 0), dtype=torch.float32)
-
 
 def _supports_buffer_protocol(obj) -> bool:
     """Check if object supports the buffer protocol for direct memory access."""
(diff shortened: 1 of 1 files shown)


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -61,6 +62,17 @@ def __init__(
         num_channels: int | None = None,
     ):
         torch._C._log_api_usage_once("torchcodec.decoders.AudioDecoder")
+
+        # Try fast WAV path
+        self._wav_decoder = WavDecoder.validate_and_init(
+            source, sample_rate, num_channels, stream_index
+        )
```
Dan-Flores: `self._wav_decoder` is only populated if WAV decoding was successful.

src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -61,6 +62,17 @@ def __init__(
...
+        if self._wav_decoder is not None:
+            self.stream_index = self._wav_decoder.stream_index
+            self.metadata = self._wav_decoder.metadata
```
Dan-Flores, line 74: Because `WavDecoder` is a python class, we can set the `AudioStreamMetadata` class easily.

src/torchcodec/decoders/_audio_decoder.py
NicolasHug: Do we have a sense of how expensive this check is? It'll be run for every single input file so it's important that it's quick.

src/torchcodec/decoders/_fast_wav.py
```diff
@@ -0,0 +1,342 @@
...
+                try:
+                    with open(path, "rb") as f:
+                        source_bytes = f.read()
+                except OSError:
+                    return None
+        elif isinstance(source, (io.RawIOBase, io.BufferedReader)) or (
+            hasattr(source, "read") and hasattr(source, "seek")
+        ):
+            source_bytes = source.read()
+            # Will reset seek position below if we can't use fast path
```
NicolasHug: So this works but it also reads / loads / downloads the *entire* content of a file-like object in memory. We should try to see how FFmpeg behaves when decoding a very long wav file for example. Does it need to load the entire file at once? If not, we might be losing that functionality here, which is something to consider.

Same for the C++ alternative, which I haven't checked yet.

src/torchcodec/decoders/_fast_wav.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def __init__(
+        self,
+        source: str | Path | io.RawIOBase | io.BufferedReader | bytes | Tensor,
+        sample_rate: int | None = None,
+        num_channels: int | None = None,
+        stream_index: int | None = None,
+    ):
+        """
+        Create a WavDecoder for the given source.
```
Dan-Flores, line 262: This function handles all input source types, and calls `_parse_wav_chunks` with a function that can read the detected input type. 
- If successful, this function initializes a `WavDecoder` object. 
- If it fails, it will raise an error. This init is designed to be called by `try_create` below, which catches the error so we can fallback to FFmpeg backend.

src/torchcodec/decoders/_fast_wav.py
```diff
@@ -0,0 +1,388 @@
...
+    Convert raw PCM bytes to a float32 tensor.
+
+    Args:
+        audio_bytes: Raw PCM audio data bytes
+        metadata: WAV metadata for format information
+    """
+    bytes_per_sample = metadata.bits_per_sample // 8
+    num_samples = len(audio_bytes) // bytes_per_sample // metadata.num_channels
+
+    # Convert to tensor based on format
```
Dan-Flores: Here various WAV formats are normalized to [-1, 1]. 
Some formats use the full range of a dtype, so `torch.iinfo` is used to get the maximum value of that type.

src/torchcodec/decoders/_fast_wav.py
```diff
@@ -0,0 +1,388 @@
...
+
+    if header[0:4] != b"RIFF" or header[8:12] != b"WAVE":
+        raise ValueError("Not a valid WAV file")
+
+    pos = 12
+    fmt_found = False
+    audio_format = 0
+    num_channels = 0
+    sample_rate = 0
+    bits_per_sample = 0
```
Dan-Flores, line 64: Once we find the `RIFF` and `WAVE` signatures, we can look for the `fmt` chunk which contains metadata, and the `data` chunk which contains the actual samples. We find the `data` chunk now to store `bytes_per_sample` and `num_samples`, which will be needed later for decoding.

src/torchcodec/decoders/_fast_wav.py
Dan-Flores: It does not need to, I've updated both implementations to only read the header at first. Thanks for the suggestion!

Issue #1164, opened 2026-01-13
misleading error when libpython is missing


cen-kalakaua:
### 📚 The doc issue

On debian 13 after installing torchcodec and ffmpeg (via apt-get install) importing torchcodec fails with "Could not load library: ... libtorchcodec_custom_ops7.so" (ops not core!). I straced it and what's actually missing is "apt-get install python3-dev" which install libpython3.13 along with it

zzaudio:
Life saving information... Worked for my Ubuntu 24.04 + Python 3.12 + torch 2.9.1 + cuda 12.8 setting.

lamikr:
Normally there should not be a need to link the libpython library directly as python interpreter can take of loading the symbols needed from it.  So we could take a away the direct dependency to libpython from the libtorchcodec_custom_ops[4,5,6,7,8].so libraries.

Originally when I build the torchcodec and I checked the dependencies, these were required:

```
ldd .venv/lib/python3.12/site-packages/torchcodec/libtorchcodec_custom_ops8.so 
	linux-vdso.so.1 (0x00007f34822cc000)
	libtorchcodec_core8.so => not found
	libpython3.12.so.1.0 => /lib/x86_64-linux-gnu/libpython3.12.so.1.0 (0x00007f3481800000)
	libavutil.so.60 => not found
	libavcodec.so.62 => not found
	libavformat.so.62 => not found
	libavdevice.so.62 => not found
	libavfilter.so.11 => not found
	libswscale.so.9 => not found
	libswresample.so.6 => not found
	libtorch.so => not found
	libtorch_cpu.so => not found
	libtorch_hip.so => not found
	libc10.so => not found
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f3481400000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f3482227000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f3481000000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f348213c000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f3482120000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f34820f4000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f34822ce000)
```
[...]

PR #1224, opened 2026-02-05
Direct link to libpython3.so is not needed

lamikr:
There is no need to link custom_ops libraries directly against the libpython as python interpreter will load the python related dependencies at runtime.

Building, linking and execution of applications using torchcodec will work without linking the custom_ops libraries directly against the libpython-library.

I have tested this fix on environment which have
python 3.12 installed but does not have separate libpython library installed.

Related to: https://github.com/meta-pytorch/torchcodec/issues/1164

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -169,7 +169,7 @@ function(make_torchcodec_libraries
     )
     set(custom_ops_dependencies
         ${core_library_name}
-        ${Python3_LIBRARIES}
+        ""
     )
     make_torchcodec_sublibrary(
         "${custom_ops_library_name}"


mollyxu:
Hi @lamikr, thanks for looking into this and creating the PR! We don't know the exact reason that `libpython3.so` was there for all environments. 

From the CI, it seems like removing this would lead to a fatal error on windows due to `cannot open file 'python310.lib' [C:\actions-runner\_work\torchcodec\torchcodec\meta-pytorch\torchcodec\build\temp.win-amd64-cpython-310\Release\src\torchcodec\_core\libtorchcodec_custom_ops6.vcxproj]` 

If you can fix the error on windows in the CI we would be happy to take another look!

PR #1226, opened 2026-02-09
[wip] Rotation support on cuda devices - torch::rot90

mollyxu:


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1405,6 +1405,25 @@ FrameOutput SingleStreamDecoder::convertAVFrameToFrameOutput(
 // docs: https://pytorch.org/docs/stable/generated/torch.permute.html
 torch::Tensor SingleStreamDecoder::maybePermuteHWC2CHW(
```
mollyxu, line 1406: An alternative to adding this logic to `maybePermuteHWC2CHW` is create a new method `maybeRotate` and call it every time we call `maybePermuteHWC2CHW`. That might be a clearer separation of concerns

PR #1231, opened 2026-02-11
Use `torch::stable::Device`

NicolasHug:
TSIA. Good luck.

### DIFF
--- a/src/torchcodec/_core/Metadata.cpp
+++ b/src/torchcodec/_core/Metadata.cpp
@@ -75,7 +75,6 @@ std::optional<double> StreamMetadata::getEndStreamSeconds(
     }
     default:
       STD_TORCH_CHECK(false, "Unknown SeekMode");
-      return std::nullopt;
   }
 }
 
--- a/src/torchcodec/_core/NVDECCache.cpp
+++ b/src/torchcodec/_core/NVDECCache.cpp
@@ -28,11 +28,15 @@ UniqueCUvideodecoder NVDECCache::getDecoder(CUVIDEOFORMAT* videoFormat) {
   CacheKey key(videoFormat);
   std::lock_guard<std::mutex> lock(cacheLock_);
 
-  auto it = cache_.find(key);
-  if (it != cache_.end()) {
-    auto decoder = std::move(it->second);
-    cache_.erase(it);
-    return decoder;
+  // Find all entries with matching key and look for one not in use
+  auto range = cache_.equal_range(key);
+  for (auto it = range.first; it != range.second; ++it) {
+    if (!it->second.inUse) {
+      // Take ownership of the decoder and remove the entry from cache
+      auto decoder = std::move(it->second.decoder);
+      cache_.erase(it);
+      return decoder;
+    }
   }
 
   return nullptr;
@@ -52,7 +56,8 @@ bool NVDECCache::returnDecoder(
     return false;
   }
 
-  cache_[key] = std::move(decoder);
+  // Add the decoder back to cache as not in use
+  cache_.emplace(key, CacheEntry(std::move(decoder), false));
   return true;
 }
 
--- a/src/torchcodec/_core/NVDECCache.h
+++ b/src/torchcodec/_core/NVDECCache.h
@@ -35,16 +35,30 @@ struct CUvideoDecoderDeleter {
 using UniqueCUvideodecoder =
     std::unique_ptr<CUvideodecoder, CUvideoDecoderDeleter>;
 
+// Cache entry that holds a decoder and tracks whether it's currently in use.
+struct CacheEntry {
+  UniqueCUvideodecoder decoder;
+  bool inUse;
+
+  CacheEntry(UniqueCUvideodecoder dec, bool used)
+      : decoder(std::move(dec)), inUse(used) {}
+};
+
 // A per-device cache for NVDEC decoders. There is one instance of this class
 // per GPU device, and it is accessed through the static getCache() method.
+// The cache supports multiple decoders with the same parameters, tracking
+// which ones are currently in use.
 class NVDECCache {
  public:
   static NVDECCache& getCache(const StableDevice& device);
 
-  // Get decoder from cache - returns nullptr if none available
+  // Get decoder from cache - returns nullptr if none available.
+  // The returned decoder is marked as "in use" until returned via
+  // returnDecoder.
   UniqueCUvideodecoder getDecoder(CUVIDEOFORMAT* videoFormat);
 
-  // Return decoder to cache - returns true if added to cache
+  // Return decoder to cache - marks the decoder as not in use.
+  // Returns true if the decoder was successfully returned to cache.
   bool returnDecoder(CUVIDEOFORMAT* videoFormat, UniqueCUvideodecoder decoder);
 
  private:
@@ -92,7 +106,7 @@ class NVDECCache {
   NVDECCache() = default;
   ~NVDECCache() = default;
 
-  std::map<CacheKey, UniqueCUvideodecoder> cache_;
+  std::multimap<CacheKey, CacheEntry> cache_;
   std::mutex cacheLock_;
 
   // Max number of cached decoders, per device


src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -22,12 +22,16 @@ PerGpuCache<NppStreamContext> g_cached_npp_ctxs(
...
   // It is important for pytorch itself to create the cuda context. If ffmpeg
   // creates the context it may not be compatible with pytorch.
   // This is a dummy tensor to initialize the cuda context.
   torch::Tensor dummyTensorForCudaInitialization = torch::zeros(
-      {1}, torch::TensorOptions().dtype(torch::kUInt8).device(device));
+      {1},
+      torch::TensorOptions()
+          .dtype(torch::kUInt8)
+          .device(torch::Device(
+              static_cast<c10::DeviceType>(device.type()), device.index())));
```
NicolasHug, line 34: We're casting our `stableDevice` into a non-stable `torch::Device`, because this is what `torch::zeros` expects. These kinds of quirks will go away once we migrate fully to the stable ABI: we won't be relying on `torch::zeros` anymore, we'll be relying on its stable counterpart, which will accept a `stableDevice`.

I'm not making this a TODO, because this will be a *forced* changed.

test/third-party-interface/ThirdPartyInterfaceTest.cpp
```diff
@@ -12,7 +12,7 @@ namespace facebook::torchcodec {
 
 class DummyDeviceInterface : public DeviceInterface {
  public:
-  DummyDeviceInterface(const torch::Device& device) : DeviceInterface(device) {}
+  DummyDeviceInterface(const StableDevice& device) : DeviceInterface(device) {}
```
NicolasHug, line 15: FYI @dvrogozh , I think we'll move forward with this, which means third-party interfaces will need to use `StableDevice`. This is a subset of the changes from https://github.com/meta-pytorch/torchcodec/pull/1188, as I'll try to do the port in stages.

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -235,6 +235,34 @@ function(make_torchcodec_libraries
...
+    # VideoStreamOptions could be in fact "hidden" as well.
+    # TODO_STABLE_ABI: do that!
+    if (LINUX)
+        target_compile_options(
+            ${core_library_name}
+            PRIVATE
+            "-Wno-attributes"
+        )
+    endif()
+
```
NicolasHug, line 265: Image

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -164,15 +164,15 @@ class DeviceInterface {
...
   SharedAVCodecContext codecContext_;
   AVMediaType avMediaType_;
 };
 
 using CreateDeviceInterfaceFn =
-    std::function<DeviceInterface*(const torch::Device& device)>;
+    std::function<DeviceInterface*(const StableDevice& device)>;
 
-bool registerDeviceInterface(
+TORCHCODEC_THIRD_PARTY_API bool registerDeviceInterface(
```
NicolasHug, line 175: The `TORCHCODEC_THIRD_PARTY_API` helper ensures the `registerDeviceInterface` symbol is properly public - which is needed because it is called by third-party libraries. Note that it was already public on `main`! Now with this PR, we need to *force* it to public, because it would otherwise get demoted to hidden. Why? Because some of its parameters `CreateDeviceInterfaceFn` *transitively* depends on a hidden type, namely `StableDevice`. Because `StableDevice` is hidden, the compiler will by default demote everything that also depends on it - that includes `registerDeviceInterface`. Hence why we have to force this back to public.

Also relevant: https://github.com/meta-pytorch/torchcodec/pull/1231/changes#r2794367843

test/third-party-interface/ThirdPartyInterfaceTest.cpp
dvrogozh: Ok, I see that's now not a draft, so I assume it's ready for review and try out?

NicolasHug: yes!

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -61,7 +59,8 @@ UniqueAVBufferRef getHardwareDeviceContext(const torch::Device& device) {
   }
 
   // Create hardware device context
-  c10::cuda::CUDAGuard deviceGuard(device);
+  c10::cuda::CUDAGuard deviceGuard(
+      c10::Device(static_cast<c10::DeviceType>(device.type()), device.index()));
```
Dan-Flores, line 63: Is this a similar case to `torch::zeros`, where we must cast to a non-stable `torch::Device` until a stable counterpart for `CUDAGuard` is implemented?

src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -285,7 +284,8 @@ void CudaDeviceInterface::convertAVFrameToFrameOutput(
       preAllocatedOutputTensor.value().copy_(cpuFrameOutput.data);
       frameOutput.data = preAllocatedOutputTensor.value();
     } else {
-      frameOutput.data = cpuFrameOutput.data.to(device_);
+      frameOutput.data = cpuFrameOutput.data.to(torch::Device(
+          static_cast<c10::DeviceType>(device_.type()), device_.index()));
```
Dan-Flores, line 288: I saw there is a `torch::stable::Tensor` [defined](https://github.com/pytorch/pytorch/blob/04bf7cc08c54e9b790d072dd67950ff6f4241d7f/torch/csrc/stable/tensor_struct.h). I assume we are saving that migration for a separate PR?

src/torchcodec/_core/StableABICompat.h
```diff
@@ -20,3 +31,19 @@
...
+
+namespace facebook::torchcodec {
+
+// Device types
+using StableDevice = torch::stable::Device;
+using StableDeviceType = torch::headeronly::DeviceType;
+using StableDeviceIndex = torch::stable::accelerator::DeviceIndex;
+
+// DeviceGuard for CUDA context management
+using StableDeviceGuard = torch::stable::accelerator::DeviceGuard;
```
Dan-Flores: We aren't using this currently. Was this meant to be used in `CudaDeviceInterface.cpp` instead of casting back to `torch::Device` in `deviceGuard`?

src/torchcodec/_core/CudaDeviceInterface.cpp
mollyxu, line 512: To double check my understanding, once we switch to `torch::stable::Tensor`, we would modify this comparison to `kStableCuda`?

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -5,6 +5,7 @@
 // LICENSE file in the root directory of this source tree.
 
 #include "DeviceInterface.h"
+#include <cstring>
```
mollyxu: I couldn't find specific instances where we were using functions from this header but I might've missed something. Maybe this is something that is related to the rest of porting to stable ABI?

src/torchcodec/_core/CudaDeviceInterface.cpp
NicolasHug: Correct!

src/torchcodec/_core/CudaDeviceInterface.cpp
NicolasHug: Yes this is probably going to be the next (and biggest) migration

src/torchcodec/_core/DeviceInterface.cpp
NicolasHug: Good catch thank you

src/torchcodec/_core/StableABICompat.h
NicolasHug: Good catch, thanks. I'll remove it for now and do a separate migration for all the "CUDA stuff" (streams sync, guards, etc.)

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -235,6 +235,34 @@ function(make_torchcodec_libraries
...
+    # We have such a pattern in the core library with the VideoStreamOptions
+    # class and its StableDevice field:
+    # - VideoStreamOptions has "default" (public) visibility because it is part
+    #   of the core library, and the core library must have public symbols
+    #   because it is depended on by other libraries (the pybind ops library and
+    #   the custom ops library)
+    # - the StableDevice field has "hidden" visibility because this is how
+    #   torch::stable exports it.
+    # This creates this mismatch where a class has "higher" visibility than its
+    # member, hence the warning.
```
Dan-Flores, line 250: Is the issue that the compiler is trying to prevent here that: A user of `VideoStreamOptions` could try to call a method of `StableDevice`, but since `StableDevice` is "hidden", that method will not be available (due to the linker not finding it when building?)

And the reason we can ignore this warning is: `StableDevice` is header only, so all the functions are inline and will be available in each built `.so` file, so no linker issues will occur.

Is this your understanding? I read [this wiki](https://gcc.gnu.org/wiki/Visibility) to try to understand the visibility attribute, but it is a little hard to apply to our situation.

NicolasHug: What's going on is along the same vein as what you're describing, but it's not quite exactly that.

With the warning, the compiler is essentially telling us "Hey, that `VideoStreamOptions` class is supposed to have public visibility, but one of its field has hidden visibility. I'm going to have to make the class hidden too!!".

We can get away with this warning, because we're fine with `VideoStreamOptions` being hidden since no other (dynamically loaded) library is relying on `VideoStreamOptions`.

The key primer here, is that:
- a public symbol can be dynamically loaded by another library
- a private symbol cannot

(see also https://claude.ai/share/0f9ce2e8-0a98-47d1-a4dc-286e618a89e2)

Here, we don't have any library that dynamically loads the core library and that relies on `VideoStreamOptions`, so it can be private.

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -29,6 +29,20 @@ std::string getDeviceType(const std::string& device) {
...
 }
 
+// Parse device type from string (e.g., "cpu", "cuda")
+// TODO_STABLE_ABI: we might need to support more device types, i.e. those from
+// https://github.com/pytorch/pytorch/blob/main/torch/headeronly/core/DeviceType.h
+// Ideally we'd remove this helper?
+StableDeviceType parseDeviceType(const std::string& deviceType) {
+  if (deviceType == "cpu") {
+    return kStableCPU;
+  } else if (deviceType == "cuda") {
```
dvrogozh, line 39: @NicolasHug , these check breaks backends registration which we have designed. I posted the quick fix which just adds XPU to the list. This unblocks me on my side. See https://github.com/meta-pytorch/torchcodec/pull/1248. Please, help to resolve the issue before v0.11.0 release of torchcodec as this will prevent us to release XPU plugin support compatible with torchcodec.

Waiting the resolution to merge in https://github.com/dvrogozh/torchcodec-xpu/pull/16 on my side (update XPU plugin to work with latest torchcodec).

PR #1235, opened 2026-02-11
Support rotation on beta cuda

mollyxu:
Add rotation support to `BetaCudaDeviceInterface` using `torch::rot90`

Parametrize over `cpu` and `cuda:beta` in `test_rotation_applied_to_frames` (note: since we are planning to move to beta cuda eventually, we've decided to currently not support this on cuda with ffmpeg backend)

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -4,7 +4,6 @@
 // This source code is licensed under the BSD-style license found in the
 // LICENSE file in the root directory of this source tree.
 
-#include <c10/cuda/CUDAStream.h>
 #include <torch/types.h>
 #include <map>
 #include <mutex>
@@ -584,8 +583,8 @@ int BetaCudaDeviceInterface::receiveFrame(UniqueAVFrame& avFrame) {
   // the NPP stream before any color conversion.
   // Re types: we get a cudaStream_t from PyTorch but it's interchangeable with
   // CUstream
-  procParams.output_stream = reinterpret_cast<CUstream>(
-      at::cuda::getCurrentCUDAStream(device_.index()).stream());
+  procParams.output_stream =
+      reinterpret_cast<CUstream>(getCurrentCudaStream(device_.index()));
 
   CUdeviceptr framePtr = 0;
   unsigned int pitch = 0;
@@ -869,8 +868,7 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
       gpuFrame->format == AV_PIX_FMT_CUDA,
       "Expected CUDA format frame from BETA CUDA interface");
 
-  at::cuda::CUDAStream nvdecStream =
-      at::cuda::getCurrentCUDAStream(device_.index());
+  cudaStream_t nvdecStream = getCurrentCudaStream(device_.index());
 
   if (rotation_ == Rotation::NONE) {
     validatePreAllocatedTensorShape(preAllocatedOutputTensor, gpuFrame);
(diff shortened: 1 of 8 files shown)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -868,13 +869,52 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
...
+      case Rotation::CW90:
+        k = 3;
+        break;
+      default:
+        STD_TORCH_CHECK(false, "Unexpected rotation value");
+        break;
+    }
+    // Apply rotation using torch::rot90 on the H and W dims of our HWC tensor.
+    // torch::rot90 returns a view, so we need to make it contiguous.
+    frameOutput.data = torch::rot90(frameOutput.data, k, {0, 1}).contiguous();
```
Dan-Flores: The code structure makes it a bit hard to follow what `preAllocatedOutputTensor` is storing, perhaps we could change it to be more explicit when changes to the preallocated tensor are needed:

```py
if rotation_ == Rotation::NONE:
  convertNV12FrameToRGB() # regular call
else:
  convertNV12FrameToRGB() # call without preallocated tensor
  applyRotation()

```
Ideally we could handle the rotation `switch` statement and call to `torch::rot90` in a separate function. 

There might be nuances to the logic I am missing that make this structure difficult, please double check me on this.

NicolasHug: Agreed it might help to have an `applyRotation()` helper.
I'm not sure about having `convertNV12FrameToRGB()` within `if/else` blocks, it might be tricky since there is also `validatePreAllocatedTensorShape` which depends on `preAllocatedOutputTensor` having the correct size (i.e. we'd still need the `savedPreAllocatedOutputTensor` hack above, I think).

test/test_decoders.py
```diff
@@ -1966,20 +1966,26 @@ def test_cpu_fallback_no_fallback_on_cpu_device(self):
         assert "No fallback required" in str(decoder.cpu_fallback)
 
     @pytest.mark.parametrize("dimension_order", ["NCHW", "NHWC"])
-    def test_rotation_applied_to_frames(self, dimension_order):
+    @pytest.mark.parametrize(
+        "device",
+        ("cpu", pytest.param("cuda:beta", marks=pytest.mark.needs_cuda)),
+    )
+    def test_rotation_applied_to_frames(self, dimension_order, device):
```
NicolasHug, line 1975: We should add a small comment here saying that we deliberately choose not to address rotation metadata for the FFmpeg CUDA interface. Otherwise it might surprising to see that we're not parametrizing over all devices/interfaces

NicolasHug (approved):
Thanks @mollyxu , this LGTM, I'll let @Dan-Flores make another pass as well

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
Dan-Flores: I'm still a little hesitant on this approach - when the hack sets `preAllocatedOutputTensor = std::nullopt`, then `validatePreAllocatedTensorShape` [does not actually do any validation](https://github.com/meta-pytorch/torchcodec/blob/49c889ce3b4658497fded0f6d2df159b9ef104b1/src/torchcodec/_core/CUDACommon.cpp#L304). Excluding the validation call in the `else` block would be more explicit about that.

This is minor though, and could be addressed when we implement the less hacky solution which I believe will use [the NPP functions](https://docs.nvidia.com/cuda/archive/10.2/npp/group__rotate__.html)?

mollyxu: Thanks for the discussion! I agree that it is clearer to be explicit in excluding the validation call and to reduce the number of variables

PR #1236, opened 2026-02-11
Resolve lint

Dan-Flores:
Resolves lint:
* Unreachable return in `Metadata.cpp`

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -781,7 +781,7 @@ void VideoEncoder::initializeEncoder(
   // TODO-VideoEncoder: (P2) Allow height and width to be set
   int outWidth = inWidth;
   int outHeight = inHeight;
-  AVPixelFormat outPixelFormat;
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -781,7 +781,7 @@ void VideoEncoder::initializeEncoder(
   // TODO-VideoEncoder: (P2) Allow height and width to be set
   int outWidth = inWidth;
   int outHeight = inHeight;
-  AVPixelFormat outPixelFormat = AV_PIX_FMT_NONE;
+  AVPixelFormat outPixelFormat;
```
NicolasHug: The uncatchable returns seems valid, but in this case I'm not sure. I get that `outPixelFormat` will be set in all 3 branches below, but I still think it's always best to initialize it (the code below may eventually change!). Would you be able to share a link to the original lint message? We might want to ignore it

Dan-Flores: After reviewing, lets ignore this lint.

PR #1238, opened 2026-02-12
Use CUDA and stable APIs instead of at::cuda

NicolasHug:


src/torchcodec/_core/CUDACommon.h
```diff
@@ -25,6 +24,17 @@ namespace facebook::torchcodec {
...
+// C++ wrapper for aoti_torch_get_current_cuda_stream.
+// Returns the current CUDA stream for the given device index.
+// This is the stable ABI way to get a cudaStream_t for use with CUDA libraries
+// (NPP, NVDEC, etc.).
+inline cudaStream_t getCurrentCudaStream(int32_t deviceIndex) {
+  void* stream = nullptr;
+  TORCH_ERROR_CODE_CHECK(
+      aoti_torch_get_current_cuda_stream(deviceIndex, &stream));
+  return static_cast<cudaStream_t>(stream);
+}
```
NicolasHug: TODO for myself, I might need to check for nullptr, and I need to see if we can get rid of `TORCH_ERROR_CODE_CHECK` - or at least understand where it comes from

src/torchcodec/_core/CUDACommon.h
```diff
@@ -6,10 +6,9 @@
 
 #pragma once
 
-#include <ATen/cuda/CUDAEvent.h>
-#include <c10/cuda/CUDAStream.h>
+#include <cuda_runtime.h>
 #include <npp.h>
-#include <torch/types.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
```
Dan-Flores: I was surprised to see this file was not in the `torch/csrc/stable` folder as other includes have been, but the [source code does indicate](https://github.com/pytorch/pytorch/blob/main/torch/csrc/inductor/aoti_torch/c/shim.h#L7-L8) that it is the stable API.

NicolasHug: Great observation! I was surprised as well, but this is what is recommended [from the docs](https://github.com/pytorch/pytorch/blob/7bc8d4b0648e1d364dce0104c3aea2e7e3c1640a/docs/cpp/source/stable.rst#getting-the-current-cuda-stream), and I recently confirmed with Jane that this file is OK and safe to include long-term.

PR #1240, opened 2026-02-13
Consolidate addstream to solve issue 1064

zy1git:
This commit consolidates the two-step API (create decoder + add audio stream) into a single constructor call for AudioDecoder. Previously, creating an AudioDecoder required two separate C++ calls: first creating the decoder, then adding the audio stream. Now this is done in one consolidated operation.

As for the video decoder, we need to get `metadata` (to obtain height and width) and compute `transform_specs `using `metadata.height` and `metadata.width `between the create decoder and add audio stream ( see [this code chunk](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/decoders/_video_decoder.py#L228-L243)), so it is hard to implement. Will continue implementing that. 

Please correct me if I misunderstand anything.


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -61,7 +60,15 @@ def __init__(
...
         torch._C._log_api_usage_once("torchcodec.decoders.AudioDecoder")
-        self._decoder = create_decoder(source=source, seek_mode="approximate")
+
+        # Use consolidated function that creates decoder and adds audio stream in one call
+        self._decoder = create_audio_decoder(
+            source=source,
+            seek_mode="approximate",
+            stream_index=stream_index,
+            sample_rate=sample_rate,
+            num_channels=num_channels,
```
zy1git, line 70: Here is to combine create decoder and add stream into one call.

zy1git:
This PR has both python and C++ which is hard to review. I close this PR and create a new PR with python wrapper first.

PR #1242, opened 2026-02-13
Use `torch::stable::Tensor` everywhere except in `custom_ops.cpp`

NicolasHug:
🫡

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -358,9 +358,9 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
   int64_t numFrames = all_frames.sizes()[0];
   streamInfos_[streamIndex].allFrames.reserve(numFrames);
   streamInfos_[streamIndex].keyFrames.reserve(numFrames);
-  auto pts_data = accessor<int64_t, 1>(all_frames);
-  auto is_key_frame_data = accessor<bool, 1>(is_key_frame);
-  auto duration_data = accessor<int64_t, 1>(duration);
+  auto pts_data = constAccessor<int64_t, 1>(all_frames);
+  auto is_key_frame_data = constAccessor<bool, 1>(is_key_frame);
+  auto duration_data = constAccessor<int64_t, 1>(duration);
 
   auto& streamMetadata = containerMetadata_.allStreamMetadata[streamIndex];
 
@@ -408,7 +408,7 @@ torch::stable::Tensor SingleStreamDecoder::getKeyFrameIndices() {
       streamInfos_[activeStreamIndex_].keyFrames;
   torch::stable::Tensor keyFrameIndices = torch::stable::empty(
       {static_cast<int64_t>(keyFrames.size())}, kStableInt64);
-  auto keyFrameIndicesAccessor = accessor<int64_t, 1>(keyFrameIndices);
+  auto keyFrameIndicesAccessor = mutableAccessor<int64_t, 1>(keyFrameIndices);
   for (size_t i = 0; i < keyFrames.size(); ++i) {
     keyFrameIndicesAccessor[i] = keyFrames[i].frameIndex;
   }
@@ -687,7 +687,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
     const torch::stable::Tensor& frameIndices) {
   validateActiveStream(AVMEDIA_TYPE_VIDEO);
 
-  auto frameIndicesData = accessor<int64_t, 1>(frameIndices);
+  auto frameIndicesData = constAccessor<int64_t, 1>(frameIndices);
 
   bool indicesAreSorted = true;
   for (int64_t i = 1; i < frameIndices.numel(); ++i) {
@@ -721,9 +721,9 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
       frameIndices.numel(), getOutputDims(), videoStreamOptions.device);
 
   auto frameBatchOutputPtsSeconds =
-      accessor<double, 1>(frameBatchOutput.ptsSeconds);
+      mutableAccessor<double, 1>(frameBatchOutput.ptsSeconds);
   auto frameBatchOutputDurationSeconds =
-      accessor<double, 1>(frameBatchOutput.durationSeconds);
+      mutableAccessor<double, 1>(frameBatchOutput.durationSeconds);
 
   auto previousIndexInVideo = -1;
   for (int64_t f = 0; f < frameIndices.numel(); ++f) {
@@ -786,9 +786,9 @@ FrameBatchOutput SingleStreamDecoder::getFramesInRange(
       numOutputFrames, getOutputDims(), videoStreamOptions.device);
 
   auto frameBatchOutputPtsSeconds =
-      accessor<double, 1>(frameBatchOutput.ptsSeconds);
+      mutableAccessor<double, 1>(frameBatchOutput.ptsSeconds);
   auto frameBatchOutputDurationSeconds =
-      accessor<double, 1>(frameBatchOutput.durationSeconds);
+      mutableAccessor<double, 1>(frameBatchOutput.durationSeconds);
   for (int64_t i = start, f = 0; i < stop; i += step, ++f) {
     FrameOutput frameOutput =
         getFrameAtIndexInternal(i, selectRow(frameBatchOutput.data, f));
@@ -858,8 +858,8 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedAt(
 
   torch::stable::Tensor frameIndices =
       torch::stable::empty({timestamps.numel()}, kStableInt64);
-  auto frameIndicesAccessor = accessor<int64_t, 1>(frameIndices);
-  auto timestampsAccessor = accessor<double, 1>(timestamps);
+  auto frameIndicesAccessor = mutableAccessor<int64_t, 1>(frameIndices);
+  auto timestampsAccessor = constAccessor<double, 1>(timestamps);
 
   for (int64_t i = 0; i < timestamps.numel(); ++i) {
     auto frameSeconds = timestampsAccessor[i];
@@ -966,9 +966,9 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
         numOutputFrames, getOutputDims(), videoStreamOptions.device);
 
     auto frameBatchOutputPtsSeconds =
-        accessor<double, 1>(frameBatchOutput.ptsSeconds);
+        mutableAccessor<double, 1>(frameBatchOutput.ptsSeconds);
     auto frameBatchOutputDurationSeconds =
-        accessor<double, 1>(frameBatchOutput.durationSeconds);
+        mutableAccessor<double, 1>(frameBatchOutput.durationSeconds);
 
     // Decode frames, reusing already-decoded frames for duplicates
     int64_t lastDecodedSourceIndex = -1;
@@ -1011,9 +1011,9 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     FrameBatchOutput frameBatchOutput(
         numFrames, getOutputDims(), videoStreamOptions.device);
     auto frameBatchOutputPtsSeconds =
-        accessor<double, 1>(frameBatchOutput.ptsSeconds);
+        mutableAccessor<double, 1>(frameBatchOutput.ptsSeconds);
     auto frameBatchOutputDurationSeconds =
-        accessor<double, 1>(frameBatchOutput.durationSeconds);
+        mutableAccessor<double, 1>(frameBatchOutput.durationSeconds);
     for (int64_t i = startFrameIndex, f = 0; i < stopFrameIndex; ++i, ++f) {
       FrameOutput frameOutput =
           getFrameAtIndexInternal(i, selectRow(frameBatchOutput.data, f));
--- a/src/torchcodec/_core/StableABICompat.h
+++ b/src/torchcodec/_core/StableABICompat.h
@@ -123,7 +123,7 @@ inline torch::stable::Tensor selectRow(
 }
 
 template <typename T, size_t N>
-torch::headeronly::HeaderOnlyTensorAccessor<T, N> accessor(
+torch::headeronly::HeaderOnlyTensorAccessor<T, N> mutableAccessor(
     torch::stable::Tensor& tensor) {
   return torch::headeronly::HeaderOnlyTensorAccessor<T, N>(
       tensor.mutable_data_ptr<T>(),
@@ -132,7 +132,7 @@ torch::headeronly::HeaderOnlyTensorAccessor<T, N> accessor(
 }
 
 template <typename T, size_t N>
-torch::headeronly::HeaderOnlyTensorAccessor<const T, N> accessor(
+torch::headeronly::HeaderOnlyTensorAccessor<const T, N> constAccessor(
     const torch::stable::Tensor& tensor) {
   return torch::headeronly::HeaderOnlyTensorAccessor<const T, N>(
       tensor.const_data_ptr<T>(),
@@ -158,6 +158,8 @@ inline const char* deviceTypeName(StableDeviceType deviceType) {
       return "cpu";
     case kStableCUDA:
       return "cuda";
+    case kStableXPU:
+      return "xpu";
     default:
       return "unknown";
   }


src/torchcodec/_core/AVIOTensorContext.cpp
```diff
@@ -105,27 +105,34 @@ int64_t seek(void* opaque, int64_t offset, int whence) {
...
 }
 
 AVIOToTensorContext::AVIOToTensorContext()
     : tensorContext_{
-          torch::empty({INITIAL_TENSOR_SIZE}, {torch::kUInt8}),
+          torch::stable::empty(
+              {INITIAL_TENSOR_SIZE},
+              kStableUInt8,
+              kStableStrided,
+              StableDevice(kStableCPU)),
```
NicolasHug: TODO: rely on defaults?

src/torchcodec/_core/AVIOTensorContext.cpp
```diff
@@ -62,18 +62,18 @@ int write(void* opaque, const uint8_t* buf, int buf_size) {
         MAX_TENSOR_SIZE,
         " bytes. If you think this should be supported, please report.");
 
-    // We double the size of the outpout tensor. Calling cat() may not be the
-    // most efficient, but it's simple.
+    // We double the size of the outpout tensor. Calling stableCat() may not be
+    // the most efficient, but it's simple.
```
NicolasHug: TODO

src/torchcodec/_core/AVIOTensorContext.h
```diff
@@ -23,19 +23,19 @@ struct TensorContext {
 
 // For Decoding: enables users to pass in the entire video or audio as bytes.
 // Our read and seek functions then traverse the bytes in memory.
-class AVIOFromTensorContext : public AVIOContextHolder {
+class TORCHCODEC_API AVIOFromTensorContext : public AVIOContextHolder {
```
NicolasHug: TODO: why is TORCHCODEC_API needed?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -908,12 +931,16 @@ void BetaCudaDeviceInterface::applyRotation(
...
+  // stableRot90 returns a view, so we need to make it contiguous.
+  frameOutput.data =
+      torch::stable::contiguous(stableRot90(frameOutput.data, k, 0, 1));
 
   if (preAllocatedOutputTensor.has_value()) {
-    preAllocatedOutputTensor.value().copy_(frameOutput.data);
+    torch::stable::copy_(
+        preAllocatedOutputTensor.value(),
+        frameOutput.data,
+        /*non_blocking=*/std::nullopt);
```
NicolasHug: TODO rely on default?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -489,31 +498,38 @@ std::optional<torch::Tensor> CpuDeviceInterface::maybeFlushAudioBuffers() {
 
   int numChannels =
       audioStreamOptions_.numChannels.value_or(getNumChannels(codecContext_));
-  torch::Tensor lastSamples =
-      torch::empty({numChannels, numRemainingSamples}, torch::kFloat32);
+  StableTensor lastSamples = torch::stable::empty(
+      {numChannels, numRemainingSamples},
+      kStableFloat32,
+      kStableStrided,
+      StableDevice(kStableCPU));
```
NicolasHug: TODO rely on default?

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -489,31 +498,38 @@ std::optional<torch::Tensor> CpuDeviceInterface::maybeFlushAudioBuffers() {
...
 
   std::vector<uint8_t*> outputBuffers(numChannels);
   for (auto i = 0; i < numChannels; i++) {
-    outputBuffers[i] = static_cast<uint8_t*>(lastSamples[i].data_ptr());
+    outputBuffers[i] = reinterpret_cast<uint8_t*>(
+        lastSamples.mutable_data_ptr<float>() + i * lastSamples.sizes()[1]);
```
NicolasHug: TODO needs comment

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -556,7 +572,11 @@ UniqueAVFrame CpuDeviceInterface::convertTensorToAVFrameForEncoding(
   inputFrame->width = inWidth;
   inputFrame->height = inHeight;
 
-  uint8_t* tensorData = static_cast<uint8_t*>(frame.data_ptr());
+  // The tensor data is read-only here (used as source for sws_scale), but
+  // AVFrame::data requires uint8_t* (non-const). We use const_data_ptr and
+  // const_cast since FFmpeg's sws_scale treats the source as const internally.
+  uint8_t* tensorData =
+      const_cast<uint8_t*>(static_cast<const uint8_t*>(frame.const_data_ptr()));
```
NicolasHug: TODO understand this

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -6,6 +6,7 @@
 
 #include <fmt/format.h>
 #include <pybind11/pybind11.h>
+#include <torch/types.h>
```
NicolasHug, line 9: TODO review entire file

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -114,19 +114,48 @@ std::unique_ptr<DeviceInterface> createDeviceInterface(
...
+
+// Deleter callback for from_blob: frees the AVFrame associated with the data.
+static void avFrameDeleter(void* data) {
+  std::lock_guard<std::mutex> lock(g_avframe_map_mutex);
+  auto it = g_avframe_map.find(data);
+  if (it != g_avframe_map.end()) {
+    UniqueAVFrame avFrameToDelete(it->second);
+    g_avframe_map.erase(it);
+  }
+}
```
NicolasHug: TODO hmmmmmmm

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -78,7 +77,7 @@ class DeviceInterface {
   // buffered samples.
   // Returns an optional tensor containing the flushed samples, or std::nullopt
   // if there are no buffered samples or audio is not supported.
-  virtual std::optional<torch::Tensor> maybeFlushAudioBuffers() {
+  virtual std::optional<StableTensor> maybeFlushAudioBuffers() {
```
NicolasHug: TODO review everything below this file.

src/torchcodec/_core/StableABICompat.h
```diff
@@ -34,7 +46,19 @@
 
 namespace facebook::torchcodec {
 
+// ============================================================================
+// Tensor types
+// ============================================================================
```
NicolasHug: Consider removing most aliases

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -343,28 +342,26 @@ void SingleStreamDecoder::readCustomFrameMappingsUpdateMetadataAndIndex(
...
   // Allocate vectors using num frames to reduce reallocations
-  int64_t numFrames = all_frames.size(0);
+  int64_t numFrames = all_frames.sizes()[0];
   streamInfos_[streamIndex].allFrames.reserve(numFrames);
   streamInfos_[streamIndex].keyFrames.reserve(numFrames);
-  // Use accessor to efficiently access tensor elements
-  auto pts_data = all_frames.accessor<int64_t, 1>();
-  auto is_key_frame_data = is_key_frame.accessor<bool, 1>();
-  auto duration_data = duration.accessor<int64_t, 1>();
+  // Use raw data pointers to efficiently access tensor elements
```
NicolasHug: TODO hm, try accessors?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -723,26 +724,29 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
...
     } else {
       FrameOutput frameOutput = getFrameAtIndexInternal(
-          indexInVideo, frameBatchOutput.data[indexInOutput]);
-      frameBatchOutput.ptsSeconds[indexInOutput] = frameOutput.ptsSeconds;
-      frameBatchOutput.durationSeconds[indexInOutput] =
-          frameOutput.durationSeconds;
+          indexInVideo,
+          torch::stable::select(frameBatchOutput.data, 0, indexInOutput));
+      ptsPtr[indexInOutput] = frameOutput.ptsSeconds;
+      durPtr[indexInOutput] = frameOutput.durationSeconds;
```
NicolasHug: TODO, consider:

- using accessors instead of pointers
- consolidate calls to torch::stable::select into helper

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -404,16 +400,17 @@ int SingleStreamDecoder::getActiveStreamIndex() const {
...
   const std::vector<FrameInfo>& keyFrames =
       streamInfos_[activeStreamIndex_].keyFrames;
-  torch::Tensor keyFrameIndices =
-      torch::empty({static_cast<int64_t>(keyFrames.size())}, {torch::kInt64});
+  StableTensor keyFrameIndices = torch::stable::empty(
+      {static_cast<int64_t>(keyFrames.size())}, kStableInt64);
+  auto keyFrameIndicesAccessor = accessor<int64_t, 1>(keyFrameIndices);
   for (size_t i = 0; i < keyFrames.size(); ++i) {
-    keyFrameIndices[i] = keyFrames[i].frameIndex;
+    keyFrameIndicesAccessor[i] = keyFrames[i].frameIndex;
```
NicolasHug, line 413: Note that we'll use accessors in a bunch of places where we would previously just directly index tensors via `[]`. That's because indexing via `[]` isn't supported anymore on stable Tensors. In this case, it's kind of a fix too, we should probably have been using an accessor in the first place for performance reason (this will only matter if there are a lot of keyframes).

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -723,25 +720,33 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
...
-          frameBatchOutput.data[previousIndexInOutput]);
-      frameBatchOutput.ptsSeconds[indexInOutput] =
-          frameBatchOutput.ptsSeconds[previousIndexInOutput];
-      frameBatchOutput.durationSeconds[indexInOutput] =
-          frameBatchOutput.durationSeconds[previousIndexInOutput];
+      copyFrame(
+          frameBatchOutput.data,
+          indexInOutput,
+          frameBatchOutput.data,
+          previousIndexInOutput);
```
NicolasHug, line 740: As mentioned earlier there is no `[]` support on stable tensors, so this is small helper that avoids the ugly calls to `torch::stable::select(...)` which is the official "replacement"

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -116,19 +117,31 @@ std::unique_ptr<DeviceInterface> createDeviceInterface(
...
-      avFrameClone->data[0], shape, strides, deleter, {torch::kUInt8});
+
+  at::Tensor tensor = at::from_blob(
+      avFrameClone->data[0], shape, strides, deleter, {at::kByte});
+
+  // We got an at::Tensor, we have to convert it to a torch::stable::Tensor.
+  // This is safe, there won't be any memory leak, i.e. the at::Tensor's deleter
+  // will properly be passed down to the torch::stable::Tensor.
+  at::Tensor* p = new at::Tensor(std::move(tensor));
+  return torch::stable::Tensor(reinterpret_cast<AtenTensorHandle>(p));
```
NicolasHug, line 144: To validate that the `deleter` is properly called and that we don't have any memory leak, I used this snippet:

```py
import resource

from torchcodec.transforms import CenterCrop
import torch
import gc


# Use a transform to force `rgbAVFrameToTensor` path via filtergraph.
dec = VideoDecoder("test/resources/testsrc2_h265.mp4", transforms=[CenterCrop((224, 224))])
for _ in range(100):
    all_frames = dec[:]
    del all_frames
    gc.collect()

    max_rss_kb = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print(f"Peak memory usage: {max_rss_kb / 1024:.1f} MB")

```

In the current branch, the memory is stable:

```
~/dev/torchcodec-cuda (stable_tensor_maybe*) » python lol.py 
Peak memory usage: 685.6 MB
Peak memory usage: 685.6 MB
Peak memory usage: 685.6 MB
Peak memory usage: 685.6 MB
Peak memory usage: 685.6 MB
Peak memory usage: 685.6 MB
...
```

When I don't pass the deleter to `from_blob`:

```C++
  at::Tensor tensor = at::from_blob(
      avFrameClone->data[0], shape, strides, {at::kByte});
```

I get a clear memory leak:

```
~/dev/torchcodec-cuda (stable_tensor_maybe*) » python lol.py 
Peak memory usage: 841.5 MB
Peak memory usage: 999.6 MB
Peak memory usage: 1158.5 MB
Peak memory usage: 1316.8 MB
Peak memory usage: 1474.9 MB
...
```

This validates that the `deleter` is properly transfered to the `StableTensor`, and that this (temporary!) hack is safe.

src/torchcodec/_core/AVIOTensorContext.h
```diff
@@ -23,19 +23,19 @@ struct TensorContext {
 
 // For Decoding: enables users to pass in the entire video or audio as bytes.
 // Our read and seek functions then traverse the bytes in memory.
-class AVIOFromTensorContext : public AVIOContextHolder {
+class FORCE_PUBLIC_VISIBILITY AVIOFromTensorContext : public AVIOContextHolder {
```
NicolasHug, line 26: You'll see a bunch of those `FORCE_PUBLIC_VISIBILITY`. We need them for very similar reasons we needed this: https://github.com/meta-pytorch/torchcodec/blob/4cf13336448f24514f6b829d1602e482f0c29ee5/src/torchcodec/_core/CMakeLists.txt#L246-L272

Basically, those APIs need to have "public" (i.e. "default") visibility for a bunch of reasons. But because they now rely on stuff from torch::stable, which *all* have private visibility, the compiler makes them private. But again: we need them to be public! This flag enforces that and overrides the compiler.

I'll be honest, I am not 100% I understand everything that's going on in there. It's complex. You'll see a TODO_STABLE_ABI in the code inlined above, and I hope to clarify all this once I tackle it.

src/torchcodec/_core/StableABICompat.h
```diff
@@ -46,4 +60,124 @@ constexpr auto kStableCPU = torch::headeronly::DeviceType::CPU;
...
+
+// TODO_STABLE_ABI: this should probably be natively supported by torch::stable.
+// Consider upstreaming.
+inline const char* deviceTypeName(StableDeviceType deviceType) {
+  switch (deviceType) {
+    case kStableCPU:
+      return "cpu";
+    case kStableCUDA:
+      return "cuda";
+    default:
```
mollyxu, line 163: should we add a case for xpu here?

```
  case kStableXPU:                                                                                                        
    return "xpu";     
```

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -62,12 +62,8 @@ void initializeCudaContextWithPytorch(const StableDevice& device) {
...
   // This is a dummy tensor to initialize the cuda context.
-  torch::Tensor dummyTensorForCudaInitialization = torch::zeros(
-      {1},
-      torch::TensorOptions()
-          .dtype(torch::kUInt8)
-          .device(torch::Device(
-              static_cast<c10::DeviceType>(device.type()), device.index())));
+  torch::stable::Tensor dummyTensorForCudaInitialization = torch::stable::empty(
+      {1}, kStableUInt8, std::nullopt, StableDevice(device));
 }
```
Dan-Flores, line 68: Here is the forced change you mentioned on [1231](https://github.com/meta-pytorch/torchcodec/pull/1231/changes#r2794186154), this is much cleaner!

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -84,6 +85,25 @@ TORCH_LIBRARY(torchcodec_ns, m) {
...
+// to/from torch::stable::Tensor, which is cheap. Those conversion helpers will
+// eventually be removed.
+inline torch::stable::Tensor toStableTensor(const at::Tensor& tensor) {
+  at::Tensor* p = new at::Tensor(tensor);
+  return torch::stable::Tensor(reinterpret_cast<AtenTensorHandle>(p));
+}
+
+inline at::Tensor toATenTensor(const torch::stable::Tensor& t) {
+  return *reinterpret_cast<at::Tensor*>(t.get());
+}
```
Dan-Flores, line 105: In https://github.com/meta-pytorch/torchcodec/pull/1186, we updated this file to use `torch::Tensor` instead of `at::Tensor`. Is there a reason we are using `at::Tensor` here and elsewhere?

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -5,6 +5,7 @@
 // LICENSE file in the root directory of this source tree.
 
 #include "DeviceInterface.h"
+#include <ATen/ops/from_blob.h>
```
Dan-Flores, line 8: Is this header only needed for the added `at::from_blob` call?

src/torchcodec/_core/StableABICompat.h
```diff
@@ -9,10 +9,24 @@
 #include <torch/csrc/stable/accelerator.h>
 #include <torch/csrc/stable/device.h>
 #include <torch/csrc/stable/library.h>
+#include <torch/csrc/stable/ops.h>
+#include <torch/csrc/stable/tensor.h>
 #include <torch/headeronly/core/DeviceType.h>
+#include <torch/headeronly/core/ScalarType.h>
 
+#include <torch/headeronly/core/TensorAccessor.h>
```
Dan-Flores, line 17: I see these `torch/headeronly` includes are also [officially](https://docs.pytorch.org/docs/stable/notes/libtorch_stable_abi.html) part of the stable ABI. This page would be a good pre-read before reviewing these stable ABI PRs.

src/torchcodec/_core/StableABICompat.h
```diff
@@ -46,4 +60,124 @@ constexpr auto kStableCPU = torch::headeronly::DeviceType::CPU;
...
+constexpr auto kStableFloat64 = torch::headeronly::ScalarType::Double;
+constexpr auto kStableBool = torch::headeronly::ScalarType::Bool;
+
+// Note: the magic use of torch_call_dispatcher is what is officially
+// recommended
+// https://github.com/pytorch/pytorch/blob/89f3759429b96a8693b698f013990240bb4e25b3/docs/source/notes/libtorch_stable_abi.md?plain=1#L221
+// It allows us to make an ABI-stable call to an op that isn't officially in the
+// stable ABI. Some of these ops currently include permute, rot90, etc.
+// See also how xformers relies on it:
+// https://github.com/facebookresearch/xformers/blob/720adff2b021f6f43957718514f5be3d10e36fb1/xformers/csrc/pt_stable_utils.h#L85
```
Dan-Flores, line 76: It would seem the downside of this magic dispatcher is that if the behavior of `aten::permute` changes between `TORCH_ABI_VERSION`, our code would silently behave differently or begin to error, right? This is an unlikely, anyhow.

src/torchcodec/_core/StableABICompat.h
```diff
@@ -46,4 +60,124 @@ constexpr auto kStableCPU = torch::headeronly::DeviceType::CPU;
...
+torch::headeronly::HeaderOnlyTensorAccessor<T, N> accessor(
+    torch::stable::Tensor& tensor) {
+  return torch::headeronly::HeaderOnlyTensorAccessor<T, N>(
+      tensor.mutable_data_ptr<T>(),
+      tensor.sizes().data(),
+      tensor.strides().data());
+}
+
+template <typename T, size_t N>
+torch::headeronly::HeaderOnlyTensorAccessor<const T, N> accessor(
```
Dan-Flores: Since we are required to use `mutable_data_ptr` and `const_data_ptr`, should these accessors be `mutable_accessor` and `const_accessor`? I think there is a small tradeoff between being explicit and being concise, but I have no strong opinions.

src/torchcodec/_core/StableABICompat.h
```diff
@@ -46,4 +60,124 @@ constexpr auto kStableCPU = torch::headeronly::DeviceType::CPU;
...
+      "aten::rot90", "", stack.data(), TORCH_ABI_VERSION));
+  return torch::stable::detail::to<torch::stable::Tensor>(stack[0]);
+}
+
+// Shorthand for torch::stable::select(tensor, 0, index), i.e. tensor[index].
+inline torch::stable::Tensor selectRow(
+    const torch::stable::Tensor& tensor,
+    int64_t index) {
+  return torch::stable::select(tensor, 0, index);
+}
```
Dan-Flores, line 123: How are you thinking about these shorthand implementations? 

I see the value in having shorthand for long types that we use frequently (ex. `kStableCPU` vs `torch::headeronly::DeviceType::CPU`.

For a small abstraction like `selectRow`, would we implement it primarily for readability?

src/torchcodec/_core/custom_ops.cpp
NicolasHug: No reason. This code, and every mention of `at::Tensor` or `torch::Tensor` will be removed as follow-ups, so I didn't think it was worth cleaning.

src/torchcodec/_core/DeviceInterface.cpp
NicolasHug: correct. We could include a more general header I think, but it's not worth it considering this will go away in one of the next PRs.

src/torchcodec/_core/StableABICompat.h
NicolasHug: I think the scenario you describe exists regardless of whether we use `torch_call_dispatcher`. I.e. even if we were to call `aten::permute` directly, if `aten::permute` changes behavior, then so does our code which can lead to problems indeed. 
But `aten::permute` having a silent change of behavior would be a critical torch bug - that is true of any API we use from any of our dependencies (e.g. FFmpeg). 
I might be misunderstanding your comment, please let me know if that's the case

src/torchcodec/_core/StableABICompat.h
NicolasHug: Yeah that's a good point, there's no strong reason *not* to be explicit here. I'll update

src/torchcodec/_core/StableABICompat.h
NicolasHug: So `selectRow` is purely for readability. It is used in a fair amount of places, including in places where it's embedded within function calls, like in:

```C
getFrameAtIndexInternal(i, selectRow(frameBatchOutput.data, f));
```
I think it's a lot easier to read than:
```C
getFrameAtIndexInternal(i, torch::stable::select(frameBatchOutput, 0, f));
```

For the types like `kStableUInt8`, I'm using aliases because those types are used fairly often, and they're quite long (`torch::headeronly::ScalarType::Byte`).  I don't feel too strongly about them and I'm happy to just inline everything if you have a preference.

For the device stuff like `kStableCPU` and `stableDevice`, I am starting to regret having introduced them as aliases. I don't think the aliases add much value, especially not for `stableDevice`. I'm considering removing them as a follow-up. LMK if you have any preference.

src/torchcodec/_core/StableABICompat.h
Dan-Flores: I think you're right. I'm trying to understand what tradeoff exists between having official 'stable' implementations of functions versus using `torch_call_dispatcher` to make an ABI-stable call.

src/torchcodec/_core/AVIOTensorContext.h
Dan-Flores: Now that we have public and private visibility APIs, I wouldn't expect to be able to remove `FORCE_PUBLIC_VISIBILITY`. Would a follow up just focus on renaming and documenting it?

PR #1243, opened 2026-02-16
[video][cuda] NVDECCache eviction policy, static implementation

vkhalidov:
# Summary

Adds the notion of statically defined eviction policy to `NVDECCache`. Eviction policy decides which entry should leave the cache when it's full and a new decoder needs to be created. This PR suggests the following API for an eviction policy:
```
 - void onInsert(CacheIterator it)      // Called when entry is added
 - void onRemove(CacheIterator it)      // Called when entry is removed
 - CacheIterator selectForEviction()    // Returns iterator to evict
 - bool empty() const                   // Returns true if nothing to evict
```
Three example policies are provided:
```
 - LRUEvictionPolicy     // Least Recently Used eviction policy
 - FIFOEvictionPolicy    // First In, First Out eviction policy
 - NoEvictionPolicy      // No eviction, rejects new entries when the cache is full (used before this PR)
```
`LRUEvictionPolicy` is proposed as the default eviction policy.

# Test Plan

Benchmarks were run on a set of 1200 diverse videos of 132 different resolutions; top 10 resolution buckets are:
```
  Count Resolution
    200 720x720
    189 720x1280
    180 640x640
    140 720x900
     70 480x480
     59 360x640
     35 640x360
     30 640x800
     21 480x600
     18 720x406
```

On the plots below, different numbers of frames at different frame steps were extracted from these videos using different numbers of worker threads by decoders using this PR (`with eviction`) and decoders before this PR (`baseline`). Throughput in terms of the number of decoded frames per second (`FPS`) is plotted as a function of the number of worker threads (`# threads`).

15 decoded frames:
Screenshot 2026-02-16 at 09 10 05

30 decoded frames:
Screenshot 2026-02-16 at 09 12 25

60 decoded frames:
Screenshot 2026-02-16 at 09 13 51

90 decoded frames:
Screenshot 2026-02-16 at 09 14 46

# Conclusion

Eviction policy brings advantage in terms of performance and scaling across almost all the settings. However, if the number of decoded frames is too small (e.g. 15 frames at frame step 1) or too big (60 frames at larger frame steps, 90 frames), cache eviction seems to no longer yield advantage.

PR #1246, opened 2026-02-17
[video] make NVDec cache size adjustable

vkhalidov:
# Summary

Introduces API for setting / getting NVDec cache size via `set_nvdec_cache_size` and `get_nvdec_cache_size`.

`set_nvdec_cache_size` is a persistent setter (not a context manager). Unlike `set_cuda_backend`, the cache size is a global property of the shared cache, not a per-decoder-creation setting.

The value of maximum cache size is set per process (`static`) and stored as `std::atomic<int>` - thread-safe since the cache itself is accessed from multiple threads.

`NVDECCacheConfig` files are separated from `NVDECCache.h` header. Indeed, `NVDECCache.h` requires CUDA headers, but `custom_ops.cpp` is compiled without CUDA support. The config is stored in a CUDA-free file that both can use.

# Usage

```
  from torchcodec.decoders import set_nvdec_cache_size, get_nvdec_cache_size

  # Check current value (default: 20)
  print(get_nvdec_cache_size())  # 20

  # Increase the cache
  set_nvdec_cache_size(50)

  # Disable caching
  set_nvdec_cache_size(0)
```

# Performance

Performance of GPU decoders was evaluated with respect to the cache size on H100 GPUs for a diverse set of 1200 videos at 132 different resolutions; top 10 resolution buckets are:
```
  Count Resolution
    200 720x720
    189 720x1280
    180 640x640
    140 720x900
     70 480x480
     59 360x640
     35 640x360
     30 640x800
     21 480x600
     18 720x406
```

The plot below shows good scaling properties of decoders with the number of worker threads as the cache size increases:

Screenshot 2026-02-18 at 09 02 24

Note: these plots were obtained with LRU eviction policy introduced in https://github.com/meta-pytorch/torchcodec/pull/1243

# Test Plan

1. Ran benchmarks for different values of NVDec cache size with `cuda:beta` device (see the plots)
2. Compiled with `ENABLE_CUDA=0` and verified that one can still call the functions `set_nvdec_cache_size` and `get_nvdec_cache_size`

### DIFF
--- a/src/torchcodec/_core/NVDECCacheConfig.cpp
+++ b/src/torchcodec/_core/NVDECCacheConfig.cpp
@@ -45,7 +45,7 @@ int getNVDECCacheCapacity() {
   return g_nvdecCacheCapacity.load();
 }
 
-int getNVDECCacheSize(int device_index) {
+int getNVDECCacheSize([[maybe_unused]] int device_index) {
 #ifdef USE_CUDA
   TORCH_CHECK(
       device_index >= 0 && device_index < MAX_CUDA_GPUS,
--- a/src/torchcodec/_core/NVDECCacheConfig.h
+++ b/src/torchcodec/_core/NVDECCacheConfig.h
@@ -25,6 +25,6 @@ int getNVDECCacheCapacity();
 // Get the current number of entries in the NVDEC decoder cache for a device.
 // This is currently only used for tests, and not publicly exposed.
 // TODO expose it?
-int getNVDECCacheSize([[maybe_unused]] int device_index);
+int getNVDECCacheSize(int device_index);
 
 } // namespace facebook::torchcodec


vkhalidov:
rebased on recent `main`

vkhalidov:
scaling curves confirmed
Screenshot 2026-03-02 at 17 27 53

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -111,3 +111,28 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
...
+    Args:
+        size (int): The maximum number of NVDEC decoders to cache per GPU
+            device. Must be non-negative. Setting to 0 disables caching. The
+            default value is 20.
+    """
+    core.set_nvdec_cache_size(size)
+
+
+def get_nvdec_cache_size() -> int:
+    """Get the current maximum number of NVDEC decoders cached per GPU device.
```
NicolasHug: Nit, just to disambiguate with "number of decoders currently in the cache"

```suggestion
    """Get the current maximum number of NVDEC decoders that can be cached per GPU device.
```

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -111,3 +111,28 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
...
+
+    The NVDEC decoder cache stores hardware decoders for reuse, avoiding the
+    overhead of creating new decoders for subsequent video decoding operations on
+    the same GPU. This function controls the maximum number of decoders that can
+    be cached per device.
+
+    Args:
+        size (int): The maximum number of NVDEC decoders to cache per GPU
+            device. Must be non-negative. Setting to 0 disables caching. The
+            default value is 20.
```
NicolasHug: There no default for the `size` parameter, which I think is fine. I'd just suggest to remove the sentence "The
            default value is 20" from here, since this is about the parameter itself, and to add above in the main description something like 

> The cache is first created with a max size of 20.

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -111,3 +111,28 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
...
+    """Set the maximum number of NVDEC decoders cached per GPU device.
+
+    The NVDEC decoder cache stores hardware decoders for reuse, avoiding the
+    overhead of creating new decoders for subsequent video decoding operations on
+    the same GPU. This function controls the maximum number of decoders that can
+    be cached per device.
+
+    Args:
+        size (int): The maximum number of NVDEC decoders to cache per GPU
+            device. Must be non-negative. Setting to 0 disables caching. The
```
NicolasHug: > Setting to 0 disables caching

Could you double check? Based on our current logic 

https://github.com/meta-pytorch/torchcodec/blob/79971409085e472503d107b45dcd0bc2fdde25af/src/torchcodec/_core/NVDECCache.cpp#L58-L69

I think we'll still be adding an entry to the cache even when `getNVDECCacheMaxSize()` is 0.

----


As a separate but related concern, now that we can have `cache_.size()` be 0, it's unclear to me what `cache_.erase(victim);` will do (since there should be no "victim" to evict). I think we can solve both issues with the same fix which is probably to just return early in the function if `getNVDECCacheMaxSize()` is 0.

src/torchcodec/_core/NVDECCache.cpp
```diff
@@ -64,8 +69,25 @@ void NVDECCache::returnDecoder(
...
+      cache_.size() <= static_cast<size_t>(getNVDECCacheMaxSize()),
       "Cache size exceeded maximum limit, please report a bug");
 }
 
+size_t NVDECCache::size() {
+  std::lock_guard<std::mutex> lock(cacheLock_);
+  return cache_.size();
+}
+
+int NVDECCache::getMaxCurrentSizeAcrossDevices() {
```
NicolasHug: I think I'm getting a bit tripped up by the terms "max size" (as in "maximum number of allowed entries") vs "current size" which is "the current number of entries in the cache". They start conflating here in this function where it becomes "Max Current Size". I think it might be worth disambiguate in the naming of the functions. 

For example here I think this function could be named `getMaxNumEntriesAcrossDevices` or something like that?

src/torchcodec/_core/NVDECCacheConfig.cpp
```diff
@@ -0,0 +1,46 @@
...
+  int currentMax = getMaxNVDECCacheCurrentSize();
+  TORCH_CHECK(
+      size >= currentMax,
+      "Cannot set NVDEC cache max size to ",
+      size,
+      " because a device cache currently holds ",
+      currentMax,
+      " entries. Decode those videos first or set a larger max.");
+  g_nvdecCacheMaxSize.store(size);
+}
```
NicolasHug, line 42: On the above, `claude` is telling me this

```
  1. Race between setNVDECCacheMaxSize check and returnDecoder

  setNVDECCacheMaxSize reads getMaxCurrentSizeAcrossDevices(), validates, then stores. Meanwhile returnDecoder could be
   inserting entries concurrently. The sequence:

  1. Thread A: setNVDECCacheMaxSize(5) — reads current max = 4, passes check
  2. Thread B: returnDecoder() — inserts, cache becomes 5, reads old max size (20), no eviction
  3. Thread A: stores g_nvdecCacheMaxSize = 5
  4. Thread B: hits the post-insert STD_TORCH_CHECK(cache_.size() <= getNVDECCacheMaxSize()) — assertion fires with
  size 6 > 5

  This is probably unlikely in practice, but the assertion at NVDECCache.cpp:72 could trip. Worth deciding whether this
   is acceptable or whether you want to document it / remove the assertion / make it a soft check.
```

I haven't dug into it too deep, but it might be worth looking into. I think it has to do with the fact that the new atomic int isn't synced with the `cacheLock_` mutex.

If it simplifies things, I think we can get away with allowing only a single call to `set_cache_size()` at the beginning of the program before any `VideoDecoder` instance gets created. That sounds like an OK tradeoff to me at least at this time (we can always extend the functionality later!)

src/torchcodec/_core/NVDECCache.cpp
vkhalidov: what about the terms `capacity` and `usage`?
I could rename:
`getMaxCurrentSizeAcrossDevices` -> `getMaxUsageAcrossDevices`
`getNVDECCacheMaxSize` -> `getNVDECCacheCapacity`
and rename `set_nvdec_cache_size` and `get_nvdec_cache_size` on Python side accordingly

vkhalidov: switched to the term "capacity" to disambiguate

vkhalidov:
rebased on recent main

vkhalidov:
Reconfirmed the numbers used for the plots above. Consider the PR as ready

PR #1247, opened 2026-02-17
[WIP] Rocm GPU enablement

Cemberk:
Utilizing RocDecode ROCM package to mirror gpu api for accelerated decode behavior

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -145,8 +145,12 @@ class CMakeBuild(build_ext):
                 cmake_args.append(f"-DCMAKE_HIP_COMPILER={rocm_clang}")
             cmake_args.append(f"-DCMAKE_PREFIX_PATH={rocm_path}")
 
-            # Resolve GPU architectures from env or PyTorch's build config.
+            # Resolve GPU architectures: HIP_ARCHITECTURES > PYTORCH_ROCM_ARCH > torch.cuda.get_arch_list()
             hip_archs = os.environ.get("HIP_ARCHITECTURES", "")
+            if not hip_archs:
+                pytorch_rocm_arch = os.environ.get("PYTORCH_ROCM_ARCH", "")
+                if pytorch_rocm_arch:
+                    hip_archs = pytorch_rocm_arch.replace(" ", ";")
             if not hip_archs:
                 try:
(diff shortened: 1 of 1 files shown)


NicolasHug:
Hi @Cemberk!

I realize the PR is still draft, but I thought it might be worth engaging early. Would you be able to share context on this work and the motivation behind it?

Generally, we want to enable third-parties to integrate with TorchCodec and extend it to more devices. We do that by enabling *out-of-core* extensions: the extension lives in a different repo, so that you can have full control over it. The idea is to implement and register your own [DeviceInterface](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/DeviceInterface.h). See a [simple example here](https://github.com/meta-pytorch/torchcodec/tree/main/test/third-party-interface), and a [real-world usage here](https://github.com/dvrogozh/torchcodec-xpu/) for XPU support.

Cemberk:
Hi @NicolasHug!
 
The context is that I have been working on running Huggingface Transformers unit tests on Rocm GPUs in ci I am consistently seeing failures due to torchcodec working by default on Nvidia gpus but defaulting to CPU on Rocm my intention was to enable parity so the transformers code using torchcodec import would behave similarly on both gpus for native support based on the example code I have created a repo here for the plugin does this look correct? https://github.com/Cemberk/torchcodec-rocm

Also in the example it seems like it is designed to first import torchcodec then import torchcodec-xpu since in the rocm case this would mean upstream transformers code need to add some compatibility logic across the repo could we instead add the following to ops.py so plugins once installed would wire in to the same import statement ?

```python
from importlib.metadata import entry_points  
# Auto-discover and load third-party device extensions.
for _ep in entry_points(group="torchcodec.device_extensions"):  
      try:  
          _ep.load()()  
      except Exception:  
          pass
```

NicolasHug:
Thanks for the context!

On plugins: yes, there's https://github.com/meta-pytorch/torchcodec/pull/1151 which we still need to review but it should be doable

AndreasKaratzas:
This will be a nice addition for ROCm vLLM ecosystem too. So far we are using this: https://github.com/vllm-project/vllm/blob/main/tools/install_torchcodec_rocm.sh

But we would like this integrated in upstream torchcodec repo so that we can use prebuilt packages and accelerate installation.

PR #1248, opened 2026-02-17
Add XPU to the list of recognizable devices

dvrogozh:
The 6f10da0 switched to torch::stable::Device and introduced a check of the device by name which broke torchcodec external plugins registration support. This PR fixes that in a straight forward way - adds the XPU in the list. @NicolasHug , you might wish to generalize this check somehow to avoid hardcoding CPU, CUDA, XPU altogether. Please, help to resolve this issue before the next torchcodec v0.11.0 release.

Fixes: 6f10da0 ("Use `torch::stable::Device` (#1231)")


dvrogozh:
Thank you, @NicolasHug.

Issue #1064, opened 2025-11-19
Refactor C++ SingleStreamDecoder to consolidate addStream into constructor


scotts:
Right now, the C++ `SingleStreamDecoder` class has two constructors: https://github.com/meta-pytorch/torchcodec/blob/1ea235aa6b71bc9d6ce070d4b849dc60974082b5/src/torchcodec/_core/SingleStreamDecoder.h#L34-L45

And separate from there, there is a public API for adding streams of different media types: https://github.com/meta-pytorch/torchcodec/blob/1ea235aa6b71bc9d6ce070d4b849dc60974082b5/src/torchcodec/_core/SingleStreamDecoder.h#L90-L97

Note that there is also a private member function that both of those call: https://github.com/meta-pytorch/torchcodec/blob/1ea235aa6b71bc9d6ce070d4b849dc60974082b5/src/torchcodec/_core/SingleStreamDecoder.h#L304-L309

This separation is a relic of when `SingleStreamDecoder` was trying to be more than just a single stream decoder. This API does not match the public Python API, and that fact causes some awkwardness, in particular with how we deal with custom frame mappings; see PR https://github.com/meta-pytorch/torchcodec/pull/1060#discussion_r2541371586 for more.

Note that one could imagine the current separation enables just getting metadata without adding a decoding stream, but that should be well supported with `SeekMode::approximate`.

scotts:
I realized that we can do this in stages:

1. Consolidate all of the core APIs related to video decoder creation and adding a stream. We don't need to actually change the C++ code to do this.
2. Then change the C++ code.

PR #1249, opened 2026-02-19
Consolidate AudioDecoder creation with python wrapper

zy1git:
This PR adds a `create_audio_decoder() `Python wrapper function that consolidates the two-step decoder creation process into a single function call. The wrapper combines `create_decoder()` and `add_audio_stream()` with validation for stream index and stream type.
This is the Python-only phase of the audio decoder consolidation for issue #1064.

### DIFF
--- a/src/torchcodec/decoders/_decoder_utils.py
+++ b/src/torchcodec/decoders/_decoder_utils.py
@@ -7,6 +7,7 @@
 
 import contextvars
 import io
+
 from collections.abc import Generator
 from contextlib import contextmanager
 from pathlib import Path
--- a/src/torchcodec/decoders/_audio_decoder.py
+++ b/src/torchcodec/decoders/_audio_decoder.py
@@ -10,6 +10,7 @@ from pathlib import Path
 
 import torch
 from torch import Tensor
+
 from torchcodec import _core as core, AudioSamples
 from torchcodec._core._decoder_utils import create_audio_decoder
 


src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -55,6 +55,42 @@ def create_decoder(
...
+) -> Tensor:
+    """Create an audio decoder and add an audio stream.
+
+    This is a Python convenience wrapper that calls the existing two-step
+    C++ API (create decoder + add audio stream) in a single Python function.
+    """
+    decoder = create_decoder(source=source, seek_mode=seek_mode)
+
+    if stream_index is not None:
+        container_metadata = core.get_container_metadata(decoder)
```
NicolasHug: Seems like this call is now duplicated within `create_audio_decoder` and also within the `AudioDecoder`. We should make sure not to duplicate logic and calls.

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -55,6 +55,42 @@ def create_decoder(
     )
 
 
+def create_audio_decoder(
```
NicolasHug: Our main goal with https://github.com/meta-pytorch/torchcodec/issues/1064 is to consolidate the core API. That is, the APIs in `torchcodec._core`. This new `create_audio_decoder` is going the in the right direction, but it should be in `_core`, not in `decoders`.

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -55,6 +55,42 @@ def create_decoder(
...
+        if stream_index >= len(container_metadata.streams):
+            raise ValueError(
+                f"The stream at index {stream_index} is not a valid stream."
+            )
+        stream_metadata = container_metadata.streams[stream_index]
+        if not isinstance(stream_metadata, core._metadata.AudioStreamMetadata):
+            raise ValueError(
+                f"The stream at index {stream_index} is not an audio stream. "
+            )
+    core.add_audio_stream(
```
mollyxu: To add onto what Nicolas said, I think we no longer want to expose `core.add_audio_stream` but have something like `core.create_audio_decoder` instead

src/torchcodec/decoders/_decoder_utils.py
zy1git: Fixed in the new commit.

src/torchcodec/decoders/_decoder_utils.py
zy1git: I moved the main logic to the `ops.py` in `_core` but still have a `create_audio_decoder` in this `_decoder_utils.py` file, because if not, in the _audio_decoder.py, we need to write in two steps:
1, self._decoder = create_decoder(...)
2, (...) = core.create_audio_decoder(...)

Or I can also move the [`create_decoder` ](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/decoders/_decoder_utils.py#L24-L55)function logic into [`create_audio_decoder` in `ops.py`](https://github.com/zy1git/torchcodec/blob/90d4a1d698c87ebc5a985f7f0cfc306dec1a1347/src/torchcodec/_core/ops.py#L544-L578), then we can get rid of the `create_audio_decoder` wrapper in `_decoder_utils.py` file completely.

Feel free to give any suggestions.

src/torchcodec/decoders/_decoder_utils.py
zy1git: Thanks for the comment. I pushed a new commit. Could you please take a look whether my new implementation addressed your comment here.

src/torchcodec/_core/ops.py
```diff
@@ -534,3 +539,40 @@ def get_ffmpeg_library_versions():
 @register_fake("torchcodec_ns::_get_backend_details")
 def _get_backend_details_abstract(decoder: torch.Tensor) -> str:
     return ""
+
+
```
NicolasHug: Let's not have that in ops.py. Ideally we want `ops.py` to be exclusively about ops that have a direct equivalent to the C++ ops. Here, we have some higher-level Python code, so it should be somewhere else e.g. in `_core._decoder_utils.py` or something similar.

src/torchcodec/decoders/_decoder_utils.py
NicolasHug: Hmmmmm. Let's just temporarily copy/paste `create_decoder()` (the one from this file) in `_core._decoder_utils.py`.

When you follow-up with the video PR, we'll be able to remove the `create_decoder()` from this file, and only rely on the one from `_core`.

src/torchcodec/decoders/_decoder_utils.py
mollyxu: Thanks for the updates! I agree with Nicolas' comments on temporarily copy/paste `create_decoder()` to `_core._decoder_utils.py`

src/torchcodec/_core/ops.py
zy1git: Addressed in the new commit.

src/torchcodec/decoders/_decoder_utils.py
zy1git: Addressed in the new commit.

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -7,7 +7,6 @@
 
 import contextvars
 import io
-
```
NicolasHug, line 10: make sure to remove unrelated changes before merging

zy1git: Fixed in the new commit.

PR #1250, opened 2026-02-19
Consolidate VideoDecoder creation with python wrapper

zy1git:
This PR introduces a `create_video_decoder() `wrapper function that consolidates the two-step video decoder creation process (create decoder + add video stream) into a single function call. 

This is the Python-only phase of the video decoder consolidation for issue https://github.com/meta-pytorch/torchcodec/issues/1064.

### DIFF
--- a/src/torchcodec/_core/_decoder_utils.py
+++ b/src/torchcodec/_core/_decoder_utils.py
@@ -101,7 +101,7 @@ def create_audio_decoder(
     return (decoder, stream_index, metadata)
 
 
-def _get_and_validate_stream_metadata(
+def _get_and_validate_video_stream_metadata(
     *,
     decoder: Tensor,
     stream_index: int | None = None,
@@ -139,10 +139,7 @@ def _get_and_validate_stream_metadata(
             "The number of frames is unknown. " + _ERROR_REPORTING_INSTRUCTIONS
         )
 
-    return (
-        metadata,
-        stream_index,
-    )
+    return (metadata, stream_index)
 
 
 def create_video_decoder(
@@ -156,32 +153,16 @@ def create_video_decoder(
     device_variant: str = "ffmpeg",
     transforms: Sequence[DecoderTransform | nn.Module] | None = None,
     custom_frame_mappings: tuple[Tensor, Tensor, Tensor] | None = None,
-) -> tuple[Tensor, VideoStreamMetadata, int]:
-    """Create a video decoder and add a video stream.
+) -> tuple[Tensor, int, VideoStreamMetadata]:
 
-    This function consolidates the creation of a decoder and adding a video stream
-    into a single operation, performing all necessary validation.
-
-    Args:
-        source: The source of the video.
-        seek_mode: The seek mode for the decoder.
-        stream_index: The stream index to decode, or None to use the best stream.
-        dimension_order: The dimension order for decoded frames.
-        num_ffmpeg_threads: Number of FFmpeg threads for CPU decoding.
-        device: The device for decoding (as a string).
-        device_variant: The CUDA backend variant to use ("ffmpeg" or "beta").
-        transforms: Optional sequence of transforms to apply.
-        custom_frame_mappings: Optional pre-processed frame mappings data.
-
-    Returns:
-        A tuple of (decoder, metadata, stream_index).
-    """
     decoder = create_decoder(source=source, seek_mode=seek_mode)
 
     (
         metadata,
         stream_index,
-    ) = _get_and_validate_stream_metadata(decoder=decoder, stream_index=stream_index)
+    ) = _get_and_validate_video_stream_metadata(
+        decoder=decoder, stream_index=stream_index
+    )
 
     transform_specs = _make_transform_specs(
         transforms,
@@ -199,8 +180,4 @@ def create_video_decoder(
         custom_frame_mappings=custom_frame_mappings,
     )
 
-    return (
-        decoder,
-        metadata,
-        stream_index,
-    )
+    return (decoder, stream_index, metadata)
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -210,8 +210,8 @@ class VideoDecoder:
             device = str(device)
         (
             self._decoder,
-            self.metadata,
             self.stream_index,
+            self.metadata,
         ) = create_video_decoder(
             source=source,
             seek_mode=seek_mode,


src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +161,136 @@ def create_audio_decoder(
...
+        end_stream_seconds,
+        num_frames,
+    ) = _get_and_validate_stream_metadata(decoder=decoder, stream_index=stream_index)
+
+    if device is None:
+        device = str(torch.get_default_device())
+    elif isinstance(device, torch_device):
+        device = str(device)
+
+    device_variant = _get_cuda_backend()
```
NicolasHug: Ideally we'd keep `_get_cuda_backend` and `set_cuda_backend()` where they were, i.e. outside of `_core` in `torchcodec.decoders._decoder_utils.py`.  
I think one way to achieve that is to let this new `create_video_decoder` accept both `device` and `device_variant` as parameter.

zy1git: Thanks for the comment. I addressed it in the new commit.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -219,29 +203,31 @@ def __init__(
...
+            source=source,
+            seek_mode=seek_mode,
+            stream_index=stream_index,
             dimension_order=dimension_order,
-            num_threads=num_ffmpeg_threads,
+            num_ffmpeg_threads=num_ffmpeg_threads,
             device=device,
-            device_variant=device_variant,
-            transform_specs=transform_specs,
+            device_variant=_get_cuda_backend(),
```
NicolasHug: This `device_variant` is now both a parameter and a returned value, I think we only need it to be a parameter to `create_video_decoder()`

zy1git: Thanks a lot for the comment. I have fixed it in the new commit.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +100,136 @@ def create_audio_decoder(
...
+    """Get and validate video stream metadata from a decoder.
+
+    Args:
+        decoder: The decoder tensor.
+        stream_index: The stream index to use, or None to use the best stream.
+
+    Returns:
+        A tuple of (metadata, stream_index, begin_stream_seconds,
+        end_stream_seconds, num_frames).
+    """
```
NicolasHug: This helper didn't have docs, I feel like it's not really necessary to add them. The interface is pretty straightforward, this isn't user-facing, and any modification to the underlying logic would mean we'd have to keep the docs up to date consequently. So I'd suggest to just remove the docstrings and keep it they it was.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +100,136 @@ def create_audio_decoder(
...
+        stream_index,
+        begin_stream_seconds,
+        end_stream_seconds,
+        num_frames,
+    ) = _get_and_validate_stream_metadata(decoder=decoder, stream_index=stream_index)
+
+    if device is None:
+        device = str(torch.get_default_device())
+    elif isinstance(device, torch_device):
+        device = str(device)
```
NicolasHug: This is now done twice, here and in `__init__`. Let's just do it once and let this function accept the string as input.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +100,136 @@ def create_audio_decoder(
...
+            "The number of frames is unknown. " + _ERROR_REPORTING_INSTRUCTIONS
+        )
+    num_frames = metadata.num_frames
+
+    return (
+        metadata,
+        stream_index,
+        begin_stream_seconds,
+        end_stream_seconds,
+        num_frames,
```
NicolasHug: I realize this was already the case in `main`, but let's use this PR as opportunity to clean this up too: all of begin_stream_seconds, begin_stream_seconds and num_frames are just fields of `metadata`, so let's not return them. We should just use the info from `metadata` whenever relevant.

We can keep the existing fields like `self._begin_stream_seconds` on the VideoDecoder` however.

src/torchcodec/_core/_decoder_utils.py
zy1git: Removed in the new commit.

src/torchcodec/_core/_decoder_utils.py
zy1git: Done in the new commit.

src/torchcodec/_core/_decoder_utils.py
zy1git: Done in the new commit.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +99,108 @@ def create_audio_decoder(
...
+        dimension_order: The dimension order for decoded frames.
+        num_ffmpeg_threads: Number of FFmpeg threads for CPU decoding.
+        device: The device for decoding (as a string).
+        device_variant: The CUDA backend variant to use ("ffmpeg" or "beta").
+        transforms: Optional sequence of transforms to apply.
+        custom_frame_mappings: Optional pre-processed frame mappings data.
+
+    Returns:
+        A tuple of (decoder, metadata, stream_index).
+    """
```
NicolasHug: Let's not add a docstring here. Same reason as in https://github.com/meta-pytorch/torchcodec/pull/1250#discussion_r2863882187

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +99,108 @@ def create_audio_decoder(
...
+        device_variant=device_variant,
+        transform_specs=transform_specs,
+        custom_frame_mappings=custom_frame_mappings,
+    )
+
+    return (
+        decoder,
+        metadata,
+        stream_index,
+    )
```
NicolasHug: Let's keep the return order consistent with `create_audio_decoder` which is returning `decoder, stream_index, metadata`.

NicolasHug: Also, as Nit, here and in other places, just make the return fit in one line instead of 3 if possible.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -91,3 +99,108 @@ def create_audio_decoder(
     )
 
     return (decoder, stream_index, metadata)
+
+
+def _get_and_validate_stream_metadata(
```
NicolasHug: Let's name this `_get_and_validate_video_stream_metadata` since this is video only now.

src/torchcodec/_core/_decoder_utils.py
zy1git: done in the new commit.

src/torchcodec/_core/_decoder_utils.py
zy1git: done in the new commit

src/torchcodec/_core/_decoder_utils.py
zy1git: done in the new commit

PR #1252, opened 2026-02-20
Make custom_ops ABI stable

NicolasHug:


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -377,31 +355,29 @@ std::vector<Transform*> makeTransforms(const std::string& transformSpecsRaw) {
...
 
 // Create a SingleStreamDecoder from file and wrap the pointer in a tensor.
-torch::Tensor create_from_file(
-    std::string_view filename,
-    std::optional<std::string_view> seek_mode = std::nullopt) {
-  std::string filenameStr(filename);
-
+torch::stable::Tensor create_from_file(
+    std::string filename,
+    std::optional<std::string> seek_mode = std::nullopt) {
```
NicolasHug, line 360: The stable ABI wants string objects, not string_view. It won't matter for perf, all our strings are super small, so copies are dead cheap.

We've been using both types inconsistently throughout the codebase anyway. This is orthogonal to the porting effort but I'll open an issue to make sure we use string and string_view in ways that make sense, and that are consistent.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -50,9 +45,9 @@ TORCH_LIBRARY(torchcodec_ns, m) {
   m.def(
       "_create_from_file_like(int file_like_context, str? seek_mode=None) -> Tensor");
   m.def(
-      "_add_video_stream(Tensor(a!) decoder, *, int? num_threads=None, str? dimension_order=None, int? stream_index=None, str device=\"cpu\", str device_variant=\"ffmpeg\", str transform_specs=\"\", (Tensor, Tensor, Tensor)? custom_frame_mappings=None, str? color_conversion_library=None) -> ()");
+      "_add_video_stream(Tensor(a!) decoder, *, int? num_threads=None, str? dimension_order=None, int? stream_index=None, str device=\"cpu\", str device_variant=\"ffmpeg\", str transform_specs=\"\", Tensor? custom_frame_mappings_pts=None, Tensor? custom_frame_mappings_duration=None, Tensor? custom_frame_mappings_keyframe_indices=None, str? color_conversion_library=None) -> ()");
```
NicolasHug, line 48: tuples aren't supported by the new stable registration mechanism, so we have to unpack the 3-tuple of tensors into 3 tensors.

src/torchcodec/_core/ops.py
```diff
@@ -68,8 +68,42 @@ def expose_ffmpeg_dlls():  # noqa: F811
 _create_from_file_like = torch._dynamo.disallow_in_graph(
     torch.ops.torchcodec_ns._create_from_file_like.default
 )
-add_video_stream = torch.ops.torchcodec_ns.add_video_stream.default
+_add_video_stream_raw = torch.ops.torchcodec_ns.add_video_stream.default
 _add_video_stream = torch.ops.torchcodec_ns._add_video_stream.default
+
+
```
NicolasHug, line 74: This new function is a consequence of the key frame mapping unpacking.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -85,39 +80,32 @@ TORCH_LIBRARY(torchcodec_ns, m) {
...
-      decoder, {sizeof(SingleStreamDecoder*)}, deleter, {at::kLong});
+  int64_t sizes[] = {static_cast<int64_t>(sizeof(SingleStreamDecoder*))};
+  int64_t strides[] = {1};
+  torch::stable::Tensor tensor = torch::stable::from_blob(
+      decoder,
+      {sizes, 1},
+      {strides, 1},
+      StableDevice(kStableCPU),
+      kStableInt64,
+      &decoderDeleter);
```
NicolasHug, line 101: The deleter is safe and doesn't leak:

```py
import resource

import gc


for _ in range(100000):
    dec = VideoDecoder("test/resources/testsrc2_h265.mp4")
    del dec
    gc.collect()

    max_rss_kb = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
    print(f"Peak memory usage: {max_rss_kb / 1024:.1f} MB")
```

In this PR:

```
~/dev/torchcodec-cuda (stable_custom_ops*) » python lol.py                 130 ↵ nicolashug@nicolashug-fedora-PW0H326Y
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB
Peak memory usage: 639.7 MB

```

commenting-out the deleter:

```
~/dev/torchcodec-cuda (stable_custom_ops*) » python lol.py                 130 ↵ nicolashug@nicolashug-fedora-PW0H326Y
Peak memory usage: 640.1 MB
Peak memory usage: 642.3 MB
Peak memory usage: 644.6 MB
Peak memory usage: 646.9 MB
Peak memory usage: 649.3 MB
Peak memory usage: 651.6 MB
Peak memory usage: 654.0 MB
Peak memory usage: 656.3 MB
Peak memory usage: 658.7 MB
Peak memory usage: 661.0 MB
Peak memory usage: 663.4 MB
Peak memory usage: 665.7 MB
Peak memory usage: 667.9 MB
```

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -476,17 +457,27 @@ void _add_video_stream(
...
-      custom_frame_mappings.has_value()
-      ? std::make_optional(makeFrameMappings(custom_frame_mappings.value()))
+  bool hasPts = custom_frame_mappings_pts.has_value();
+  bool hasDuration = custom_frame_mappings_duration.has_value();
+  bool hasKeyframeIndices = custom_frame_mappings_keyframe_indices.has_value();
+  STD_TORCH_CHECK(
+      (hasPts == hasDuration) && (hasDuration == hasKeyframeIndices),
+      "custom_frame_mappings_pts, custom_frame_mappings_duration, and "
+      "custom_frame_mappings_keyframe_indices must all be provided or all be "
+      "None. This is a bug in TorchCodec, please report it.");
```
NicolasHug, line 474: On the above: we have 3 optional Tensors now, instead of having 1 Optional 3-tuple of tensors.
In theory, one could now pass a non-null duration and a null pts for frame mappings, which makes no sense. The previous interface prevented that, the new one cannot. 

This `STD_TORCH_CHECK` exist to catch at runtime what could have previously been caught statically (and during compilation).

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -85,39 +80,32 @@ TORCH_LIBRARY(torchcodec_ns, m) {
...
-// migrated as a follow-up.  For now, we can work around this by converting
-// to/from torch::stable::Tensor, which is cheap. Those conversion helpers will
-// eventually be removed.
-inline torch::stable::Tensor toStableTensor(const at::Tensor& tensor) {
-  at::Tensor* p = new at::Tensor(tensor);
-  return torch::stable::Tensor(reinterpret_cast<AtenTensorHandle>(p));
+// TODO_STABLE_ABI: use previous deleter pattern with a lambda, once
+// https://github.com/pytorch/pytorch/pull/175089 is available.
+void decoderDeleter(void* data) {
+  delete static_cast<SingleStreamDecoder*>(data);
```
Dan-Flores, line 86: Since there is no memory leak associated with this deleter, would we migrate to using a lambda pattern for consistency?

NicolasHug: The main reason is that this new `decoderDeleter` assumes that `data` *is* the decoder, i.e. the data that was passed to `from_blob()`. That's a guarantee from `from_blob()`, but using the lambda would allow us to not even make that assumption. Honestly, this is a very nit-picky todo!

Issue #1253, opened 2026-02-20
Expose FFmpeg thread count on `AudioDecoder`


HaukurPall:
### 🚀 The feature

Expose a `num_ffmpeg_threads` parameter on `AudioDecoder`, mirroring the `num_threads` parameter that video streams already have, and default it to `1`:

```python
AudioDecoder(
    source,
    *,
    stream_index=None,
    sample_rate=None,
    num_channels=None,
    num_ffmpeg_threads=1,  # new parameter, default 1
)
```

Currently, the video path in [`custom_ops.cpp`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/custom_ops.cpp) exposes `num_threads` on `add_video_stream` and forwards it to `VideoStreamOptions.ffmpegThreadCount`. The audio path (`add_audio_stream`) has no such parameter. In [`SingleStreamDecoder.cpp`](https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/SingleStreamDecoder.cpp), `addAudioStream` calls `addStream` without a thread count, so it defaults to `std::nullopt`, which resolves to `thread_count=0` (FFmpeg auto-detect):

```cpp
streamInfo.codecContext->thread_count = ffmpegThreadCount.value_or(0);
```

FFmpeg interprets `thread_count=0` as "use as many threads as there are CPU cores".

### Motivation, pitch

When multiple `AudioDecoder` objects exist concurrently (common in data-loading pipelines), each one spawns a full FFmpeg thread pool sized to the CPU count. This quickly hits the OS thread limit, crashing with `[Errno 11] Resource temporarily unavailable`.

This is particularly problematic in containerized environments (Docker, Kubernetes) where the OS reports the **host machine's hardware CPU count** rather than the allocated vCPUs, causing far more threads to be spawned than expected.

**Reproduction:**

```python
import io
import os

import numpy as np
import psutil
import soundfile as sf
from torchcodec.decoders import AudioDecoder


def count_threads():
    return psutil.Process(os.getpid()).num_threads()


# Generate a small FLAC file in memory
buf = io.BytesIO()
sf.write(buf, np.zeros(16000, dtype=np.float32), 16000, format="FLAC")
flac_bytes = buf.getvalue()

print(f"Before: {count_threads()} threads")

# Simulate a data pipeline holding multiple decoders concurrently
decoders = []
for i in range(100):
    d = AudioDecoder(flac_bytes)
[...]

PR #1254, opened 2026-02-20
Set num_ffmpeg_threads to 1 on AudioDecoder

mollyxu:
Set `num_ffmpeg_threads` parameter to 1 on AudioDecoder to avoid oversubscription

Related issue: #1253

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -615,6 +615,9 @@ void SingleStreamDecoder::addAudioStream(
         *audioStreamOptions.numChannels);
   }
 
+  // We hardcode ffmpegThreadCount=1 for audio, see
+  // https://github.com/pytorch/torchcodec/issues/1253 and
+  // https://github.com/pytorch/torchcodec/pull/1254
   addStream(
       streamIndex, AVMEDIA_TYPE_AUDIO, StableDevice(kStableCPU), "ffmpeg", 1);
 


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -612,7 +615,12 @@ void SingleStreamDecoder::addAudioStream(
         *audioStreamOptions.numChannels);
   }
 
-  addStream(streamIndex, AVMEDIA_TYPE_AUDIO);
+  addStream(
+      streamIndex,
+      AVMEDIA_TYPE_AUDIO,
+      StableDevice(kStableCPU),
```
mollyxu: Adding default `device` and `deviceVariant` here to pass in `ffmpegThreadCount`

NicolasHug (commented):
Thanks for the PR @mollyxu . 

I wasn't  aware that `AVCodecContext.threadCount` was used by some audio decoders, but it turns out it is used by [FLAC](https://github.com/FFmpeg/FFmpeg/blob/fba9fc0c6b349348a2374c2619ed71e2be5ca85d/libavcodec/flacdec.c#L829), [TAK](https://github.com/FFmpeg/FFmpeg/blob/fba9fc0c6b349348a2374c2619ed71e2be5ca85d/libavcodec/takdec.c#L955C25-L955C41) and [wavpack](https://github.com/FFmpeg/FFmpeg/blob/fba9fc0c6b349348a2374c2619ed71e2be5ca85d/libavcodec/wavpack.c#L1719C42-L1720).

I definitely think we should set it to 1 to avoid the oversubscriptions mentioned by @HaukurPall in https://github.com/meta-pytorch/torchcodec/issues/1253.

But I'm wondering whether we need to expose it as a parameter to `AudioDecoder`?

For the `VideoDecoder`, `num_ffmpeg_thread` is relevant for almost all codecs. It still has this inconvenient characteristic that it's a backend-specific parameter (which we try to avoid!), but at least it's relevant for all codecs.
However in `AudioDecoder`, `num_ffmpeg_thread` would be both backend-specific *and* codec-specific (pretty niche codecs as well, except maybe for FLAC).

In addition, my understanding of `num_ffmpeg_threads` for FLAC is that it would allocate each thread to decode individual portions of the file - which is basically what can be achieved in Python [as e.g. in our tutorial](https://meta-pytorch.org/torchcodec/stable/generated_examples/decoding/parallel_decoding.html#method-4-joblib-multithreading).

So, I'm tempted not to expose the parameter right now, since it wouldn't be relevant for the vast majority of users. I suggested we just set it to 1, as a bugfix. If really there is demand to expose it for specific use-case, we can reconsider and figure out what would be a good way to expose it.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -612,7 +615,8 @@ void SingleStreamDecoder::addAudioStream(
         *audioStreamOptions.numChannels);
   }
 
-  addStream(streamIndex, AVMEDIA_TYPE_AUDIO);
+  addStream(
+      streamIndex, AVMEDIA_TYPE_AUDIO, StableDevice(kStableCPU), "ffmpeg", 1);
```
NicolasHug, line 622: Might be worth adding a link to both https://github.com/meta-pytorch/torchcodec/issues/1253 and https://github.com/meta-pytorch/torchcodec/pull/1254#pullrequestreview-3842783497 to share context about why we're passing 1 here.

PR #1255, opened 2026-02-23
Beta CUDA interface minor cleanups

NicolasHug:
Two minor cleanups, see comments below

### DIFF
--- a/src/torchcodec/_core/NVDECCache.cpp
+++ b/src/torchcodec/_core/NVDECCache.cpp
@@ -39,12 +39,10 @@ UniqueCUvideodecoder NVDECCache::getDecoder(CUVIDEOFORMAT* videoFormat) {
   return nullptr;
 }
 
-bool NVDECCache::returnDecoder(
+void NVDECCache::returnDecoder(
     CUVIDEOFORMAT* videoFormat,
     UniqueCUvideodecoder decoder) {
-  if (!decoder) {
-    return false;
-  }
+  STD_TORCH_CHECK(decoder != nullptr, "decoder must not be null");
 
   CacheKey key(videoFormat);
   std::lock_guard<std::mutex> lock(cacheLock_);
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/NVDECCache.cpp
```diff
@@ -39,12 +39,10 @@ UniqueCUvideodecoder NVDECCache::getDecoder(CUVIDEOFORMAT* videoFormat) {
   return nullptr;
 }
 
-bool NVDECCache::returnDecoder(
+void NVDECCache::returnDecoder(
```
NicolasHug, line 42: We don't return `bool` anymore. The `bool` was supposed to indicate whether we were able to properly return a decoder to the cache, but now that we have the LRU eviction mechanism, we *always* return a decoder to the cache.

Also added a validation for decoder being non-null

src/torchcodec/_core/NVDECCache.h
```diff
@@ -70,13 +69,15 @@ class NVDECCache {
...
-          numDecodeSurfaces(videoFormat->min_num_decode_surfaces) {}
+    explicit CacheKey(CUVIDEOFORMAT* videoFormat) {
+      STD_TORCH_CHECK(videoFormat != nullptr, "videoFormat must not be null");
+      codecType = videoFormat->codec;
+      width = videoFormat->coded_width;
+      height = videoFormat->coded_height;
+      chromaFormat = videoFormat->chroma_format;
+      bitDepthLumaMinus8 = videoFormat->bit_depth_luma_minus8;
+      numDecodeSurfaces = videoFormat->min_num_decode_surfaces;
+    }
```
NicolasHug, line 80: Above: added explicit `STD_TORCH_CHECK`. We were still safe before because this check was enforced earlier in any caller of the `CacheKey` constructor, but it's always safer to validate where things are used.

mollyxu (approved):

PR #1258, opened 2026-02-23
Add `py.typed`

Samoed:
Add `py.typed` to make library checkable by mypy


Samoed:
Hi @Dan-Flores. Can you merge this, please? This would help to do proper typechecking in other libraries

NicolasHug (approved):
For my own ref, pytorch has this file too https://github.com/pytorch/pytorch/blob/main/torch/py.typed

We've been pretty good at consistently typing the public APIs in torchcodec, so I think this is a surface that we can expose.

Samoed:
Yes, I looked a lot at your code and I think typing is great!

Issue #1223, opened 2026-02-05
torchcodec decode extremely slow at large mkv file(42x slower than mp4)


MilkClouds:
### 🐛 Describe the bug

On following code, decoding frames on mkv is 42x slower than mp4. It occurs when **start time>>0.** (work normally at start_time=0)

```
Benchmarking 5.0s window @ 10.0Hz
Start: 250.0s
MKV: /mnt/testfs/example-mcaps/woozy-ruby-ostrich-f153ac423f61-20220224-172750.mkv
MP4: /mnt/testfs/example-mcaps/woozy-ruby-ostrich-f153ac423f61-20220224-172750.mp4
------------------------------------------------------------
MKV: 5.8444s (50 frames)
MP4: 0.1376s (50 frames)
------------------------------------------------------------
MKV/MP4 ratio: 42.47x
```

Following is code I used for experiment. Full video is 300s length.

```py
#!/usr/bin/env python3
"""Benchmark script comparing torchcodec read times for MKV vs MP4 at 10Hz, 5-second window."""

import time
from pathlib import Path

from torchcodec.decoders import VideoDecoder

# Video paths
BASE_NAME = "woozy-ruby-ostrich-f153ac423f61-20220224-172750"
VIDEO_DIR = Path("/mnt/testfs/example-mcaps")
MKV_PATH = VIDEO_DIR / f"{BASE_NAME}.mkv"
MP4_PATH = VIDEO_DIR / f"{BASE_NAME}.mp4"

# Sampling parameters
FPS = 10.0  # 10Hz
WINDOW_SECONDS = 5.0  # 5-second window
START_SECONDS = 250.0  # start at 250s


def benchmark_read(video_path: Path, start_seconds: float = START_SECONDS) -> tuple[float, int]:
    """Read a 5-second window at 10Hz and return (elapsed_time, num_frames)."""
    decoder = VideoDecoder(video_path, seek_mode="approximate")

    # Generate timestamps at 10Hz for 5 seconds: 0.0, 0.1, 0.2, ..., 4.9
    num_frames = int(WINDOW_SECONDS * FPS)
    timestamps = [start_seconds + i / FPS for i in range(num_frames)]

    t0 = time.perf_counter()
    frames = decoder.get_frames_played_at(seconds=timestamps)
    elapsed = time.perf_counter() - t0

    return elapsed, frames.data.shape[0]


def main():
[...]

PR #1259, opened 2026-02-24
Detect missing index in mkv + approximate seek mode

Dan-Flores:
This PR resolves https://github.com/meta-pytorch/torchcodec/issues/1223.

#### The issue:
The issue occurred because the `mkv` container format lazily creates the "cues" or index we use in `canWeAvoidSeeking`. Without that index, we do not seek forward when we should!

#### Details:

The reproducing script starts at `START_SECONDS = 250.0`, so torchcodec should seek to the correct frame in the video before decoding.

* While decoding the `mp4`, we correctly seek forward because `av_index_search_timestamp` indicates that the desired `pts` to decode has a different keyframe than `lastDecodedAvFramePts_` (which is initialized to 0).

* While decoding the `mkv`, we do not seek forward because `av_index_search_timestamp` indicates that the desired `pts` to decode **has the same keyframe as `lastDecodedAvFramePts_`** (which is initialized to 0). So we sequentially seek from 0 -> the target pts.
  * `mkv` files can defer the creation of this index until it is required during decoding. This means we will skip the initial seek and instead read all the frames from the start until the target pts.
  * this is evident by logging `streamInfo.stream->nb_index_entries` during decoding. As an additional source, the [matroska decoder source code](https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/matroskadec.c#L425-L426) repeatedly indicates that parsing 'cues' / building the index is deferred until needed.

#### Testing:
Using the benchmark I was able to verify the fix improves performance:
```
Benchmarking 5.0s window @ 10.0Hz
Start: 250.0s
MKV: /home/dev/torchcodec/woozy-ruby-ostrich-25d10fc097fe-20220411-050631.mkv
MP4: /home/dev/torchcodec/woozy-ruby-ostrich-25d10fc097fe-20220411-050631.mp4
------------------------------------------------------------

====== benchmarking MKV ====== 
MKV: 0.1260s (50 frames)

====== benchmarking Mp4 ====== 
MP4: 0.1168s (50 frames)
------------------------------------------------------------
MKV/MP4 ratio: 1.08x
```

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.h
+++ b/src/torchcodec/_core/SingleStreamDecoder.h
@@ -358,13 +358,14 @@ class FORCE_PUBLIC_VISIBILITY SingleStreamDecoder {
   int64_t cursor_ = INT64_MIN;
   bool cursorWasJustSet_ = false;
   // Initialized to INT64_MIN instead of 0. With 0, canWeAvoidSeeking() could
-  // incorrectly skip a seek when the index had not yet been built, as some
-  // formats (mkv, webm) delay building the index until the first seek.
+  // incorrectly skip a seek when the internal FFmpeg frame index (used by
+  // av_index_search_timestamp() in approximate mode) had not yet been built,
+  // as some formats (mkv, webm) delay building it until the first seek.
   // With INT64_MIN, we always seek when retrieving the first frame. This
   // means we correctly seek when the first requested frame is far into the
   // video, at the cost of an unnecessary (likely cheap) seek when the first
   // requested frame is near the start.
-  // See: https://github.com/meta-pytorch/torchcodec/issues/1223
+  // See: https://github.com/meta-pytorch/torchcodec/pull/1259
   int64_t lastDecodedAvFramePts_ = INT64_MIN;
   int64_t lastDecodedAvFrameDuration_ = 0;
   int64_t lastDecodedFrameIndex_ = INT64_MIN;


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1205,6 +1205,23 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
...
+  // After decoding a frame, the index will be built, so this condition will
+  // only be hit once and only applies to approximate seek_mode.
+  if (formatContext_->iformat && formatContext_->iformat->name &&
+      strcmp(formatContext_->iformat->name, "matroska,webm") == 0) {
+    // FFmpeg does not expose any flags to check if the index has been built.
+    // As a heuristic, checking if there is only 1 index entry works.
+    if (getStreamIndexEntryCount(streamInfo.stream) <= 1) {
+      return false;
+    }
+  }
```
NicolasHug: Thanks a lot for the investigations @Dan-Flores !

IIUC, here we *force* a seek, because not seeking we cause us to decode frames sequentially up until the target frame, which is slow.

I think the new logic here will address the problem for the very first frame the user requests. However, once that first frame has been returned, we'd still potentially have the same problem for any future frame because at that point, `getStreamIndexEntryCount(streamInfo.stream)` would be > 1 and we wouln't be able to avoid the long sequencial decoding? Basically:

```py
dec = VideoDecoder(...)

dec.get_frame_at(500)  # getStreamIndexEntryCount(streamInfo.stream) <= 1 so we seek and avoid a long decoding sequence

# By now, the lazy index has been built up until frame 500 (but not futher)

dec.get_frame_at(1000)  # This is still potentially super slow because our logic won't realize we should seek
```

Dan-Flores: My understanding is that the full index gets built once triggered, so after we force a seek to frame 250, the index gets built, then decoding frame 299 would trigger our seek logic as normal. 

I checked this locally by printing the decoded `pts` and the `nb_index_entries`:
* The decoded `pts` starts near the second target `pts`, indicating we found the correct keyframe.
* The `nb_index_entries` remains static after decoding frame 250, no new index entries are added when decoding later frames.

Dan-Flores: @NicolasHug The `mkv` [function that parses cues](https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/matroskadec.c#L2022-L2039) also does not take in a timestamp, it builds the index for the whole file.

mollyxu (approved):
Thanks for the PR! These changes LGTM. I'll let @NicolasHug take another pass

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1205,6 +1205,23 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
...
     return false;
   }
+  // Some mkv / webm files lazily create "cues" or index entries.
+  //   https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/matroskadec.c#L425-L426
+  // As a result, getKeyFrameIdentifier which relies on
+  // av_index_search_timestamp can return the same keyframe for any two
+  // timestamps, even one very far from the target timestamp. In that case, we
+  // sequentially decode to the target timestamp when we should seek instead.
+  // For example, see: https://github.com/meta-pytorch/torchcodec/issues/1223
+  // After decoding a frame, the index will be built, so this condition will
```
NicolasHug: Note: my understanding after asking `claude` to run me through the mkv demuxer code is that the index is built upon the first *seek*, but not upon frame decoding.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1205,6 +1205,23 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
...
+  // only be hit once and only applies to approximate seek_mode.
+  if (formatContext_->iformat && formatContext_->iformat->name &&
+      strcmp(formatContext_->iformat->name, "matroska,webm") == 0) {
+    // FFmpeg does not expose any flags to check if the index has been built.
+    // As a heuristic, checking if there is only 1 index entry works.
+    if (getStreamIndexEntryCount(streamInfo.stream) <= 1) {
+      return false;
+    }
```
NicolasHug: I just realized that this `getStreamIndexEntryCount(streamInfo.stream) <= 1` could be problematic for mkv files that have only one key frame (at frame 0):

```py
dec = VideoDecoder("video_with_only_one_keyframe.mkv")
dec[250]  # seeks to frame 0 - that's fine, we're already there
dec[253]  # The new logic forces a seek back to fame 0  !!!! We should just decode forward
```

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: Separately: if we decide to go for this fix, we should also check for the seekmode. My understanding is that we're only hitting this issue in approximate mode.

Dan-Flores:
Thanks for the suggestion @NicolasHug! I was able to implement your suggested fix as described, and saw `mkv` performance improve:
```
====== benchmarking MKV ====== 
MKV: 0.0464s (2 frames)

====== benchmarking Mp4 ====== 
MP4: 0.0425s (2 frames)
------------------------------------------------------------
MKV/MP4 ratio: 1.09x
```

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: Yes, that is correct:
```
    /* Parse the CUES now since we need the index data to seek. */
    if (matroska->cues_parsing_deferred > 0) {
        matroska->cues_parsing_deferred = 0;
        matroska_parse_cues(matroska);
    }
```
From https://github.com/FFmpeg/FFmpeg/blob/release/8.0/libavformat/matroskadec.c#L4328-L4332

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -357,7 +357,15 @@ class FORCE_PUBLIC_VISIBILITY SingleStreamDecoder {
...
   bool cursorWasJustSet_ = false;
-  int64_t lastDecodedAvFramePts_ = 0;
+  // Initialized to INT64_MIN instead of 0. With 0, canWeAvoidSeeking() could
+  // incorrectly skip a seek when the index had not yet been built, as some
+  // formats (mkv, webm) delay building the index until the first seek.
+  // With INT64_MIN, we always seek when retrieving the first frame. This
+  // means we correctly seek when the first requested frame is far into the
+  // video, at the cost of an unnecessary (likely cheap) seek when the first
+  // requested frame is near the start.
+  // See: https://github.com/meta-pytorch/torchcodec/issues/1223
```
NicolasHug: Let link to https://github.com/meta-pytorch/torchcodec/pull/1259 instead (this PR), I think it'll be more useful to provide context to future us. There are more info about what's going on, especially in your (great!) PR description.

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -357,7 +357,15 @@ class FORCE_PUBLIC_VISIBILITY SingleStreamDecoder {
...
+  int64_t lastDecodedAvFramePts_ = INT64_MIN;
```
NicolasHug, line 369: `INT64_MIN` makes sense but out of curiosity, did you try `AV_NOPTS_VALUE`?

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -357,7 +357,15 @@ class FORCE_PUBLIC_VISIBILITY SingleStreamDecoder {
   // pts to the user when they request a frame.
   int64_t cursor_ = INT64_MIN;
   bool cursorWasJustSet_ = false;
-  int64_t lastDecodedAvFramePts_ = 0;
+  // Initialized to INT64_MIN instead of 0. With 0, canWeAvoidSeeking() could
+  // incorrectly skip a seek when the index had not yet been built, as some
+  // formats (mkv, webm) delay building the index until the first seek.
```
NicolasHug: Nit: it's not clear what "index" refers to here. We could interpret that as our `keyFrames` and `allFrames` vecs of `FrameInfo`, but it's not that. I'm not sure how to best describe this, but we should try to disambiguate.

What we mean by "index" here is "the ffmpeg-internal container-specific frame index that we rely on in approximate mode when we call `av_index_search_timestamp`" but that's a bit long.

It's also worth clarifying that all this only happens in approximate mode.

src/torchcodec/_core/SingleStreamDecoder.h
Dan-Flores: I did not, but it seems that `AV_NOPTS_VALUE` [is defined ](https://ffmpeg.org/doxygen/5.1/group__lavu__time.html#ga2eaefe702f95f619ea6f2d08afa01be1) to be equal to `INT64_MIN`.

PR #1260, opened 2026-02-24
Finalize stable ABI migration

NicolasHug:
This should be the last PR we need to make TorchCodec ABI stable. We enforce ABI stability by defining `TORCH_TARGET_VERSION` which targets 2.11.

Yes, 2.11! Not 2.12, as originally assumed. It turns out that https://github.com/pytorch/pytorch/pull/175089, which I thought was missing from 2.11, was actually cherry-picked for 2.11 in https://github.com/pytorch/pytorch/pull/176440.

So effectively, TorchCodec 0.12 will be ABI stable from torch 2.11. I have verified that this works by building and testing this PR against the 2.11 wheels in https://github.com/meta-pytorch/torchcodec/pull/1328 (the usual jobs are failing, but we're building and testing fine otherwise).

We can merge this PR now. We shouldn't cherry-pick it though (not worth it). It'll be out with Torchcodec 0.12.

See comments below for other details.

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -57,8 +57,8 @@ function(make_torchcodec_sublibrary
     # torch, we want to rely on the "header-only" mode of these headers and hit
     # all the `#ifdef FMT_HEADER_ONLY` paths, so we define FMT_HEADER_ONLY here.```
     target_compile_definitions(${library_name} PRIVATE FMT_HEADER_ONLY)
-    # TC is ABI-stable w.r.t. torch since torch 2.12
-    target_compile_definitions(${library_name} PRIVATE TORCH_TARGET_VERSION=0x020c000000000000)
+    # TC is ABI-stable w.r.t. torch since torch 2.11
+    target_compile_definitions(${library_name} PRIVATE TORCH_TARGET_VERSION=0x020b000000000000)
 
     # Avoid adding the "lib" prefix which we already add explicitly.
     set_target_properties(${library_name} PROPERTIES PREFIX "")


src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -126,22 +125,17 @@ torch::stable::Tensor rgbAVFrameToTensor(const UniqueAVFrame& avFrame) {
...
-  // will properly be passed down to the torch::stable::Tensor.
-  at::Tensor* p = new at::Tensor(std::move(tensor));
-  return torch::stable::Tensor(reinterpret_cast<AtenTensorHandle>(p));
+  return torch::stable::from_blob(
+      avFrameClone->data[0],
+      shape,
+      strides,
+      StableDevice(kStableCPU),
+      kStableUInt8,
+      deleter);
```
NicolasHug, line 138: Can confirm the above is correct and leak-free by running the exact same experiment as in https://github.com/meta-pytorch/torchcodec/pull/1242/changes#r2822852062

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -98,7 +99,7 @@ torch::stable::Tensor wrapDecoderPointerToTensor(
       {strides, 1},
       StableDevice(kStableCPU),
       kStableInt64,
-      &decoderDeleter);
+      deleter);
```
NicolasHug, line 120: Can confirm this is safe and leak-free by running the same experiment as in https://github.com/meta-pytorch/torchcodec/pull/1252/changes#r2833300819

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -57,6 +57,8 @@ function(make_torchcodec_sublibrary
     # torch, we want to rely on the "header-only" mode of these headers and hit
     # all the `#ifdef FMT_HEADER_ONLY` paths, so we define FMT_HEADER_ONLY here.```
     target_compile_definitions(${library_name} PRIVATE FMT_HEADER_ONLY)
+    # TC is ABI-stable w.r.t. torch since torch 2.12
+    target_compile_definitions(${library_name} PRIVATE TORCH_TARGET_VERSION=0x020c000000000000)
```
NicolasHug: See https://github.com/pytorch/pytorch/blob/f4de0c9271ec35094b1d10d8361d062da17ba407/docs/source/notes/libtorch_stable_abi.md?plain=1#L247C1-L248C1  and equivalent use in xformers: https://github.com/facebookresearch/xformers/blob/03b91d7d9ff295ae68a320e2e733dd6c2ef8f342/setup.py#L261

`0x0209` was 2.9, `0x020a` was 2.10, `0x020c` is 2.12.

src/torchcodec/_core/AVIOFileLikeContext.h
```diff
@@ -6,8 +6,14 @@
...
 #pragma once
 
+// pybind11 headers from torch error out if TORCH_TARGET_VERSION is defined,
+// so we temporarily undefine it.
+// See https://github.com/pytorch/pytorch/pull/174372 for context
+#pragma push_macro("TORCH_TARGET_VERSION")
+#undef TORCH_TARGET_VERSION
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
+#pragma pop_macro("TORCH_TARGET_VERSION")
```
NicolasHug, line 16: To explain what the above does and why it's needed, we first need to understand what `TORCH_TARGET_VERSION` is.

`TORCH_TARGET_VERSION` is a macro that we now define (see `Cmakelists.txt` just below) when we compile torchcodec. It is the torch-blessed way of ensuring that we will never be including headers that aren't ABI-stable w.r.t. torch 2.12.

The way it works is defined in this [function in torch](https://github.com/pytorch/pytorch/blob/f4de0c9271ec35094b1d10d8361d062da17ba407/setup.py#L1102): every non-stable header that torch exposes is wrapped with [this error](https://github.com/pytorch/pytorch/blob/f4de0c9271ec35094b1d10d8361d062da17ba407/setup.py#L1145-L1151) which triggers whenever a downstream library (like TC) tries include these headers while `TORCH_TARGET_VERSION` is defined.

That is, if you (in TC) define `TORCH_TARGET_VERSION` and try to include a non-stable header like  `torch/types.h` (as we previously did), you get a compile-time error telling you that you cannot do that.

The only torch headers that *do not* raise such a compile time error are the "stable ABI headers" defined in this [allow-list](https://github.com/pytorch/pytorch/blob/f4de0c9271ec35094b1d10d8361d062da17ba407/setup.py#L1119-L1124).


-----


Now, the problem is that the `pybind` (and `fmt`!) headers we use directly come form torch. And they get wrapped with that error message.

Ideally, we'd want them to be in the allow-list: they're all ABI-stable by definition because they're header-only. That's what I suggested in https://github.com/pytorch/pytorch/pull/174372, but after offline discussion with Jane and Mikayla, there was a preference to be extra safe and not make that the default behavior.

So here, to avoid the compile-time error, we unset `TORCH_TARGET_VERSION` just before including `pybind` and `fmt` headers, and re-set it just after.

Again, this is safe: neither `pybind` nor `fmt` have anything to do with torch, so they're torch-ABI-stable by definition. On top of that, they're header-only.

mollyxu: Thanks for the explanation! To check my understanding, header-only`pybind` is `torch-ABI-stable` meaning it won't affect us if we were to exclude it from the warnings (but it should not be added to the allow-list because it's not `python-ABI-stable`). I'm curious what exactly would be a footgun situation caused by it not being `python-ABI-stable`

src/torchcodec/_core/CMakeLists.txt
```diff
@@ -57,6 +57,8 @@ function(make_torchcodec_sublibrary
     # torch, we want to rely on the "header-only" mode of these headers and hit
     # all the `#ifdef FMT_HEADER_ONLY` paths, so we define FMT_HEADER_ONLY here.```
     target_compile_definitions(${library_name} PRIVATE FMT_HEADER_ONLY)
+    # TC is ABI-stable w.r.t. torch since torch 2.11
+    target_compile_definitions(${library_name} PRIVATE TORCH_TARGET_VERSION=0x020b000000000000)
```
NicolasHug, line 61: See [pytorch/pytorch@f4de0c9/docs/source/notes/libtorch_stable_abi.md?plain=1#L247C1-L248C1](https://github.com/pytorch/pytorch/blob/f4de0c9271ec35094b1d10d8361d062da17ba407/docs/source/notes/libtorch_stable_abi.md?plain=1#L247C1-L248C1) and equivalent use in xformers: [facebookresearch/xformers@03b91d7/setup.py#L261](https://github.com/facebookresearch/xformers/blob/03b91d7d9ff295ae68a320e2e733dd6c2ef8f342/setup.py#L261)

0x0209 was 2.9, 0x020a was 2.10, 0x020b is 2.11.

src/torchcodec/_core/custom_ops.cpp
NicolasHug: There are two new use of the previous deleter pattern in custom_ops now: the wav decoder, and the multi-stream-encoder. I'll update them as a follow-up PR, we don't need to update them *now*. There are corresponding TODO_STABLE_ABI, which I plan to address/cleanup.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -98,16 +104,11 @@ STABLE_TORCH_LIBRARY(torchcodec_ns, m) {
 
 namespace {
 
-// TODO_STABLE_ABI: use previous deleter pattern with a lambda, once
```
Dan-Flores, line 101: We have two more instances of this `TODO_STABLE_ABI` on [main in custom_ops.cpp](https://github.com/search?q=repo%3Ameta-pytorch%2Ftorchcodec%20TODO_STABLE_ABI&type=code) for `multiStreamEncoderDeleter` and `wavDecoderDeleter` that we should clean up. Perhaps you need to pull the latest main to see those changes?

mollyxu: I saw another comment that could be related here https://github.com/meta-pytorch/torchcodec/pull/1260/changes#r3072689563

NicolasHug: Correct, I plan to address those as follow-ups.

Issue #1170, opened 2026-01-15
Allow user to know when a video is HDR


NicolasHug:
Users currently have no way to know whether the video they are about to decode is an HDR video (10bit, 12bit, etc.). We probably want to expose something in our metadata attribute. Maybe the pixel format? I'm not sure.

bhack:
The new Fffmpeg 8.1 branch cut for the next release has color transformation support with https://code.ffmpeg.org/FFmpeg/FFmpeg/pulls/21611

NicolasHug:
We're exposing the relevant info in the medata now, this can be closed

https://github.com/meta-pytorch/torchcodec/blob/f3b5983893bac720490322d69befb19ffa1dbcf1/src/torchcodec/_core/_metadata.py#L120-L130

PR #1261, opened 2026-02-24
Expose pixel format on VideoStreamMetadata 

mollyxu:
## Expose pixel format on video stream metadata

Related: #1170 

Users currently have no way to determine from metadata whether a video contains HDR content (e.g. 10-bit `yuv420p10le`). This exposes the source [pixel format](https://ffmpeg.org/doxygen/1.1/pixfmt_8h.html#a9a8e335cf3be472042bc9f0cf80cd4c5) on `VideoStreamMetadata`, letting users detect bit depth and make informed decisions about how to process the video.

#### `Metadata.h`
Added `std::optional<std::string> pixelFormat` to the video-only section of `StreamMetadata`.

#### `SingleStreamDecoder.cpp`
Extract the pixel format from `avStream->codecpar->format` using `av_get_pix_fmt_name()`, following the same pattern as the audio `sampleFormat` extraction. Added `#include <libavutil/pixdesc.h>` for the function declaration.

#### `custom_ops.cpp`
Serialize `pixelFormat` as a quoted JSON string in `get_stream_json_metadata()`.

#### `_metadata.py`
Added `pixel_format: str | None` field to `VideoStreamMetadata`

#### `test_metadata.py`
- Added `pixel_format == "yuv420p"` assertion in `test_get_metadata` (runs across all seek modes)
- Updated `test_repr` expected output to include the new field

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -171,9 +171,13 @@ void SingleStreamDecoder::initializeDecoder() {
       streamMetadata.sampleAspectRatio =
           avStream->codecpar->sample_aspect_ratio;
 
-      AVPixelFormat pixFmt =
+      AVPixelFormat pixelFormat =
           static_cast<AVPixelFormat>(avStream->codecpar->format);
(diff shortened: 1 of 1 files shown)


NicolasHug (approved):
Thanks @mollyxu !

Issue #1262, opened 2026-02-25
Bug: Incorrect ColorTwist matrix offsets cause color corruption on CUDA decode for BT.709 full-range videos


ladhiet:
### 🐛 Describe the bug

Hi, thanks for the great work on torchcodec! We love using it :) I believe however in recent versions >= 0.7.0 a bug in decoding BT.709 full range videos has been introduced:

## Description

Videos with `colorspace=bt709` and `color_range=pc` (full/JPEG range) are decoded with a severe pink/magenta overlay on CUDA GPU, while CPU decode is correct. Mean pixel diff between CPU decoding and GPU decoding is ~90–98 (out of 255).

### Root cause

I am not entirely sure this is a definitive fix, but I believe the `bt709FullRangeColorTwist` matrix (`CUDACommon.cpp`, lines 154–157) has incorrect 4th-column offsets:

```cpp
const Npp32f bt709FullRangeColorTwist[3][4] = {
    {1.0f, 0.0f, 1.5748f, 0.0f},        // R row
    {1.0f, -0.187324273f, -0.468124273f, -128.0f},  // G row
    {1.0f, 1.8556f, 0.0f, -128.0f}};     // B row
```

The 3×3 coefficients are correct (BT.709 full-range YUV→RGB). The problem is the **4th column (bias)**.

The comment at lines 138–141 describes the offset as a per-input-channel centering of U and V by −128. However, NPP's ColorTwist appears to perform a straightforward affine transform on raw [0, 255] values:

$$\text{output}_i = a_{i,0} \cdot Y + a_{i,1} \cdot U + a_{i,2} \cdot V + a_{i,3}$$

The 4th column is a **per-output-row** bias, not a per-input-channel offset. Since the conversion requires centered chroma $(U - 128)$, $(V - 128)$, expanding gives:

$$a_{i,3} = -(c_{U,i} + c_{V,i}) \times 128$$

| Row | c_U | c_V | Correct offset | Current offset |
|-----|-----|-----|----------------|----------------|
| R | 0.0 | 1.5748 | **−201.57** | **0.0** |
| G | −0.1873 | −0.4681 | **+83.90** | **−128.0** |
| B | 1.8556 | 0.0 | **−237.52** | **−128.0** |

Sanity check with neutral gray (Y=128, U=128, V=128):
- **Current matrix** → R=255, G=0, B=237 (bright pink — matches observed corruption)
- **Corrected matrix** → R=128, G=128, B=128 (correct)

[...]

Issue #1210, opened 2026-01-29
Bug in CUDA decoder on GoPro MP4 files


urbste:
### 🐛 Describe the bug

I run the cuda example basic_cuda_example.py on different GoPro MP4 files.
I also resampled the files with ffmpeg to another resolution to check if there was an issue with the files. But no matter which file from the GoPro I take the CUDA decoder produces that:

```
torch.__version__='2.9.1+cu128'
torch.cuda.is_available()=True
torch.cuda.get_device_properties(0)=_CudaDeviceProperties(name='NVIDIA GeForce RTX 4090', major=8, minor=9, total_memory=24077MB, multi_processor_count=128, uuid=6ab5479f-c3f6-e48b-c6b8-7d5b7cb8e9c5, pci_bus_id=1, pci_device_id=0, pci_domain_id=0, L2_cache_size=72MB)
torch.Size([3, 360, 640]) torch.uint8
cuda:0
frames_equal=False
mean_abs_diff=tensor(108.0210, device='cuda:0')
max_abs_diff=tensor(212., device='cuda:0')
```

<img width="2098" height="1277" alt="Image" src="https://github.com/user-attachments/assets/75cdbf2c-a65d-4647-b384-18b8bd1b9b56" />

It looks like the decoder is using a different color format? Or something else off? How could I debug what is going wrong?

### Versions

Collecting environment information...
PyTorch version: 2.9.1+cu128
Is debug build: False
CUDA used to build PyTorch: 12.8
ROCM used to build PyTorch: N/A

OS: Ubuntu 22.04.5 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0
Clang version: 14.0.0-1ubuntu1.1
CMake version: version 3.22.1
Libc version: glibc-2.35

Python version: 3.11.14 (main, Oct 21 2025, 18:31:21) [GCC 11.2.0] (64-bit runtime)
Python platform: Linux-6.8.0-90-generic-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.8.93
CUDA_MODULE_LOADING set to: 
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 4090
Nvidia driver version: 580.126.09
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.9.18.1
/usr/lib/x86_64-linux-gnu/libcudnn_adv.so.9.18.1
[...]

PR #1265, opened 2026-02-25
Fix BT.709 full range CUDA color conversion on CUDA 12

NicolasHug:
This PR fixes the  BT.709 full range color conversion for both CUDA interfaces. The logic we had works for CUDA 13, but not for CUDA 12. Why? I don't know. 

For ref these are the [CUDA 13 docs](https://docs.nvidia.com/cuda/npp/image_color_conversion.html) and these are [the 12.6 docs](https://docs.nvidia.com/cuda/archive/12.6.0/npp/image_color_conversion.html). They're both awful, and they're both the same.

Fixes https://github.com/meta-pytorch/torchcodec/issues/1262
Fixes https://github.com/meta-pytorch/torchcodec/issues/1210

### DIFF
--- a/src/torchcodec/_core/CUDACommon.cpp
+++ b/src/torchcodec/_core/CUDACommon.cpp
@@ -195,9 +195,9 @@ const Npp32f bt709FullRangeColorTwist[3][4] = {
     {1.0f, -0.187324273f, -0.468124273f, -128.0f},
     {1.0f, 1.8556f, 0.0f, -128.0f}};
 #else
-// Just kidding! The note above about nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx
+// The note above about nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx
 // offsets is actually only true for CUDA 13+. For CUDA 12, the behavior is
-// different - and, of course, still undocumented.
+// different and still undocumented.
 // See https://github.com/meta-pytorch/torchcodec/issues/1262#issue-3989049538
 // for how these offsets need to be derived.
 const Npp32f bt709FullRangeColorTwist[3][4] = {


test/test_decoders.py
```diff
@@ -1462,12 +1461,7 @@ def test_full_and_studio_range_bt709_video(self, asset):
             gpu_frame = decoder_gpu.get_frame_at(frame_index).data.cpu()
             cpu_frame = decoder_cpu.get_frame_at(frame_index).data
 
-            if cuda_version_used_for_building_torch() >= (13, 0):
-                torch.testing.assert_close(gpu_frame, cpu_frame, rtol=0, atol=3)
-            elif cuda_version_used_for_building_torch() >= (12, 9):
-                torch.testing.assert_close(gpu_frame, cpu_frame, rtol=0, atol=2)
-            elif cuda_version_used_for_building_torch() == (12, 8):
-                assert psnr(gpu_frame, cpu_frame) > 20
+            torch.testing.assert_close(gpu_frame, cpu_frame, rtol=0, atol=3)
```
NicolasHug, line 1464: We can use stricter checks for all CUDA versions now.

We only test on 12.6 and 13.0 on PRs, but I also validated 12.8 here https://github.com/meta-pytorch/torchcodec/pull/1265/changes/80ce23c707a323effe374df36108c8b36447c716

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -184,15 +184,27 @@ void initializeCudaContextWithPytorch(const StableDevice& device) {
...
 const Npp32f bt709FullRangeColorTwist[3][4] = {
     {1.0f, 0.0f, 1.5748f, 0.0f},
     {1.0f, -0.187324273f, -0.468124273f, -128.0f},
     {1.0f, 1.8556f, 0.0f, -128.0f}};
+#else
+// Just kidding! The note above about nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx
+// offsets is actually only true for CUDA 13+. For CUDA 12, the behavior is
+// different - and, of course, still undocumented.
+// See https://github.com/meta-pytorch/torchcodec/issues/1262#issue-3989049538
+// for how these offsets need to be derived.
```
NicolasHug, line 202: I will tone down the sassiness upon reviewer request. This is already a heavily self-censored comment.

I hate this *so much*.

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -184,15 +184,27 @@ void initializeCudaContextWithPytorch(const StableDevice& device) {
 //         [ 1.0000e+00, -1.8732e-01, -4.6812e-01,     -128]
 //         [ 1.0000e+00,  1.8556e+00,  4.6231e-09 ,    -128]])
 //
-// And that's what we need to pass for BT701, full range.
+// And that's what we need to pass for BT709, full range.
 /* clang-format on */
 
 // BT.709 full range color conversion matrix for YUV to RGB conversion.
 // See Note [YUV -> RGB Color Conversion, color space and color range]
+#if CUDART_VERSION >= 13000
```
NicolasHug, line 192: This is available from cuda_runtime.h, see similar uses in torch https://github.com/search?q=repo%3Apytorch%2Fpytorch%20CUDART_VERSION&type=code

src/torchcodec/_core/CUDACommon.cpp
mollyxu: Thanks for looking into this! This is very frustrating for there to be undocumented behavior change across updates. For the comment, perhaps we could do something like 

```
// The note above about nppiNV12ToRGB_8u_ColorTwist32f_P2C3R_Ctx
// offsets is actually only true for CUDA 13+. For CUDA 12, the behavior is
// different and still undocumented.
// See https://github.com/meta-pytorch/torchcodec/issues/1262#issue-3989049538
// for how these offsets need to be derived.
```

PR #1267, opened 2026-02-27
Add bt2020 color conversion support on Beta Cuda

mollyxu:
This PR is a step towards better support for HDR videos on TorchCodec.

Added BT.2020 (NCL/CL) YUV→RGB color conversion support on the CUDA decode path, with ColorTwist matrices for both full and limited range, and for both CUDA >= 13 and CUDA < 13 NPP conventions.

**Known limitation**:  
The added test is not passing because there is an up to ~73 pixel value difference vs the CPU path, with ~5.6% mismatched elements. It does not seem to be caused by the changes in the PR. In fact, before the PR, there is an up to 71 pixel value difference, with ~59.2% mismatched elements.

This is reflected in a TODO to investigate the root cause of this difference


src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -276,10 +316,27 @@ torch::stable::Tensor convertNV12FrameToRGB(
...
+        dst.mutable_data_ptr<Npp8u>(),
+        dst.stride(0),
+        oSizeROI,
+        matrix,
+        *nppCtx);
   } else {
     // TODO we're assuming BT.601 color space (and probably limited range) by
-    // calling nppiNV12ToRGB_8u_P2C3R_Ctx. We should handle BT.601 full range,
-    // and other color-spaces like 2020.
+    // calling nppiNV12ToRGB_8u_P2C3R_Ctx. We should handle BT.601 full range.
```
NicolasHug, line 339: As a drive-by, let's also add that we should probable have a heuristic to actually dispatch to the 709 case e.g. when color space info is missing but the resolution is high enough.

test/utils.py
```diff
@@ -564,6 +564,26 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
     frames={0: {}},  # Not needed for now
 )
 
+# BT.2020 10-bit video with limited range (tv), generated with:
```
NicolasHug, line 567: Instead of a limited range video, can you try to check with a full-range one? I am hoping that our results will be closer to the CPU ones on a full range video, because the color-conversion matrix is more straightforward for full range.

test/test_decoders.py
```diff
@@ -1463,6 +1464,30 @@ def test_full_and_studio_range_bt709_video(self, asset):
 
             torch.testing.assert_close(gpu_frame, cpu_frame, rtol=0, atol=3)
 
+    @needs_cuda
+    @pytest.mark.skip("TODO investigate CPU vs BetaCUDA mismatch on BT.2020 10-bit")
```
NicolasHug: Agreed with the TODO, but maybe we can unskip this test and just loosen the tolerance check a bit. If we now have a ~5.6% mismatch with CPU instead of the previous 59%, then that's still a great improvement that we can assert against by relying on `assert_tensor_close_on_at_least`. This will ensure that this test will catch any regression - and we can still make progress as follow-ups according to the TODO.

mollyxu: Thanks for the suggestion! I tested with a full range bt2020 video and we have a ~12% mismatch with CPU with these changes instead of 42% mismatch. Would it be better to keep the limited range video for the non-regression test?

NicolasHug: Interesting. Since the full range one isn't better, let's not bother, we can keep the current one.

test/utils.py
```diff
@@ -564,6 +564,26 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
+# -pix_fmt yuv420p10le -color_primaries bt2020 -color_trc smpte2084 \
+# -colorspace bt2020nc -color_range tv bt2020_10bit.mp4
+#
+# Confirm color space with:
+# ffprobe -v quiet -select_streams v:0 -show_entries stream=color_space,color_transfer,color_primaries,color_range -of default=noprint_wrappers=1 test/resources/bt2020_10bit.mp4
+# color_range=tv
+# color_space=bt2020nc
+# color_transfer=smpte2084
+# color_primaries=bt2020
+BT2020_10BIT = TestVideo(
```
NicolasHug: Nit, I think we already have `BT709_FULL_RANGE` so it's useful to disambiguate

```suggestion
BT2020_LIMTED_RANGE_10BIT = TestVideo(
```

NicolasHug (approved):
Thanks for the PR @mollyxu . Ill approve now to unblock, let's still try to unskip the new test as previously mentioned.

PR #1270, opened 2026-03-03
Fix bt601 color conversion on cuda

mollyxu:
## Summary
Addressed TODO in codebase related to handling videos with undefined color space on Cuda.

Previously, the BT.601 branch in convertNV12FrameToRGB used the pre-defined `nppiNV12ToRGB_8u_P2C3R_Ctx`, which didn't closely match CPU results and didn't distinguish between full and limited range. 

This PR:                                                               
  - Adds custom color twist matrices for BT.601 full range and BT.601 limited range (with CUDART_VERSION >= 13000 variants, matching the existing BT.709 pattern)
  - Checks color_range for BT.601 (just like we already do for BT.709), dispatching to the appropriate matrix

Reproduction script from #1262 

## Reproduction

<details>
<summary>Minimal reproduction script (test.py)</summary>

```python
"""Minimal reproducible example: torchcodec GPU decode color mismatch.

Generates synthetic test videos with different pixel format / colorspace
combinations, then decodes on CPU and GPU via torchcodec and reports
per-frame RGB differences.
"""

import math
import pathlib
import subprocess

import torch
import torchcodec
import torchcodec.decoders

OUTPUT_DIR = pathlib.Path("/tmp/torchcodec_repro")
NUM_FRAMES_TO_CHECK = 10
TOTAL_FRAMES = 300  # 10s @ 30fps
FPS = 30

VIDEOS: dict[str, dict] = {
    "yuvj420p_bt709_fullrange": {
        "path": OUTPUT_DIR / "yuvj420p_bt709_fullrange.mp4",
        "ffmpeg_args": [
            "-f", "lavfi", "-i",
            "testsrc=duration=10:size=1920x1080:rate=30",
            "-c:v", "libx264", "-profile:v", "high",
            "-pix_fmt", "yuvj420p",
            "-colorspace", "bt709", "-color_trc", "bt709",
            "-color_primaries", "bt709", "-y",
        ],
    },
    "yuvj420p_unknown": {
        "path": OUTPUT_DIR / "yuvj420p_unknown.mp4",
        "ffmpeg_args": [
            "-f", "lavfi", "-i",
            "testsrc=duration=10:size=1920x1080:rate=30",
            "-c:v", "libx264", "-profile:v", "high",
            "-pix_fmt", "yuvj420p", "-y",
        ],
    },
    "yuv420p_bt709_fullrange": {
        "path": OUTPUT_DIR / "yuv420p_bt709_fullrange.mp4",
        "ffmpeg_args": [
            "-f", "lavfi", "-i",
            "testsrc=duration=10:size=1920x1080:rate=30",
            "-c:v", "libx264", "-profile:v", "high",
            "-pix_fmt", "yuv420p", "-color_range", "pc",
            "-colorspace", "bt709", "-color_trc", "bt709",
            "-color_primaries", "bt709", "-y",
        ],
    },
    "yuv420p_unknown": {
        "path": OUTPUT_DIR / "yuv420p_unknown.mp4",
        "ffmpeg_args": [
            "-f", "lavfi", "-i",
            "testsrc=duration=10:size=1920x1080:rate=30",
            "-c:v", "libx264", "-profile:v", "baseline",
            "-pix_fmt", "yuv420p", "-y",
        ],
    },
}


def generate_videos() -> None:
    OUTPUT_DIR.mkdir(parents=True, exist_ok=True)
    for name, spec in VIDEOS.items():
        path = spec["path"]
        if path.exists():
            continue
        cmd = ["ffmpeg"] + spec["ffmpeg_args"] + [str(path)]
        subprocess.run(cmd, check=True, capture_output=True)


def decode_frames(video_path, frame_indices, device):
    decoder = torchcodec.decoders.VideoDecoder(
        video_path.as_posix(), device=device, num_ffmpeg_threads=0
    )
    raw = decoder.get_frames_at(frame_indices)
    return [frame.data.cpu() for frame in raw]


def compare_video(name, path):
    step = max(1, TOTAL_FRAMES // NUM_FRAMES_TO_CHECK)
    frame_indices = list(range(0, TOTAL_FRAMES, step))[:NUM_FRAMES_TO_CHECK]
    cpu_frames = decode_frames(path, frame_indices, device="cpu")
    gpu_frames = decode_frames(path, frame_indices, device="cuda")
    per_frame_diffs = [
        (cpu_frames[i].float() - gpu_frames[i].float()).abs().mean().item()
        for i in range(len(frame_indices))
    ]
    mean_diff = sum(per_frame_diffs) / len(per_frame_diffs)
    std_diff = math.sqrt(
        sum((d - mean_diff) ** 2 for d in per_frame_diffs) / len(per_frame_diffs)
    )
    return {"mean": mean_diff, "std": std_diff}


def main():
    generate_videos()
    print(f"torch {torch.__version__}  |  torchcodec {torchcodec.__version__}\n")
    results = {}
    for name, spec in VIDEOS.items():
        results[name] = compare_video(name, spec["path"])
    name_width = max(len(n) for n in results)
    header = f"{'Video':<{name_width}}  {'Mean Diff':>10}  {'Std Diff':>10}"
    print("=" * len(header))
    print(header)
    print("-" * len(header))
    for name, stats in results.items():
        print(f"{name:<{name_width}}  {stats['mean']:>10.2f}  {stats['std']:>10.2f}")
    print("=" * len(header))


if __name__ == "__main__":
    main()
```

</details>

Before:
```
torch 2.11.0.dev20260211+cu126  |  torchcodec 0.11.0a0+39793ba

================================================
Video                      Mean Diff    Std Diff
------------------------------------------------
yuvj420p_bt709_fullrange        0.28        0.00
yuvj420p_unknown                8.13        0.09
yuv420p_bt709_fullrange         0.28        0.00
yuv420p_unknown                18.69        0.11

```

After:
```
torch 2.11.0.dev20260211+cu126  |  torchcodec 0.11.0a0+63714dd

================================================
Video                      Mean Diff    Std Diff
------------------------------------------------
yuvj420p_bt709_fullrange        0.28        0.00
yuvj420p_unknown                0.14        0.00
yuv420p_bt709_fullrange         0.28        0.00
yuv420p_unknown                 0.42        0.00
================================================
```


Dan-Flores:
We should add / update some tests to document and track this improved accuracy! I believe this TODO is related:
https://github.com/meta-pytorch/torchcodec/blob/900979ede9bb878bd922cc6735e961da4ca65fe6/test/test_decoders.py#L1669-L1672

We might be able to reduce the skip condition now that the beta cuda interface does not default to BT.709, which seems to have been an issue with `TEST_SRC_2_720P_MPEG4`
https://github.com/meta-pytorch/torchcodec/blob/900979ede9bb878bd922cc6735e961da4ca65fe6/test/test_decoders.py#L1712-L1713

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -240,6 +275,11 @@ torch::stable::Tensor convertNV12FrameToRGB(
 
   // For background, see
   // Note [YUV -> RGB Color Conversion, color space and color range]
+  //
+  // When the colorspace is unspecified, we default to BT.601. This matches
+  // FFmpeg's swscale behavior: sws_getCoefficients(SWS_CS_DEFAULT) returns
+  // BT.601 coefficients
```
NicolasHug: Let's move this comment within the `BT.601` section itself, it'll be easier to follow

mollyxu:
Thanks for the reviews! I added non-regression test for both full and limited range. 

For the following code, I investigated a bit and it seems like Beta Cuda is getting colorspace from NVCUVID which reports the video as BT.709 whereas FFmpeg is getting colorspace from the metadata, which in this case is undefined, resulting in a fallback to BT.601. On the Beta Cuda path, the fallback condition was never triggered because NVCUVID reports this video's matrix coefficient as 1 (BT.709). As such, I don't think the current PR will have any effects on this TODO but I modified the comment to make it more clear. Let me know what you think @Dan-Flores 
https://github.com/meta-pytorch/torchcodec/blob/900979ede9bb878bd922cc6735e961da4ca65fe6/test/test_decoders.py#L1669-L1672

test/utils.py
```diff
@@ -575,6 +575,44 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
+# color_primaries=unknown
+UNSPECIFIED_LIMITED_RANGE = TestVideo(
+    filename="unspecified_limited_range.mp4",
+    default_stream_index=0,
+    stream_infos={
+        0: TestVideoStreamInfo(width=320, height=240, num_color_channels=3),
+    },
+    frames={0: {}},  # Not needed for now
+)
+
```
Dan-Flores, line 628: This block should probably go after `BT2020_LIMITED_RANGE_10BIT`, so the description on lines 567-577 can be next to the `TestVideo` object

test/test_decoders.py
```diff
@@ -1668,8 +1688,8 @@ def test_get_frames_at_tensor_indices(self):
     #   assert_tensor_close_on_at_least or something like that.
     # - unskip equality assertion checks for MPEG4 asset. The frames are decoded
     #   fine, it's the color conversion that's different. The frame from the
-    #   BETA interface is assumed to be 709 while the one from the default
-    #   interface is 601.
```
Dan-Flores, line 1672: Thanks for looking into this! This comment is helpful, I see why this TODO not affected by the changes in `convertNV12FrameToRGB` - the `matrix_coefficients` field is used in the beta cuda interface, so colorspace is not unspecified.

It's interesting that for `TEST_SRC_2_720P_MPEG4` which does not report color parameters, beta cuda is able to determine that the video is in BT.709, while ffmpeg defaults to treating it as BT.601. Claude is indicating this might be decoder or container-format specific, so I don't think this is blocking. 

```
ffprobe -v quiet -select_streams v:0 -show_entries  stream=color_space,color_transfer,color_primaries,color_range -of default=noprint_wrappers=1 test/resources/testsrc2_mpeg4.avi                               
color_range=unknown
color_space=unknown
color_transfer=unknown
color_primaries=unknown
```

test/utils.py
```diff
@@ -584,6 +584,44 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
+# color_transfer=unknown
+# color_primaries=unknown
+UNSPECIFIED_FULL_RANGE = TestVideo(
+    filename="unspecified_full_range.mp4",
+    default_stream_index=0,
+    stream_infos={
+        0: TestVideoStreamInfo(width=320, height=240, num_color_channels=3),
+    },
+    frames={0: {}},  # Not needed for now
+)
```
Dan-Flores, line 606: We will need to check in these new videos to the repo. Based on the ffmpeg commands used to generate them they should be reasonably small.

test/utils.py
```diff
@@ -584,6 +584,44 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
...
+# ffmpeg -f lavfi -i testsrc2=duration=2:size=320x240:rate=30 -c:v libx264
+# -profile:v high -pix_fmt yuv420p -color_range pc unspecified_full_range.mp4
+#
+# Confirm color space with:
+# ffprobe -v quiet -select_streams v:0 -show_entries stream=color_space,color_transfer,color_primaries,color_range -of default=noprint_wrappers=1 test/resources/unspecified_full_range.mp4
+# color_range=pc
+# color_space=unknown
+# color_transfer=unknown
+# color_primaries=unknown
+UNSPECIFIED_FULL_RANGE = TestVideo(
```
NicolasHug: I'd just suggest to name this one `BT601_FULL_RANGE` and explicitly have it be BT 601 with corresponding metadata fields (same for the other one).

Ultimately, this PR is about properly discriminating between BT601 full range vs BT601 limited range, so I think the videos we check-in should reflect that.

Sure, we fallback to that BT601 path when the color spaces are unknown, but that's not the essence of what this PR is addressing.

If we wanted to test the fallback mechanism when metadata is unknown, we would have a slightly different test (e.g. we compare the output of `BT601_FULL_RANGE` vs `UNSPECIFIED_FULL_RANGE` and confirm that they're bitwise equivalent) - but I think that's out of scope for this PR.

test/test_decoders.py
```diff
@@ -1491,6 +1493,24 @@ def test_bt2020_10bit_video(self):
...
+        # decoder on videos with unspecified colorspace metadata.
+        with set_cuda_backend("beta"):
+            decoder_gpu = VideoDecoder(asset.path, device="cuda")
+        decoder_cpu = VideoDecoder(asset.path, device="cpu")
+
+        for frame_index in (0, 10, 20, 5):
+            gpu_frame = decoder_gpu.get_frame_at(frame_index).data.cpu()
+            cpu_frame = decoder_cpu.get_frame_at(frame_index).data
+
+            torch.testing.assert_close(gpu_frame, cpu_frame, rtol=0, atol=3)
```
NicolasHug, line 1512: Just checking: you verified that these tests are properly failing on `main` on all CUDA versions? (The simple way to check is to submit a dummy PR with that test, and without the cpp changes)

test/utils.py
NicolasHug: I think it's OK to leave it as-is for now, we have [various instances](https://github.com/meta-pytorch/torchcodec/blob/1699f5eedba202f2989e9d5668283294b15f4192/test/utils.py#L512) of that already. We only need the `frames` field when we query metadata or when we compare against checked-in references, but for now that's not needed on these videos.

NicolasHug (approved):
Thanks @mollyxu , LGTM, I left some non-blocking suggestions

test/test_decoders.py
mollyxu: Thanks for the tips! I've verified that the updated tests are properly failing on `main` on all CUDA versions with this dummy PR #1278

PR #1271, opened 2026-03-04
Expose HDR Color metadata

mollyxu:
## Add HDR color metadata to VideoStreamMetadata

Related: #1170

Builds on #1261. This PR exposes the color primaries, colorspace, and color transfer characteristic from FFmpeg on VideoStreamMetadata.

### Changes Made

  Metadata.h
  Added `std::optional<AVColorPrimaries>`, `std::optional<AVColorSpace>`, and `std::optional<AVColorTransferCharacteristic>` to the video-only section of `StreamMetadata`. 

  SingleStreamDecoder.cpp
  Extract color primaries, color space, and color transfer characteristic from avStream->codecpar. 

  custom_ops.cpp
  Serialize the three color fields as quoted JSON strings in get_stream_json_metadata(), using FFmpeg's `av_color_primaries_name()`, `av_color_space_name()`, and `av_color_transfer_name()`
  for enum-to-string conversion.

  _metadata.py
  Added `color_primaries`, `color_space`, and `color_transfer_characteristic` fields (all str | None) to VideoStreamMetadata.

  test_metadata.py
  Added test_color_metadata covering a BT.2020 10-bit PQ video (bt2020_10bit.mp4) and the NASA BT.709 video. Updated test_repr expected output to include the new fields.

### DIFF
--- a/src/torchcodec/_core/_metadata.py
+++ b/src/torchcodec/_core/_metadata.py
@@ -123,7 +123,7 @@ class VideoStreamMetadata(StreamMetadata):
     """Color space as reported by FFmpeg. E.g. ``"bt709"``,
     ``"bt2020nc"``."""
     color_transfer_characteristic: str | None
-    """Color transfer characteristic (OETF) as reported by FFmpeg
+    """Color transfer characteristic as reported by FFmpeg
     E.g. ``"bt709"``, ``"smpte2084"`` (PQ), ``"arib-std-b67"`` (HLG)."""
     pixel_format: str | None
     """The source pixel format of the video as reported by FFmpeg.


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -983,6 +988,18 @@ std::string get_stream_json_metadata(
...
+        quoteValue(av_color_primaries_name(*streamMetadata.colorPrimaries));
+  }
+  if (streamMetadata.colorSpace.has_value()) {
+    map["colorSpace"] =
+        quoteValue(av_color_space_name(*streamMetadata.colorSpace));
+  }
+  if (streamMetadata.colorTransferCharacteristic.has_value()) {
+    map["colorTransferCharacteristic"] = quoteValue(
+        av_color_transfer_name(*streamMetadata.colorTransferCharacteristic));
+  }
```
Dan-Flores, line 1003: The [docs indicate](https://www.ffmpeg.org/doxygen/8.0/pixdesc_8c.html#a811dbb6cf48bbcd66a1243bdd6d89909) `av_color_primaries_name` and similar functions can return `NULL`, so we should check that these values are valid before passing it to `quoteValue`.

NicolasHug: Let's also do all that in some new method of the `StreamMetadata`, e.g. `getColorPrimariesName()` or something like that. I think we'll want to keep custom_ops.cpp as logic-free as possible.

src/torchcodec/_core/_metadata.py
```diff
@@ -117,8 +117,17 @@ class VideoStreamMetadata(StreamMetadata):
         or -90 degree rotation, this means width and height are swapped
         compared to the raw encoded dimensions in the container.
     """
+    color_primaries: str | None
+    """Color primaries as reported by FFmpeg. E.g. ``"bt709"``, ``"bt2020"``."""
+    color_space: str | None
+    """Color space as reported by FFmpeg. E.g. ``"bt709"``,
+    ``"bt2020nc"``."""
+    color_transfer_characteristic: str | None
+    """Color transfer characteristic (OETF) as reported by FFmpeg
```
Dan-Flores: TIL `OETF` stands for "Opto-Electronic Transfer Function"!
I could not find that acronym in [FFmpeg docs](https://trac.ffmpeg.org/wiki/colorspace), but perhaps it will be recognizable to other media engineers.

mollyxu: I agree that it is confusing because in this sentence it sounds like OETF would be a part of FFmpeg docs. I removed it for clarity

PR #1272, opened 2026-03-05
Streaming encoder ops to open and close

Dan-Flores:
This PR implements ops to open and close `MultiStreamEncoder` for writing to a file. Writing to file-like will come later. 

### Changes:
* The logic used in `MultiStreamEncoder` is copied from `VideoEncoder`'s constructor (see PR comment).

Some points to bikeshed on: 
* It's currently named `MultiStreamEncoder`. 
  * Alternatives could be `Encoder`, `StreamEncoder`, open to suggestions.
* It's currently implemented in `Encoder.cpp`. 
  * There's significant overlap with `VideoEncoder` so adding it here was easiest, though `Encoder.cpp` is becoming a large file.
  

### Open Questions:
* Currently, there's no support for outputting a tensor.
  * `VideoEncoder::encodeToTensor` returns the output tensor once encoding is complete. This is different from the streaming encoder use case, which should make the output available at all times, not just at the end of writing.
  * If we needed to make this work, we could accept and write to a user allocated tensor, but I don't think thats a great UX. I am open to suggestions here.

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -160,6 +160,12 @@ SingleStreamDecoder* unwrapTensorToGetDecoder(torch::stable::Tensor& tensor) {
   return decoder;
 }
 
+// TODO_STABLE_ABI: use previous deleter pattern with a lambda, once
+// https://github.com/pytorch/pytorch/pull/175089 is available.
+void multiStreamEncoderDeleter(void* data) {
+  delete static_cast<MultiStreamEncoder*>(data);
+}
+
 torch::stable::Tensor wrapMultiStreamEncoderPointerToTensor(
     std::unique_ptr<MultiStreamEncoder> uniqueEncoder) {
   MultiStreamEncoder* encoder = uniqueEncoder.release();
@@ -171,7 +177,7 @@ torch::stable::Tensor wrapMultiStreamEncoderPointerToTensor(
       {strides, 1},
       StableDevice(kStableCPU),
       kStableInt64,
-      [encoder](void*) { delete encoder; });
+      &multiStreamEncoderDeleter);
   auto multiStreamEncoder =
       static_cast<MultiStreamEncoder*>(tensor.mutable_data_ptr());
   STD_TORCH_CHECK(


src/torchcodec/_core/custom_ops.cpp
```diff
@@ -114,6 +119,40 @@ SingleStreamDecoder* unwrapTensorToGetDecoder(torch::stable::Tensor& tensor) {
   return decoder;
 }
 
+// TODO_STABLE_ABI: use previous deleter pattern with a lambda, once
+// https://github.com/pytorch/pytorch/pull/175089 is available.
```
NicolasHug: It's available, let's use it. You might just need to update your local nightly version of torch.

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
...
+            f = open(tmp_path / "test.mp4", "wb")
+            handle = create_streaming_encoder(dest=f)
+        else:
+            f = open(tmp_path / "test", "wb")
+            handle = create_streaming_encoder(dest=f, format="mp4")
+        streaming_encoder_close(handle)
+        streaming_encoder_close(handle)  # double close is a no-op
+
+    def test_create_invalid_path(self):
+        with pytest.raises(RuntimeError):
```
NicolasHug: make sure to match the error message. It's more robust, and also acts as documentation for the reviewer / code reader.

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
             )
 
 
+class TestStreamingEncoderOps:
+    @pytest.mark.parametrize(
+        "output", ("file", "file_like", "file_like_explicit_format")
+    )
+    def test_create_and_close(self, tmp_path, output):
+        if output == "file":
+            handle = create_streaming_encoder(str(tmp_path / "test.mp4"))
```
NicolasHug: is that a "handle" as in "file handle" or a "tensor encoder" object? I think it's just an encoder tensor?

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
...
+        elif output == "file_like":
+            f = open(tmp_path / "test.mp4", "wb")
```
NicolasHug: We'll want to make sure `f` is closed, same below. Or, maybe I'm missing something, but we should understand and test against calling `f.close()` before and after `streaming_encoder_close(handle)`

Dan-Flores: Good catch! Let me add `f.close()`.

ATM, we are not writing to the file, so the close order is not important, but I'll keep this in mind for following PRs.

test/test_ops.py
Dan-Flores: Yes thats correct, updated for clarity.

src/torchcodec/_core/ops.py
```diff
@@ -240,6 +247,25 @@ def encode_video_to_file_like(
     )
 
 
+def create_streaming_encoder(
```
mollyxu: Quick question on the PR summary bullet point about supporting writing frames to URLs in the future: Would we be modifying the logic here to check if `format` is a url before dispatching to the corresponding method?

Dan-Flores: In this particular function, we would always call `_create_streaming_encoder`, but add the `format` arg. Whether or not an input string is a `url` or `filename` is determined later by FFmpeg, we just need to provide `avio_open` with that string and an `AVFormatContext` (which is where the `format` arg is used).

This commit where I undid these changes might be helpful, specifically `Encoder.cpp` where `avformat_alloc_output_context2` is called and `avio_open` is called later: https://github.com/meta-pytorch/torchcodec/pull/1272/changes/8daa3048aec9e5ba3c789adfecd4695702c6d832

src/torchcodec/_core/ops.py
```diff
@@ -240,6 +247,25 @@ def encode_video_to_file_like(
...
+) -> torch.Tensor:
+    if isinstance(dest, str):
+        return _create_streaming_encoder(dest)
+    else:
+        assert _pybind_ops is not None
+        format = format or Path(getattr(dest, "name", "")).suffix.lstrip(".")
+        if not format:
+            raise ValueError(
+                "format must be specified when it cannot be inferred from dest"
+            )
```
NicolasHug: Let's keep things simple: this logic is relying on the potential existence of the `.name` attribute of file-like objects.

It exists for File objects coming from `open()`, but it won't exist in general.

Importantly, this "feature" isn't supported in the current VideoEncoder, so we don't need to support it now. See it as some kind of syntaxic sugar that we could add at the very, very end of the development process. Now's not the time for these kind of quality-of-life features.

NicolasHug: Also, let's not define this `create_streaming_encoder` helper now. We probably don't need to? If we did, it'd be better in a separate file, not in `ops.py`. We want to keep the ops.py file to be mostly a smaller C++ <-> Python binding layer.

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
...
+class TestStreamingEncoderOps:
+    def test_create_and_close_file(self, tmp_path):
+        encoder_tensor = create_streaming_encoder(str(tmp_path / "test.mp4"))
+        streaming_encoder_close(encoder_tensor)
+        streaming_encoder_close(encoder_tensor)  # double close is a no-op
+
+    # TODO: Test if closing the file before closing the streaming encoder causes any issues.
+    @pytest.mark.parametrize("format", (None, "mp4"))
+    def test_create_and_close_file_like(self, tmp_path, format):
+        filename = "test.mp4" if format is None else "test"
```
NicolasHug: I got tripped up for a pretty long time about this `format=None` input in file-like contexts. See my other comment above, I don't think we need to support this now.

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
...
+        f = open(tmp_path / filename, "wb")
```
NicolasHug: Use `open()` in a context manager setting for now, we can call `close()` manually later when we address the TODO but for now it's not needed. Same below

test/test_ops.py
```diff
@@ -1140,5 +1142,43 @@ def test_bad_input(self, tmp_path):
...
+        encoder_tensor = create_streaming_encoder(dest=f, format=format)
+        streaming_encoder_close(encoder_tensor)
```
NicolasHug: Call this twice to mirror the other test above?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1012,4 +1028,34 @@ void VideoEncoder::flushBuffers() {
...
+  setFFmpegLogLevel();
+  allocateAVIOFormatContext(
+      formatName, avioContextHolder_.get(), avFormatContext_);
+}
+
+void StreamingEncoder::close() {
+  if (closed_) {
+    return;
+  }
+  closed_ = true;
```
NicolasHug, line 1049: The other encoders do not have a `closed_` attribute, so it's not immediately obvious why this one is needed.
I think it's OK to have it in, but since the logic is likely to change considerably, it's worth adding a TODO_STREAMING_ENCODER to re-evaluate the utility of this field down the road.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -118,6 +123,34 @@ SingleStreamDecoder* unwrapTensorToGetDecoder(torch::stable::Tensor& tensor) {
...
+  StreamingEncoder* encoder = uniqueEncoder.release();
+  int64_t sizes[] = {static_cast<int64_t>(sizeof(StreamingEncoder*))};
+  int64_t strides[] = {1};
+  torch::stable::Tensor tensor = torch::stable::from_blob(
+      encoder,
+      {sizes, 1},
+      {strides, 1},
+      StableDevice(kStableCPU),
+      kStableInt64,
+      [encoder](void*) { delete encoder; });
```
NicolasHug: Oh, I suggested to use the lambda deleter because https://github.com/pytorch/pytorch/pull/175089 is available in the nightlies, but we'd still have the same problem as with https://github.com/meta-pytorch/torchcodec/pull/1260: if we merge this PR, it'd be relying on a torch 2.12 feature, so we'd have to be super careful not to cherry-pick it in our releases branches before we release in sync with 2.12.

So let's use the same logic as for `decoderDeleter` and avoid the lambda deleter. Sorry about the churn.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -701,28 +692,23 @@ VideoEncoder::VideoEncoder(
...
+void allocateAVIOFormatContext(
     std::string_view formatName,
-    std::unique_ptr<AVIOContextHolder> avioContextHolder,
-    const VideoStreamOptions& videoStreamOptions)
-    : frames_(validateFrames(frames)),
-      inFrameRate_(frameRate),
-      avioContextHolder_(std::move(avioContextHolder)) {
-  setFFmpegLogLevel();
+    AVIOContextHolder* avioContextHolder,
+    UniqueEncodingAVFormatContext& avFormatContextHolder) {
```
NicolasHug: There are a few issues with these 2 new helpers, mainly:

- `allocateAVIOFormatContext` takes both a raw pointer and a reference as input, both are in-out parameters. We should try to avoid passing raw pointers whenever possible
- `avFormatContextHolder` is a weird name because there is no such thing as an "AVFormatContext holder". There is only `AVFormatContext`.
- `allocateFileFormatContext` is a bit of a weird name since there is no such thing as a "file format context"
- They are used to refactor the video encoder, but not the audio encoder (and they could!)
- Lastly, I understand this wasn't intended, but they make the diff harder to review because they moved the definition of the the `VideoEncoder` constructors in non-linear ways.

Honestly, I'd suggest to remove these helpers and just duplicate the same logic in the new constructors. The logic is currently already duplicated within the AudioEncoder and VideoEncoder. If really we want to refactor, then let's do it as a separate follow-up PR that would also handle the AudioEncoder part. But I suspect this will be largely not needed because we'll probably want to remove both the VideoEncoder and the AudioEncoder down the line.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1012,4 +1012,44 @@ void VideoEncoder::flushBuffers() {
...
+
+  status = avio_open(&avFormatContext_->pb, fileName.data(), AVIO_FLAG_WRITE);
+  STD_TORCH_CHECK(
+      status >= 0,
+      "avio_open failed. The destination file is ",
+      fileName,
+      ", make sure it's a valid path? ",
+      getFFMPEGErrorStringFromErrorCode(status));
+}
+
```
Dan-Flores, line 1043: The lines above are taken from `VideoEncoder`'s file constructor: https://github.com/meta-pytorch/torchcodec/blob/d66806c1fd22243790680b2540eab08bf6ba41dd/src/torchcodec/_core/Encoder.cpp#L690-L712

src/torchcodec/_core/custom_ops.cpp
NicolasHug: I think the comment above is still relevant, The comment above about the deleter is still relevant I think. I think it's safer to use the separate deleter, because that gives us the option to push our next release *before* 2.12 is out.

Dan-Flores: Good point! Fixed

Issue #649, opened 2025-04-23
Migrate docs to new pytorch theme


NicolasHug:
Like: https://pytorch.org/docs/main/index.html

Guide: https://github.com/pytorch/pytorch_sphinx_theme/blob/pytorch_sphinx_theme2/pytorch_sphinx_theme2_migration_guide.md

Internal post: https://fb.workplace.com/groups/pytorch.dev/posts/1777934519451687

mollyxu:
Closed by #1275

PR #1275, opened 2026-03-05
Migrate torchcodec docs to sphinx theme2

mollyxu:
Related issue: #649 

Migrating TorchCodec docs to sphinx_theme2 following the guide here: https://github.com/pytorch/pytorch_sphinx_theme/blob/pytorch_sphinx_theme2/pytorch_sphinx_theme2_migration_guide.md

image

image

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -195,7 +195,7 @@ html_theme_options = {
         },
         {
             "name": "GitHub",
-            "url": "https://github.com/pytorch/torchcodec",
+            "url": "https://github.com/meta-pytorch/torchcodec",
             "icon": "fa-brands fa-github",
         },
         {
@@ -251,9 +251,9 @@ html_context = {
     "theme_variables": theme_variables,
     "display_github": True,
     "github_url": "https://github.com",
-    "github_user": "pytorch",
+    "github_user": "meta-pytorch",
     "github_repo": "torchcodec",
-    "feedback_url": "https://github.com/pytorch/torchcodec",
+    "feedback_url": "https://github.com/meta-pytorch/torchcodec",
     "github_version": "main",
     "doc_path": "docs/source",
     "library_links": [],
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -25,7 +25,7 @@ Installation instructions
 
      .. grid-item-card:: :octicon:`file-code;1em`
         Installation instructions
-        :link: https://github.com/pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec
+        :link: https://github.com/meta-pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec
         :link-type: url
 
         How to install TorchCodec
@@ -115,13 +115,7 @@ Encoding
    :maxdepth: 1
    :hidden:
 
-   Home <self>
-
-.. toctree::
-   :maxdepth: 1
-   :hidden:
-
-   Installation <https://github.com/pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec>
+   Installation <https://github.com/meta-pytorch/torchcodec?tab=readme-ov-file#installing-torchcodec>
 
 .. toctree::
    :maxdepth: 1
--- a/examples/decoding/basic_cuda_example.py
+++ b/examples/decoding/basic_cuda_example.py
@@ -21,7 +21,7 @@ CUDA decoding also uses less PCI-e bandwidth.
 Installing TorchCodec with CUDA Enabled
 ---------------------------------------
 
-Refer to the installation guide in the `README <https://github.com/pytorch/torchcodec#installing-cuda-enabled-torchcodec>`_.
+Refer to the installation guide in the `README <https://github.com/meta-pytorch/torchcodec#installing-cuda-enabled-torchcodec>`_.
 
 """
 


docs/source/conf.py
```diff
@@ -219,8 +245,25 @@ def __call__(self, filename):
...
+    "github_version": "main",
+    "doc_path": "docs/source",
+    "library_links": theme_variables.get("library_links", []),
+    "community_links": theme_variables.get("community_links", []),
+    "language_bindings_links": html_theme_options.get("language_bindings_links", []),
+}
+
 # sitemap config
-html_baseurl = "https://meta-pytorch.org/torchcodec/stable/"
+html_baseurl = f"https://meta-pytorch.org/torchcodec/{version}/"
```
mollyxu: From the migration guidelines, it is suggested that we have a reliable way to acquire the version and structure the url with the version. We could keep using `stable` if the above method is not robust enough

docs/source/conf.py
```diff
@@ -177,26 +176,44 @@ def __call__(self, filename):
...
+        },
+        {
+            "name": "PyPi",
+            "url": "https://pypi.org/project/torchcodec/",
+            "icon": "fa-brands fa-python",
+        },
+    ],
+    "use_edit_page_button": True,
+    "navbar_center": "navbar-nav",
+    "navbar_start": ["navbar-logo", "pytorch_version"],
```
mollyxu: We can consider a version switcher instead of displaying the version

docs/requirements.txt
```diff
@@ -1,13 +1,15 @@
-sphinx-gallery>0.11
-sphinx==5.0.0
-sphinx_design
+sphinx==7.2.6
+-e git+https://github.com/pytorch/pytorch_sphinx_theme.git@pytorch_sphinx_theme2#egg=pytorch_sphinx_theme2
+sphinx-gallery==0.14.0
```
Dan-Flores: I needed to upgrade `sphinx-gallery` to 0.15.0 or 0.16.0 to avoid an error:
`exception: module 'ast' has no attribute 'Str`.

Were you able to build with `sphinx-gallery==0.14.0`? Did you see a similar error?

docs/source/conf.py
Dan-Flores: I have a slight preference for the version switcher, it seems helpful to navigate between versions in the [executorch docs](https://docs.pytorch.org/executorch/stable/index.html).

docs/source/conf.py
```diff
@@ -137,13 +131,18 @@ def __call__(self, filename):
...
+    "_templates",
+    os.path.join(os.path.dirname(pytorch_sphinx_theme2.__file__), "templates"),
+]
 
 # The suffix(es) of source filenames.
 # You can specify multiple suffix as a list of string:
 #
 source_suffix = [".rst"]
 
+version = ".".join(torchcodec.__version__.split(".")[:2])
```
Dan-Flores, line 144: Is this variable currently needed? Is it related to the version switching functionality?

Dan-Flores:
Thanks for working on this @mollyxu! I left some comments, but generally this looks good!

docs/requirements.txt
NicolasHug: It might be something else, I'm able to using 0.14. Although ideally we wouldn't enforce a hard pin with `==`, and instead just enforce a lower bound with `>=` on most dependencies

```
~/dev/torchcodec-cuda/docs (migrate-docs*) » pip freeze | grep galler 
sphinx-gallery==0.14.0
```

docs/source/api_ref.rst
```diff
@@ -0,0 +1,11 @@
+API Reference
+=============
+
+.. toctree::
+   :maxdepth: 1
+
+   api_ref_torchcodec
+   api_ref_decoders
```
NicolasHug, line 8: Make sure to sync with `main`, we recently added some cuda helpers like `set_nvdec_cache_capacity` and others that I do not see being displayed

mollyxu: Thanks for the review! In our current docs, we have the PyTorch logo and text in our navbar. Would having a different style be something we want out of this migration? 
image

docs/requirements.txt
mollyxu: Thanks for the tips! I think the discrepancy might be related to python version. I will change this to a lower bound

docs/source/conf.py
mollyxu: Thanks for the review! I left that in for flexibility when we are switching to navbar with version switcher. It is now in use

mollyxu:
For our next release, we would have to update the version switcher in our navbar by doing the following
  1. Add "0.11 (stable)" with "preferred": true to `docs/source/_static/torchcodec-versions.json`                                 
  2. Remove preferred from the 0.10 entry

NicolasHug:
@mollyxu just flagging https://github.com/meta-pytorch/torchcodec/pull/1276 which will likely cause small merge conflicts, but it should be easy to address

Dan-Flores:
> For our next release, we would have to update the version switcher in our navbar

Thanks for flagging this, it presents an unusual order of operations problem:

1. The `release/0.X` branch (and `main`) must include a PR to update the version switcher `preferred` variable. 
2. The [`upload-docs`](https://github.com/meta-pytorch/torchcodec/blob/21c6a8a0606dde2e816513b01a8014469db7e16c/.github/workflows/linux_cuda_wheel.yaml#L227) job will kick off when the PR lands, which updates the `main/` docs to use the new `preferred` version.
3. The last step in a release is to update the `gh-pages` branch to publish the newly built docs. 

Between 2 and 3, there is a period of time where the docs page's `preferred` version does not exist. I'm not sure how big a problem that is in reality, but I think we can avoid it by hosting the `versions.json` in `gh-pages`, similar to [how executorch does](https://github.com/pytorch/executorch/blob/gh-pages/executorch-versions.json). 

This way, the step of updating the `preferred` version happens at the same time as the other `gh-pages` updates. Let me know if this makes sense!

docs/source/conf.py
```diff
@@ -177,26 +176,52 @@ def __call__(self, filename):
...
+    "show_lf_header": True,
+    "icon_links": [
+        {
+            "name": "X",
+            "url": "https://x.com/PyTorch",
+            "icon": "fa-brands fa-x-twitter",
+        },
+        {
+            "name": "GitHub",
+            "url": "https://github.com/pytorch/torchcodec",
```
NicolasHug:
```suggestion
            "url": "https://github.com/meta-pytorch/torchcodec",
```

docs/source/conf.py
```diff
@@ -219,6 +244,23 @@ def __call__(self, filename):
...
+# html_context for theme2
+theme_variables = pytorch_sphinx_theme2.get_theme_variables()
+
+html_context = {
+    "theme_variables": theme_variables,
+    "display_github": True,
+    "github_url": "https://github.com",
+    "github_user": "pytorch",
+    "github_repo": "torchcodec",
+    "feedback_url": "https://github.com/pytorch/torchcodec",
```
NicolasHug:
```suggestion
    "feedback_url": "https://github.com/meta-pytorch/torchcodec",
```

PR #1276, opened 2026-03-06
Minor doc improvements

NicolasHug:
Just moved `set_cuda_backend` and other CUDA-related utils into a new section under the docs.


src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -99,7 +99,6 @@ def get_nvdec_cache_capacity() -> int:
     See also :func:`~torchcodec.decoders.set_nvdec_cache_capacity`.
 
     Returns:
-        int: The maximum number of NVDEC decoders that can be cached per GPU
-            device.
+        int: The maximum number of NVDEC decoders that can be cached per GPU device.
```
NicolasHug, line 102: Driveby, rendering was off

docs/source/api_ref_decoders.rst
```diff
@@ -22,17 +22,21 @@ For an audio decoder tutorial, see: :ref:`sphx_glr_generated_examples_decoding_a
 .. autosummary::
     :toctree: generated/
     :nosignatures:
-    :template: function.rst
+    :template: dataclass.rst
+
+    VideoStreamMetadata
+    AudioStreamMetadata
+    CpuFallbackStatus
```
Dan-Flores: It seems `CpuFallbackStatus` should also go under the new CUDA header, since its only used during CUDA decoding.

docs/source/api_ref_decoders.rst
```diff
@@ -33,6 +45,4 @@ For an audio decoder tutorial, see: :ref:`sphx_glr_generated_examples_decoding_a
     :nosignatures:
     :template: dataclass.rst
 
-    VideoStreamMetadata
-    AudioStreamMetadata
     CpuFallbackStatus
```
Dan-Flores, line 48: I have no strong opinon, just checking that this is rendering as intended with a gap between the `cuda`/`nvdec` functions and  `CpuFallbackStatus`?
Image

NicolasHug: It's a bit annoying but I don't know how to avoid it. It's related to the fact that the first 3 instances are functions and can thus rely on the same rst template (for functions), while CPUFallbackStatus is a class that needs the class template.

I hope the new sphinx theme will help make all this look nicer

PR #1281, opened 2026-03-09
WavDecoder parses WAV headers

Dan-Flores:
This PR implements header parsing to validate if a file is a WAV, and store its metadata.
* Header parsing is based on the standard WAV format, [see this doc](https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html). It covers:
  * `RIFF` header fields and lengths
  * `fmt` chunk fields and lengths 
* `WavDecoder` is technically private, since its not exported from the [top level decoders library](https://github.com/meta-pytorch/torchcodec/blob/b66b35fa54a18f50fb4d4ae86009f6d3528df590/src/torchcodec/decoders/__init__.py#L8).
  * We could use the name `_WavDecoder`, but IMO its not a strong defense against usage, and it adds the step of renaming it later.
  
See inline comments for additional details.

### DIFF
--- a/src/torchcodec/_core/WavDecoder.cpp
+++ b/src/torchcodec/_core/WavDecoder.cpp
@@ -64,12 +64,12 @@ bool matchesFourCC(
     const uint8_t* data,
     size_t dataSize,
     size_t offset,
-    const char* expected) {
+    std::string_view expected) {
   STD_TORCH_CHECK(
       dataSize >= 4 && offset <= dataSize - 4,
       "Data array too small for FourCC comparison at offset ",
       offset);
-  return std::memcmp(data + offset, expected, 4) == 0;
+  return std::memcmp(data + offset, expected.data(), 4) == 0;
 }
 
 template <typename Container>
@@ -196,7 +196,7 @@ void WavDecoder::validateHeader() const {
 // Given a chunkId, read through each chunk until we find a match, then return
 // its offset and size.
 WavDecoder::ChunkInfo WavDecoder::findChunk(
-    const char* chunkId,
+    std::string_view chunkId,
     uint64_t startPos,
     uint64_t fileSize) {
   if (fileSize < CHUNK_HEADER_SIZE) {
@@ -212,15 +212,8 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
     uint32_t chunkSize = readValue<uint32_t>(chunkHeader, 4);
 
     if (matchesFourCC(chunkHeader.data(), chunkHeader.size(), 0, chunkId)) {
-      STD_TORCH_CHECK(
-          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
-          "File position arithmetic would overflow");
       return {startPos + CHUNK_HEADER_SIZE, chunkSize};
     }
-    STD_TORCH_CHECK(
-        chunkSize <= UINT64_MAX - CHUNK_HEADER_SIZE - (chunkSize % 2),
-        "Chunk size would cause overflow: ",
-        chunkSize);
     // Skip this chunk and continue searching (odd chunks are padded)
     uint64_t numBytesToSkip =
         CHUNK_HEADER_SIZE + static_cast<uint64_t>(chunkSize) + (chunkSize % 2);
--- a/src/torchcodec/_core/WavDecoder.h
+++ b/src/torchcodec/_core/WavDecoder.h
@@ -9,6 +9,7 @@
 #include <cstdint>
 #include <fstream>
 #include <string>
+#include <string_view>
 #include "StableABICompat.h"
 
 namespace facebook::torchcodec {
@@ -37,10 +38,14 @@ class WavDecoder {
   struct ChunkInfo {
     uint64_t offset;
     uint32_t size;
+
+    ChunkInfo(uint64_t offset, uint32_t size) : offset(offset), size(size) {}
   };
 
-  ChunkInfo
-  findChunk(const char* chunkId, uint64_t startPos, uint64_t fileSizeLimit);
+  ChunkInfo findChunk(
+      std::string_view chunkId,
+      uint64_t startPos,
+      uint64_t fileSizeLimit);
   void parseHeader(uint64_t actualFileSize);
   void validateHeader() const;
 


src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+  return static_cast<int64_t>(bytesRead);
+}
+
+int64_t WavFileReader::seek(int64_t position) {
+  STD_TORCH_CHECK(file_ != nullptr, "WAV file handle is null");
+#ifdef _WIN32
+  if (_fseeki64(file_, position, SEEK_SET) != 0) {
+#else
+  if (fseeko(file_, static_cast<off_t>(position), SEEK_SET) != 0) {
+#endif
```
Dan-Flores: The [Windows builds were failing](https://github.com/meta-pytorch/torchcodec/actions/runs/22873974634/job/66360825023?pr=1281) since `fseeko` is not defined on Windows, so I added an `#ifdef` here, but I am open to relocating this elsewhere.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+    return -1;
+  }
+  return position;
+}
+
+void WavDecoder::parseHeader() {
```
Dan-Flores: This function is where the structure described in the [sections "Wave File Format" and "fmt Chunk"](https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html) are relevant.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+WavDecoder::WavDecoder(std::unique_ptr<WavReader> reader)
+    : reader_(std::move(reader)) {
+  parseHeader();
+  validate();
+}
+
+uint16_t WavDecoder::getEffectiveFormat() const {
+  return (header_.audioFormat == WAV_FORMAT_EXTENSIBLE) ? header_.subFormat
+                                                        : header_.audioFormat;
```
Dan-Flores: This function is needed because when `WAVE_FORMAT_EXTENSIBLE` is set as the audio format, we must check the subformat field instead.
See the ["Extensible Format" header here](https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html)

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+// Given a chunkId, read through each chunk until we find a match, then return
+// its offset and size.
+WavDecoder::ChunkInfo WavDecoder::findChunk(
```
Dan-Flores, line 198: There are only two chunks we will need to find: `"fmt "` (in this PR) and `"data"` (later). This function abstracts the chunk searching logic out. 
We need to search because the positions of these chunks are not fixed, there are optional chunks that can change the positions of `fmt` and `data`.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,276 @@
...
+      effectiveFormat != WAV_FORMAT_IEEE_FLOAT) {
+    STD_TORCH_CHECK(
+        false,
+        "Unsupported WAV format: ",
+        effectiveFormat,
+        ". Only PCM and IEEE float formats are supported.");
+  }
+
+  if (effectiveFormat == WAV_FORMAT_PCM) {
+    // TODO: support 8, 16, 24 bits
```
Dan-Flores: For now, we are only testing 32 bit formats. In later PRs, we can check in additional WAV files and extend the `validate()` function.

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,79 @@
...
+#include <string>
+#include "StableABICompat.h"
+
+namespace facebook::torchcodec {
+
+constexpr uint16_t WAV_FORMAT_PCM = 1;
+constexpr uint16_t WAV_FORMAT_IEEE_FLOAT = 3;
+constexpr uint16_t WAV_FORMAT_EXTENSIBLE = 0xFFFE;
+
+struct WavHeader {
```
NicolasHug: It's worth linking to a ref like https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html (or a relevant section in there) to document the types and their respective lengths.

src/torchcodec/decoders/_wav_decoder.py
```diff
@@ -0,0 +1,65 @@
...
+    Args:
+        source (str or ``Pathlib.path``): Path to a WAV file.
+
+    Attributes:
+        metadata (AudioStreamMetadata): Metadata of the audio stream.
+        stream_index (int): Always 0 for WAV files (single stream).
+
+    Raises:
+        ValueError: If the source is not a supported uncompressed WAV file.
+    """
```
NicolasHug: I'd recommend to leave out the user-facing docstrings for now. It'll be easier to review them all at once at the end once we focus on the user-facing docs, rather than incrementally in each PR, as they provide too much of an opportunity for bike-shedding.

src/torchcodec/_core/_metadata.py
```diff
@@ -158,6 +158,30 @@ def __repr__(self):
         return super().__repr__()
 
 
+@dataclass
+class WavStreamMetadata(AudioStreamMetadata):
```
NicolasHug: From the tests, it looks like this struct and its fields are meant to be public. If that's the case, then we have a design question on our hands now: do we want / need to expose these fields?


I don't know the answer to that question, but do we need an answer _now_ ? I suspect the answer to that is _no_: we don't need to think about what kind of extra metadata we'll want to expose (if any). We can probably answer that question waaay down the line, possibly even after the first release of the wav decoder.

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,79 @@
...
+  uint16_t subFormat = 0; // Extracted from SubFormat GUID (first 2 bytes)
+  uint16_t validBitsPerSample = 0;
+};
+
+class WavReader {
+ public:
+  virtual ~WavReader() = default;
+  virtual int64_t read(void* buffer, int64_t size) = 0;
+  virtual int64_t seek(int64_t position) = 0;
+};
```
NicolasHug: The methods of this `WavReader` base class are very close to what the `AVIOContextHolder` was made for https://github.com/meta-pytorch/torchcodec/blob/c5b29cf6b86961b889b1349bdfded4b842f6faf2/src/torchcodec/_core/AVIOContextHolder.h#L50-L51

I suspect we should be able to reuse the functionality of `AVIOContextHolder` when we'll want to support other kinds of inputs. I don't know if we'll _want_ to, but it sounds like a natural fit.

However, since this is quickly going to become a design discussion, I think it's worth taking a step back and asking a similar question to the one about exposing the new metadata fields: we _probably_ do not need to think about this now. We don't have to prioritize support for file-like and tensor inputs right now, this can come later.

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -1089,6 +1117,20 @@ void scan_all_streams_to_update_metadata(torch::stable::Tensor& decoder) {
...
+  auto decoder = std::make_unique<WavDecoder>(std::move(reader));
+  return wrapWavDecoderPointerToTensor(std::move(decoder));
+}
+
+torch::stable::Tensor get_wav_all_samples(torch::stable::Tensor& decoder) {
+  auto wavDecoder = unwrapTensorToGetWavDecoder(decoder);
+  (void)wavDecoder; // Suppress unused variable warning
+  STD_TORCH_CHECK(false, "get_wav_all_samples not yet implemented");
+  return torch::stable::empty({0}, kStableFloat32);
+}
```
Dan-Flores, line 1150: `get_wav_all_samples()` is not implemented yet, but its used to call `unwrapTensorToGetWavDecoder()` to avoid an unused function linter. 

I do this so that `wrapWavDecoderPointerToTensor()` and `unwrapTensorToGetWavDecoder()` can be landed together in this PR.

src/torchcodec/_core/WavDecoder.h
Dan-Flores: Thanks for this feedback, I've removed the `WavReader` abstraction. 
The current design has `WavDecoder` accept a `WavFileReader` so we can change the behavior of `read()` and seek()` later on with a light refactor.

src/torchcodec/_core/_metadata.py
Dan-Flores: Yeah, I might have got a bit carried away with the metadata fields!

For now, I'll remove metadata from the Python API. I would suggest we expose the same metadata as `AudioDecoder`, but we can have that discussion separately.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+      return {startPos + 8, chunkSize};
+    }
+
+    // Skip this chunk and continue searching
+    startPos += 8 + chunkSize;
+    STD_TORCH_CHECK(
+        static_cast<uint64_t>(startPos) <= header_.fileSize,
+        "Chunk extends beyond file bounds at position: ",
+        startPos);
+    fseek(file_, static_cast<long>(startPos), SEEK_SET);
```
Dan-Flores: `fseek` expects `long offset`, so we cast here and above. We could use `fseeko` and avoid the cast, but that requires an `#ifdef` to use an equivalent function in windows (`_fseeki64`).

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+  uint16_t effectiveFormat = getEffectiveFormat();
+  // Only uncompressed formats are supported
+  if (effectiveFormat != WAV_FORMAT_PCM &&
+      effectiveFormat != WAV_FORMAT_IEEE_FLOAT) {
+    STD_TORCH_CHECK(
+        false,
+        "Unsupported WAV format: ",
+        effectiveFormat,
+        ". Only PCM and IEEE float formats are supported.");
+  }
```
NicolasHug, line 204: Just use `STD_TORCH_CHECK(cond)` instead of `if(cond) {STD_TORCH_CHECK(false)}`

src/torchcodec/decoders/_wav_decoder.py
```diff
@@ -0,0 +1,43 @@
...
+    # TODO: Docstrings
+    def __init__(self, source: str | Path):
+        torch._C._log_api_usage_once("torchcodec.decoders.WavDecoder")
+
+        self._decoder = _try_create_wav_decoder(source)
+        if self._decoder is None:
+            raise ValueError(
+                "Source is not a supported uncompressed WAV file. "
+                "For compressed audio formats or non-WAV files, use AudioDecoder instead."
+            )
```
NicolasHug: The `_try_create_wav_decoder` and the way we raise exceptions above are a bit unconventional (also, we shouldn't `print()` anything). Is that a relic of when we'd try to transparently dispatch to the WavDecoder within AudioDecoder? In other classes, we don't have a "try" logic, we just let each core op raise its own exceptions. Could we simplify this?

src/torchcodec/_core/custom_ops.cpp
```diff
@@ -109,6 +111,32 @@ torch::stable::Tensor wrapDecoderPointerToTensor(
...
+
+  int64_t sizes[] = {static_cast<int64_t>(sizeof(WavDecoder*))};
+  int64_t strides[] = {1};
+  torch::stable::Tensor tensor = torch::stable::from_blob(
+      decoder,
+      {sizes, 1},
+      {strides, 1},
+      StableDevice(kStableCPU),
+      kStableInt64,
+      [](void* data) { delete static_cast<WavDecoder*>(data); });
```
NicolasHug: We'll have to do the same thing as in https://github.com/meta-pytorch/torchcodec/pull/1272#discussion_r2917782769

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,52 @@
...
+#include <cstdint>
+#include <memory>
+#include <optional>
+#include <string>
+#include "StableABICompat.h"
+
+namespace facebook::torchcodec {
+constexpr uint16_t WAV_FORMAT_PCM = 1;
+constexpr uint16_t WAV_FORMAT_IEEE_FLOAT = 3;
+constexpr uint16_t WAV_FORMAT_EXTENSIBLE = 0xFFFE;
```
NicolasHug: These values too seem to be part of the standard - maybe have a joint comment linking to https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html for all of those

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,52 @@
...
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
+
+#pragma once
+
+#include <cstdint>
+#include <memory>
+#include <optional>
```
NicolasHug: double-check the headers please, claude tells me `memory` and `optional` aren't needed

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: `claude` said this - is this valid?

```
  2. findChunk doesn't handle odd-sized chunks (WavDecoder.cpp:119): Per the RIFF spec, chunks are padded to even byte
  boundaries. If chunkSize is odd, you need to skip chunkSize + 1 bytes (the padding byte). Without this, you'll
  misalign and fail to find subsequent chunks in some valid WAV files. Fix: startPos += 8 + ((chunkSize + 1) & ~1); or
  simply startPos += 8 + chunkSize + (chunkSize % 2).
```

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,52 @@
...
+    int64_t offset;
+    uint32_t size;
+  };
+
+  uint16_t getEffectiveFormat() const;
+  ChunkInfo findChunk(const char* chunkId, int64_t startPos);
+  void parseHeader();
+  void validate() const;
+
+  std::FILE* file_;
```
NicolasHug: We might want to use the C++ `std::ifstream` rather than the C-like FILE stuff. We'd get RAII semantics and all the good stuff that comes with it.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+    startPos += 8 + chunkSize;
+    STD_TORCH_CHECK(
+        static_cast<uint64_t>(startPos) <= header_.fileSize,
+        "Chunk extends beyond file bounds at position: ",
+        startPos);
+    fseek(file_, static_cast<long>(startPos), SEEK_SET);
+  }
+}
+
+void WavDecoder::validate() const {
```
NicolasHug:
```suggestion
void WavDecoder::validateHeader() const {
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+// Given a chunkId, read through each chunk until we find a match, then return
+// its offset and size.
+WavDecoder::ChunkInfo WavDecoder::findChunk(
+    const char* chunkId,
+    int64_t startPos) {
+  fseek(file_, static_cast<long>(startPos), SEEK_SET);
+
+  while (true) {
+    // Read chunk header (4B ID + 4B size)
+    uint8_t chunkHeader[8];
```
NicolasHug: hardcode 8 as CHUNK_HEADER_SIZE (or something that makes sense).

We want to give a name to all the instances of `8` there are here.


Try to do that for other instances (e.g. 12 is `RIFF_HEADER_SIZE` (I think?))

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+    uint32_t chunkSize = readLittleEndian<uint32_t>(chunkHeader + 4);
+
+    if (checkFourCC(chunkHeader, chunkId)) {
+      return {startPos + 8, chunkSize};
+    }
+
+    // Skip this chunk and continue searching
+    startPos += 8 + chunkSize;
+    STD_TORCH_CHECK(
+        static_cast<uint64_t>(startPos) <= header_.fileSize,
```
NicolasHug: This is trusting `header_.fileSize` to be correct, right? We'll also want to make sure we're not seeking past the *actual* file size, in case `header_.fileSize` is wrong or corrupted.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+namespace facebook::torchcodec {
+namespace {
+
+template <typename T>
+T readLittleEndian(const uint8_t* data) {
+  T value;
+  std::memcpy(&value, data, sizeof(T));
+  return value;
+}
```
NicolasHug, line 61: So this doesn't really "read a little endian value" or "read a value and put make it little endian"

It "reads raw value regardless of endianness" and since wav files store values in little endian, that raw value will only be interpreted correctly on little-endian architectures.

Let's just call it `readValue`.

Separately, for now we can probably enforce little endianness of the host machine. In the future we'll want to handle big endian archs, but that can come later. We can figure out endianness at runtime with

```
bool is_little_endian() {
  uint32_t x = 1;
  return *(uint8_t*)&x;
}
```

which is something we have in torchvision as well.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+    STD_TORCH_CHECK(
+        fmtChunk.size >= 40, "WAVE_FORMAT_EXTENSIBLE fmt chunk too small");
+    // SubFormat GUID starts at offset 24, first 2 bytes are the format code
+    header_.subFormat = readLittleEndian<uint16_t>(fmtData.data() + 24);
+  }
+
+  // TODO: Find data chunk
+}
+
+WavDecoder::WavDecoder(const std::string& path) : file_(nullptr) {
```
NicolasHug: Nit: move constructors and destructors at the top of the file, and try to put related methods next to each other

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: Does it need to be a function though? It seems that it could be inlined? We're only calling it once

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+      bytesRead,
+      ")");
+
+  STD_TORCH_CHECK(checkFourCC(riffHeader, "RIFF"), "Missing RIFF header");
+  STD_TORCH_CHECK(
+      checkFourCC(riffHeader + 8, "WAVE"), "Missing WAVE format identifier");
+
+  header_.fileSize = readLittleEndian<uint32_t>(riffHeader + 4) + 8;
+
+  // Find and parse fmt chunk (start search after 12-byte RIFF header)
```
NicolasHug: Make a pass on some of these comments - I find some of them to be a bit claude-y (i.e. purely descriptive of what's happening just below, and thus not useful)

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: In addition to the above: let's not rely on `header_` at all in this method. When `findChunk` is called, `header_` is not fully initialized yet, so the method can currently be used in contexts that are invalid states. We should design methods that prevent that from happening, in this case we shouldn't rely on the assumption that `header_` is initialized with just the right amount of info for `findChunk` to work. Any info that is needed for `findChunk` could just be passed as parameter.

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: You can avoid this fseek call as well as the first one by having a single one at the very beginning of the `while(true)` loop, I think. While we're at it, we should probably have the STD_TORCH_CHECK just before as well.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,169 @@
...
+  ChunkInfo fmtChunk = findChunk("fmt ", 12);
+  STD_TORCH_CHECK(
+      fmtChunk.size >= 16, "Invalid fmt chunk: size must be at least 16 bytes");
+
+  // Use ChunkInfo to seek to and read the fmt chunk data
+  fseek(file_, static_cast<long>(fmtChunk.offset), SEEK_SET);
+  std::vector<uint8_t> fmtData(fmtChunk.size);
```
NicolasHug: We probably want to check `fmtChunk.size` to something small before we allocate a vector of that size, otherwise a maliciouly crafter file could cause us to allocated many GBs.

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: In addition to that, we should generally be extremely paranoid about anything we read: that includes correctness of the value from the headers, but also just generally making sure we're not reading past files or past buffers.

Here's there's no validation that the `data` buffer is large enough to include a value of type `T`. We should probably pass the buffer size as input and validate that offset + sizeof(T) <= bufferSize.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Good point - a lot of the claude-y comments are me trying to explain what these magic numbers are, but they're better documented as constants.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Thanks, I had not thought of this, it seems like this is an existing example of this type of check?
https://github.com/meta-pytorch/torchcodec/blob/21c6a8a0606dde2e816513b01a8014469db7e16c/src/torchcodec/_core/AVIOTensorContext.cpp#L15

Is it our responsibility to defend against maliciously crafted files?

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Not at the moment, lets inline it.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: It would seem so, I haven't encountered odd byte sizes yet but there are many chunk types. The suggested fix is clever, I've added it.

src/torchcodec/_core/WavDecoder.h
Dan-Flores: Thanks for raising this early, I'll make the switch since it means we no longer need to explicitly call `std::fclose` in the destructor, which is safer.

Would you know why `FILE*` stuff is still used elsewhere? I was referring to [this torchvision file](https://github.com/pytorch/vision/blob/b8667f0eea20b9c6429a9324d418921a6e546fd2/torchvision/csrc/io/image/cpu/read_write_file.cpp#L65C3-L65C8) as a reference for IO reading functions.

src/torchcodec/_core/WavDecoder.cpp
NicolasHug: It's a similar check yeah, we don't want to allocate something that's too big. In both cases, the allocated size depends on something that was provided by the user.

Yes, the moment we start reading user-provided data, we have to be defensive and we are responsible for avoiding crashes, segfaults, memory corruptions, invalid accesses, DOS attacks (etc.)

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,183 @@
...
+      ". Only PCM and IEEE float formats are supported.");
+
+  if (effectiveFormat == WAV_FORMAT_PCM) {
+    // TODO WavDecoder: support 8, 16, 24 bits
+    if (header_.bitsPerSample != 32) {
+      STD_TORCH_CHECK(
+          false,
+          "Unsupported PCM bit depth: ",
+          header_.bitsPerSample,
+          ". Currently supported bit depths are: 32)");
```
mollyxu: nit: extra parenthesis here and on line 148

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,183 @@
...
+      RIFF_HEADER_SIZE,
+      " bytes, got ",
+      file_.gcount(),
+      ")");
+
+  STD_TORCH_CHECK(matchesFourCC(riffHeader, "RIFF"), "Missing RIFF header");
+  STD_TORCH_CHECK(
+      matchesFourCC(riffHeader + 8, "WAVE"), "Missing WAVE format identifier");
+
+  header_.fileSize = readValue<uint32_t>(riffHeader, 4, RIFF_HEADER_SIZE) + 8;
```
mollyxu: Will `fileSize` be used somewhere later in followup PRs? Do we plan on checking it against `actualFileSize`?

Dan-Flores: This is not needed now that we pass `actualFileSize` so I'll remove this, thanks!

We could verify that the reported file size in the header is accurate, but that might be overly strict without much benefit.

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,61 @@
...
+constexpr size_t CHUNK_HEADER_SIZE = 8; // chunkID + chunkSize
+constexpr size_t MIN_FMT_CHUNK_SIZE = 16;
+constexpr size_t MIN_WAVEX_FMT_CHUNK_SIZE = 40;
+constexpr uint32_t MAX_FMT_CHUNK_SIZE = 1000; // 1 KB
+
+// See standard format codes and Wav file format used in WavHeader:
+// https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html
+constexpr uint16_t WAV_FORMAT_PCM = 1;
+constexpr uint16_t WAV_FORMAT_IEEE_FLOAT = 3;
+constexpr uint16_t WAV_FORMAT_EXTENSIBLE = 0xFFFE;
```
NicolasHug: All of the constants above are currently declared as fully public, they'll be visible to the whole torchcodec namespace. Since they're only used within the cpp file, we should encapsulate them there, in a private namespace.

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,61 @@
...
+
+struct WavHeader {
```
NicolasHug: Similarly to the comment just above: this `WavHeader` class is not used anywhere outside of this file. But since it's currently publicly visible, if we make a change to WavHeader, then any other file that depends on `WavDecoder.h` will need to be recompiled.

Let's move it as a private struct within `WavDecoder`. We might have to revisit that in the future if we ever want to expose some extra wav-specific metadata, but we don't know that it'll be the case just yet.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+#include <fstream>
+#include <vector>
+
+namespace facebook::torchcodec {
+namespace {
+
+bool is_little_endian() {
+  uint32_t x = 1;
+  return reinterpret_cast<const uint8_t*>(&x)[0];
+}
```
NicolasHug: Now that we are on C++ 20 (https://github.com/meta-pytorch/torchcodec/pull/1279), we can use https://en.cppreference.com/w/cpp/types/endian.html and just have something like 

```C
static_assert(std::endian::native == std::endian::little,
                "WAV decoder requires little-endian architecture");
```

which is a compile-time error instead of a runtime error.

If we decide to keep the runtime error, this should be `isLittleEndian`, and claude tells me there's a slightly safer way (worth double checking!):

```
● The current implementation uses reinterpret_cast which is technically undefined behavior (type punning through pointer cast violates strict aliasing). Using memcpy is the  
  well-defined way:                                   
                                                                                                                                                                              
  bool is_little_endian() {                                                                                                                                                   
    uint32_t x = 1;                                                                                                                                                           
    uint8_t first_byte;                               
    std::memcpy(&first_byte, &x, 1);
    return first_byte == 1;
  }

  memcpy for type punning is explicitly blessed by the standard, and compilers optimize it away completely — it'll compile to the same thing (or just a constant) in practice.
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+      ": exceeds buffer length ",
+      bufferSize);
+  T value;
+  std::memcpy(&value, data + offset, sizeof(T));
+  return value;
+}
+
+bool matchesFourCC(const uint8_t* data, const char* expected) {
+  return std::memcmp(data, expected, 4) == 0;
+}
```
NicolasHug: There's no bound check in `matchesFourCC`. For now, we only call it on buffers that have been implicitly validated to have enough data, but we should at least document here that the caller should ensure `data` have at least 4 bytes.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+  uint64_t actualFileSize = std::filesystem::file_size(path);
+  parseHeader(actualFileSize);
+  validateHeader();
+}
+
+void WavDecoder::parseHeader(uint64_t actualFileSize) {
+  file_.seekg(0, std::ios::beg);
+
+  uint8_t riffHeader[RIFF_HEADER_SIZE];
+  file_.read(reinterpret_cast<char*>(riffHeader), RIFF_HEADER_SIZE);
```
NicolasHug: We could consider wrapping all the `read() + STD_TORCH_CHECK(file(), gcount())` calls into a `safeReadFile()` helper

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+      "Invalid fmt chunk: size must be at least ",
+      MIN_FMT_CHUNK_SIZE,
+      " bytes");
+  STD_TORCH_CHECK(
+      fmtChunk.size <= MAX_FMT_CHUNK_SIZE,
+      "We tried to allocate fmt chunk of ",
+      fmtChunk.size,
+      " bytes, but maximum allowed is ",
+      MAX_FMT_CHUNK_SIZE,
+      " bytes.");
```
NicolasHug: Assuming we'll be reusing `findChunk()` in other places, it's best to validate `chunkSize` within `findChunk` itself, before returning, instead of leaving the validation up to the callers.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+          file_.gcount() == static_cast<std::streamsize>(fmtChunk.size),
+      "WAV: unexpected end of data (expected ",
+      fmtChunk.size,
+      " bytes, got ",
+      file_.gcount(),
+      ")");
+
+  header_.audioFormat = readValue<uint16_t>(fmtData.data(), 0, fmtChunk.size);
+  header_.numChannels = readValue<uint16_t>(fmtData.data(), 2, fmtChunk.size);
+  header_.sampleRate = readValue<uint32_t>(fmtData.data(), 4, fmtChunk.size);
```
NicolasHug: Above and in other relevant places, we should be passing `fmtData.size()`, not `fmtChunk.size`.

`fmtData.size()` *is* the size of the buffer we're passing, i.e. it's the size of `fmtData.data()`, which is what needs to be validated.

It happens to be equal to `fmtChunk.size`, because of the logic we have defined above, but that's not an invariant we want to rely on when validating the sizes.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+  uint64_t actualFileSize = std::filesystem::file_size(path);
+  parseHeader(actualFileSize);
+  validateHeader();
+}
+
+void WavDecoder::parseHeader(uint64_t actualFileSize) {
+  file_.seekg(0, std::ios::beg);
+
+  uint8_t riffHeader[RIFF_HEADER_SIZE];
```
NicolasHug: We're reading into an array here, whereas below we're reading into an std::vector. For simplicity, let's just always read into std::vector everywhere. This way, we can then simplify `readValue` by not passing the `bufferSize` parameter, as we can just run the validation within `readValue` itslef, something like

```C++
  template <typename T>                                                                                                                                                       
  T readValue(const std::vector<uint8_t>& data, size_t offset) {                                                                                                              
    static_assert(std::is_trivially_copyable_v<T>);                                                                                                                           
    STD_TORCH_CHECK(                                  
        offset + sizeof(T) <= data.size(),
        "Reading ", sizeof(T), " bytes at offset ", offset,
        ": exceeds buffer length ", data.size());
    T value;
    std::memcpy(&value, data.data() + offset, sizeof(T));
    return value;
  }
```


(BTW, note the `static_assert(std::is_trivially_copyable_v<T>);` which we should do regardless)

NicolasHug: EDIT: actually we could still use an std::array here (and in `findChunk`), since it would avoid a dynamic allocation (although not super critical).

This would become something like this, ultimately the goal is the same, i.e. to simplify `readValue()`:

```C++

std::array<uint8_t, RIFF_HEADER_SIZE> riffHeader;

//... 
  template <typename T, typename Container>                                                                                                                                   
  T readValue(const Container& data, size_t offset) { 
    static_assert(std::is_trivially_copyable_v<T>);
    STD_TORCH_CHECK(
        offset + sizeof(T) <= data.size(), ...);
    T value;
    std::memcpy(&value, data.data() + offset, sizeof(T));
    return value;
  }
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+  STD_TORCH_CHECK(header_.sampleRate > 0, "Invalid WAV: zero sample rate");
+}
+
+// Given a chunkId, read through each chunk until we find a match, then return
+// its offset and size.
+WavDecoder::ChunkInfo WavDecoder::findChunk(
+    const char* chunkId,
+    int64_t startPos,
+    uint64_t fileSizeLimit) {
+  while (startPos + CHUNK_HEADER_SIZE <= fileSizeLimit) {
```
NicolasHug: Make sure to double check all the integer comparison and assignment. This one is comparing a signed startPos to the unsigned fileSizeLimit. We should have a static_cast somewhere (which I think the internal linter would have caught). There are probably other instances of this

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+    file_.seekg(startPos, std::ios::beg);
+
+    uint8_t chunkHeader[CHUNK_HEADER_SIZE];
+    file_.read(reinterpret_cast<char*>(chunkHeader), CHUNK_HEADER_SIZE);
+    STD_TORCH_CHECK(
+        !file_.fail() && file_.gcount() == CHUNK_HEADER_SIZE,
+        "Chunk not found: ",
+        chunkId);
+    // Read chunk size which immediately follows the chunk ID
+    uint32_t chunkSize = readValue<uint32_t>(chunkHeader, 4, CHUNK_HEADER_SIZE);
```
NicolasHug: I think we only need `chunkSize` if we have found the right chunk, so this could just be within the `if (matchesFourCC(chunkHeader, chunkId)) {` block? I was a bit surprised by the order of operations here.

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,61 @@
...
+#include <fstream>
+#include <string>
+#include "StableABICompat.h"
+
+namespace facebook::torchcodec {
+constexpr size_t RIFF_HEADER_SIZE = 12; // "RIFF" + fileSize + "WAVE"
+constexpr size_t CHUNK_HEADER_SIZE = 8; // chunkID + chunkSize
+constexpr size_t MIN_FMT_CHUNK_SIZE = 16;
+constexpr size_t MIN_WAVEX_FMT_CHUNK_SIZE = 40;
+constexpr uint32_t MAX_FMT_CHUNK_SIZE = 1000; // 1 KB
```
NicolasHug: All 3 above, it's worth documenting how these were derived. I think I can figure out `MIN_FMT_CHUNK_SIZE` from https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html, but not the rest.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+    if (matchesFourCC(chunkHeader, chunkId)) {
+      return {startPos + static_cast<int64_t>(CHUNK_HEADER_SIZE), chunkSize};
+    }
+
+    // Skip this chunk and continue searching (odd chunks are padded)
+    startPos += CHUNK_HEADER_SIZE + chunkSize + (chunkSize % 2);
```
NicolasHug: Here as well we're having a bunch of implicit type conversions that are worth clearing out. We also want to make sure the `CHUNK_HEADER_SIZE + chunkSize` addition doesn't overflow in uint32.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Thanks for the suggestion! Since the endianness of the users machine is only important for `WavDecoder`, let's keep the runtime error.  

If I understand your suggestion correctly, the compile-time error would appear when building `torchcodec` from source on a big endian machine, but not when installing pre-built wheels via `pip` (which accounts for most users I imagine).

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Since we know the size of the array, lets opt to use `std::array`.

It took me a minute to figure out why we need to template over `Container` here even if we always pass in `std::array` - the number of elements is [part of the type](https://en.cppreference.com/w/cpp/container/array.html) of an `std::array`.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Implicitly fixed by dropping the `bufferSize` arg in `readValue`.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: I'll update `startPos` to be `uint64_t`, since there's never going to be a negative starting position. I've updated other operators to explicitly cast, and added `validateUint64ToStreampos` to ValidationUtils.cpp, since that is the one instance where we do a narrowing cast.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: We also use `chunkSize` to move past this chunk if it was the incorrect chunk, in the `startPos` calculation below

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: The minimum size check is specific to the `fmt` chunk. The other case where we will use this function, finding the `data` chunk, has no minimum size.

To keep the max chunk size `TORCH_CHECK` near the actual allocation, I've also kept it here in `parseHeader()`. If this becomes repetitive in the future, maybe we can refactor to have the `TORCH_CHECK` and vector allocation within a helper function.

src/torchcodec/_core/WavDecoder.h
Dan-Flores: Added! For `MAX_FMT_CHUNK_SIZE` the value is a bit arbitrary so I left a comment reflecting that. I am open to suggestions on that value.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+constexpr uint32_t RIFF_HEADER_SIZE = 12; // "RIFF" + fileSize + "WAVE"
+constexpr uint32_t CHUNK_HEADER_SIZE = 8; // chunkID + chunkSize
+// Standard WAV fmt chunk is at least 16 bytes:
+// audioFormat(2) + numChannels(2) + sampleRate(4) + byteRate(4) + blockAlign(2)
+// + bitsPerSample(2)
+constexpr uint32_t MIN_FMT_CHUNK_SIZE = 16;
+// WAVE_FORMAT_EXTENSIBLE adds: cbSize(2) + wValidBitsPerSample(2) +
+// dwChannelMask(4) + SubFormat GUID(16) = 24 more bytes, total 40
+constexpr uint32_t MIN_WAVEX_FMT_CHUNK_SIZE = 40;
```
NicolasHug: There's something I don't understand here: 

> cbSize(2) + wValidBitsPerSample(2) + dwChannelMask(4) + SubFormat GUID(16) = 24

that I agree with, but then were do the extra 40 - 24 = 16 bytes come from?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+bool is_little_endian() {
+  uint32_t x = 1;
+  uint8_t first_byte;
+  std::memcpy(&first_byte, &x, 1);
+  return first_byte == 1;
+}
+
+// Container template requires .data() and .size() methods.
+// We currently call this function on std::array.
```
NicolasHug: The fact that we call it one two different types is the reason we needed to template over Container.
IMHO the comment could be removed, but no strong opinion. If we keep it:

```suggestion
// We currently call this function on std::array and std::vector
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+  STD_TORCH_CHECK(header_.sampleRate > 0, "Invalid WAV: zero sample rate");
+}
+
+// Given a chunkId, read through each chunk until we find a match, then return
+// its offset and size.
+WavDecoder::ChunkInfo WavDecoder::findChunk(
+    const char* chunkId,
+    uint64_t startPos,
+    uint64_t fileSizeLimit) {
+  while (startPos <= fileSizeLimit - CHUNK_HEADER_SIZE) {
```
NicolasHug: This `fileSizeLimit - CHUNK_HEADER_SIZE` unsigned subtraction can underflow if `fileSizeLimit` if less than `CHUNK_HEADER_SIZE`. We should check against that.

Side note: let's try to have single name for `fileSizeLimit`. Currently we have `fileSizeLimit` and `actualFileSize`, which denote the same thing

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+    if (matchesFourCC(chunkHeader.data(), chunkId)) {
+      STD_TORCH_CHECK(
+          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
+          "File position arithmetic would overflow");
+      return {startPos + CHUNK_HEADER_SIZE, chunkSize};
+    }
+
+    STD_TORCH_CHECK(
+        startPos <= UINT64_MAX - CHUNK_HEADER_SIZE - chunkSize - 1,
```
NicolasHug: This seems off by one depending on the evenness of `chunkSize` (since we add `chunkSize % 2`).

It will also be easier to understand and verify the intent of this check by defining the increment variable separately. Something like this

```Cpp
  uint64_t advance = CHUNK_HEADER_SIZE + static_cast<uint64_t>(chunkSize) + (chunkSize % 2);
  STD_TORCH_CHECK(startPos <= UINT64_MAX - advance, ...);
  startPos += advance;
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+  return value;
+}
+
+// The caller should ensure that 'data' has at least 4 bytes
+bool matchesFourCC(const uint8_t* data, const char* expected) {
+  return std::memcmp(data, expected, 4) == 0;
+}
+
+template <typename Container>
+void safeReadFile(std::ifstream& file, Container& buffer, size_t bytesToRead) {
```
NicolasHug, line 76: We should validate that `bytesToRead <= buffer.size()`

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+WavDecoder::ChunkInfo WavDecoder::findChunk(
+    const char* chunkId,
+    uint64_t startPos,
+    uint64_t fileSizeLimit) {
+  while (startPos <= fileSizeLimit - CHUNK_HEADER_SIZE) {
+    file_.seekg(validateUint64ToStreampos(startPos, "startPos"), std::ios::beg);
+
+    std::array<uint8_t, CHUNK_HEADER_SIZE> chunkHeader;
+    safeReadFile(file_, chunkHeader, CHUNK_HEADER_SIZE);
+    // Read chunk size which immediately follows the chunk ID
```
NicolasHug, line 211: remove this comment?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+          effectiveFormat == WAV_FORMAT_IEEE_FLOAT,
+      "Unsupported WAV format: ",
+      effectiveFormat,
+      ". Only PCM and IEEE float formats are supported.");
+
+  if (effectiveFormat == WAV_FORMAT_PCM) {
+    // TODO WavDecoder: support 8, 16, 24 bits
+    if (header_.bitsPerSample != 32) {
+      STD_TORCH_CHECK(
+          false,
```
NicolasHug: here and below, avoid TORCH_CHECK(false)

src/torchcodec/_core/ValidationUtils.cpp
```diff
@@ -32,4 +32,21 @@ std::optional<int> validateOptionalInt64ToInt(
   }
 }
 
+std::streampos validateUint64ToStreampos(
+    uint64_t value,
+    const std::string& parameterName) {
+  // std::streampos is typically std::streamoff, which is signed, so converting
+  // it to uuint64_t is a widening cast.
```
NicolasHug:
```suggestion
  // it to uint64_t is a widening cast.
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+// Arbitrary max for fmt chunk allocation - set to 5x extended format size
+constexpr uint32_t MAX_FMT_CHUNK_SIZE = 200;
+
+// See standard format codes and Wav file format used in WavHeader:
+// https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html
+constexpr uint16_t WAV_FORMAT_PCM = 1;
+constexpr uint16_t WAV_FORMAT_IEEE_FLOAT = 3;
+constexpr uint16_t WAV_FORMAT_EXTENSIBLE = 0xFFFE;
+
+bool is_little_endian() {
```
NicolasHug:
```suggestion
bool isLittleEndian() {
```

src/torchcodec/_core/ValidationUtils.cpp
NicolasHug: Also, this comment is a bit confusing, I'd just remove it. The point of the function is to convert a uint64 to a smaller `std::streampos` type, but the comment confusingly mentions that `std::streamoff` is converted into a uint64

I understand the conversion to uint64 is needed for proper comparison within `STD_TORCH_CHECK`, but that's still confusing, and there seem to be a mismatch between `streamoff` and `streampos`?

src/torchcodec/_core/ValidationUtils.cpp
```diff
@@ -32,4 +32,21 @@ std::optional<int> validateOptionalInt64ToInt(
...
+  // https://en.cppreference.com/w/cpp/io/streamoff.html
+  STD_TORCH_CHECK(
+      value <=
+          static_cast<uint64_t>(std::numeric_limits<std::streamoff>::max()),
```
NicolasHug, line 43: Should this be streampos instead of streamoff?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,219 @@
...
+  return first_byte == 1;
+}
+
+// Container template requires .data() and .size() methods.
+// We currently call this function on std::array.
+template <typename T, typename Container>
+T readValue(const Container& data, size_t offset) {
+  static_assert(std::is_trivially_copyable_v<T>);
+  STD_TORCH_CHECK(
+      data.size() >= sizeof(T) && offset <= data.size() - sizeof(T),
```
NicolasHug, line 51: That second condition is `data.size() >= sizeof(T) + offset`, which seems to make the first one obsolete?

NicolasHug: Also, this is assuming the Container stores uint8 data. We could statically validate against it, or at least document it

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: `WAVEX` fields always come after the standard WAV header, so the 16 bytes are the ones calcualted just above. 
I'll updated "adds" to be more explicitly "WAVE_FORMAT_EXTENSIBLE adds to the standard WAV fmt chunk"

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: I would say its helpful when reading the code, its not immediately obvious why we choose to read 4 bytes.

src/torchcodec/_core/ValidationUtils.cpp
Dan-Flores: `std::streampos` is a typedef of `std::fpos`, neither of which provide `std::numeric_limits` for us to validate against.
The docs for [`streampos`](https://en.cppreference.com/w/cpp/io/fpos.html) indicate that [`streamoff`](https://en.cppreference.com/w/cpp/io/streamoff.html) is typically used to hold the position in the stream, which makes sense given that `streamoff`: "is an implementation-defined signed integral(since C++11) type of sufficient size to represent the maximum possible file size supported by the operating system"

https://stackoverflow.com/questions/24437016/is-stdstreampos-guaranteed-to-be-unsigned-long-long

The OS specific dtypes are a bit of a pain. `claude` indicates that `streamoff` is the correct type to use here in most cases.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Thanks! The first condition was intended to protect against underflow, but I agree that its obsolete.

I've added a static assert to check that the container has 1 byte elements, so this requirement is documented.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+constexpr uint32_t MAX_FMT_CHUNK_SIZE = 200;
+
+// See standard format codes and Wav file format used in WavHeader:
+// https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html
+constexpr uint16_t WAV_FORMAT_PCM = 1;
+constexpr uint16_t WAV_FORMAT_EXTENSIBLE = 0xFFFE;
+
+bool isLittleEndian() {
+  uint32_t x = 1;
+  uint8_t first_byte;
```
NicolasHug:
```suggestion
  uint8_t firstByte;
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+  } catch (const std::filesystem::filesystem_error& e) {
+    STD_TORCH_CHECK(
+        false, "Failed to get file size for: ", path, ". Error: ", e.what());
+  }
+  parseHeader(fileSize);
+  validateHeader();
+}
+
+void WavDecoder::parseHeader(uint64_t fileSize) {
+  file_.seekg(0, std::ios::beg);
```
NicolasHug: Here and in other places, we should check whether seekg succeeded. By e.g. calling `file_.fail()` (which could be in a separate `seek()` helper).

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+      ": exceeds buffer length ",
+      data.size());
+  T value;
+  std::memcpy(&value, data.data() + offset, sizeof(T));
+  return value;
+}
+
+template <size_t N>
+bool matchesFourCC(
+    const std::array<std::byte, N>& data,
```
NicolasHug: We use `uint8_t` everywhere else in the codebase, and this `std::byte` will be cast to `uint8_t` in most callsites as well, so let's stick to `uint8_t`.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+    size_t offset,
+    const char* expected) {
+  static_assert(N >= 4);
+  STD_TORCH_CHECK(
+      offset <= N - 4,
+      "Data array too small for FourCC comparison at offset ",
+      offset);
```
NicolasHug: Having N as a template parameter seems overkill. `N` is just data.size(). This check and the `static_assert` above are doing the same check that `readValue()` does, but in a more obfuscated way. Let's simplify it.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+}
+
+template <typename T, typename Container>
+T readValue(const Container& data, size_t offset) {
+  static_assert(std::is_trivially_copyable_v<T>);
+  static_assert(
+      sizeof(typename Container::value_type) == 1,
+      "Container value_type must be a 1-byte type for safe byte access");
+  STD_TORCH_CHECK(
+      offset <= data.size() - sizeof(T),
```
NicolasHug: This can underflow. I was hoping we could use the simpler `data.size() >= sizeof(T) + offset` but this can overflow too. The way you had it before seems to be the most robust check, despite being annoying cryptic


```suggestion
      data.size() >= sizeof(T) && offset <= data.size() - sizeof(T),
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+      STD_TORCH_CHECK(
+          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
+          "File position arithmetic would overflow");
+      return {startPos + CHUNK_HEADER_SIZE, chunkSize};
+    }
+    STD_TORCH_CHECK(
+        chunkSize <= UINT64_MAX - CHUNK_HEADER_SIZE - (chunkSize % 2),
+        "Chunk size would cause overflow: ",
+        chunkSize);
+    uint64_t chunkLen =
```
NicolasHug: We already defined chunkSize above, which conflicts with chunkLen in terms of names. `numBytesToSkip` or something along these lines would be less ambiguous

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,223 @@
...
+        CHUNK_HEADER_SIZE + static_cast<uint64_t>(chunkSize) + (chunkSize % 2);
+    STD_TORCH_CHECK(
+        startPos <= UINT64_MAX - chunkLen,
+        "File position arithmetic would overflow");
+    // Skip this chunk and continue searching (odd chunks are padded)
```
NicolasHug: The "odd chunks are padded" comment needs to be above

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Good point, I see our `AVIOFileLikeContext` and related classes all use `uint8_t`, I'll revert the  `std::byte ` change.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Yes it was a bit much. I've updated this function to accept a `uint_t*` data ptr and `size_t` arg directly.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,235 @@
...
+    uint32_t chunkSize = readValue<uint32_t>(chunkHeader, 4);
+
+    if (matchesFourCC(chunkHeader.data(), chunkHeader.size(), 0, chunkId)) {
+      STD_TORCH_CHECK(
+          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
+          "File position arithmetic would overflow");
+      return {startPos + CHUNK_HEADER_SIZE, chunkSize};
+    }
+    STD_TORCH_CHECK(
+        chunkSize <= UINT64_MAX - CHUNK_HEADER_SIZE - (chunkSize % 2),
```
NicolasHug: This is mostly dead code, `chunkSize` is uint32 so it cannot ever exceed `UINT64_MAX - CHUNK_HEADER_SIZE - (chunkSize % 2)`

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+    std::array<uint8_t, CHUNK_HEADER_SIZE> chunkHeader;
+    safeReadFile(file_, chunkHeader, CHUNK_HEADER_SIZE);
+    // Read chunk size which immediately follows the chunk ID
+    uint32_t chunkSize = readValue<uint32_t>(chunkHeader, 4);
+
+    if (matchesFourCC(chunkHeader.data(), chunkHeader.size(), 0, chunkId)) {
+      STD_TORCH_CHECK(
+          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
+          "File position arithmetic would overflow");
```
NicolasHug: I think this check is dead code too. `startPos` was already validated to be `startPos <= fileSize - CHUNK_HEADER_SIZE`

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -0,0 +1,235 @@
...
+      ". Currently supported bit depths are: 32");
+
+  STD_TORCH_CHECK(header_.numChannels > 0, "Invalid WAV: zero channels");
+  STD_TORCH_CHECK(header_.sampleRate > 0, "Invalid WAV: zero sample rate");
+}
+
+// Given a chunkId, read through each chunk until we find a match, then return
+// its offset and size.
+WavDecoder::ChunkInfo WavDecoder::findChunk(
+    const char* chunkId,
```
NicolasHug: Consider passing the more c++ idiomatic `std:string_view` for all existing uses of `char *`

src/torchcodec/_core/WavDecoder.h
```diff
@@ -0,0 +1,51 @@
...
+    uint16_t numChannels = 0;
+    uint32_t sampleRate = 0;
+    uint16_t bitsPerSample = 0;
+    // Extended format fields (WAVE_FORMAT_EXTENSIBLE)
+    uint16_t subFormat = 0; // Extracted from SubFormat GUID (first 2 bytes)
+  };
+
+  struct ChunkInfo {
+    uint64_t offset;
+    uint32_t size;
```
NicolasHug, line 40: Best not to leave anything un-initialized. We should initialize those to 0.

Or better, delete the default constructor by defining a new one:

```Cpp
  struct ChunkInfo {                                                                                                   
      uint64_t offset;                                                                                                 
      uint32_t size;                                  
      ChunkInfo(uint64_t offset, uint32_t size)
          : offset(offset), size(size) {}
  };
```

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Yes, you are right - since `fileSize` is `uint64`, the earlier validation effectively prevents overflow.

PR #1283, opened 2026-03-11
Address internal lint

mollyxu:
Address some internal lints from https://www.internalfb.com/diff/D95819469, mainly to avoid implicit conversion

### DIFF
--- a/src/torchcodec/_core/CUDACommon.cpp
+++ b/src/torchcodec/_core/CUDACommon.cpp
@@ -8,6 +8,7 @@
 #include <torch/csrc/inductor/aoti_torch/c/shim.h>
 #include "Cache.h" // for PerGpuCache
 #include "StableABICompat.h"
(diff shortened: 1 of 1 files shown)


src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -395,7 +395,7 @@ torch::stable::Tensor convertNV12FrameToRGB(
           yuvData,
           srcStep,
           dst.mutable_data_ptr<Npp8u>(),
-          dst.stride(0),
+          static_cast<int>(dst.stride(0)),
```
Dan-Flores: We could leverage the existing [`validateInt64ToInt`](https://github.com/meta-pytorch/torchcodec/blob/13ffad0d4510aef6689311baec5c06defe7d07bd/src/torchcodec/_core/ValidationUtils.cpp#L13) function here and above, to be extra cautious?

PR #1284, opened 2026-03-11
Create torchcodec-versions.json

mollyxu:
Add torchcodec-versions.json to work with the pytorch_sphinx_theme2.


NicolasHug:
We should figure out what it means for https://github.com/meta-pytorch/torchcodec/blob/gh-pages/versions.html, ideally we could remove it, otherwise we'd have to update versions in different places.

PR #1287, opened 2026-03-12
Improve example gallery thumbnails

NicolasHug:
Our example gallery looks a bit sad:

image

Introducing... Grumps.

image

As mentioned offline, I don't feel strongly about any of these, I'm happy to consider alternative. I just think we really should make the example gallery look nicer, and having a thumbnail for each really improves it IMHO.

### DIFF
--- a/examples/decoding/performance_tips.py
+++ b/examples/decoding/performance_tips.py
@@ -252,4 +252,3 @@ to increase performance.
 # - :class:`torchcodec.decoders.VideoDecoder` - Full API reference
 
 # sphinx_gallery_thumbnail_path = '_static/thumbnails/grumps_brrrr.jpg'
-


examples/decoding/approximate_mode.py
```diff
@@ -33,6 +33,7 @@
 from time import perf_counter_ns
 
 
+# sphinx_gallery_thumbnail_path = '_static/thumbnails/grumps_seek_mode.jpg'
```
NicolasHug, line 36: All the new `sphinx_gallery_thumbnail_path = ` should ideally be comments *in the code*, as done here. Otherwise they'll render as an "empty block" which can be seen in the "performance tips" tutorial right now (this one doesn't have code). See the [docs](https://sphinx-gallery.github.io/stable/configuration.html#providing-thumbnail):

> the comment should be somewhere below the docstring (ideally in a code block,

The comment gets removed by `sphinx-gallery` automatically thanks to:

https://github.com/meta-pytorch/torchcodec/blob/a8a257ab67975e0ab14fa12cae17e2363b4d9819/docs/source/conf.py#L105

docs/source/_static/thumbnails/grumps_6.jpg
mollyxu, line 1: Thanks for looking into this @NicolasHug ! I think even if we wanted to diversify the pictures we should at least keep the streaming grumps and audio grumps

examples/decoding/performance_tips.py
```diff
@@ -17,6 +17,7 @@
 to increase performance.
 """
 
+# sphinx_gallery_thumbnail_path = '_static/thumbnails/grumps_brrrr.jpg'
```
Dan-Flores: Locally I was able to avoid the empty block by moving this comment to the very end of the file after the `Conclusion` section. The docs rendered the thumbnail without adding the comment to the tutorial, could you check if that also renders correctly on the CI?

examples/encoding/video_encoding.py
```diff
@@ -24,6 +24,7 @@
 from IPython.display import Video
 
 
+# sphinx_gallery_thumbnail_path = '_static/thumbnails/grumps_encoding_video.jpg'
```
Dan-Flores: Maybe we can use the cat in the tutorial's video as the thumbnail? No offense to `grumps`, but its more accurate to what the user can expect if they click this tutorial.

PR #1290, opened 2026-03-13
Remove low-level decoding functions from _core exports

zy1git:
Remove low-level decoding functions from _core exports. Those functions removal should not break Videodecoder/Audiodecoder.


src/torchcodec/decoders/_audio_decoder.py
```diff
@@ -10,7 +10,6 @@
 
 import torch
 from torch import Tensor
-
```
NicolasHug, line 13: Here and in other places, please avoid unrelated lint changes

benchmarks/decoders/gpu_benchmark.py
```diff
@@ -4,11 +4,10 @@
 from pathlib import Path
 
 import torch
-
 import torch.utils.benchmark as benchmark
-
-import torchcodec
 import torchvision.transforms.v2.functional as F
+from torchcodec._core import get_next_frame
```
NicolasHug, line 11: Let's please have a follow-up PR to entirely remove `get_next_frame` from the core ops. It's a relic of the past. We should just delete test and benchmarks related to it as well. Please be extra sure there are no more internal uses of it as well.

src/torchcodec/decoders/_audio_decoder.py
zy1git: Fixed in the new commit.

benchmarks/decoders/gpu_benchmark.py
zy1git: Sure, I'll create a follow-up PR to remove `get_next_frame` from core ops, along with related tests and benchmarks. And I will detect the internal uses of it using the similar trick as this PR.

PR #1292, opened 2026-03-13
[wip] uint16 poc

mollyxu:
This is a proof-of-concept for 10-bit HDR video decoding across both CPU and Beta CUDA paths. Instead of truncating to 8-bit, >8-bit sources now output uint16 tensors preserving the full bit depth.

To review:
  1. Read FrameDims and Frame.cpp changes. FrameDims gains a bitDepth field (default 8), which flows through the entire pipeline to determine the output tensor dtype
  (uint8 vs uint16) and pixel format (RGB24 vs RGB48LE).
  2. Read the CPU path changes in CpuDeviceInterface.cpp.
    - New helpers getBitDepthFromAVPixelFormat, getOutputPixelFormat, and getFormatFilterString determine the output format dynamically.
    - SwScale now accepts an explicit outputFormat parameter (AV_PIX_FMT_RGB24 or AV_PIX_FMT_RGB48LE) and computes bytesPerPixel_ accordingly. The double-swscale path
  (color conversion then resize) works the same way but with 16-bit data when needed.
    - The filtergraph path dynamically prepends format=rgb48le, or format=rgb24, to user transform filters based on source bit depth.
  3. Read the Beta CUDA path in BetaCudaDeviceInterface.cpp and CUDACommon.cpp.
    - NVDEC now requests P016 surface format for >8-bit videos (preserving full bit depth) instead of forcing NV12 which truncated to 8-bit.
    - New transferCpuFrameToGpuP016 handles the CPU-fallback case for 10-bit content.
    - New convertP016FrameToRGB uses NPP's nppiNV12ToRGB_16u_ColorTwist32f_P2C3R_Ctx for 16-bit YUV→RGB conversion.
    - 16-bit color twist matrices are provided for BT.601/BT.709/BT.2020 in both full and limited range, with CUDA 13 vs <13 offset conventions 
  4. Skim the test changes.
    - Two HDR test assets: nasa_13013_hdr.mp4 and testsrc2_hdr.mp4 (10-bit H.265 with BT.2020 + SMPTE2084/PQ).
    - HDR videos are added to existing test_10bit_videos, resize, center crop, random crop, color conversion, and transform pipeline tests. Tests are skipped on FFmpeg 4 due to an unknown divergence in decoded results (plain 10-bit without HDR metadata is unaffected).
    - Tolerances are scaled by 256 for uint16 comparisons. The CPU-vs-GPU BT.2020 comparison right-shifts by 6 to recover 10-bit values before asserting.

  Open questions / known limitations:
  - Would we want to have actual HDR test videos as assets in the actual PR? Or would having videos with HDR metadata be sufficient for testing purposes?
  - The FFmpeg 4 discrepancy with HDR transform results needs investigation.
  - output dtype?

### DIFF
--- a/test/test_decoders.py
+++ b/test/test_decoders.py
@@ -41,6 +41,7 @@ from .utils import (
     H265_10BITS,
     H265_VIDEO,
     in_fbcode,
+    IS_WINDOWS,
     make_video_decoder,
     NASA_AUDIO,
     NASA_AUDIO_MP3,
@@ -1605,9 +1606,27 @@ class TestVideoDecoder:
         "asset",
         (
             H264_10BITS,
-            H265_10BITS,
-            NASA_VIDEO_HDR,
-            TEST_SRC_2_720P_HDR,
+            pytest.param(
+                H265_10BITS,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="uint8 vs uint16 HDR color conversion differs on Windows + FFmpeg 4",
+                ),
+            ),
+            pytest.param(
+                NASA_VIDEO_HDR,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="uint8 vs uint16 HDR color conversion differs on Windows + FFmpeg 4",
+                ),
+            ),
+            pytest.param(
+                TEST_SRC_2_720P_HDR,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="uint8 vs uint16 HDR color conversion differs on Windows + FFmpeg 4",
+                ),
+            ),
         ),
     )
     def test_output_dtype_uint8_vs_uint16_on_10bit(self, asset):
@@ -1692,7 +1711,23 @@ class TestVideoDecoder:
 
     @pytest.mark.parametrize(
         "asset",
-        (H264_10BITS, H265_10BITS, NASA_VIDEO_HDR),
+        (
+            H264_10BITS,
+            pytest.param(
+                H265_10BITS,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="uint8 vs uint16 HDR color conversion differs on Windows + FFmpeg 4",
+                ),
+            ),
+            pytest.param(
+                NASA_VIDEO_HDR,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="uint8 vs uint16 HDR color conversion differs on Windows + FFmpeg 4",
+                ),
+            ),
+        ),
     )
     @pytest.mark.parametrize(
         "transforms",


src/torchcodec/_frame.py
```diff
@@ -70,7 +70,7 @@ class FrameBatch(Iterable):
     """
 
     data: Tensor
-    """The frames data (``torch.Tensor`` of uint8)."""
+    """The frames data (``torch.Tensor`` of uint8 for 8-bit sources, uint16 for 10-bit HDR sources)."""
```
mollyxu, line 73: Would we want to output something that is more well supported for 10-bit HDR sources? `uint16` doesn't have a lot of supported operations. Or would this be something we would want the user to handle?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -580,8 +580,18 @@ void SingleStreamDecoder::addVideoStream(
   // Set preRotationDims_ for the active stream. These are the raw encoded
   // dimensions from FFmpeg, used as a fallback for tensor pre-allocation when
   // no resize/rotation transforms are applied.
+  int bitDepth = 8;
+  {
+    const AVPixFmtDescriptor* desc = av_pix_fmt_desc_get(
```
mollyxu, line 585: might want to consolidate logic in here and in `CpuDeviceInterface` to get bit depth for the actual PR

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -860,38 +864,169 @@ UniqueAVFrame BetaCudaDeviceInterface::transferCpuFrameToGpuNV12(
...
 void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
     UniqueAVFrame& avFrame,
     FrameOutput& frameOutput,
     std::optional<torch::stable::Tensor> preAllocatedOutputTensor) {
-  UniqueAVFrame gpuFrame =
-      cpuFallback_ ? transferCpuFrameToGpuNV12(avFrame) : std::move(avFrame);
+  bool is10Bit = (bitDepth_ > 8);
+
+  UniqueAVFrame gpuFrame;
+  if (cpuFallback_) {
```
mollyxu, line 1015: this can probably be refactored into a call to `transferCpuFrameToGpu`

src/torchcodec/_frame.py
NicolasHug: I don't know yet, but that will be one of the most fundamental and important design question we'll need to answer.

The choice of the output dtype won't affect the complexity of the work that much, but it will be a key factor in the UX.

Roughly, I think we can consider:

- `uint16`. Users can't do anything with a uint16 tensor other than convert it to another dtype like int32 or float. Note that torchvision's `decode_png` does just that: https://docs.pytorch.org/vision/main/generated/torchvision.io.decode_png.html, but it doesn't mean we have to do exactly the same (see also https://fburl.com/workplace/lvj7u3jo).
  - one drawback with that is that if most users just end up wanting float32 (or int32), then we're forcing them to do an extra conversion (i.e. extra copy).
- `int32`?
- `float32`?

src/torchcodec/_core/Frame.h
```diff
@@ -16,10 +16,14 @@ namespace facebook::torchcodec {
 struct FrameDims {
   int height = 0;
   int width = 0;
+  // Bit depth per channel of the source video. 8 for standard video,
+  // 10 or 12 for HDR. Used to determine output tensor dtype:
+  // uint8 for bitDepth <= 8, uint16 for bitDepth > 8.
+  int bitDepth = 8;
```
NicolasHug, line 22: This is OK for a POC, but we'll want to rethink how to pass around the bitDepth info. At the very least, FrameDims should be renamed since it's not just about the dims anymore. Not something we need to worry about now, I'm just flagging it as a future discussion / design point.

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: Agreed, also in general we'll want to be super careful about getting the bitDepth from different sources, and making sure we gracefully handle a potential mismatch. We don't have to worry about that now in the POC stage though.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -6,7 +6,36 @@
...
+  const AVPixFmtDescriptor* desc = av_pix_fmt_desc_get(format);
+  if (desc && desc->nb_components > 0) {
+    return desc->comp[0].depth;
+  }
+  return 8;
+}
+
+// Returns the appropriate RGB output format based on source bit depth.
+AVPixelFormat getOutputPixelFormat(int bitDepth) {
+  return (bitDepth > 8) ? AV_PIX_FMT_RGB48LE : AV_PIX_FMT_RGB24;
```
NicolasHug: We should rationalize the use of `AV_PIX_FMT_RGB48LE` for this, in contrast to e.g. `AV_PIX_FMT_RGB48BE`, or any other appropriate pixel format.

(I'm not saying `AV_PIX_FMT_RGB48LE` isn't the right thing, I'm just saying we should justify why we're using it)

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -86,16 +115,21 @@ void CpuDeviceInterface::initializeVideo(
...
-    // conversion to ensure the output frame matches the pixel format
-    // specified in the sink. But by default, it will insert it after the user
-    // filters. We need an explicit format conversion to get the behavior we
-    // want.
-    filters_ = "format=rgb24," + filters.str();
+    // We store the user transforms without the format prefix here. The format
+    // prefix is added dynamically in convertAVFrameToTensorUsingFilterGraph()
+    // based on the source bit depth.
+    userTransformFilters_ = filters.str();
+    filters_ = "format=rgb24," + userTransformFilters_;
```
NicolasHug: I'm missing the bigger picture and I haven't looked at the rest of the code yet, so take this with a grain of salt, but I'm confused by seeing `rgb24`  here, knowing that we might actually want to convert to rgb48

mollyxu: Thanks for pointing this out! I removed "format=rgb24" completely because it will be dynamically added in `convertVideoAVFrameToFrameOutput`

src/torchcodec/_core/CpuDeviceInterface.cpp
mollyxu: Thanks for the feedback! After looking into this some more, I think it's probably best to use `AV_PIX_FMT_RGB48`

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -6,7 +6,38 @@
...
+  }
+  return 8;
+}
+
+// Returns the appropriate RGB output format based on source bit depth.
+// RGB24 is 8 bits per channel, RGB48 is 16 bits per channel. For >8-bit
+// sources (e.g. 10-bit HDR), we need RGB48 to preserve the full bit depth;
+// RGB24 would lose the extra precision.
+AVPixelFormat getOutputPixelFormat(int bitDepth) {
+  return (bitDepth > 8) ? AV_PIX_FMT_RGB48 : AV_PIX_FMT_RGB24;
```
mollyxu, line 31: it seems like the `uint8` path and `uint16 `path would not necessarily produce "the same results but bitshifted "because they have different pixel formats which would lead to different compute paths in `sws_scale`

test/test_decoders.py
```diff
@@ -1546,14 +1550,274 @@ def test_10bit_gpu_fallsback_to_cpu(self):
...
+        frame_at_pts = decoder.get_frame_played_at(pts)
+        assert frame_at_pts.data.shape == (expected_c, expected_h, expected_w)
+
+    @pytest.mark.parametrize(
+        "asset",
+        (
+            H264_10BITS,
+            pytest.param(
+                H265_10BITS,
+                marks=pytest.mark.skipif(
```
mollyxu, line 1611: Similar to the skip in `test_transform_ops` where the test is only failing on ffmpeg4+windows. Might be something about color conversion on FFmpeg4

PR #1293, opened 2026-03-18
WavDecoder metadata 

Dan-Flores:
This PR adds the plumbing needed to add a `metadata` field to `WavDecoder`. 
It reuses the `AudioStreamMetadata` class so the available metadata is the same as `AudioDecoder`.

To keep `WavHeader` private to `WavDecoder.cpp`, `getStreamMetadata()` converts the relevant fields into `StreamMetadata`. 
This adds some complexity, but makes it clear that `WavDecoder.metadata` has parity with `AudioDecoder.metadata`. If we decide to expose additional WAV metadata in the future, we can refactor to use a new class.

* Note: Addressed feedback in earlier commit: https://github.com/meta-pytorch/torchcodec/commit/bdab09995428e0fbc987fd88657975db0f5a2f5d
* Note 2: Addressed feedback in second commit: 26b0f8ccaa4fefa6ba222d9fe3234ba6e8131c84

### DIFF
--- a/src/torchcodec/_core/WavDecoder.cpp
+++ b/src/torchcodec/_core/WavDecoder.cpp
@@ -172,7 +172,7 @@ void WavDecoder::parseHeader(uint64_t fileSize) {
   header_.dataSize = dataChunk.size;
 }
 
-void WavDecoder::validateHeader() const {
+void WavDecoder::validateHeader() {
   uint16_t effectiveFormat = (header_.audioFormat == WAV_FORMAT_EXTENSIBLE)
       ? header_.subFormat
       : header_.audioFormat;


test/test_decoders.py
```diff
@@ -2613,3 +2614,46 @@ def test_num_channels_errors(self, asset):
...
                 decoder.get_all_samples()
+
+
+class TestWavDecoder:
+    def test_metadata(self):
+        asset = SINE_MONO_S32
+        wav_decoder = WavDecoder(asset.path)
+        audio_decoder = AudioDecoder(asset.path)
+
+        assert isinstance(wav_decoder.metadata, AudioStreamMetadata)
```
NicolasHug, line 2625: Could we just do `assert wav_decoder.metadata == audio_decoder.metadata` instead of manually checking every field?

src/torchcodec/_core/WavDecoder.h
```diff
@@ -33,6 +38,8 @@ class WavDecoder {
     uint16_t bitsPerSample = 0;
     // Extended format fields (WAVE_FORMAT_EXTENSIBLE)
     uint16_t subFormat = 0; // Extracted from SubFormat GUID (first 2 bytes)
+    // Fields from data chunk
```
NicolasHug: Do we plan to add more "data chunk" fields? For now, the "Size of audio data in bytes" is self-explanatory so I'd suggest to remove this one

```suggestion
```

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -225,4 +237,69 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+      "Unsupported format after validation. That's unexpected, please report this to the TorchCodec repo.");
+}
+
+StreamMetadata WavDecoder::getStreamMetadata() const {
+  StreamMetadata metadata;
+  metadata.streamIndex = 0; // WAV files have single audio stream
+  metadata.sampleRate = static_cast<int64_t>(header_.sampleRate);
+  metadata.numChannels = static_cast<int64_t>(header_.numChannels);
+  metadata.sampleFormat = getSampleFormat();
+  metadata.codecName = getCodecName();
```
NicolasHug: If we keep them these two should be private methods.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -225,4 +237,69 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+      " * ",
+      bytesPerSample);
+  uint32_t bytesPerSecond = samplesPerSecond * bytesPerSample;
+
+  metadata.durationSecondsFromHeader =
+      static_cast<double>(header_.dataSize) / bytesPerSecond;
+  metadata.beginStreamPtsSecondsFromContent = 0.0;
+  metadata.bitRate = static_cast<double>(header_.sampleRate) *
+      static_cast<double>(header_.numChannels) *
+      static_cast<double>(header_.bitsPerSample);
```
NicolasHug, line 249: Flagging https://github.com/meta-pytorch/torchcodec/commit/26b0f8ccaa4fefa6ba222d9fe3234ba6e8131c84#r179912350 which may have been overridden in a merge commit. I think it's worth simplifying the logic above and take a "shortest path" to the computation of both bit rate and duration.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -212,8 +217,15 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
     if (matchesFourCC(chunkHeader.data(), chunkHeader.size(), 0, chunkId)) {
+      STD_TORCH_CHECK(
+          startPos <= UINT64_MAX - CHUNK_HEADER_SIZE,
+          "File position arithmetic would overflow");
       return {startPos + CHUNK_HEADER_SIZE, chunkSize};
     }
+    STD_TORCH_CHECK(
+        chunkSize <= UINT64_MAX - CHUNK_HEADER_SIZE - (chunkSize % 2),
+        "Chunk size would cause overflow: ",
+        chunkSize);
```
NicolasHug: remove both checks above?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -225,4 +237,69 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+
+  if (effectiveFormat == WAV_FORMAT_PCM) {
+    if (header_.bitsPerSample == 32)
+      return "pcm_s32le";
+  }
+
+  STD_TORCH_CHECK(
+      false,
+      "Unsupported format after validation. That's unexpected, please report this to the TorchCodec repo.");
+}
```
NicolasHug: On both `getter` methods above: if the final `STD_TORCH_CHECK(false)` *must* never be reached provided we did a correct header validation, then I think we should move the inference of the sample format and the codec name in the header validation method. This way we can avoid validating the same logic twice.

Dan-Flores: Thanks, I'll incorporate this.

I had these separate functions in my POC because the complete WavDecoder will have a lot of branches in this function to handle all the combinations. But we can decide the best way to handle that later.

PR #1297, opened 2026-03-19
Fix macos test with long python version

atalman:
Use full python version during testing


NicolasHug:
Thanks for the PR @atalman , after discussing offling this has been addressed at the source in `test-infra`. Thanks for following up!

PR #1299, opened 2026-03-19
Add fps example to doc

mollyxu:
Add an example of resampling using FPS to our decoding example

### DIFF
--- a/examples/decoding/basic_example.py
+++ b/examples/decoding/basic_example.py
@@ -13,14 +13,13 @@ In this example, we'll learn how to decode a video using the
 :class:`~torchcodec.decoders.VideoDecoder` class.
 """
 
-import requests
-
 # %%
 # First, a bit of boilerplate: we'll download a video from the web, and define a
 # plotting utility. You can ignore that part and jump right below to
 # :ref:`creating_decoder`.
 
 import torch
+import requests
 
 
 # Video source: https://www.pexels.com/video/dog-eating-854132/
@@ -35,14 +34,14 @@ raw_video_bytes = response.content
 
 def plot(frames: torch.Tensor, title: str | None = None):
     try:
-        import matplotlib.pyplot as plt
-        from torchvision.transforms.v2.functional import to_pil_image
         from torchvision.utils import make_grid
+        from torchvision.transforms.v2.functional import to_pil_image
+        import matplotlib.pyplot as plt
     except ImportError:
         print("Cannot plot, please run `pip install torchvision matplotlib`")
         return
 
-    plt.rcParams["savefig.bbox"] = "tight"
+    plt.rcParams["savefig.bbox"] = 'tight'
     fig, ax = plt.subplots()
     ax.imshow(to_pil_image(make_grid(frames)))
     ax.set(xticklabels=[], yticklabels=[], xticks=[], yticks=[])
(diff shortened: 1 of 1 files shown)


examples/decoding/basic_example.py
```diff
@@ -177,3 +179,39 @@ def plot(frames: torch.Tensor, title: str | None = None):
...
 plot(other_frames.data, "Other frames")
+
+# %%
+# Retrieving frames in a time range
+# ----------------------------------
+#
+# We can also retrieve all frames within a time range using
+# :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_in_range`. This
+# returns all frames with a :term:`pts` in the half-open range
+# [start_seconds, stop_seconds).
```
Dan-Flores: The phrasing here is a bit odd, lets just say:
```py
# This returns all frames with :term:`pts` in [start_seconds, stop_seconds).
```

This is how we explain exclusive range in [`get_frames_in_range`](https://meta-pytorch.org/torchcodec/stable/generated/torchcodec.decoders.VideoDecoder.html#torchcodec.decoders.VideoDecoder.get_frames_in_range), which is a bit simpler.

examples/decoding/basic_example.py
```diff
@@ -177,3 +179,39 @@ def plot(frames: torch.Tensor, title: str | None = None):
...
+
+frames_in_range = decoder.get_frames_played_in_range(start_seconds=1, stop_seconds=3)
+print(f"Frames in [1, 3) at source fps: {frames_in_range.data.shape[0]} frames")
```
Dan-Flores:
```suggestion
print(f"{type(frames_in_range) = }")
print(frames_in_range)
```

Elsewhere in this tutorial we print that the output is of type `FrameBatch`, and print its fields, lets stick to that here.

Since printing 50 frame PTS is quite verbose, we could also reduce the `frames_in_range` calculation to be 1 second long (for example, `(start_seconds=1, stop_seconds=2)`. 

This also has the benefit of making the resampling example below very simple to understand!

examples/decoding/basic_example.py
```diff
@@ -177,3 +179,39 @@ def plot(frames: torch.Tensor, title: str | None = None):
...
+# The source video is 25 fps, so a 2-second range contains ~50 frames.
+# Resampling to 5 fps gives us only 10 frames:
+frames_5fps = decoder.get_frames_played_in_range(start_seconds=1, stop_seconds=3, fps=5)
+print(f"Frames in [1, 3) at 5 fps: {frames_5fps.data.shape[0]} frames")
+
+# %%
+plot(frames_5fps.data, "2-second range resampled at 5 fps")
+
+# %%
+# :meth:`~torchcodec.decoders.VideoDecoder.get_all_frames` also supports the
```
Dan-Flores: We should definitely mention that `get_all_frames` also supports resampling, but I'm not sure this example covers new ground. Instead, lets mention both functions under the "Resampling with the fps parameter" header, something like:
```
  # Both :meth:`~torchcodec.decoders.VideoDecoder.get_frames_played_in_range` and                                                                                                                    
  # :meth:`~torchcodec.decoders.VideoDecoder.get_all_frames` support resampling                                                                                                                      
  # with the ``fps`` parameter to output to a different frame rate.                                                                                                                                        
```

examples/decoding/basic_example.py
```diff
@@ -115,9 +116,10 @@ def plot(frames: torch.Tensor, title: str | None = None):
...
 
 for frame in decoder:
-    assert (
-        isinstance(frame, torch.Tensor)
-        and frame.shape == (3, decoder.metadata.height, decoder.metadata.width)
+    assert isinstance(frame, torch.Tensor) and frame.shape == (
+        3,
+        decoder.metadata.height,
+        decoder.metadata.width,
     )
```
Dan-Flores, line 121: nit: Are all the changes unrelated to the new example from the linter? If not, lets undo them, it makes this PR a little noisier to review.

PR #1305, opened 2026-03-24
Remove get_next_frame and seek_to_pts from core ops

zy1git:
Summary:

`get_next_frame` is a legacy cursor-based API that is not exposed by the public `VideoDecoder` class. All usages have been replaced with the index-based and PTS-based public APIs (`get_frame_at_index`, `get_frame_at_pts`, `get_frames_at_indices`).
`seek_to_pts` was only used together with `get_next_frame`, so it is also removed.

Test plans:

- Removed all tests that reference to`get_next_frame `.
- CI will validate all tests


test/test_ops.py
```diff
@@ -347,103 +308,6 @@ def test_get_frames_in_range(self, device):
...
-        seek_to_pts(decoder, 12.979633 + 1.0e-4)
-        with pytest.raises(IndexError, match="no more frames"):
-            get_next_frame(decoder)
-
-    @pytest.mark.skipif(
-        get_python_version() >= (3, 14),
-        reason="torch.compile is not supported on Python 3.14+",
-    )
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_compile_seek_and_next(self, device):
```
NicolasHug, line 387: these compile tests we should try to keep and just migrate to other decoding ops

test/test_ops.py
```diff
@@ -347,103 +308,6 @@ def test_get_frames_in_range(self, device):
...
-        )
-        assert_frames_equal(frame0, reference_frame0.to(device))
-        assert_frames_equal(frame_time6, reference_frame_time6.to(device))
-
-    @pytest.mark.parametrize("device", all_supported_devices())
-    @pytest.mark.parametrize(
-        "create_from",
-        ("file", "tensor", "bytes", "file_like_rawio", "file_like_bufferedio"),
-    )
-    def test_create_decoder(self, create_from, device):
```
NicolasHug, line 417: For this test and others, as discussed offline, let's just port them to use other decoding utils instead of removing them entirely. We should only remove those that were exclusively testing `get_next_frame`.

NicolasHug: Let's also try to import this PR internally to see what breaks.

test/test_ops.py
zy1git: Add it back in the new commit.

test/test_ops.py
zy1git: Fixed in the new commit and imported the PR.

test/test_ops.py
```diff
@@ -550,63 +515,6 @@ def test_seek_mode_custom_frame_mappings_fails(self):
                 decoder, stream_index=0, custom_frame_mappings=different_lengths
             )
 
-    @needs_ffmpeg_cli
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_seek_mode_custom_frame_mappings(self, device):
```
zy1git, line 556: should not remove.

test/test_ops.py
```diff
@@ -13,9 +13,7 @@
 
 import numpy as np
 import pytest
-
```
NicolasHug, line 16: avoid

test/test_ops.py
```diff
@@ -69,37 +64,6 @@
 
 
 class TestVideoDecoderOps:
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_seek_and_next(self, device):
```
NicolasHug, line 74: ok to remove

test/test_ops.py
```diff
@@ -69,37 +64,6 @@
...
-        assert_frames_equal(frame1, reference_frame1.to(device))
-        seek_to_pts(decoder, 6.0)
-        frame_time6, _, _ = get_next_frame(decoder)
-        reference_frame_time6 = NASA_VIDEO.get_frame_data_by_index(
-            INDEX_OF_FRAME_AT_6_SECONDS
-        )
-        assert_frames_equal(frame_time6, reference_frame_time6.to(device))
-
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_seek_to_negative_pts(self, device):
```
NicolasHug, line 92: try to replace with `get_frame_played_at()`, but I'm not sure we'd be hitting the same code paths, we might be erroring earlier - in which case, remove

test/test_ops.py
```diff
@@ -352,32 +316,6 @@ def test_get_frames_in_range(self, device):
         empty_frame, *_ = get_frames_in_range(decoder, start=5, stop=5)
         assert_frames_equal(empty_frame, NASA_VIDEO.empty_chw_tensor.to(device))
 
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_throws_exception_at_eof(self, device):
```
NicolasHug, line 357: same as above about negative pts test

test/test_ops.py
```diff
@@ -352,32 +316,6 @@ def test_get_frames_in_range(self, device):
...
-        reference_last_frame = NASA_VIDEO.get_frame_data_by_index(389)
-        assert_frames_equal(last_frame, reference_last_frame.to(device))
-        with pytest.raises(IndexError, match="no more frames"):
-            get_next_frame(decoder)
-
-        with pytest.raises(IndexError, match="no more frames"):
-            get_frame_at_pts(decoder, seconds=1000.0)
-
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_throws_exception_if_seek_too_far(self, device):
```
NicolasHug, line 373: same

test/test_ops.py
```diff
@@ -550,63 +515,6 @@ def test_seek_mode_custom_frame_mappings_fails(self):
...
-            INDEX_OF_FRAME_AT_6_SECONDS, stream_index=stream_index
-        )
-        assert_frames_equal(frame6, reference_frame6.to(device))
-
-        ref_frames0_9 = NASA_VIDEO.get_frame_data_by_range(0, 9)
-        bulk_frames0_9, *_ = get_frames_in_range(decoder, start=0, stop=9)
-        assert_frames_equal(bulk_frames0_9, ref_frames0_9.to(device))
-
-    @pytest.mark.parametrize("color_conversion_library", ("filtergraph", "swscale"))
-    def test_color_conversion_library(self, color_conversion_library):
```
NicolasHug, line 595: don't remove

test/test_ops.py
```diff
@@ -646,19 +554,6 @@ def test_color_conversion_library_with_dimension_order(
         assert frames.shape[1:] == expected_shape
         assert_frames_equal(frames[0], frame0_ref)
 
-    @needs_cuda
-    def test_cuda_decoder(self):
```
NicolasHug, line 651: don't remove

test/test_ops.py
```diff
@@ -933,77 +808,6 @@ def get_all_frames(asset, sample_rate=None, stop_seconds=None):
 
         torch.testing.assert_close(frames_downsampled_to_8000, frames_8000_native)
 
-    @pytest.mark.parametrize("buffering", (0, 1024))
-    @pytest.mark.parametrize("device", all_supported_devices())
-    def test_file_like_decoding(self, buffering, device):
```
NicolasHug, line 939: don't remove

test/test_ops.py
zy1git: Addressed in the new commit.

test/test_ops.py
zy1git: address in the new commit.

test/test_ops.py
zy1git: Addressed in the new commit.

test/test_ops.py
zy1git: Addressed in the new commit.

test/test_ops.py
zy1git: Addressed in the new commit.

test/test_ops.py
zy1git: Addressed in the new commit.

PR #1307, opened 2026-03-24
Change show_lf_header option to False

svekars:


mollyxu:
Thanks for the PR! I ended up removing `show_lf_header` completely in #1308

PR #1311, opened 2026-03-25
[wip] Lossless hdr

mollyxu:


s-deeper:
Small correctness concern in the new P016 path:

`convertP016FrameToRGB()` always falls back to `bt601LimitedRange16ColorTwist` for non-BT.709 / non-BT.2020 content. The 8-bit path doesnt do that: when colorspace is unspecified, it still branches on `color_range` and uses BT.601 full-range for `AVCOL_RANGE_JPEG`.

Should the 16-bit fallback mirror that behavior? As written, a >8-bit full-range or partially tagged source could be silently decoded as limited-range.

PR #1319, opened 2026-04-01
Add filelike support and tests to streaming encoder

Dan-Flores:
This PR adds a `MultiStreamEncoder` constructor that accepts an `AVIOContextHolder` to support writing to a file-like. 
Tests are updated to call both `create_streaming_encoder_to_file_like` and `create_streaming_encoder_to_file`.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1154,19 +1154,22 @@ class TestMultiStreamEncoderOps:
     @staticmethod
     def _create_encoder(method, tmp_path, format):
         if method == "to_file":
-            output = tmp_path / f"test.{format}"
-            return create_streaming_encoder_to_file(str(output)), output
+            encoder_output = tmp_path / f"test.{format}"
+            return create_streaming_encoder_to_file(str(encoder_output)), encoder_output
         elif method == "to_file_like":
-            output = io.BytesIO()
-            return create_streaming_encoder_to_file_like(format, output), output
+            encoder_output = io.BytesIO()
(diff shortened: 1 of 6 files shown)


test/test_ops.py
```diff
@@ -1150,19 +1151,37 @@ def test_bad_input(self, tmp_path):
...
+            output = tmp_path / f"test.{format}"
+            return create_streaming_encoder_to_file(str(output)), output
+        elif method == "to_file_like":
+            output = io.BytesIO()
+            return create_streaming_encoder_to_file_like(format, output), output
+        else:
+            raise ValueError(f"Unknown method: {method}")
+
+    @staticmethod
+    def _get_decoder_source(output):
```
NicolasHug: seeing source and output in the same context confused me, I didn't know which one was which. I think renaming `output` to `encode_output` would help. Same above in `create_encoder`, to keep consistent names across functions / helpers.

NicolasHug: That being said, as soon as we can, we probably want to get rid of all this and just use the same 'scaffolding' we already have for `TestVideoEncoder` in `test_encoders.py`?

NicolasHug (approved):
Thanks @Dan-Flores !

test/test_ops.py
Dan-Flores: For sure, once we add a Python API we can migrate these tests to `test_encoders.py` and reuse the testing helper functions.

PR #1320, opened 2026-04-02
Optimize WAV decoding with buffered read

Dan-Flores:
This PR enables the optimized file reading from this TODO:
> TODO WavDecoder: Optimize decoding and converison by processing samples in
  fixed size buffer. For now, we use a simpler implementation: create a
  tensor from raw bytes and convert tensor to float32
  
The optimization: read samples and convert them through a fixed size buffer, rather than reading all values at once.
* In WavDecoder::getSamplesInRange, we create a buffer close to `DEFAULT_CHUNK_BUFFER_SIZE`  bytes. 
  * The actual buffer size is adjusted based on the `numBytesPerSample`, to ensure we do not read partial samples.
* In WavDecoder::getSamplesInRange, we process samples according to the buffer size and create the output `samples` tensor.
* WavDecoder::convertSamplesToFloat does the actual conversion to float32.

Additional changes:
`readValue` is split into `safeReadValue` and `readValue`. This is necessary for performance, see inline comments for more details.

Benchmarks:
<details>

[Benchmarking code](https://www.internalfb.com/phabricator/paste/view/P2259734221)

Audio Decoding Benchmark

test/resources/sine_stereo_s32_10min.wav
Warmup: 3, Trials: 10

AudioDecoder       0.249s
WavDecoder         0.148s
SoundFile          0.146s

</details>


src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -250,10 +271,47 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
 AudioFramesOutput WavDecoder::getSamplesInRange(
     double startSeconds,
     std::optional<double> stopSecondsOptional) {
   // Calculate the range of samples to decode
+  STD_TORCH_CHECK(
+      startSeconds <= INT64_MAX / header_.sampleRate,
+      "startSample calculation would overflow: startSeconds * sampleRate");
   // Sample boundary alignment: round to nearest sample to avoid partial samples
   // See corresponding logic in AudioDecoder:
   // https://github.com/meta-pytorch/torchcodec/blob/910005cf5328d9d44ff8123ad540a51db9ce15b5/src/torchcodec/decoders/_audio_decoder.py#L142
```
Dan-Flores, line 317: Driveby fix: since `startSeconds` is a `double`, we need to validate that it will fit into `int64_t` when multiplied by `sampleRate`.
Same below with `stopSecondsOptional`.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -250,10 +271,47 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+  STD_TORCH_CHECK(
+      header_.bitsPerSample == 32,
+      "Unsupported PCM bit depth in conversion: ",
+      header_.bitsPerSample,
+      ". Currently supported: 32");
+
+  // Normalize 32-bit PCM samples to [-1.0, 1.0] range
+  constexpr float scale =
+      1.0f / static_cast<float>(std::numeric_limits<int32_t>::max());
+  const int32_t* intData = getAlignedData<int32_t>(bufferData);
```
Dan-Flores, line 301: `getAlignedData()` returns a typed pointer which allows us to use direct array access instead of `readValue()`, which ends up being slower for many repeated reads.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -108,6 +110,25 @@ void safeSeek(
   STD_TORCH_CHECK(!file.fail(), "Failed to seek to ", pos, " in WAV file");
 }
 
+template <typename T>
```
Dan-Flores, line 113: Later we will add support for other dtypes than `int32`, so this function is templated.

src/torchcodec/_core/WavDecoder.cpp
mollyxu: Thanks for the driveby fix! Would we want to consider creating a helper function for safe multiplication?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -299,51 +360,54 @@ AudioFramesOutput WavDecoder::getSamplesInRange(
...
+    const int64_t samplesInBuffer =
+        bytesToReadThisIteration / header_.numBytesPerSample;
+    float* outputPtr = samples.mutable_data_ptr<float>() +
+        (samplesProcessed * header_.numChannels);
+    convertSamplesToFloat(buffer, samplesInBuffer, outputPtr);
+    samplesProcessed += samplesInBuffer;
+    totalBytesRead += bytesToReadThisIteration;
+  }
 
+  if (samplesProcessed < numSamples) {
```
mollyxu: Would `samplesProcessed` be equal to `numSamples` when it reaches this condition?

Since we are reading `min(bytesToRead - totalBytesRead, alignedBufferSize)`, it seems like when we exit the loop `totalBytesRead = bytesToRead`.  

As such, `samplesProcessed = totalBytesRead / numBytesPerSample = bytesToRead / numBytesPerSample = numSamples`

Dan-Flores: Yes I think you are right, I'll delete this condition since we always end the loop exactly at `numSamples`.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: Currently we have 3 multiplications guarded by a `STD_TORCH_CHECK` with the same error message (`z calculation would overflow: x * y`), and 2 additions guarded by a `STD_TORCH_CHECK` with a similar error. 

We will probably add more checks of this type, but at the moment a safe multiply function would not be used often. Let me know if you think it would improve readability!

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -250,10 +257,45 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+      "Unsupported PCM bit depth in conversion: ",
+      header_.bitsPerSample,
+      ". Currently supported: 32");
+
+  // Normalize 32-bit PCM samples to [-1.0, 1.0] range.
+  // We use readValue because the buffer size is already validated above.
+  constexpr float scale =
+      1.0f / static_cast<float>(std::numeric_limits<int32_t>::max());
+  for (int64_t i = 0; i < totalSamples; ++i) {
+    int32_t sample = readValue<int32_t>(
```
Dan-Flores, line 288: Here we use `readValue()` without the additional `TORCH_CHECK`s in `safeReadValue()` for performance.

It is safe to read values without the bounds checks because the `TORCH_CHECK` above ensures that `bufferData` is large enough for `totalSamples`.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -250,10 +257,37 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
 
+void WavDecoder::convertSamplesToFloat(
+    const std::vector<uint8_t>& bufferData,
+    int64_t samplesInBuffer,
+    float* outputPtr) const {
+  int64_t totalSamples = samplesInBuffer * header_.numChannels;
+
+  // Currently only supporting 32-bit PCM
+  STD_TORCH_CHECK(
+      header_.bitsPerSample == 32,
```
mollyxu: Would we need a torch check here if we check for 32-bit in `validateHeader`?

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -299,50 +336,46 @@ AudioFramesOutput WavDecoder::getSamplesInRange(
...
+      header_.numBytesPerSample,
+      ") exceeds buffer size (",
+      DEFAULT_CHUNK_BUFFER_SIZE,
+      ")");
+
+  // Allocate buffer and read samples in chunks
+  std::vector<uint8_t> buffer(alignedBufferSize);
+
+  int64_t samplesProcessed = 0;
+  auto samples =
```
mollyxu, line 376: From my understanding, `numSamples` = `dataSize / numBytesPerSample`.  Would we want to validate that `dataSize` in case the headerValue is some arbitrarily large number due to corruption?

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: No, this check is redundant, I'll remove it.

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: For integer overflow, we do not need to validate before doing that division, because `int32_t  / uint16_t` will always fit inside `int64_t`.

For general overallocation concerns, this is a bit trickier to validate. Since `dataSize` is `int32_t`, the maximum valid value it can have is substantial but not always incorrect. I'll add a clamp to ensure the tensor allocated is not larger than the actual file, to prevent an incorrect header from causing us to overallocate.

Dan-Flores: To address the other dimension of our allocated tensor, `header_.numChannels`: there is no equivalent ground truth as the filesystem's `fileSize` is to `dataSize`.

Instead, I've added a validation in `validateHeader()` for `numBytesPerSample` that the WAV specification already requires ([labelled here as `BytePerBloc`](https://en.wikipedia.org/wiki/WAV#WAV_file_header)):
```py
numBytesPerSample == numChannels * bitsPerSample / 8
```
We can substitute the values used to calculate `numSamples` to see that the number of elements in the tensor is bounded by the file size:
```py
samples tensor elements = numSamples * header_.numChannels
	= (std::min(header_.dataSize, fileSize_) / header_.numBytesPerSample)) * header_.numChannels
	= (fileSize_ / header_.numBytesPerSample) * header_.numChannels
	= (fileSize_ / (numChannels * bitsPerSample / 8)) * header_.numChannels
	= fileSize_ / (bitsPerSample / 8)
```

As an aside: If the header is corrupted such that `numSamples` is calculated incorrectly, we were already going to fail in `WavDecoder::getSamplesInRange()` when we call `safeReadFile()` to read data beyond what exists. Clamping `dataSize` to `fileSize`  doesn't prevent that failure either, but it does prevent us from overallocating before failing.

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -28,7 +28,9 @@ constexpr int64_t MIN_FMT_CHUNK_SIZE = 16;
 constexpr int64_t MIN_WAVEX_FMT_CHUNK_SIZE = 40;
 // Arbitrary max for fmt chunk allocation - set to 5x extended format size
 constexpr int64_t MAX_FMT_CHUNK_SIZE = 200;
-constexpr int64_t MAX_TENSOR_SIZE = 320'000'000; // 320 MB
+// Soundfile's default chunk size. See
+// https://github.com/libsndfile/libsndfile/blob/master/src/common.h#L77
+constexpr size_t DEFAULT_CHUNK_BUFFER_SIZE = 8192;
```
NicolasHug, line 33: I think we should avoid naming this a "chunk".

A "chunk" in the context of wav files refer to the "format chunk" or "data chunk". This thing is just a temporary buffer that 'views' the data chunk, but it's not a chunk of its own (i.e. it's not a part of the wav specs).

src/torchcodec/_core/WavDecoder.cpp
```diff
@@ -250,18 +272,44 @@ WavDecoder::ChunkInfo WavDecoder::findChunk(
...
+// Callers must ensure outputPtr has space for at least
+// samplesInBuffer * numChannels floats.
+void WavDecoder::convertSamplesToFloat(
+    const std::vector<uint8_t>& bufferData,
+    int64_t samplesInBuffer,
+    float* outputPtr) const {
+  int64_t totalSamples = samplesInBuffer * header_.numChannels;
+
+  // Normalize 32-bit PCM samples to [-1.0, 1.0] range.
+  // We use readValue because the buffer size is already validated above.
```
NicolasHug, line 284: Can you show me where this is validated?

By 'above' I think you mean 'by the caller', but I don't see where. There should also be a comment about this expectation at the top of the function, similar to the one that exists for `outputPtr`.

NicolasHug (approved):

src/torchcodec/_core/WavDecoder.cpp
Dan-Flores: The chain of intermediate variables passed to `readValue()` from `getSamplesInRange()` is complex, so I am less confident in this claim. Since we create the buffer and determine the number of samples we read, we should be safe to read from the buffer without checking the length each time, but this could probably use some refactoring to make this more clear.

PR #1321, opened 2026-04-02
CUDA support on streaming encoder

Dan-Flores:
This PR adds support for GPU encoding in the streaming encoder, copying existing logic from `VideoEncoder`.

* `MultiStreamEncoder::initializeVideoStream()` is updated to detect the device from the passed in frames, and handle CUDA options accordingly. This function is parallel with [VideoEncoder::initializeEncoder](https://github.com/meta-pytorch/torchcodec/blob/910005cf5328d9d44ff8123ad540a51db9ce15b5/src/torchcodec/_core/Encoder.cpp#L746)
  * See [this diffing](https://www.internalfb.com/intern/diffing/?paste_number=2275591167) for a side by side comparison
  
### Testing
These tests are updated to ensure encoding succeeds on CUDA:
* `test_add_video_stream_and_encode_frames` and `test_fragmented_mp4`
These tests are updated to ensure expected errors are raised on CUDA:
* `test_add_frames_different_sizes_errors`
* `test_add_frames_without_stream_errors`
This test is added to ensure adding frames on different devices raises an error:
* `test_add_frames_different_devices_errors`

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -540,7 +540,8 @@ namespace {
 
 torch::stable::Tensor validateFrames(
     const torch::stable::Tensor& frames,
-    const AVCodecContext* avCodecContext = nullptr) {
+    const AVCodecContext* avCodecContext = nullptr,
+    DeviceInterface* deviceInterface = nullptr) {
   STD_TORCH_CHECK(
       frames.scalar_type() == kStableUInt8,
       "frames must have uint8 dtype, got ",
@@ -553,6 +554,22 @@ torch::stable::Tensor validateFrames(
       frames.sizes()[1] == 3,
       "frame must have 3 channels (R, G, B), got ",
       frames.sizes()[1]);
+  if (deviceInterface != nullptr) {
+    auto& expectedDevice = deviceInterface->device();
+    auto framesDevice = frames.device();
+    STD_TORCH_CHECK(
+        static_cast<StableDeviceType>(framesDevice.type()) ==
+                expectedDevice.type() &&
+            framesDevice.index() == expectedDevice.index(),
+        "All frames must be on the same device. Expected device type=",
+        deviceTypeName(expectedDevice.type()),
+        " index=",
+        expectedDevice.index(),
+        ", got device type=",
+        deviceTypeName(static_cast<StableDeviceType>(framesDevice.type())),
+        " index=",
+        framesDevice.index());
+  }
   if (avCodecContext) {
     STD_TORCH_CHECK(
         static_cast<int>(frames.sizes()[2]) == avCodecContext->height &&
@@ -1101,9 +1118,6 @@ void MultiStreamEncoder::addVideoStream(
 void MultiStreamEncoder::initializeVideoStream(
     const torch::stable::Tensor& frames) {
   auto tensorDevice = frames.device();
-  // TODO MultiStreamEncoder: Enable CUDA support
-  STD_TORCH_CHECK(
-      tensorDevice.is_cpu(), "Only CPU tensors are supported for encoding.");
   deviceInterface_ = createDeviceInterface(StableDevice(
       static_cast<StableDeviceType>(tensorDevice.type()),
       tensorDevice.index()));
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -553,6 +554,22 @@ torch::stable::Tensor validateFrames(
...
+            framesDevice.index() == expectedDevice.index(),
+        "All frames must be on the same device. Expected device type=",
+        deviceTypeName(expectedDevice.type()),
+        " index=",
+        expectedDevice.index(),
+        ", got device type=",
+        deviceTypeName(static_cast<StableDeviceType>(framesDevice.type())),
+        " index=",
+        framesDevice.index());
+  }
```
Dan-Flores, line 572: `validateFrames()` is shared by `VideoEncoder` and `MultiStreamEncoder`. 

This added logic is necessary because `MultiStreamEncoder` will need to call `validateFrames()` multiple times, and each batch of frames should be on the same device.

There is no change to `VideoEncoder`, since `validateFrames()` is only called once.

test/test_ops.py
```diff
@@ -1281,19 +1302,40 @@ def test_add_video_stream_twice_errors(self, tmp_path, method):
...
         streaming_encoder_add_video_stream(encoder, frame_rate=30.0)
-        frames_64 = torch.randint(0, 256, (2, 3, 64, 64), dtype=torch.uint8)
-        frames_128 = torch.randint(0, 256, (2, 3, 128, 128), dtype=torch.uint8)
-        streaming_encoder_add_frames(encoder, frames_64)
+        frames_256 = torch.randint(0, 256, (2, 3, 256, 256), dtype=torch.uint8).to(
+            device
+        )
+        frames_512 = torch.randint(0, 256, (2, 3, 512, 512), dtype=torch.uint8).to(
+            device
+        )
```
Dan-Flores, line 1316: small fix, the NVENC codecs have a minimum frame size so I increased the size of these generated tensors.

test/test_ops.py
```diff
@@ -1224,30 +1232,43 @@ def test_create_invalid_format(self, tmp_path, method):
...
+    @pytest.mark.parametrize(
+        "device", ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
+    )
+    def test_fragmented_mp4(self, format, tmp_path, method, device):
+        source_decoder = VideoDecoder(str(TEST_SRC_2_720P.path))
+        source_frames = source_decoder.get_frames_in_range(start=0, stop=10).data.to(
+            device
+        )
         frame_rate = source_decoder.metadata.average_fps
+        percentage, atol = (96, 2) if device == "cuda" else (99, 2)
```
Dan-Flores, line 1244: This matches the tolerance on the [VideoEncoder nvenc test](https://github.com/meta-pytorch/torchcodec/blob/598aaf65e2d6125af34484e1f9b9f5e38ca9a4ca/test/test_encoders.py#L1476)

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1154,18 +1175,32 @@ void MultiStreamEncoder::initializeVideoStream(
   int outWidth = inWidth;
   int outHeight = inHeight;
   AVPixelFormat outPixelFormat = AV_PIX_FMT_NONE;
+
   if (videoStreamOptions_.pixelFormat.has_value()) {
+    // TODO-VideoEncoder: (P2) Enable pixel formats to be set by user on GPU
```
mollyxu: Was this meant to be a multistream encoder todo?

src/torchcodec/_core/Encoder.cpp
mollyxu: Would we want to use `operator==` and do something like the following?

```
    STD_TORCH_CHECK(                                                                                                                                                                                                                                                                    
        framesDevice == expectedDevice,
        "All frames must be on the same device. Expected ",                                                                                                                                                                                                                             
        deviceTypeName(expectedDevice.type()), ":", expectedDevice.index(),
        ", got ",                                                                                                                                                                                                                                                                       
        deviceTypeName(framesDevice.type()), ":", framesDevice.index());
```

test/test_ops.py
```diff
@@ -1281,19 +1302,40 @@ def test_add_video_stream_twice_errors(self, tmp_path, method):
...
+        cuda_frames = cpu_frames.to("cuda")
+        streaming_encoder_add_frames(encoder, cpu_frames)
+        with pytest.raises(RuntimeError, match="same device"):
+            streaming_encoder_add_frames(encoder, cuda_frames)
+
+    @pytest.mark.parametrize("method", ("to_file", "to_file_like"))
+    @pytest.mark.parametrize(
+        "device", ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
+    )
+    def test_add_frames_without_stream_errors(self, tmp_path, method, device):
```
mollyxu, line 1336: Would we also want to have a test for when users try to add a video stream with an invalid `pixel_format`?

mollyxu:
Thanks for making great progress on the multi-stream encoder @Dan-Flores! I left a few questions on the PR

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Yes, fixed!

test/test_ops.py
Dan-Flores: Both `VideoEncoder` and `MultiStreamEncoder` go through `validatePixelFormat()`, so I think the codepath for this error is covered in the existing [`test_pixel_format_errors()` test](https://github.com/meta-pytorch/torchcodec/blob/598aaf65e2d6125af34484e1f9b9f5e38ca9a4ca/test/test_encoders.py#L783).

Once we add a Python API for `MultiStreamEncoder`, we should be able to merge these tests.

mollyxu (approved):

PR #1322, opened 2026-04-04
Support passing extra options to decoder

mavenlin:
Several features can be supported:

- Tuning the probing size when the video url is an http link;
- Support `protocol_whitelist` etc.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <iostream>
+#include <map>
 #include <string_view>
 #include "Metadata.h"
 #include "StableABICompat.h"
@@ -33,6 +34,14 @@ int64_t getPtsOrDts(const UniqueAVFrame& avFrame) {
   return avFrame->pts == INT64_MIN ? avFrame->pkt_dts : avFrame->pts;
 }
 
+void populateInputOptionsDictionary(
+    UniqueAVDictionary& options,
+    const std::map<std::string, std::string>& inputOptions) {
+  for (const auto& [key, value] : inputOptions) {
+    av_dict_set(options.getAddress(), key.c_str(), value.c_str(), 0);
+  }
+}
+
 } // namespace
 
 // --------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


NicolasHug:
Hi @mavenlin , thanks for the PR. Can you share more on which use-cases these features are meant to enable?

mavenlin:
> Hi @mavenlin , thanks for the PR. Can you share more on which use-cases these features are meant to enable?

Hi @NicolasHug, sure

For example, the subfile protocol allows you to stream from a offset in a file.
```
extra_options={"protocol_whitelist": "subfile"},
```

There are more options here at https://ffmpeg.org/ffmpeg-protocols.html#Protocol-Options

NicolasHug:
Thanks for the details @mavenlin . My understanding of the `subfile` protocol is that it allows to treat a byte range within a larger file as if it were a standalone file. I think we already support this natively through Python via our `file-like` API?

Would you be able to share more exactly on how this feature would unblock your specific use-cases?

mavenlin:
> Thanks for the details @mavenlin . My understanding of the `subfile` protocol is that it allows to treat a byte range within a larger file as if it were a standalone file. I think we already support this natively through Python via our `file-like` API?
> 
> Would you be able to share more exactly on how this feature would unblock your specific use-cases?

For local files yes, but what I am trying to turn on is to use subfile to stream directly from a http link.

scotts:
@mavenlin,

> For local files yes, but what I am trying to turn on is to use subfile to stream directly from a http link.

Luckily, that is also supported by our file-like support! :) See our tutorial: [Streaming data through file-like support](https://meta-pytorch.org/torchcodec/stable/generated_examples/decoding/file_like.html
)

mavenlin:
> @mavenlin,
> 
> > For local files yes, but what I am trying to turn on is to use subfile to stream directly from a http link.
> 
> Luckily, that is also supported by our file-like support! :) See our tutorial: [Streaming data through file-like support](https://meta-pytorch.org/torchcodec/stable/generated_examples/decoding/file_like.html)

Yes file like object over the network connection works in this case, but it would require careful tuning to handle the seek, in terms of when to start a new byte ranged request and when to continue streaming with the existing one, while ffmpeg is more directly out of the box.

PR #1325, opened 2026-04-09
[WIP] Create separate so for CUDA 

NicolasHug:


dvrogozh (commented):
@NicolasHug , thank you for working on this. This looks good to me. I hope I will have a chance to verify the PR for XPU before it will be merged. Can you suggest if I can pull the pre-built wheels from this PR somehow for my test as I currently don't have readily available machine with NVidia GPU?

NicolasHug:
Hi @dvrogozh

I actually think I might drop this PR and rely on https://github.com/meta-pytorch/torchcodec/pull/1371 instead, as it seems to be a simpler solution. I'm happy to wait for you to try it out before landing it.

> Can you suggest if I can pull the pre-built wheels from this PR somehow for my test as I currently don't have readily available machine with NVidia GPU?

Do you have access to [this page](https://github.com/meta-pytorch/torchcodec/actions/runs/25755258548?pr=1371)? You can download the built wheels at the bottom.

dvrogozh:
> rely on https://github.com/meta-pytorch/torchcodec/pull/1371 instead, as it seems to be a simpler solution.

@NicolasHug Just #1371 does not work for me as it still leaves dependencies on CUDA specific torch and other CUDA libraries (`libnvrtc.so`). Note that I am installing XPU enabled torch alongside torchcodec, i.e. with:
```
$ uv pip install torch torchcodec --index https://download.pytorch.org/whl/xpu
$ python -c "import torchcodec"
OSError: libnvrtc.so.13: cannot open shared object file: No such file or directory
$ ldd ./.venv/lib/python3.10/site-packages/torchcodec/libtorchcodec_core7.so | grep cuda
        libc10_cuda.so => not found
        libtorch_cuda.so => not found
        libcudart.so.13 => not found
```
At the moment torch is coming from the XPU index at https://download.pytorch.org/whl/xpu and torchcodec is coming from PyPI.

I think we have 2 options:
1. Either go with #1371 and address #1376 (upload torchodec CPU to XPU index)
2. Or go with the #1325 which I think might work for all the cases

It seems you incline towards #1371 which I am fine with, but let's resolve #1376 first.

NicolasHug:
Things work if you make sure to install a CPU-only version of TC, right? I understand the UX isn't perfect, but this does not seem like a hard blocker?

dvrogozh:
> Things work if you make sure to install a CPU-only version of TC, right? I understand the UX isn't perfect, but this does not seem like a hard blocker?

Well, things can be made working by installing CPU-only version of TC, but it's an absolute nightmare to manage. I afraid you might be underestimating complexity as it's not just about `pip install torchcodec --index FROM CPU`. This might needed to be done in multiple locations such as `pyproject.toml` build dependencies configurations, project requirements, etc., etc. I tried to apply such workaround for my XPU ci and project management and stopped doing that - it's just overcomplex. Thus, I do consider that to be a blocker for the torchcodec plugins enabling.

NicolasHug:
My understanding is that, in all the places where you are already specifying `--index https://download.pytorch.org/whl/xpu` to install `torch`, you'd now have to separately install torchcodec by specifying  `--index https://download.pytorch.org/whl/cpu`?

pyproject.toml typically doesn't support specifying any channel, does it? So how are you specifying the XPU index for torch there anyway?

dvrogozh:
> pyproject.toml typically doesn't support specifying any channel, does it? So how are you specifying the XPU index for torch there anyway?

That's the point. Projects typically list dependencies one by one with `torch` followed by `torchcodec`. And you can specify externally the index to fetch from. Specifying 2 indexes should be possible, but I really would like to avoid specifying 2 torch indexes at the same time as I don't want to step into any potential conflicts. To install XPU pytorch plugin we are doing this (building from sources):
```
 uv pip install torch~=2.11.0 -e . \
  --index https://download.pytorch.org/whl/xpu -vv
```

@NicolasHug , it seems we are discussing a workaround. Can we, please, focus on discussing permanent solution? Is anything wrong with doing #1376?

NicolasHug:
Regarding:

> To install XPU pytorch plugin we are doing this (building from sources):
> uv pip install torch~=2.11.0 -e . \
>  --index https://download.pytorch.org/whl/xpu -vv


If you're already running this command, then you'd just need to add a separate one for torchcodec with a different index.

----

I don't underestimate the complexity of the python packaging space. I understand the UX is bad and that you'd like to make it better for your users. I want the exact same thing. I'm dealing with constraints of my own on the TorchCodec side, and what I'm trying to figure out right now is whether we have a blocking issue before merging https://github.com/meta-pytorch/torchcodec/pull/1371, or whether this is something we can live with for a little while.

dvrogozh:
> I'm trying to figure out right now is whether we have a blocking issue before merging https://github.com/meta-pytorch/torchcodec/pull/1371

@NicolasHug , #1371 just improves a CUDA wheel and per se does not impact anything else. It depends on the next steps after PR will be merged to determine if it's a blocking issue. My thoughts are:
1. If you ask me if merging #1371 right now blocks anything on my side in terms of development, then no it does not as we do not rely on non-released wheels from TC (we always rely on the latest released version).
2. The merged #1371 becomes a problem only if you plan to replace CPU wheels with CUDA wheels in 0.13 as it resurrects already reported problem and directly impacts XPU plugin install instruction plus gives significant complexity to XPU development workflow
3. However, if #1376 will be additionally addressed and CPU TC wheels will be uploaded to XPU index, then I believe we will mitigate the issue and effectively decouple XPU path from using PyPI TC.

NicolasHug:
Thanks for clarifying the timeline. It sounds like we can land https://github.com/meta-pytorch/torchcodec/pull/1371 now and, as long as we find a way to ship CPU wheels to the XPU channel before we release 0.13 (where CUDA wheels will be on PyPI), you'll be fine.

I worked on that in https://github.com/meta-pytorch/torchcodec/pull/1384 and https://github.com/pytorch/test-infra/pull/8073.

If https://github.com/meta-pytorch/torchcodec/pull/1384 is green, I'll merge it, then we'll be able to observe tomorrow if our CPU nightly wheels were pushed to `/xpu`.

dvrogozh:
Thanks. I think https://github.com/meta-pytorch/torchcodec/pull/1325#issuecomment-4441932814 should resolve the concerns.

PR #1332, opened 2026-04-14
Add Cpp and Python logging (private APIs)

mollyxu:
## Summary
- Add a unified logging system for torchcodec, with both C++ and Python sides controlled by a single user-facing API: `torchcodec.set_log_level("OFF" / "ALL")`
- Logging is disabled by default (`"OFF"`) — no output unless the user explicitly opts in
- Add CUDA CPU-fallback as the first use case, logged at both the C++ and Python layers
- While both C++ and Python log the same event (CPU fallback), the C++ `TC_LOG` in `BetaCudaDeviceInterface` reports *where* the fallback happens (NVDEC unavailable or codec unsupported). The Python log in `VideoDecoder` reports *that* a fallback happened, at the user-facing API level. These serve different audiences — this specific C++ message is useful for developers debugging internals, while this Python message is useful for users monitoring their pipeline. Currently both fire when logging is enabled. With the planned introduction of log levels, these will naturally separate, so users see only the detail level they need.

## Design

**User-facing API:**
```python
import torchcodec
torchcodec.set_log_level("ALL")
torchcodec.get_log_level()  # returns "OFF" or "ALL"
```

**C++ side:**
Inspired by ExecuTorch's `ET_LOG`, we introduce a `TC_LOG(...)` macro that prints to stderr with `[torchcodec file:line]` formatting. Logging can be toggled on/off at runtime. The `log()` implementation lives in an `internal` namespace (matching ExecuTorch's convention) to signal it shouldn't be called directly.

**Python side:**
Uses the standard `logging` library with a `"torchcodec"` logger. The stderr handler format matches the C++ output for visual consistency. The handler is only attached when logging is enabled.

**Future work:**
Add granular log levels (DEBUG, INFO, WARNING, ERROR) and a `TORCHCODEC_LOG_LEVEL` environment variable, similar to the existing `TORCHCODEC_FFMPEG_LOG_LEVEL`.

## Test plan
- [ ] `test_cpp_logging_cuda_fallback` — C++ TC_LOG fires when NVDEC falls back
- [ ] `test_python_logging_cuda_fallback` — Python logger fires when CPU fallback is detected
- [ ] `test_both_logs_cuda_fallback` — both messages appear together
- [ ] `test_logging_disabled_by_default` — no output when logging is not enabled


test/test_logging.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        )
+        result = subprocess.run(
+            [sys.executable, "-c", script],
+            capture_output=True,
+            text=True,
+        )
+        assert result.returncode == 0, result.stderr
+        assert "falling back to CPU" not in result.stderr
+        assert "CUDA decoding fell back to CPU" not in result.stderr
```
NicolasHug: can we just `assert not result.stderr`?

test/test_logging.py
```diff
@@ -0,0 +1,101 @@
...
+        """
+        )
+        result = subprocess.run(
+            [sys.executable, "-c", script],
+            capture_output=True,
+            text=True,
+        )
+        assert result.returncode == 0, result.stderr
+        assert "falling back to CPU" in result.stderr
+        assert "CUDA decoding fell back to CPU" in result.stderr
```
NicolasHug: This test seems to be the union of both `test_cpp_logging_cuda_fallback` and `test_python_logging_cuda_fallback`. Let's just have one single test!

src/torchcodec/__init__.py
```diff
@@ -25,3 +25,4 @@
 # Similarly, these are exposed for downstream builds that use torchcodec as a
 # dependency.
 from ._core import core_library_path, ffmpeg_major_version  # usort:skip
+from ._core._logging import is_logging_enabled, set_logging_enabled  # usort:skip
```
NicolasHug: Just to be on the safe side, let's not expose these publicly just yet. If we expose them publicly now, and if we're not able to release them safely with 2.12 for whatever reason (unlikely, but still), then we'd have to revert the changes in the release branch, or remember to put them back as private.

If we keep them private from the beginning, we don't have to worry about that, and we can make them public at the very end, once we write the docs / tutorials.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -260,6 +263,7 @@ def cpu_fallback(self) -> CpuFallbackStatus:
                 self._cpu_fallback.status_known = True
 
                 if "CPU fallback" in backend_details:
+                    _LG.info("CUDA decoding fell back to CPU.")
```
NicolasHug, line 270: I understand that logging this info *here* in Python is useful for the tests, where we want to test both the Python and C++ logging paths.

But I'm not sure we'll want to keep this log here. I think in general, we should log stuff at their source, as soon as the relevant info is available. In this case, I suspect the "CPU fallback" behavior should only be logged from within C++, because this is where the fallback happens, and we have all the info we need directly at the C++ layer.

One quirky consequence of logging this here in Python is that we have to "invoke" the logger manually via

```
            _ = decoder.cpu_fallback
```

which isn't super intuitive.


I think it's OK to temporarily keep this for testing purposes, but we should add a TODO to remove it before the release, and consider updating the tests a bit once we have removed this. This can probably happen once we log the "import torchcodec" machinery!

NicolasHug (commented):
Thanks for the great progress @mollyxu !

test/test_logging.py
```diff
@@ -0,0 +1,57 @@
...
+        """Test that both C++ and Python logs fire for CUDA fallback."""
+        script = textwrap.dedent(
+            """\
+            from torchcodec._core._logging import set_log_level
+            set_log_level("ALL")
+            from torchcodec.decoders import VideoDecoder, set_cuda_backend
+            from test.utils import H265_VIDEO
+            with set_cuda_backend("beta"):
+                decoder = VideoDecoder(H265_VIDEO.path, device="cuda")
+            _ = decoder.cpu_fallback
```
NicolasHug, line 30: Let's make sure we remove that line once we address the new TODO, since it shouldn't be needed anymore (same below). Maybe add another TODO so we don't forget!

test/test_logging.py
```diff
@@ -0,0 +1,57 @@
...
+        """
```
NicolasHug, line 35: Can you add `set_log_level("OFF")` in this test and validate that we stop logging anything?

Related: instead of having multiple `assert some_str in result.stdedd`, can we validate `assert result.stderr == the_entire_expected_stderr` ? This should be doable with a regexp that catches all the line numbers in the log (claude would know!)

src/torchcodec/_core/Logging.h
```diff
@@ -0,0 +1,26 @@
...
+void setLogLevel(int level);
+int getLogLevel();
+
+namespace internal {
+void log(const char* file, int line, const char* fmt, ...)
+    __attribute__((format(printf, 3, 4)));
+} // namespace internal
+} // namespace facebook::torchcodec
+
+// NOLINTNEXTLINE(cppcoreguidelines-macro-usage)
```
NicolasHug: Can you verify this and share why this is needed (as a comment in the code as well!)

src/torchcodec/_core/_logging.py
NicolasHug, line 1: I think this entire file should go within `torchcodec/`, not within `torchcodec/_core`. It's similar in nature to `_frame.py`

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+        if _handler not in _LG.handlers:
+            _LG.addHandler(_handler)
+    else:
+        _LG.removeHandler(_handler)
+
+
+def get_log_level() -> str:
+    """Return the current torchcodec log level as a string."""
+    level_int = _get_log_level()
+    return _LEVEL_INT_TO_NAME.get(level_int, f"UNKNOWN({level_int})")
```
NicolasHug: let's raise if `level_int` isn't in `_LEVEL_INT_TO_NAME`: that would be a bug on our part!

test/test_logging.py
```diff
@@ -0,0 +1,57 @@
...
+        result = subprocess.run(
+            [sys.executable, "-c", script],
+            capture_output=True,
+            text=True,
+        )
+        assert result.returncode == 0, result.stderr
+        assert "[torchcodec" in result.stderr
+        assert "falling back to CPU" in result.stderr
+        assert "CUDA decoding fell back to CPU" in result.stderr
+
```
NicolasHug, line 59: We should have a small test for `get_log_level()`!

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+# Currently only OFF and ALL are supported; more levels will be added later.
+_LEVEL_NAME_TO_INT = {
+    "OFF": 0,
+    "ALL": 1,
+}
+
+_LEVEL_INT_TO_NAME = {v: k for k, v in _LEVEL_NAME_TO_INT.items()}
+
+
+def set_log_level(level: str) -> None:
```
NicolasHug: let's type `level` with a `Literal`

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+    """Set the torchcodec log level (both Python and C++ sides).
+
+    Supported levels: "OFF", "ALL".
+    """
+    level = level.upper()
+    if level not in _LEVEL_NAME_TO_INT:
+        raise ValueError(
+            f"Invalid log level: {level!r}. Supported levels: {list(_LEVEL_NAME_TO_INT.keys())}"
```
NicolasHug: TIL about `!r` !

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+_LEVEL_NAME_TO_INT = {
+    "OFF": 0,
+    "ALL": 1,
+}
+
+_LEVEL_INT_TO_NAME = {v: k for k, v in _LEVEL_NAME_TO_INT.items()}
+
+
+def set_log_level(level: str) -> None:
+    """Set the torchcodec log level (both Python and C++ sides).
```
NicolasHug: Whether the log comes from Python or C++ is an implementation detail, so we don't need to document where it comes from. The docs are public-facing!

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+_handler.setFormatter(
+    logging.Formatter("[torchcodec %(filename)s:%(lineno)d] %(message)s")
+)
+
+# Maps user-facing level names to C++ log level ints.
+# Currently only OFF and ALL are supported; more levels will be added later.
+_LEVEL_NAME_TO_INT = {
+    "OFF": 0,
+    "ALL": 1,
+}
```
NicolasHug: This is replicating the functionality of [`enum.Enum`](https://docs.python.org/3/library/enum.html), let's use that instead, and document that we should keep that `Enum` in sync with its corresponding C++ enum (see my other comment about this).

We'll probably still have to pass those as int64 through the custom_ops layer, but they can be enums before and before that.

src/torchcodec/_core/Logging.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <cstdarg>
+#include <cstdio>
+#include <cstring>
+
+namespace facebook::torchcodec {
+
+// Log level: 0 = OFF (default), >0 = enabled.
+// Currently only OFF and ALL (1) are used; more granular levels can be added
+// later without changing this interface.
+static std::atomic<int> gLogLevel{0};
```
NicolasHug: Let's make the level a C++ enum (which is just a non-ordinal int with a name), it'll be easier to keep track of.

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+
+    Supported levels: "OFF", "ALL".
+    """
+    level = level.upper()
+    if level not in _LEVEL_NAME_TO_INT:
+        raise ValueError(
+            f"Invalid log level: {level!r}. Supported levels: {list(_LEVEL_NAME_TO_INT.keys())}"
+        )
+    level_int = _LEVEL_NAME_TO_INT[level]
+    _set_log_level(level_int)
```
NicolasHug: I think we might want to call this one `_set_cpp_log_level()`, to disambiguate.


The existing `_core._get_log_level()` can stay as-is, because it's the source of truth of the log level for both Python and C++, but this `_core._set_log_level()` is exclusively about setting the log level on C++ logic. The Python log level is set separately just below.

src/torchcodec/_core/_logging.py
```diff
@@ -0,0 +1,51 @@
...
+    if level_int > 0:
+        if _handler not in _LG.handlers:
+            _LG.addHandler(_handler)
+    else:
+        _LG.removeHandler(_handler)
```
NicolasHug: This is not thread safe I think. But it's probably OK, at least for now. At least, we should add a small comment pointing this out:

> probably not thread safe, and probably OK still.

mollyxu: Is it ok for now because it is not common to frequently change logging configuration while running?

src/torchcodec/_core/Logging.h
```diff
@@ -0,0 +1,35 @@
...
+  ALL = 1,
+};
+
+void setLogLevel(LogLevel level);
+LogLevel getLogLevel();
+
+namespace internal {
+void log(const char* file, int line, const char* fmt, ...)
+#ifndef _WIN32
+    __attribute__((format(printf, 3, 4)))
```
mollyxu, line 23: `__attribute__((format(printf, 3, 4)))` is not supported on windows

src/torchcodec/_core/Logging.h
```diff
@@ -0,0 +1,35 @@
...
+} // namespace internal
+} // namespace facebook::torchcodec
+
+#define TC_LOG(...)                                                           \
+  do {                                                                        \
+    if (::facebook::torchcodec::getLogLevel() !=                              \
+        ::facebook::torchcodec::LogLevel::OFF) {                              \
+      ::facebook::torchcodec::internal::log(__FILE__, __LINE__, __VA_ARGS__); \
+    }                                                                         \
+  } while (0)
```
Dan-Flores, line 35: TIL the term variadic functions!

src/torchcodec/_logging.py
```diff
@@ -0,0 +1,57 @@
...
+# LICENSE file in the root directory of this source tree.
+
+import enum
+import logging
+from typing import Literal
+
+from torchcodec._core.ops import _get_log_level, _set_cpp_log_level
+
+_LG = logging.getLogger("torchcodec")
+_LG.setLevel(logging.DEBUG)
```
Dan-Flores, line 14: Why do we call `setLevel` here?

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -303,6 +304,8 @@ void BetaCudaDeviceInterface::initialize(
   STD_TORCH_CHECK(avStream != nullptr, "AVStream cannot be null");
   rotation_ = rotationFromDegrees(getRotationFromStream(avStream));
   if (!nvcuvidAvailable_ || !nativeNVDECSupport(device_, codecContext)) {
+    TC_LOG(
+        "NVDEC not available or codec not supported; falling back to CPU decoding.");
```
Dan-Flores, line 308: We could split this into two errors, since the debugging path for each is different. No strong opinion if we need to do that in this PR or later.

src/torchcodec/_core/Logging.cpp
```diff
@@ -0,0 +1,48 @@
...
+namespace internal {
+
+void log(const char* file, int line, const char* fmt, ...) {
+  // Strip path prefix to show only the filename.
+  const char* basename = std::strrchr(file, '/');
+  if (basename != nullptr) {
+    basename += 1;
+  } else {
+    basename = file;
+  }
```
Dan-Flores, line 35: I think we can replace this with a `<filesystem>` function call, which is easier to read and doesn't require us to manually move the `char*` pointer.

```suggestion
auto basename = std::filesystem::path(file).filename().string();
```

src/torchcodec/_logging.py
mollyxu: Thanks for bringing attention to this! I think since the logging level of Python's root logger is `warning`, our example log would not be visible because it is logged at `info` level unless we set the logger to `debug` level here. I've made some changes (including getting rid of the adding/removing handler hack) that should make this logic more clear.

test/test_logging.py
```diff
@@ -0,0 +1,80 @@
...
+            _ = decoder.cpu_fallback
+        """
+        )
+        result = subprocess.run(
+            [sys.executable, "-c", script],
+            capture_output=True,
+            text=True,
+        )
+        assert result.returncode == 0, result.stderr
+        assert not result.stderr
```
Dan-Flores: We could also add a test for passing an invalid log level, to test the error in `set_log_level()`!

test/test_logging.py
```diff
@@ -0,0 +1,80 @@
...
+        assert re.fullmatch(expected, result.stderr), result.stderr
+
+    def test_get_set_log_level(self):
+        """Test that get_log_level reflects set_log_level changes."""
+        from torchcodec._logging import get_log_level, set_log_level
+
+        assert get_log_level() == "OFF"
+        set_log_level("ALL")
+        assert get_log_level() == "ALL"
+        set_log_level("OFF")
```
Dan-Flores: To avoid altering the log level between tests, this approach of calling `set_log_level("OFF")` at the end of tests works when all test assertions pass (and is probably fine for this test).

If we want to be extra careful, we can wrap the test in a `try` block, and move `set_log_level("OFF")` to the  `finally` block to ensure it always runs, regardless of any test failures:
```suggestion
        assert get_log_level() == "OFF"                                                         
        try:                                                                                    
            set_log_level("ALL")
            assert get_log_level() == "ALL"
        finally:
            set_log_level("OFF")
```

src/torchcodec/_logging.py
```diff
@@ -0,0 +1,58 @@
...
+
+def get_log_level() -> str:
+    """Return the current torchcodec log level as a string."""
+    level_int = _get_log_level()
+    try:
+        return _LogLevel(level_int).name
+    except ValueError:
+        raise RuntimeError(
+            f"C++ log level {level_int} has no matching Python level name. "
+        )
```
Dan-Flores, line 59:
```suggestion
def get_log_level() -> str:
    """Return the current torchcodec log level as a string."""
    return _LogLevel(_get_log_level()).name
```

Since `_LogLevel(level_int)` will raise a `ValueError`, we can rely on that error here and simplify this function.

Dan-Flores (commented):
Thanks for updating this @mollyxu! Lets clean up the TODOs so we can review this PR in the final state and merge.

NicolasHug (approved):
Thanks a ton for the great work @mollyxu ! I made some minor changes (mainly removed the fallback logging from Python) and slightly changed the tests. I'll follow up by logging more stuff and exposing this publicly eventually.

PR #1334, opened 2026-04-15
Use lambda deleter in custom_ops

NicolasHug:
Addresses a few TODO_STABLE_ABI, follow-up to https://github.com/meta-pytorch/torchcodec/pull/1260#discussion_r3073733768

I didn't bother verifying that there's no leak since I already made plenty of these checks in the past, and this is just replicating what we're already using.


Dan-Flores (approved):
Thanks for cleaning this up @NicolasHug !

PR #1335, opened 2026-04-15
streaming audio+video and fragmented test

Dan-Flores:
Many lines of code are copied from AudioEncoder and I have not reviewed them yet. 
Please focus on the comments relating to the design decisions!

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -1056,39 +1056,72 @@ MultiStreamEncoder::MultiStreamEncoder(std::string_view fileName) {
       getFFMPEGErrorStringFromErrorCode(status));
 }
 
-void MultiStreamEncoder::addVideoStream(
+int MultiStreamEncoder::addVideoStream(
     double frameRate,
     std::optional<std::string> codec,
     std::optional<std::string> pixelFormat,
     std::optional<double> crf,
     std::optional<std::string> preset,
     std::optional<std::map<std::string, std::string>> extraOptions) {
+  STD_TORCH_CHECK(!closed_, "Cannot operate on a closed encoder.");
   STD_TORCH_CHECK(
-      inFrameRate_ == 0,
+      !hasReceivedData_,
+      "Cannot add streams after data has been written. Register all streams before writing samples or frames.");
+  STD_TORCH_CHECK(
+      videoStreams_.empty(),
       "A video stream has already been added. Cannot add another.");
   STD_TORCH_CHECK(frameRate > 0, "frame_rate must be > 0, got ", frameRate);
-  inFrameRate_ = frameRate;
-  videoStreamOptions_.codec = std::move(codec);
-  videoStreamOptions_.pixelFormat = std::move(pixelFormat);
-  videoStreamOptions_.crf = crf;
-  videoStreamOptions_.preset = std::move(preset);
-  videoStreamOptions_.extraOptions = std::move(extraOptions);
+  VideoStream videoStream;
+  videoStream.inFrameRate = frameRate;
+  videoStream.options.codec = std::move(codec);
+  videoStream.options.pixelFormat = std::move(pixelFormat);
+  videoStream.options.crf = crf;
+  videoStream.options.preset = std::move(preset);
+  videoStream.options.extraOptions = std::move(extraOptions);
+  videoStreams_.push_back(std::move(videoStream));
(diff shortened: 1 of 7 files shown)


src/torchcodec/_core/Encoder.h
```diff
@@ -188,30 +190,94 @@ class FORCE_PUBLIC_VISIBILITY MultiStreamEncoder {
...
+  bool hasReceivedData_ = false;
+  enum class StreamType { Video, Audio };
+
+  struct InitialWrite {
+    StreamType type;
+    int streamIndex;
+    torch::stable::Tensor data;
+  };
+
+  std::vector<InitialWrite> initialWrites_;
```
Dan-Flores, line 280: `initialWrites_` is the buffer that stores the `samples` or `frames` that are written before every stream has had a write. See usage in `MultiStreamEncoder::maybeWriteHeaderAndFlushInitialWrites`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1309,6 +1840,10 @@ void MultiStreamEncoder::close() {
   // TODO MultiStreamEncoder: Revisit if "closed_" flag is useful
   closed_ = true;
 
+  STD_TORCH_CHECK(
+      headerWritten_ || initialWrites_.empty(),
+      "Cannot close encoder before all registered streams have received at least one chunk of data.");
+
```
Dan-Flores, line 1846: This `STD_TORCH_CHECK` is a side effect of automatically writing the header once all streams are initialized. 

We need to check for the case where a stream is added but never written to, in which case all the added `samples` or `frames` are held in the `initialWrites_` buffer. 

Since we never received a tensor for one of the streams containing `height` / `width` / `num_channels`, we cannot write the container header.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1146,121 +1179,314 @@ void MultiStreamEncoder::initializeVideoStream(
...
+  auto initialWrites = std::move(initialWrites_);
+  initialWrites_.clear();
+  for (auto& write : initialWrites) {
+    if (write.type == StreamType::Video) {
+      encodeVideoFrames(write.data, write.streamIndex);
+    } else {
+      encodeAudioSamples(write.data, write.streamIndex);
+    }
+  }
```
Dan-Flores, line 1386: Once the header is written, we can finally write all the `samples` and `frames` that were buffered in `initialWrites_`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1146,121 +1179,314 @@ void MultiStreamEncoder::initializeVideoStream(
...
+    auto avAudioFifo = av_audio_fifo_alloc(
+        audioStream.avCodecContext_->sample_fmt,
+        audioStream.outNumChannels,
+        audioStream.avCodecContext_->frame_size * 2);
+    STD_TORCH_CHECK(avAudioFifo != nullptr, "Couldn't create AVAudioFifo.");
+    audioStream.avAudioFifo_.reset(avAudioFifo);
+  }
+}
+
+void MultiStreamEncoder::maybeWriteHeaderAndFlushInitialWrites() {
```
Dan-Flores, line 1355: This function automatically calls `avformat_write_header()` to write the container header once all streams stored in `videoStreams_` and `audioStreams_` have received data. 

The alternative would be to require the frame `height`/`width`/`num_channels` when adding a video or audio stream, and have the user call an `open()` function that calls `avformat_write_header()`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1056,39 +1056,72 @@ MultiStreamEncoder::MultiStreamEncoder(std::string_view fileName) {
...
     std::optional<std::string> preset,
     std::optional<std::map<std::string, std::string>> extraOptions) {
+  STD_TORCH_CHECK(!closed_, "Cannot operate on a closed encoder.");
   STD_TORCH_CHECK(
-      inFrameRate_ == 0,
+      !hasReceivedData_,
+      "Cannot add streams after data has been written. Register all streams before writing samples or frames.");
+  STD_TORCH_CHECK(
+      videoStreams_.empty(),
       "A video stream has already been added. Cannot add another.");
```
Dan-Flores, line 1072: `videoStreams_` and `audioStreams_` are stored in `std::vector`, but we enforce that only one of each can be added for now.

PR #1342, opened 2026-04-22
Add output type

mollyxu:
## Summary

First of three PRs adding lossless HDR decoding support. This PR adds only the plumbing — no C++ output behavior changes. The default stays `uint8`, so all existing tests pass unchanged.

 - New `OutputDtype` enum (`UINT8` / `FLOAT32` / `AUTO`) in `StreamOptions.h`, plus a matching field on `VideoStreamOptions` (default `UINT8`, so existing callers are unchanged).        
  - `_add_video_stream` / `add_video_stream` schemas in `custom_ops.cpp` gain an optional `output_dtype: str?` arg. Parses `"uint8"` / `"float32"` / `"auto"` into the enum and raises     
  `"Invalid output_dtype=..."` on anything else.                                                                                                                                           
  - New `SingleStreamDecoder::maybeConvertToFloat32` helper, called on every public frame-returning path: `getNextFrame`, `getFrameAtIndex`, `getFramesAtIndices`, `getFramesInRange`,     
  `getFramePlayedAt`, and both branches of `getFramesPlayedInRange`. Returns the tensor unchanged for `UINT8`; for `FLOAT32` it casts to float32 and normalizes by `maxVal`.
  - `torchcodec._core.ops.add_video_stream` gains an `output_dtype: str | None = None` kwarg that forwards to the custom op. That's the only Python surface that changes; `VideoDecoder`,
  `_frame.py`, and `test_decoders.py` are untouched on this branch.     

Follow-ups:
- CPU HDR output — `SwScale` / `FilterGraph` emit RGB48 for >8-bit sources, tensor allocation handles uint16
- Beta CUDA HDR output — NVDEC emits P016, NPP 16-bit color-twist converts to uint16 RGB, and `NVDECCache` keys on surface format.
- Modify API

## Test plan

- New unit tests in `test/test_ops.py` until we modify our API:
  - `test_output_dtype_float32_sdr` — float32 on an 8-bit source returns `float32` in `[0, 1]` and matches `uint8_ref / 255.0`.
  - `test_output_dtype_auto_sdr_is_uint8` — `"auto"` on SDR keeps `uint8` (no behavior change until step 2).
  - `test_output_dtype_invalid` — rejects unsupported dtypes (`"not_a_dtype"`, `blah`).
- Existing decoder tests pass on cpu / cuda / cuda:beta × exact / approximate seek modes.
- No change to default behavior: `VideoDecoder(path)[0].dtype == torch.uint8` remains true.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -8,6 +8,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <iostream>
+#include <limits>
 #include <string_view>
 #include "Metadata.h"
 #include "StableABICompat.h"
@@ -665,6 +666,7 @@ FrameOutput SingleStreamDecoder::getNextFrame() {
   auto output = getNextFrameInternal();
   if (streamInfos_[activeStreamIndex_].avMediaType == AVMEDIA_TYPE_VIDEO) {
     output.data = maybePermuteHWC2CHW(output.data);
+    output.data = maybeConvertToFloat32(output.data);
   }
   return output;
 }
@@ -681,6 +683,7 @@ FrameOutput SingleStreamDecoder::getNextFrameInternal(
 FrameOutput SingleStreamDecoder::getFrameAtIndex(int64_t frameIndex) {
   auto frameOutput = getFrameAtIndexInternal(frameIndex);
   frameOutput.data = maybePermuteHWC2CHW(frameOutput.data);
+  frameOutput.data = maybeConvertToFloat32(frameOutput.data);
   return frameOutput;
 }
 
@@ -782,6 +785,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesAtIndices(
     previousIndexInVideo = indexInVideo;
   }
   frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+  frameBatchOutput.data = maybeConvertToFloat32(frameBatchOutput.data);
   return frameBatchOutput;
 }
 
@@ -826,6 +830,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesInRange(
     frameBatchOutputDurationSeconds[f] = frameOutput.durationSeconds;
   }
   frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+  frameBatchOutput.data = maybeConvertToFloat32(frameBatchOutput.data);
   return frameBatchOutput;
 }
 
@@ -867,6 +872,7 @@ FrameOutput SingleStreamDecoder::getFramePlayedAt(double seconds) {
   // Convert the frame to tensor.
   FrameOutput frameOutput = convertAVFrameToFrameOutput(avFrame);
   frameOutput.data = maybePermuteHWC2CHW(frameOutput.data);
+  frameOutput.data = maybeConvertToFloat32(frameOutput.data);
   return frameOutput;
 }
 
@@ -952,6 +958,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     FrameBatchOutput frameBatchOutput(
         0, getOutputDims(), videoStreamOptions.device);
     frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+    frameBatchOutput.data = maybeConvertToFloat32(frameBatchOutput.data);
     return frameBatchOutput;
   }
 
@@ -1019,6 +1026,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
     }
 
     frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+    frameBatchOutput.data = maybeConvertToFloat32(frameBatchOutput.data);
     return frameBatchOutput;
   } else {
     // Note that we look at nextPts for a frame, and not its pts or duration.
@@ -1051,6 +1059,7 @@ FrameBatchOutput SingleStreamDecoder::getFramesPlayedInRange(
       frameBatchOutputDurationSeconds[f] = frameOutput.durationSeconds;
     }
     frameBatchOutput.data = maybePermuteHWC2CHW(frameBatchOutput.data);
+    frameBatchOutput.data = maybeConvertToFloat32(frameBatchOutput.data);
 
     return frameBatchOutput;
   }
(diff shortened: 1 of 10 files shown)


src/torchcodec/decoders/_video_decoder.py
```diff
@@ -123,6 +123,11 @@ class VideoDecoder:
             :class:`~torchcodec.transforms.DecoderTransform` and
             :class:`~torchvision.transforms.v2.Transform`
             objects. Read more about this parameter in: TODO_DECODER_TRANSFORMS_TUTORIAL.
+        output_dtype (torch.dtype or str, optional): The dtype of decoded frame
+            tensors. One of ``torch.uint8`` (default), ``torch.float32``, or
+            ``"auto"``. ``float32`` returns tensors normalized to ``[0, 1]``.
+            ``auto`` returns ``uint8`` for SDR sources; support for higher bit
+            depths will be added in a follow-up.
```
Dan-Flores: We probably don't need to promise more bit depths here, its not relevant to a user trying to understand our existing API.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -271,11 +292,19 @@ def cpu_fallback(self) -> CpuFallbackStatus:
 
         return self._cpu_fallback
 
+    def _maybe_to_float32(self, tensor: Tensor) -> Tensor:
+        if self._output_dtype == torch.float32 or (
+            self._output_dtype == "auto" and tensor.dtype == torch.uint16
+        ):
+            max_val = 65535.0 if tensor.dtype == torch.uint16 else 255.0
```
Dan-Flores: We can programatically get a dtype's max value using [torch.iinfo](https://docs.pytorch.org/docs/stable/type_info.html), which might be cleaner whan expanding this ternary if we ever support more dtypes:
```
torch.iinfo(tensor.dtype).max
```

test/test_decoders.py
```diff
@@ -184,6 +184,29 @@ def test_create_fails(self):
...
 
+    def test_output_dtype_float32_sdr(self):
+        # float32 on an 8-bit source should be exactly (uint8_output / 255.0).
+
+        uint8_frame = VideoDecoder(NASA_VIDEO.path)[0]
+        float_frame = VideoDecoder(NASA_VIDEO.path, output_dtype=torch.float32)[0]
+
+        assert uint8_frame.dtype == torch.uint8
+        assert float_frame.dtype == torch.float32
+        torch.testing.assert_close(float_frame, uint8_frame.to(torch.float32) / 255.0)
```
Dan-Flores: Nice test!

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -271,11 +292,19 @@ def cpu_fallback(self) -> CpuFallbackStatus:
...
+            return tensor.to(torch.float32).div_(max_val)
+        return tensor
+
     def _getitem_int(self, key: int) -> Tensor:
         assert isinstance(key, int)
 
         frame_data, *_ = core.get_frame_at_index(self._decoder, frame_index=key)
-        return frame_data
+        return self._maybe_to_float32(frame_data)
```
Dan-Flores: The Python API should be a very thin layer over the C++ backend code that implements the complex logic to make our code easier to reason about. Lets try to avoid adding `_maybe_to_float32()` here, since it requires us to update many callsites. 
Could `_maybe_to_float32()` live in `SingleStreamDecoder`, similar to `maybePermuteHWC2CHW`?

src/torchcodec/_core/StreamOptions.h
```diff
@@ -21,6 +21,27 @@ enum ColorConversionLibrary {
...
+  switch (outputDtype) {
+    case OutputDtype::UINT8:
+      return 8;
+    case OutputDtype::FLOAT32:
+      return sourceBitDepth;
+    case OutputDtype::AUTO:
+      return (sourceBitDepth <= 8) ? 8 : sourceBitDepth;
+  }
+  return 8;
+}
```
Dan-Flores: This isn't used until https://github.com/meta-pytorch/torchcodec/pull/1343, lets remove it from this PR.

src/torchcodec/decoders/_video_decoder.py
mollyxu: Thanks for the suggestion! I think it is cleaner to have `_maybe_to_float32` follow the pattern of `maybePermuteHWC2CHW`

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -382,13 +400,17 @@ def get_frames_in_range(self, start: int, stop: int, step: int = 1) -> FrameBatc
...
             start=start,
             stop=stop,
             step=step,
         )
-        return FrameBatch(*frames)
+        return FrameBatch(
+            data=data,
+            pts_seconds=pts_seconds,
+            duration_seconds=duration_seconds,
+        )
```
Dan-Flores: nit: lets reduce the diff here, I think this change was previously needed but not anymore. Same below on the other returned `FrameBatch`.

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -169,6 +173,7 @@ def __init__(
         custom_frame_mappings: (
             str | bytes | io.RawIOBase | io.BufferedReader | None
         ) = None,
+        output_dtype: "torch.dtype | str | None" = None,
```
Dan-Flores: Since this PR only adds the plumbing, lets not expose this option in the public API yet. Adding the option to the public API should come last, after we've fully enabled the feature and tested that it works.

Lets also remove the updated docstrings, since we do not want this to be user facing just yet.

src/torchcodec/_frame.py
```diff
@@ -70,7 +70,7 @@ class FrameBatch(Iterable):
     """
 
     data: Tensor
-    """The frames data (``torch.Tensor`` of uint8)."""
+    """The frames data (``torch.Tensor`` of uint8 by default; float32 in ``[0, 1]`` if ``output_dtype`` is ``torch.float32`` or ``"auto"``)."""
```
Dan-Flores: Lets hold off on these docstring changes as well until this feature is nearly complete.

Dan-Flores (commented):
Nice work on this! Let's hold off on exposing this in the public API until this feature is complete. I think you will need to move the tests to `test_ops.py` once we remove the changes to `VideoDecoder`.

src/torchcodec/_core/_decoder_utils.py
```diff
@@ -166,6 +166,7 @@ def create_video_decoder(
     device_variant: str = "ffmpeg",
     transforms: Sequence[DecoderTransform | nn.Module] | None = None,
     custom_frame_mappings: tuple[Tensor, Tensor, Tensor] | None = None,
+    output_dtype: str | None = None,
```
Dan-Flores: Can we clean this up and add it as part of the public API updates PR, or is it needed now?

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1476,6 +1485,27 @@ torch::stable::Tensor SingleStreamDecoder::maybePermuteHWC2CHW(
   }
 }
 
+torch::stable::Tensor SingleStreamDecoder::maybeConvertToFloat32(
+    torch::stable::Tensor& tensor) {
+  OutputDtype outputDtype =
+      streamInfos_[activeStreamIndex_].videoStreamOptions.outputDtype;
+  bool isUInt16 = tensor.scalar_type() == torch::headeronly::ScalarType::UInt16;
+  bool shouldConvert = (outputDtype == OutputDtype::FLOAT32) ||
+      (outputDtype == OutputDtype::AUTO && isUInt16);
```
Dan-Flores, line 1494: I wonder if we could resolve `OutputDtype::AUTO` to an actual `OutputDtype` earlier, since `auto` just exists for user convenience. This might be easier once we get the `bitDepth` from frames in future PRs?

Dan-Flores (approved):

src/torchcodec/_core/SingleStreamDecoder.cpp
mollyxu: Yeah I think we can resolve `OutputDtype` earlier and I can do it in the next PR when we have `bitDepth` from frames

PR #1343, opened 2026-04-22
HDR CPU decoding

mollyxu:
Second of three PRs adding lossless HDR decoding support. Builds on step 1's `output_dtype` plumbing to make it functional for HDR sources on the CPU path. With this PR, decoding a 10-bit or 12-bit video with `output_dtype=torch.float32` or `"auto"` returns `float32` tensors in `[0, 1]` on CPU — byte-exact against `ffmpeg -pix_fmt rgb48le` references.

- New `kStableUInt16` scalar type and shared `getBitDepthFromAVPixelFormat()` helper (`FFMPEGCommon`).
- `allocateEmptyHWCTensor` and `FrameBatchOutput` take a `bitDepth` parameter; `>8` → uint16 tensor.
- `rgbAVFrameToTensor` handles both RGB24 (uint8) and RGB48 (uint16).
- `SwScale` accepts an `outputFormat` parameter and caches on it.
- `CpuDeviceInterface` uses the `outputBitDepth` it receives via `initializeVideo` to pick the output pixel format (RGB24 vs RGB48) and threads it through the sws and filtergraph paths — user transforms now run in the correct output color space.   

Follow-up:
- **Step 3:** Beta CUDA HDR output — NVDEC emits P016, NPP 16-bit color-twist converts to uint16 RGB, and `NVDECCache` keys on surface format.
- **Public API PR:** re-introduce `output_dtype` on `VideoDecoder.__init__` (and collapse the temporary if/else branches in the transform tests, see below).

### Tests

All `output_dtype` testing in step 2 goes through the internal ops layer (`add_video_stream(decoder, output_dtype=...)`) since it is not exposed as a parameter on `VideoDecoder` yet. This leads to some tests that we cannot parametrize yet.

- Three HDR assets are added plus matching `.rgb48.pt` references generated via ffmpeg CLI.
   - `testsrc2_hdr.mp4` 10-bit
   - `nasa_13013_hdr.mp4`  10-bit
   - `testsrc2_12bit_hdr.mp4` 12-bit


src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -538,6 +539,14 @@ UniqueAVFrame convertAudioAVFrameSamples(
   return convertedAVFrame;
 }
 
+int getBitDepthFromAVPixelFormat(AVPixelFormat format) {
+  const AVPixFmtDescriptor* desc = av_pix_fmt_desc_get(format);
+  if (desc && desc->nb_components > 0) {
+    return desc->comp[0].depth;
+  }
+  return 8;
```
Dan-Flores: When would this case occur?

Dan-Flores: When would this case occur?

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -53,10 +53,14 @@ class DeviceInterface {
...
   // Initialize the device with parameters generic to all kinds of decoding.
+  // outputDtype is passed here (rather than in initializeVideo) so that
+  // implementations can use it during initialization, e.g. to decide whether
+  // hardware decode supports the required output format.
   virtual void initialize(
       const AVStream* avStream,
       const UniqueDecodingAVFormatContext& avFormatCtx,
-      const SharedAVCodecContext& codecContext) = 0;
+      const SharedAVCodecContext& codecContext,
+      OutputDtype outputDtype = OutputDtype::UINT8) = 0;
```
Dan-Flores: nit: I dont think we need a default argument here, the DeviceInterface implementations should override this. Please double check me on this!

test/test_transform_ops.py
```diff
@@ -39,26 +42,88 @@ class TestPublicVideoDecoderTransformOps:
...
+    @pytest.mark.parametrize("output_dtype", [None, torch.float32])
     def test_resize_torchvision(
-        self, video, height_scaling_factor, width_scaling_factor
+        self, video, output_dtype, height_scaling_factor, width_scaling_factor
     ):
         height = int(video.get_height() * height_scaling_factor)
         width = int(video.get_width() * width_scaling_factor)
+        is_hdr = video in (NASA_VIDEO_HDR, TEST_SRC_2_720P_HDR)
+
+        # TODO: Relocate + parametrize over output_dtype once output_dtype is
```
mollyxu: These tests cannot be parametrized yet because `output_dtype` is not exposed on `VideoDecoder` yet. We will get rid of these `if/else` blocks in a follow-up PR.

src/torchcodec/_core/FFMPEGCommon.cpp
mollyxu: Thanks for flagging this! ~~I think `av_pix_fmt_desc_get` would not return null here~~

Previously, decoding didn't depend on the pixel format being resolvable. Because of the new changes with HDR, we are now checking pixel format to resolve `bitDepth`. However, we would still want to ensure that decoding doesn't break in cases of pixel format not being resolvable, thus the default to 8.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -213,8 +237,10 @@ void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
         outputDims.width,
         outputDims.height);
 
-    if (!swScale_ || swScale_->getConfig() != swsConfig) {
-      swScale_ = std::make_unique<SwScale>(swsConfig, swsFlags_);
+    if (!swScale_ || swScale_->getConfig() != swsConfig ||
+        swScale_->getOutputFormat() != outputPixelFormat) {
```
Dan-Flores: Seeing this updated condition made me ask - can `outputPixelFormat` be a part of `swsConfig`?  If so, checking if the format updated will naturally fit into the config comparison, which is simpler.

Then in `SwScale`, we can access the format through the existing `config` field and avoid an additional member variable `outputFormat_`.

Please double check me if this change is possible, but it seems it would reduce the added complexity!

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -117,25 +121,43 @@ std::unique_ptr<DeviceInterface> createDeviceInterface(
...
+  } else {
+    std::vector<int64_t> strides = {avFrameClone->linesize[0], 3, 1};
+    return torch::stable::from_blob(
+        avFrameClone->data[0],
+        shape,
+        strides,
+        StableDevice(kStableCPU),
+        kStableUInt8,
+        deleter);
+  }
```
Dan-Flores: Since these two `torch::stable::from_blob` calls are mostly the same, we could define `strides` and the divisor based on the condition, similar to `bytesPerPixel_` in `SwScale.cpp`:
```
int bytesPerElement = format == AV_PIX_FMT_RGB48 ? 2 : 1;
std::vector<int64_t> strides = {avFrameClone->linesize[0] / bytesPerElement, 3, 1};
return torch::stable::from_blob( 
...
```

src/torchcodec/_core/FFMPEGCommon.cpp
```diff
@@ -538,6 +539,10 @@ UniqueAVFrame convertAudioAVFrameSamples(
   return convertedAVFrame;
 }
 
+int getBitDepthFromAVPixelFormat(AVPixelFormat format) {
+  return av_pix_fmt_desc_get(format)->comp[0].depth;
```
Dan-Flores: The docs for [`av_pix_fmt_desc_get`](https://ffmpeg.org/doxygen/4.1/pixdesc_8h.html#afe0c3e8aef5173de28bbdaea4298f5f0) indicate it can return null, lets add a torch check to ensure a value is returned before accessing the returned value, as we do here: 
https://github.com/meta-pytorch/torchcodec/blob/5cb1582d1002b6740fb9cc9d7f5c633c903f2868/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L206-L209

src/torchcodec/_core/DeviceInterface.h
```diff
@@ -59,12 +59,14 @@ class DeviceInterface {
       const SharedAVCodecContext& codecContext) = 0;
 
   // Initialize the device with parameters specific to video decoding. There is
-  // a default empty implementation.
+  // a default empty implementation. outputBitDepth is the resolved output
+  // bit depth (single source of truth, set once at stream setup).
```
Dan-Flores: I'm not sure this is the right place to document `outputBitDepth`, it seems a bit out of place. Maybe we can move this comment to the location where `outputBitDepth` is resolved?

test/test_ops.py
```diff
@@ -677,19 +682,124 @@ def test_output_dtype_float32_sdr(self):
...
+    def test_10bit_batch_apis(self, asset):
+        # Validate batch APIs on 10-bit float32 content against ffmpeg rgb48
+        # references.
+        decoder = create_from_file(str(asset.path))
+        add_video_stream(decoder, output_dtype="float32")
+
+        def assert_frame_correct(frame, frame_index):
+            frame_as_uint16 = (frame * 65535).round().to(torch.uint16)
+            ref = asset.get_frame_data_by_index_rgb48(frame_index)
+            torch.testing.assert_close(frame_as_uint16, ref, rtol=0, atol=0)
```
Dan-Flores: In the above tests, we use similar frame conversion logic (scale + convert frame, get reference frame, assert frames are close). Since we need to reuse this logic in other tests, lets move this function to be a helper method in TestVideoDecoderOps.
If we do that, we should also rename `assert_frame_correct` to be more specific to the `output_dtype` problem

test/test_ops.py
```diff
@@ -677,19 +682,124 @@ def test_output_dtype_float32_sdr(self):
...
+        # HDR source: auto -> float32 matching ffmpeg rgb48
+        decoder_hdr = create_from_file(str(NASA_VIDEO_HDR.path))
+        add_video_stream(decoder_hdr, output_dtype="auto")
+        for frame_index in [0, 5, 10]:
+            frame, *_ = get_frame_at_index(decoder_hdr, frame_index=frame_index)
+            assert frame.dtype == torch.float32
+
+            if IS_WINDOWS and ffmpeg_major_version < 5:
+                # swscale YUV10->RGB48 differs on Windows + FFmpeg 4
+                continue
```
Dan-Flores: Just curious, what is the difference here? Could we adjust the tolerance, and still assert that the frames are close?

test/test_ops.py
```diff
@@ -677,19 +682,124 @@ def test_output_dtype_float32_sdr(self):
...
+            ),
+            pytest.param(
+                TEST_SRC_2_12BIT_HDR,
+                marks=pytest.mark.skipif(
+                    IS_WINDOWS and ffmpeg_major_version < 5,
+                    reason="swscale YUV12->RGB48 differs on Windows + FFmpeg 4",
+                ),
+            ),
+        ),
+    )
```
Dan-Flores, line 747: I think this can be a condition in the code, similar to `test_output_dtype_auto` to simplify the parametrization here:
`if (windows and ffmpeg4) { continue }`

test/test_ops.py
```diff
@@ -677,19 +682,124 @@ def test_output_dtype_float32_sdr(self):
...
+    def test_10bit_float32_against_ffmpeg_rgb48(self, asset):
```
Dan-Flores: This test name took me a second to reason through, but the test is essentially the same as `test_output_dtype_auto`, but uses `output_dtype=float32`. Could we rename it `test_output_dtype_float32`?

src/torchcodec/_core/FFMPEGCommon.cpp
mollyxu: Thanks for bringing this up! I added a bit more context [here](https://github.com/meta-pytorch/torchcodec/pull/1343#discussion_r3144890216) with my reasoning on why we might want to fallback to 8bits.

test/test_ops.py
mollyxu: I believe the color conversion here is completely different such that they are not close at all

src/torchcodec/_core/SwScale.h
```diff
@@ -39,10 +42,14 @@ class SwScale {
   int swsFlags_;
   bool needsResize_;
 
-  // Color conversion context (input format -> RGB24 at original resolution).
+  // Bytes per pixel for the output format (3 for RGB24, 6 for RGB48).
+  int bytesPerPixel_;
```
Dan-Flores: Since `bytesPerPixel_` is only used in `SwScale::convert()`, we should make it a local variable similar to `bitDepth`.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -86,16 +102,21 @@ void CpuDeviceInterface::initializeVideo(
...
+    // RGB48), as we specify the sink node with the appropriate format.
+    // Filtergraph will automatically insert a format conversion to ensure the
+    // output frame matches the pixel format specified in the sink. But by
+    // default, it will insert it after the user filters. We need an explicit
+    // format conversion to get the behavior we want.
+    //
+    // Build the final filters_ string with the format prefix based on the
+    // resolved output bit depth. The format prefix ensures user transforms
+    // run in the correct output color space (RGB24 or RGB48).
+    AVPixelFormat outputPixelFormat = getOutputPixelFormat(outputBitDepth);
```
Dan-Flores: In this file, we call `getOutputPixelFormat()` 3 times with the same input (`outputBitDepth_`). Lets compute this value once and store it as a member variable `outputPixelFormat_`.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -179,6 +200,11 @@ void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
   // FrameBatchOutputs based on the the stream metadata. But single-frame APIs
   // can still work in such situations, so they should.
   auto inputDims = FrameDims(avFrame->height, avFrame->width);
+  auto avFrameFormat = static_cast<AVPixelFormat>(avFrame->format);
+  // Use outputBitDepth_ (set at stream setup) so output dtype matches the
+  // pre-allocated buffer; avFrameFormat is for input format identification.
+  int bitDepth = outputBitDepth_;
```
Dan-Flores: We can just use `outputBitDepth_` in each function, rather than define this local variable.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -276,7 +302,7 @@ CpuDeviceInterface::convertAVFrameToTensorUsingFilterGraph(
       avFrame->sample_aspect_ratio,
       outputDims.width,
       outputDims.height,
-      /*outputFormat=*/AV_PIX_FMT_RGB24,
+      /*outputFormat=*/outputFormat,
```
Dan-Flores: nit: The argument comment was helpful when we hardcoded `RGB24`, but its redundant now.

src/torchcodec/_core/CpuDeviceInterface.h
```diff
@@ -65,14 +66,17 @@ class CpuDeviceInterface : public DeviceInterface {
...
+      const FrameDims& outputDims,
+      int bitDepth);
 
   ColorConversionLibrary getColorConversionLibrary(
       const FrameDims& inputDims,
       const FrameDims& outputDims) const;
 
   VideoStreamOptions videoStreamOptions_;
   AVRational timeBase_;
+  // Resolved output bit depth, set once via initializeVideo().
```
Dan-Flores: nit: this comment is not super helpful, lets remove it. Also, if we change the location where `outputBitDepth_` gets set, we will have to keep this comment in sync.

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -117,24 +121,35 @@ std::unique_ptr<DeviceInterface> createDeviceInterface(
 }
 
 torch::stable::Tensor rgbAVFrameToTensor(const UniqueAVFrame& avFrame) {
-  STD_TORCH_CHECK(avFrame->format == AV_PIX_FMT_RGB24, "Expected RGB24 format");
+  auto format = static_cast<AVPixelFormat>(avFrame->format);
+  STD_TORCH_CHECK(
+      format == AV_PIX_FMT_RGB24 || format == AV_PIX_FMT_RGB48,
+      "Expected RGB24 or RGB48 format, got ",
+      av_get_pix_fmt_name(format));
```
Dan-Flores: `av_get_pix_fmt_name()` can return null, let's add a ternary as this `STD_TORCH_CHECK` does:
https://github.com/meta-pytorch/torchcodec/blob/cbd2b687f328e51429d892a8a6ff1ccfa74ed820/src/torchcodec/_core/CudaDeviceInterface.cpp#L311-L315

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -578,9 +578,22 @@ void SingleStreamDecoder::addVideoStream(
         activeStreamIndex_, customFrameMappings.value());
   }
 
-  // Set preRotationDims_ for the active stream. These are the raw encoded
-  // dimensions from FFmpeg, used as a fallback for tensor pre-allocation when
-  // no resize/rotation transforms are applied.
```
Dan-Flores, line 583: Was this comment deleted intentionally? If not, lets keep it.

src/torchcodec/_core/SingleStreamDecoder.h
```diff
@@ -405,6 +405,8 @@ class FORCE_PUBLIC_VISIBILITY SingleStreamDecoder {
   std::vector<std::unique_ptr<Transform>> transforms_;
   std::optional<FrameDims> resizedOutputDims_;
   FrameDims preRotationDims_;
+  // Resolved output bit depth, set once in addVideoStream.
+  int outputBitDepth_;
```
Dan-Flores: nit: This comment seems out of place. From a code maintainers perspective, we can ctrl+F and find the same information, so it probably does not need to be a comment.

src/torchcodec/_core/StreamOptions.h
```diff
@@ -27,6 +27,21 @@ enum ColorConversionLibrary {
...
+inline int resolvedBitDepth(int sourceBitDepth, OutputDtype outputDtype) {
+  switch (outputDtype) {
+    case OutputDtype::UINT8:
+      return 8;
+    case OutputDtype::FLOAT32:
+      return sourceBitDepth;
+    case OutputDtype::AUTO:
+      return (sourceBitDepth <= 8) ? 8 : sourceBitDepth;
+  }
+  return 8;
```
Dan-Flores: `OutputDtype` is only defined for these 3 values, so this final return value is not needed. Lets add a `STD_TORCH_CHECK` if somehow we get an unexpected value

test/generate_reference_resources.py
```diff
@@ -80,6 +88,47 @@ def generate_frame_by_index(
     convert_image_to_tensor(output_bmp)
 
 
+def generate_frame_by_index_rgb48(
```
Dan-Flores, line 91: Ideally, this function would be an offshoot of `generate_frame_by_index()`. 

My understanding is that it needs to be separate because `PIL` will truncate the 16bit image to 8bits, so we generate the frame from ffmpeg using `rawvideo`, and pass it directly to numpy instead, and because of all this none of the existing code is reusable?

test/test_transform_ops.py
```diff
@@ -103,6 +102,85 @@ def test_resize_torchvision(
...
+                # FFmpeg 4 (root cause unknown).
+                continue
+
+            # Compare in 8-bit-equivalent rounded space so absolute tolerances
+            # match those of the SDR uint8 test (atol=1 ≈ 1 code value).
+            fr, ft, ft_na = (
+                (t * 255).round()
+                for t in (frame_resize, frame_tv, frame_tv_no_antialias)
+            )
+
```
Dan-Flores: I've never seen the term "code value" before. Is this rounding necessary? 

We could alternatively adjust the `atol` set below:
```
assert_tensor_close_on_at_least(
  frame_resize, frame_tv, percentage=99.7, atol=1/255
)
```

test/test_transform_ops.py
```diff
@@ -103,6 +102,85 @@ def test_resize_torchvision(
...
 
+    @pytest.mark.parametrize(
+        "height_scaling_factor, width_scaling_factor",
+        ((1.5, 1.31), (0.5, 0.71), (0.7, 1.31), (1.5, 0.71), (1.0, 1.0), (2.0, 2.0)),
+    )
+    @pytest.mark.parametrize(
+        "video",
+        [NASA_VIDEO, TEST_SRC_2_720P, NASA_VIDEO_HDR, TEST_SRC_2_720P_HDR],
+    )
+    def test_resize_torchvision_float32(
```
Dan-Flores: Sharing this [diffing](https://www.internalfb.com/intern/diffing/?paste_number=2307011531) I used to see that this test is mostly copied from `test_resize_torchvision`.

Dan-Flores: These tests are quite large, and mostly duplicate. This makes me think, do we need to test transforms + HDR?

My understanding is that enabling HDR mainly affects transforms during config set up, for ex when we call `createSwsContext()` we now pass `config_.outputFormat` instead of `AV_PIX_FMT_RGB24`. 

For the most part, this is tested by the new HDR tests.

In the future we should extend the existing tests to be parametrized over HDR videos (in case swscale's outputs differ significantly from torchvision, we would want to know), but for now I would argue these duplicate tests are not worth adding. Let me know your thoughts on this @mollyxu!

test/generate_reference_resources.py
mollyxu: Yeah PIL will truncate the image to 8bits. While we could re-use some of the code, the additional indirection might make it not worth it to extract it into a helper.

test/test_transform_ops.py
NicolasHug: I share Daniel's thoughts. We'll want to have these tests eventually, and we might want to find ways to mitigate the potential discrepancies with TV, but we don't necessarily need them just now at this early development stage. They will be easier to write and to integrate once that feature is available via `VideoDecoder`, too.

If we keep these tests, let's at least add TODOs mentioning we should port them to the `VideoDecoder` class and just fuse them with the existing ones through `@pytest.mark.parametrize`.

test/test_ops.py
```diff
@@ -73,6 +78,12 @@
 
 
 class TestVideoDecoderOps:
+    @staticmethod
+    def _assert_float32_matches_rgb48_ref(frame, asset, frame_index):
+        frame_as_uint16 = (frame * 65535).round().to(torch.uint16)
+        ref = asset.get_frame_data_by_index_rgb48(frame_index)
+        torch.testing.assert_close(frame_as_uint16, ref, rtol=0, atol=0)
```
NicolasHug, line 85: I am surprised but please to see that we can have `atol=0` even across all platforms. In the SDR case, the checked-in references that were computed on Linux need a slightly higher tol when comparing with Windows or MacOS (hence why we have `assert_frames_equal`).

test/test_ops.py
NicolasHug: Nit: move this condition at the top of the test and make the test as [`xfail`](https://docs.pytest.org/en/stable/how-to/skipping.html) instead of skipping it. (same for the other instances below).

test/test_ops.py
```diff
@@ -679,19 +690,79 @@ def test_output_dtype_float32_sdr(self):
...
+        indices = [0, 5, 10]
+        frames, *_ = get_frames_at_indices(decoder, frame_indices=indices)
+        for i, idx in enumerate(indices):
+            self._assert_float32_matches_rgb48_ref(frames[i], asset, idx)
+
+        # get_frames_in_range — validate frames 5 and 10 against refs
+        frames_range, *_ = get_frames_in_range(decoder, start=5, stop=11)
+        self._assert_float32_matches_rgb48_ref(frames_range[0], asset, 5)
+        self._assert_float32_matches_rgb48_ref(frames_range[5], asset, 10)
+
```
NicolasHug, line 786: QQ: I may have missed it, but do we test explicitly anywhere what happens when we pass

```py
        decoder = create_from_file(some_hdr_video)
        add_video_stream(decoder, output_dtype="uint8")
```

?

test/test_ops.py
```diff
@@ -679,19 +690,79 @@ def test_output_dtype_float32_sdr(self):
...
+        for frame_index in [0, 5, 10]:
+            frame, *_ = get_frame_at_index(decoder, frame_index=frame_index)
+            assert frame.dtype == torch.float32
+            if IS_WINDOWS and ffmpeg_major_version < 5:
+                # swscale YUV->RGB48 differs on Windows + FFmpeg 4
+                continue
+            self._assert_float32_matches_rgb48_ref(frame, asset, frame_index)
+
+    @pytest.mark.parametrize("asset", (NASA_VIDEO_HDR, TEST_SRC_2_720P_HDR))
+    def test_10bit_batch_apis(self, asset):
```
NicolasHug: Why not check `TEST_SRC_2_12BIT_HDR` here as well?

Let's also test the time-based APIs, as we'll want full coverage.

test/test_ops.py
NicolasHug: Similarly, what happens with:

```py
        decoder = create_from_file(some_sdr_video)
        add_video_stream(decoder, output_dtype="float32")
```
?

src/torchcodec/_core/DeviceInterface.cpp
```diff
@@ -9,6 +9,10 @@
 #include <mutex>
 #include "StableABICompat.h"
 
+extern "C" {
+#include <libavutil/pixdesc.h>
+}
```
NicolasHug: We probably don't need this explicit include. We tend to rely on transitive imports for FFmpeg headers. Here, it should be transitively included via DeviceInterface.h -> FFMPEGCommon.h

(Not saying this is a better practice, but this would be consistent with th rest).

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -578,6 +578,23 @@ void SingleStreamDecoder::addVideoStream(
...
+    streamInfo.videoStreamOptions.outputDtype =
+        (sourceBitDepth > 8) ? OutputDtype::FLOAT32 : OutputDtype::UINT8;
+  }
+
+  // Single source of truth for output bit depth: resolved here once at stream
+  // setup, then passed to the device interface and used for all downstream
+  // frame allocation and format conversion.
+  outputBitDepth_ = resolvedBitDepth(
+      sourceBitDepth, streamInfo.videoStreamOptions.outputDtype);
+
```
NicolasHug: We have a few members and values that all have a `1:1` relationship with each other:

- ouptutBitDepth_ of SingleStreamDecoder and CpuDeviceInterface
- outputPixelFormat_ of the interface
- the ouputDtype class
- the bitDepth parameter of some helpers


They almost all define the exact same thing, with the nuance that:

- OutputDtype has 3 values
- outputPixelFormat_ has many values, only two of which are valid
- outputBitDepth_ has `range_of_int` values, only 2-3 of which are valid


I don't exactly know *how* we can simplify this, but we should at least try to (generally, if a bunch of different things relate to the same concept, there's an opportunity for simplification).

I *think* we can get away by just passing around values of `OutputDtype` only. Maybe we'll still need to pass around AVPixelFormat values, but I think we can get rid of the `int bitDepth` intermediate layer.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -578,6 +578,23 @@ void SingleStreamDecoder::addVideoStream(
...
+  int sourceBitDepth = getBitDepthFromAVPixelFormat(
+      static_cast<AVPixelFormat>(streamInfo.stream->codecpar->format));
+
+  // Resolve AUTO once based on the source bit depth, so downstream code
+  // that reads streamInfo.videoStreamOptions.outputDtype only has to
+  // handle UINT8 / FLOAT32.
+  if (streamInfo.videoStreamOptions.outputDtype == OutputDtype::AUTO) {
+    streamInfo.videoStreamOptions.outputDtype =
+        (sourceBitDepth > 8) ? OutputDtype::FLOAT32 : OutputDtype::UINT8;
+  }
```
NicolasHug, line 592: The resolution of AUTO should probably be within `resolvedBitDepth`, since it's called just after and `resolvedBitDepth` itself assumes the AUTO resolution was just done. 
(this comment may or may not be relevant anymore based on my other comment just below)

test/test_ops.py
mollyxu: Thanks for the feedback! I added a test to cover the first case. As for the second one, I believe it would be covered by `test_output_dtype_float32_sdr`

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -179,6 +195,8 @@ void CpuDeviceInterface::convertVideoAVFrameToFrameOutput(
   // FrameBatchOutputs based on the the stream metadata. But single-frame APIs
   // can still work in such situations, so they should.
   auto inputDims = FrameDims(avFrame->height, avFrame->width);
+  auto avFrameFormat = static_cast<AVPixelFormat>(avFrame->format);
+
```
NicolasHug: Nit: keep this where it was. It's only needed in that specific `if(){}` block, and it's good practice to declare/define variables near to where they're first used.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -578,6 +578,16 @@ void SingleStreamDecoder::addVideoStream(
         activeStreamIndex_, customFrameMappings.value());
   }
 
+  int sourceBitDepth = getBitDepthFromAVPixelFormat(
+      static_cast<AVPixelFormat>(streamInfo.stream->codecpar->format));
```
NicolasHug: Since `getBitDepthFromAVPixelFormat` is only used here, maybe just inline it? We might even be able to get rid of the intermediate `sourceBitDepth` variable, I'm not sure.

src/torchcodec/_core/Frame.h
```diff
@@ -67,6 +68,7 @@ struct AudioFramesOutput {
 torch::stable::Tensor allocateEmptyHWCTensor(
     const FrameDims& frameDims,
     const StableDevice& device,
+    AVPixelFormat pixelFormat,
```
NicolasHug: It feels weird to to accept an `AVPixelFormat`, because `allocateEmptyHWCTensor` should ideally be a "purely pytorch" function, while `AVPixelFormat` is about FFmpeg.

Purely in terms of interface, it might make more sense to accept an OuputDtype instead of a AVPixelFormat. I'm not sure if it'll simplify the call-sites, as we might have to convert AVPixelFormat to OuputDtype, but it's worth trying.

src/torchcodec/_core/Frame.cpp
```diff
@@ -17,35 +17,37 @@ FrameDims::FrameDims(int height, int width) : height(height), width(width) {
...
     const FrameDims& outputDims,
-    const StableDevice& device)
+    const StableDevice& device,
+    OutputDtype outputDtype)
     : ptsSeconds(torch::stable::empty({numFrames}, kStableFloat64)),
       durationSeconds(torch::stable::empty({numFrames}, kStableFloat64)) {
-  data = allocateEmptyHWCTensor(outputDims, device, numFrames);
+  auto pixelFormat =
+      outputDtype == OutputDtype::FLOAT32 ? AV_PIX_FMT_RGB48 : AV_PIX_FMT_RGB24;
+  data = allocateEmptyHWCTensor(outputDims, device, pixelFormat, numFrames);
```
NicolasHug: Following up on my comment above about trying to get `allocateEmptyHWCTensor` to accept an `OutputDtype` instead: this is at least one place where it would make the code simpler.

src/torchcodec/_core/CpuDeviceInterface.cpp
```diff
@@ -86,16 +98,20 @@ void CpuDeviceInterface::initializeVideo(
...
+    // The output frame will always be in the output RGB format (RGB24 or
+    // RGB48), as we specify the sink node with the appropriate format.
+    // Filtergraph will automatically insert a format conversion to ensure the
+    // output frame matches the pixel format specified in the sink. But by
+    // default, it will insert it after the user filters. We need an explicit
+    // format conversion to get the behavior we want.
+    //
+    // Build the final filters_ string with the format prefix based on the
+    // resolved output bit depth. The format prefix ensures user transforms
+    // run in the correct output color space (RGB24 or RGB48).
```
NicolasHug: These 3 new lines seem to just re-hash what was explained above? I'm not sure.

src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -578,6 +578,16 @@ void SingleStreamDecoder::addVideoStream(
...
+
+  // Resolve AUTO once based on the source bit depth, so downstream code
+  // only has to handle UINT8 / FLOAT32.
+  if (streamInfo.videoStreamOptions.outputDtype == OutputDtype::AUTO) {
+    streamInfo.videoStreamOptions.outputDtype =
+        (sourceBitDepth > 8) ? OutputDtype::FLOAT32 : OutputDtype::UINT8;
+  }
```
NicolasHug, line 592: Please add a TODO in the code that we should cleanly establish the invariant that any OutputDtype value is "resolved" and cannot be AUTO past that point. Right now, we have a bunch of comments in the code indicating that it's the case, but it's always best if we can enforce it programmatically.

One obvious way to programmatically do it is to have yet another enum for it where AUTO doesn't exist. But that's annoying. Another way is to remove AUTO and move the  AUTO semantic into `nullopt` by making this an optional parameter - but that only gives us errors at runtime.

I'm rambling. There's very likely a  simpler way, but we can leave it out for later.

test/test_transform_ops.py
NicolasHug: ping on this @mollyxu :)

NicolasHug (approved):
Thanks for the great work @mollyxu !

PR #1344, opened 2026-04-23
Add open() function to streaming encoder

Dan-Flores:
As discussed offline, streaming encoder will need an explicit `open()` function once all the streams have been added.
Some additional consequences of this change:
* When calling `addVideoStream()`, `height` / `width` must be provided. 
* When calling `addVideoStream()`, `device` must also be provided and defaults to CPU.
  * See in line comment, this requires a small tweak in `CudaDeviceInterface`.
* The encoder is more stateful, so errors are raised when the encoding functions are not called in the correct order (`addVideoStream()` -> `open()` -> `addFrames()`).

In preparation of adding a second stream, other changes:
* `deviceInterface` is now a member of the `VideoStream` struct. Each video stream will need its own device interface, and audio streams do not use device interface.

### Testing
The ops tests are updated to use the new `open()` function:
```py
encoder = self._create_encoder(...)
streaming_encoder_add_video_stream(encoder, height=64, width=64, frame_rate=30)
streaming_encoder_open(encoder)
streaming_encoder_add_frames(encoder, frames)
streaming_encoder_close(encoder)
```

### DIFF
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -95,7 +95,7 @@ CudaDeviceInterface::CudaDeviceInterface(const StableDevice& device)
       device_.type() == kStableCUDA, "Unsupported device: must be CUDA");
 
   // Resolve unspecified device index (-1) to the actual current CUDA device.
-  device_ = StableDevice(kStableCUDA, getDeviceIndex(device_));
+  device_.set_index(getDeviceIndex(device_));
 
   initializeCudaContextWithPytorch(device_);
 
--- a/src/torchcodec/_core/CudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/CudaDeviceInterface.cpp
@@ -94,6 +94,9 @@ CudaDeviceInterface::CudaDeviceInterface(const StableDevice& device)
   STD_TORCH_CHECK(
       device_.type() == kStableCUDA, "Unsupported device: must be CUDA");
 
+  // Resolve unspecified device index (-1) to the actual current CUDA device.
+  device_ = StableDevice(kStableCUDA, getDeviceIndex(device_));
+
   initializeCudaContextWithPytorch(device_);
 
   hardwareDeviceCtx_ = getHardwareDeviceContext(device_);
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -1106,16 +1106,16 @@ void MultiStreamEncoder::addVideoStream(
     std::optional<std::string> preset,
     std::optional<std::map<std::string, std::string>> extraOptions) {
   STD_TORCH_CHECK(
-      !videoStream_.has_value(), "Only one video stream is supported.");
+      !videoStream_.has_value(),
+      "A video stream has already been added. Cannot add another.");
   STD_TORCH_CHECK(height > 0, "height must be > 0, got ", height);
   STD_TORCH_CHECK(width > 0, "width must be > 0, got ", width);
   STD_TORCH_CHECK(frameRate > 0, "frame_rate must be > 0, got ", frameRate);
-
   videoStream_.emplace();
   videoStream_->deviceInterface =
       createDeviceInterface(StableDevice(std::move(device)));
-  videoStream_->height = height;
-  videoStream_->width = width;
+  videoStream_->inHeight = height;
+  videoStream_->inWidth = width;
   videoStream_->inFrameRate = frameRate;
   videoStream_->options.codec = std::move(codec);
   videoStream_->options.pixelFormat = std::move(pixelFormat);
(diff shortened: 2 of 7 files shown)


src/torchcodec/_core/CudaDeviceInterface.cpp
```diff
@@ -94,6 +94,9 @@ CudaDeviceInterface::CudaDeviceInterface(const StableDevice& device)
   STD_TORCH_CHECK(
       device_.type() == kStableCUDA, "Unsupported device: must be CUDA");
 
+  // Resolve unspecified device index (-1) to the actual current CUDA device.
+  device_.set_index(getDeviceIndex(device_));
```
Dan-Flores, line 98: This change is needed when users pass in `device=cuda` without a specific device index:
```
streaming_encoder_add_video_stream(encoder, height=h, width=w, frame_rate=fps, device="cuda")
```

[By default](https://github.com/pytorch/pytorch/blob/main/c10/core/Device.h#L36-L39), a `Device` index is -1. This creates an issue in `validateFrames()` when we assert that the newly received `frames` tensor device must match the existing `deviceInterface` device:
```cpp
auto& expectedDevice = deviceInterface->device();
auto framesDevice = frames.device();
STD_TORCH_CHECK(
framesDevice == expectedDevice,
"All frames must be on the same device. Expected ",
```
This will fail because the `frames` device will correctly resolve to an actual index, while our deviceInterface stored the default value -1.

This code updates `device_`'s index to the correct device index using [getDeviceIndex()](https://github.com/meta-pytorch/torchcodec/blob/2b49b20677183ccfa4d111c9b870d1d36f6c2505/src/torchcodec/_core/CUDACommon.cpp#L474). 

When the index is -1, it will be correctly set to the user's default torch device. When the index is any actual index value, the same number will be returned, effectively doing nothing.

mollyxu: Would this be an issue for the `BetaCuda` interface as well?

Dan-Flores: The -1 as default behavior exists in the `BetaCuda` interface, but its not an issue there.

`VideoDecoder` is currently the only user of the `BetaCuda` interface, and it does not compare device index in the same way. When a user passes `device=cuda` without a device index, the index is resolved when it is needed using  [`getDeviceIndex`](https://github.com/search?q=repo%3Ameta-pytorch%2Ftorchcodec%20getDeviceIndex&type=code) or by directly passing [`device_.index()`](https://github.com/search?q=repo%3Ameta-pytorch%2Ftorchcodec+device_.index&type=code) to [`getCurrentCudaStream`](https://github.com/pytorch/pytorch/blob/main/c10/cuda/CUDAStream.cpp#L384), which handles -1 correctly.

For `MultiStreamEncoder`, this delayed device index resolution does not suffice.  By the time we call `validateFrames()` in `Encoder.cpp` to compare the stored device index against a tensor with an actual device index, the -1 default index will not be resolved and will cause an error.

I considered calling `getDeviceIndex` in `Encoder.cpp`, but that would introduce CUDA dependencies to that file, which we want to avoid.

We could add a similar `set_index()` call in the Beta Cuda interface, but its not needed yet, and its not a bug. I am open to suggestions on this.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1277,8 +1278,16 @@ void MultiStreamEncoder::initializeVideoStream(
       status == AVSUCCESS,
       "avcodec_parameters_from_context failed: ",
       getFFMPEGErrorStringFromErrorCode(status));
+}
+
+void MultiStreamEncoder::open() {
+  STD_TORCH_CHECK(!headerWritten_, "open() was already called.");
```
mollyxu, line 1284: I'm curious what are the considerations for naming this `headerWritten` vs `opened`. Would they be essentially the same?

test/test_ops.py
```diff
@@ -1333,12 +1361,35 @@ def test_add_frames_different_devices_errors(self, tmp_path, method):
     @pytest.mark.parametrize(
         "device", ("cpu", pytest.param("cuda", marks=pytest.mark.needs_cuda))
     )
-    def test_add_frames_without_stream_errors(self, tmp_path, method, device):
+    def test_add_frames_without_open_errors(self, tmp_path, method, device):
```
mollyxu, line 1393: When we have a followup PR exposing these changes to the python api would we want to have a section explaining the order dependence of 
```
streaming_encoder_add_video_stream()
streaming_encoder_open()
streaming_encoder_add_frames()
```

mollyxu (approved):
Thanks for the great progress @Dan-Flores !

test/test_ops.py
Dan-Flores: Yes definitely in the tutorial, and likely in the function docstrings as well!

src/torchcodec/_core/Encoder.cpp
Dan-Flores: For the user facing API, my thinking was it is easiest to understand a function named `open()` as it completes the pair with the `close()` function that a user calls once encoding is done.

Internally, variable names are just for us. Since we already have the `headerWritten_` variable and  `open()` calls `avformat_write_header()`, I did not rename it here. Let me know if you think `opened_` would be more clear.

Issue #1339, opened 2026-04-20
`libx265` stale frame returned after backward seek


CarolinePascal:
### 🐛 Describe the bug

Hi !

I was benchmarking video decoding performance with `torchcodec` on HEVC-encoded datasets and noticed that `VideoDecoder` silently returns the wrong frame after seeking backward when the decoder had previously decoded a frame near the end of the video. Here is a minimal reproducible example:

```python
import ctypes
import tempfile
from pathlib import Path

import av
from torchcodec.decoders import VideoDecoder

FPS = 10
NUM_FRAMES = 161   # Frames 0-160, 16.0s at 10 fps
PRIME_IDX = 159    # First seek near EOF to trigger the error
TARGET_IDX = 35    # Second backward seek earlier in the file


def make_hevc_video(path: Path) -> None:
    with av.open(str(path), "w") as container:
        stream = container.add_stream("hevc", FPS, options={"g": "3", "crf": "0"})
        stream.pix_fmt = "yuv420p"
        stream.width = stream.height = 64
        for i in range(NUM_FRAMES):
            frame = av.VideoFrame(64, 64, "yuv420p")
            for plane in frame.planes:
                ctypes.memset(plane.buffer_ptr, i % 256, plane.buffer_size)
            pkt = stream.encode(frame)
            if pkt:
                container.mux(pkt)
        pkt = stream.encode()
        if pkt:
            container.mux(pkt)


with tempfile.TemporaryDirectory() as tmp:
    path = Path(tmp) / "hevc.mp4"
    make_hevc_video(path)

    dec = VideoDecoder(str(path), seek_mode="approximate") # The issue also happens with seek_mode="exact" !

    dec.get_frames_at(indices=[PRIME_IDX]) # First seek - works.
    loaded = dec.get_frames_at(indices=[TARGET_IDX]).pts_seconds[0].item() # Second seek - wrong timestamp !

    expected = TARGET_IDX / FPS
    ok = abs(loaded - expected) < 0.5
    print(f"first={PRIME_IDX}  second={TARGET_IDX}  "
          f"expected={expected:.1f}s  loaded={loaded:.1f}s  "
          f"{'OK' if ok else 'BUG: stale frame returned'}")
    raise SystemExit(0 if ok else 1)
```

The script should print :
```bash
first=159  second=35  expected=3.5s  loaded=16.0s  BUG: stale frame returned
```

[...]

PR #1349, opened 2026-04-28
Manually empty decoder buffer if justSeeked

Dan-Flores:
This PR resolves the bug in https://github.com/meta-pytorch/torchcodec/issues/1339.

### The bug 
While decoding a HEVC video, if the decoder receives an EOF signal followed by a seek, the next frame output by the decoder will be stale, even though we always [flush after seeking](https://github.com/meta-pytorch/torchcodec/blob/08b86ea42bed555c2de52440417b651530b5e71a/src/torchcodec/_core/SingleStreamDecoder.cpp#L1319).

This behavior is detected in the code when:
* `maybeSeekToBeforeDesiredPts` returns true in the current `decodeAVFrame()` call.
* `deviceInterface_->receiveFrame(avFrame)` returns `AVSUCCESS`, indicating a frame was returned **before** any new packets have been sent. 
  * Since we seeked, this frame must be stale and was buffered by the codec at the previous location.
  
The added test reproduces this behavior on an H265 video with these two steps:
* Call `get_frame_at()` requesting a video's penultimate frame.
* Call `get_frame_at()` requesting a video's first frame. 

Before the fix, the second `get_frame_at()`call incorrectly returns the video's last frame.
After the fix, the second `get_frame_at()` call correctly returns the video's first frame.

### The fix
When the above case occurs, we want to ignore those frames that are received after a seek, but before packets are received. 
```py
if (justSeeked && status == AVSUCCESS) {
    continue;
}

...

status = deviceInterface_->sendPacket(packet);
justSeeked = false; # Once a packet is sent, disable this flag to stop discarding frames.
```
Here, we continue to receive frames from the decoder until it returns `EAGAIN`, indicating that the decoder is finally empty. Now, future loop iterations will sending new packets will decode frames at the post-seek location.

### DIFF
--- a/src/torchcodec/_core/SingleStreamDecoder.cpp
+++ b/src/torchcodec/_core/SingleStreamDecoder.cpp
@@ -1277,14 +1277,14 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
 // This method looks at currentPts and desiredPts and seeks in the
 // AVFormatContext if it is needed. We can skip seeking in certain cases. See
 // the comment of canWeAvoidSeeking() for details.
-void SingleStreamDecoder::maybeSeekToBeforeDesiredPts() {
+bool SingleStreamDecoder::maybeSeekToBeforeDesiredPts() {
   validateActiveStream();
   StreamInfo& streamInfo = streamInfos_[activeStreamIndex_];
 
   decodeStats_.numSeeksAttempted++;
   if (canWeAvoidSeeking()) {
     decodeStats_.numSeeksSkipped++;
-    return;
+    return false;
   }
 
   int64_t desiredPts = cursor_;
@@ -1317,6 +1317,7 @@ void SingleStreamDecoder::maybeSeekToBeforeDesiredPts() {
 
   decodeStats_.numFlushes++;
   deviceInterface_->flush();
+  return true;
 }
 
 // --------------------------------------------------------------------------
@@ -1329,7 +1330,7 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
 
   resetDecodeStats();
 
-  maybeSeekToBeforeDesiredPts();
+  bool justSeeked = maybeSeekToBeforeDesiredPts();
   cursorWasJustSet_ = false;
 
   UniqueAVFrame avFrame(av_frame_alloc());
@@ -1342,6 +1343,15 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
   // hardware-specific optimizations.
   while (true) {
     status = deviceInterface_->receiveFrame(avFrame);
+
+    if (justSeeked && status == AVSUCCESS) {
+      // If we just seeked and have not sent any new packets after seeking,
+      // but a frame was returned, the frame is stale.
+      // We should continue to receive frames until we get EAGAIN indicating
+      // the decoder is not holding any more stale frames and is raedy for new
+      // packets.
+      continue;
+    }
 
     if (status != AVSUCCESS && status != AVERROR(EAGAIN)) {
       // Non-retriable error
@@ -1408,6 +1418,7 @@ UniqueAVFrame SingleStreamDecoder::decodeAVFrame(
         "Could not push packet to decoder: ",
         getFFMPEGErrorStringFromErrorCode(status));
 
+    justSeeked = false;
     decodeStats_.numPacketsSentToDecoder++;
   }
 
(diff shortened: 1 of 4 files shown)


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1277,14 +1277,14 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
 // This method looks at currentPts and desiredPts and seeks in the
 // AVFormatContext if it is needed. We can skip seeking in certain cases. See
 // the comment of canWeAvoidSeeking() for details.
-void SingleStreamDecoder::maybeSeekToBeforeDesiredPts() {
+bool SingleStreamDecoder::maybeSeekToBeforeDesiredPts() {
```
mollyxu: Maybe I am missing a bit of context but would it be possible to handle the draining else where like in `DeviceInterface::flush`?  Setting `justSeeked` is introducing a new state and could potentially become hard to track.

mollyxu (commented):
Thanks for looking into this bug @Dan-Flores! I left a question on how we are handling the emptying of the decoder buffer

src/torchcodec/_core/SingleStreamDecoder.cpp
Dan-Flores: I agree tracking `justSeeked` as a new state is not ideal, but the bug is stateful: a frame is only stale if we just seeked and no packets have been sent since that seek.
 
Currently, we call both `maybeSeekToBeforeDesiredPts()` and `flush()` after seeking but before sending packets, so those would be valid places to drain the decoder.

Wherever this draining logic is added, we would need to call `receiveFrame()` and have error handling logic similar to `decodeAVFrame`. So there's a tradeoff in where we add complexity: 
* add a second loop that calls and handles errors from `receiveFrame()` in `flush()`, or
* track `justSeeked` in the existing `receiveFrame()` loop in `decodeAVFrame`

test/test_decoders.py
```diff
@@ -678,6 +678,28 @@ def test_get_frame_played_at_h265(self):
         ref_frame6 = H265_VIDEO.get_frame_data_by_index(5)
         assert_frames_equal(ref_frame6, decoder.get_frame_played_at(0.5).data)
 
+    @pytest.mark.parametrize("device", all_supported_devices())
+    @pytest.mark.parametrize("seek_mode", ("exact", "approximate"))
+    def test_get_frames_at_backward_seek_after_eof(self, seek_mode, device):
```
NicolasHug, line 683: Good test, thank you. Interestingly, this test passes on `main` for the BETA CUDA interface - which is expected, since the bug is that *ffmpeg* isn't fully flushing its  internal buffers, and the BETA CUDA interface doesn't rely on FFmpeg for that part!

src/torchcodec/_core/SingleStreamDecoder.cpp
NicolasHug: I think we should address this within `DeviceInterface::flush` for two reasons:

- conceptual reason. What we're observing is a bug in FFmpeg. By convention, when we call `avcodec_flush_buffers()`, we expect every single next call to `avcodec_receive_frame()` to return `AVERROR(EAGAIN)`. The bug is that, for whatever eason,  `avcodec_receive_frame()` does *not* return `AVERROR(EAGAIN)` in some cases for the HEVC format. `claude` identified that the bug was introduced by https://git.ffmpeg.org/gitweb/ffmpeg.git/commit/79afc45c03abfb7f1d9dc2cc758ec162bfade1a0, although I didn't verify this.

  Since the bug is that  `avcodec_flush_buffers()` isn't doing its job, and since we call `avcodec_flush_buffers()` within `DeviceInterface::flush`, then it makes sense to correct its behavior there.


- practical reason: the fix within `avcodec_flush_buffers()` is actually simpler, and mainly stateless. We just need to manually drain the decoder by calling `avcodec_receive_frame()` and discarding all the frames.

```C++
  // Flush remaining frames from decoder
  virtual void flush() {
    STD_TORCH_CHECK(
        codecContext_ != nullptr,
        "Codec context not available for default flushing");
    avcodec_flush_buffers(codecContext_.get());

    // We also manually flush any remaining frames in the decoder buffer. We
    // shouldn't have to do this, because avcodec_flush_buffers should handle
    // it, but some codecs like HEVC may still have frames buffered internally
    // in weird cases as observed in
    // https://github.com/meta-pytorch/torchcodec/issues/1339.
    UniqueAVFrame tmp(av_frame_alloc());
    while (avcodec_receive_frame(codecContext_.get(), tmp.get()) == AVSUCCESS) {
    }
  }
```

NicolasHug (commented):
Thanks a ton for the investigation and for the fix @Dan-Flores ! I think there' s a slightly simpler way to fix this, I left a comment below. LMK what you think.

NicolasHug (approved):
LGTM if green, thank you!!

Dan-Flores:
mypy lint failure is unrelated: https://github.com/meta-pytorch/torchcodec/actions/runs/25500288007/job/74830989054?pr=1349

Issue #1336, opened 2026-04-16
Make the BETA CUDA interface the default


NicolasHug:
We announced in  [0.11](https://github.com/meta-pytorch/torchcodec/releases/tag/v0.11.0) that the BETA CUDA interface will become the default. Let's do it.

In the following I'm going to refer to the 'Beta' interface as the 'default' interface. The 'ffmpeg' interface is still the 'ffmpeg' interface.

In terms of UX, I think we want:

```py
dec = VideoDecoder(..., device="cuda")  # now relies on the 'default' interface (i.e. 'beta'). Previously this relied on 'ffmpeg'

# to manually select it:
with set_cuda_backend("default"): 
    VideoDecoder(...)

# still works for BC, but we should stop document this as officially supported:
with set_cuda_backend("beta"): 
    VideoDecoder(...)


# still works. This is now the only way to get the ffmpeg interface.
with set_cuda_backend("ffmpeg"):  
    VideoDecoder(...)
```

We'll have to do some small plumbing changes too, particularly to the tests. And we should also update our docs to stop mentioning "beta". This can be done in a few different PRs.

Eventually, we might want to deprecate and stop exposing the 'ffmpeg' interface publicly, and only have it for tests, but that's a discussion that's out of scope for this issue.

NicolasHug:
Closing, this was addressed in https://github.com/meta-pytorch/torchcodec/pull/1350 and https://github.com/meta-pytorch/torchcodec/pull/1359. Thanks @mollyxu !

PR #1350, opened 2026-04-28
Switch default cuda backend to betacuda

mollyxu:
Related Issue: #1336

Switches the public default for CUDA decoding from the FFmpeg-based interface to the NVDEC-based one (previously called "beta"). Renames `"beta"` → `"default"` through every layer of the stack; `"beta"` still works as a BC alias on the public API.

```python
VideoDecoder("...", device="cuda")            # BetaCuda (was: FFmpeg CUDA)
with set_cuda_backend("ffmpeg"):              # only way to get FFmpeg CUDA now
    VideoDecoder("...", device="cuda")
```

### What changed

- Contextvar + `_core` defaults switched to `"default"`; `set_cuda_backend()` accepts `"default"`/`"ffmpeg"`/`"beta"` (alias).
- C++ registry: NVDEC under `(CUDA, "default")`, ffmpeg CUDA explicit under `(CUDA, "ffmpeg")`, CPU under `(CPU, "default")`. Registry default-variant switched to `"default"` so internal callers (encoder, audio path, `createDeviceInterface()`) resolve correctly.
- Fallback-status string `"Beta CUDA"` → `"CUDA"`.
- Tests: `_CUDA_BETA_DEVICE_STR` → `_CUDA_FFMPEG_DEVICE_STR`; cross-backend refs now use `set_cuda_backend("ffmpeg")`; `test_beta_cuda_interface_*` → `test_default_cuda_interface_*`.

### Followup PRs
- Modify docs (`examples/decoding/*.py`)

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -892,8 +892,8 @@ void BetaCudaDeviceInterface::convertAVFrameToFrameOutput(
     // preAllocatedOutputTensor has post-rotation dimensions, but NV12->RGB
     // conversion outputs pre-rotation dimensions, so we can't use it as the
     // conversion destination or validate it against the frame shape.
-    // Once we support native transforms on the beta CUDA interface, rotation
-    // should be handled as part of the transform pipeline instead.
+    // Once we support native transforms on the default CUDA interface,
+    // rotation should be handled as part of the transform pipeline instead.
     frameOutput.data = convertNV12FrameToRGB(
         gpuFrame,
         device_,
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -68,8 +68,8 @@ static DecoderCapsCache& getDecoderCapsCache() {
   return cache;
 }
 
-static bool g_cuda_beta = registerDeviceInterface(
-    DeviceInterfaceKey(kStableCUDA, /*variant=*/"beta"),
+static bool g_cuda_default = registerDeviceInterface(
+    DeviceInterfaceKey(kStableCUDA, /*variant=*/"default"),
     [](const StableDevice& device) {
       return new BetaCudaDeviceInterface(device);
     });
@@ -265,7 +265,8 @@ void cudaBufferFreeCallback(void* opaque, [[maybe_unused]] uint8_t* data) {
 
 BetaCudaDeviceInterface::BetaCudaDeviceInterface(const StableDevice& device)
     : DeviceInterface(device) {
-  STD_TORCH_CHECK(g_cuda_beta, "BetaCudaDeviceInterface was not registered!");
+  STD_TORCH_CHECK(
+      g_cuda_default, "BetaCudaDeviceInterface was not registered!");
   STD_TORCH_CHECK(
       device_.type() == kStableCUDA, "Unsupported device: must be CUDA");
 
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -68,6 +68,7 @@ static DecoderCapsCache& getDecoderCapsCache() {
   return cache;
 }
 
+// TODO: rename private variant "default" to "nvdec" to match public name.
 static bool g_cuda_default = registerDeviceInterface(
     DeviceInterfaceKey(kStableCUDA, /*variant=*/"default"),
     [](const StableDevice& device) {
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -706,6 +706,14 @@ void sortCodecOptions(
     }
   }
 }
+
+// The default CUDA interface is decode-only; encoders need the FFmpeg-based
+// one.
+std::unique_ptr<DeviceInterface> createEncoderDeviceInterface(
+    const StableDevice& device) {
+  return createDeviceInterface(
+      device, device.type() == kStableCUDA ? "ffmpeg" : "default");
+}
 } // namespace
 
 VideoEncoder::~VideoEncoder() {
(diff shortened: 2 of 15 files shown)


test/test_decoders.py
```diff
@@ -238,6 +238,7 @@ def test_getitem_numpy_int(self):
     @pytest.mark.parametrize("device", all_supported_devices())
     @pytest.mark.parametrize("seek_mode", ("exact", "approximate"))
     def test_getitem_slice(self, device, seek_mode):
+        device_param = device  # make_video_decoder shadows `device` below
```
mollyxu, line 241: Drive-by  fix: `device` from the parametrization was overridden by line 242 and only matched line 391's check accidentally because `device=="cuda"` happened to be true. With the switch to `betacuda` as the new default backend, line 391 was no longer skipping properly

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -68,8 +68,8 @@ static DecoderCapsCache& getDecoderCapsCache() {
   return cache;
 }
 
-static bool g_cuda_beta = registerDeviceInterface(
-    DeviceInterfaceKey(kStableCUDA, /*variant=*/"beta"),
+static bool g_cuda_default = registerDeviceInterface(
+    DeviceInterfaceKey(kStableCUDA, /*variant=*/"default"),
     [](const StableDevice& device) {
       return new BetaCudaDeviceInterface(device);
```
mollyxu, line 75: I didn't think it fit the scope of this PR but would we want to rename `BetaCudaDeviceInterface` eventually as well?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -706,6 +706,14 @@ void sortCodecOptions(
     }
   }
 }
+
+// The default CUDA interface is decode-only; encoders need the FFmpeg-based
```
mollyxu: As an alternative to having this helper, we could also consider dual-registering `CpuDeviceInterface` under both `CPU` and `"default"`

src/torchcodec/decoders/_decoder_utils.py
```diff
@@ -24,14 +24,8 @@ def set_cuda_backend(backend: str) -> Generator[None, None, None]:
...
     This context manager allows you to specify which CUDA backend implementation
     to use when creating :class:`~torchcodec.decoders.VideoDecoder` instances
-    with CUDA devices.
-
-    .. note::
-        **We recommend trying the "beta" backend instead of the default "ffmpeg"
-        backend!** The beta backend is faster, and will eventually become the
-        default in future versions. It may have rough edges that we'll polish
-        over time, but it's already quite stable and ready for adoption. Let us
-        know what you think!
```
Dan-Flores, line 34: Let's handle the public facing docstring changes in a second PR, and reduce the changes in this PR to our plumbing + tests + relevant internal facing comments.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Earlier I suggested moving this to DeviceInterface.h to be near [createDeviceInterface()](https://github.com/meta-pytorch/torchcodec/blob/cbd2b687f328e51429d892a8a6ff1ccfa74ed820/src/torchcodec/_core/DeviceInterface.cpp#L98). 

Now that I've had more time to think about it, perhaps it would be simpler to apply this ternary or something similar at the two `createDeviceInterface` callsites in `Encoder.cpp`. That way, we avoid having another entry point to creating a device interface.

test/test_decoders.py
```diff
@@ -1715,14 +1721,14 @@ def test_get_frames_at_tensor_indices(self):
...
-    def test_beta_cuda_interface_get_frame_at(
+    def test_default_cuda_interface_get_frame_at(
         self, asset, contiguous_indices, seek_mode
     ):
-        ref_decoder = VideoDecoder(asset.path, device="cuda", seek_mode=seek_mode)
-        with set_cuda_backend("beta"):
-            beta_decoder = VideoDecoder(asset.path, device="cuda", seek_mode=seek_mode)
+        with set_cuda_backend("ffmpeg"):
+            ref_decoder = VideoDecoder(asset.path, device="cuda", seek_mode=seek_mode)
+        default_decoder = VideoDecoder(asset.path, device="cuda", seek_mode=seek_mode)
```
Dan-Flores, line 1729: We should establish a pattern internally for referring to the previous "Beta Cuda" backend. 

I think we need a name, because `default_decoder` does not make it obvious that the decoder is GPU enabled. Also, if we were to change the default backend, which backend "default" refers to in a test variable would become ambiguous. 

In this test, `nvdec_decoder` or even `cuda_decoder` would be more clear. 

Using multiple names across the code might also be confusing, though. I wonder if we really want to expose `default` as the user facing string, or if we should pick `nvdec` or some higher signal name and use that across the codebase.

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
NicolasHug: Yes, and agreed to leave it for another PR!

src/torchcodec/_core/Encoder.cpp
```diff
@@ -777,9 +778,12 @@ VideoEncoder::VideoEncoder(
...
   auto tensorDevice = frames_.device();
-  deviceInterface_ = createDeviceInterface(StableDevice(
-      static_cast<StableDeviceType>(tensorDevice.type()),
-      tensorDevice.index()));
+  StableDevice stableDevice(
+      static_cast<StableDeviceType>(tensorDevice.type()), tensorDevice.index());
+  // The default CUDA interface is decode-only; encoders need the FFmpeg-based
+  // one.
+  deviceInterface_ = createDeviceInterface(
+      stableDevice, stableDevice.type() == kStableCUDA ? "ffmpeg" : "default");
```
NicolasHug, line 786: I remember discussing this with you yesterday and suggested to potentially leave a TODO to clean that up, but after re-looking into it, I think the way it's done is just fine. I can't think of a much clearer way of doing it. So, no need for a TODO IMHO.

NicolasHug (approved):
Thanks for the PR @mollyxu !

Following up on Daniel's https://github.com/meta-pytorch/torchcodec/pull/1350#discussion_r3185979774 on naming, we'll want to consider two related to but separate names:

- the public facing one (the parameter value to `set_cuda_backend(...)`, in error messages, etc.)
- the private facing one (in our C++ code, in tests, etc.)


Considering this week is quite special in terms of deadline, I suggest we only care about the *public facing* name at this moment. We can handle the private-facing names later.

So, for the public facing name, I previously suggested 'default' in https://github.com/meta-pytorch/torchcodec/issues/1336 and this is what this PR is doing. I agree that 'default' can become ambiguous.

I'm happy to go for `'nvdec'` or `'nvcuvid'` as well (I think users are more familiar with 'nvdec' than 'nvcuvid'). Your call @mollyxu

PR #1356, opened 2026-05-04
Enable open, initialization of audio stream

Dan-Flores:
This PR extends `MultiStreamEncoder` to open and initialize an audio stream. It does not support encoding samples yet.

* `MultiStreamEncoder::addAudioStream()`: 
  * This function stores audio stream parameters in the member variable `audioStream_`. 
  * We currently enforce that only one audio stream can be added.
* `MultiStreamEncoder::initializeAudioStream()`: 
  * This function is mostly duplicated from [`AudioEncoder::initializeEncoder()`](https://github.com/Dan-Flores/torchcodec/blob/3c865b251bc90678621d5249083ea651c477ee8d/src/torchcodec/_core/Encoder.cpp#L176), see [this diffing view](https://www.internalfb.com/intern/diffing/?paste_number=2304181997) for a direct comparison.
  * It validates the input encoding parameters, calls `avcodec_open2` and `avformat_new_stream` to enable encoding.

Testing is limited at the moment, since we do not support writing samples yet. 
* `test_add_audio_stream_and_open_close` checks that the new function calls work. In future PRs we will update this test to write samples similar to `test_add_video_stream_and_encode_frames`.

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -1124,6 +1124,33 @@ void MultiStreamEncoder::addVideoStream(
   videoStream_->options.extraOptions = std::move(extraOptions);
 }
 
+void MultiStreamEncoder::addAudioStream(
+    int sampleRate,
+    int numChannels,
+    std::optional<int> bitRate,
+    std::optional<int> desiredNumChannels,
+    std::optional<int> desiredSampleRate) {
+  STD_TORCH_CHECK(
+      !audioStream_.has_value(),
+      "An audio stream has already been added. Cannot add another.");
+  STD_TORCH_CHECK(sampleRate > 0, "sample_rate must be > 0, got ", sampleRate);
+  STD_TORCH_CHECK(
+      numChannels > 0, "num_channels must be > 0, got ", numChannels);
+  STD_TORCH_CHECK(
+      numChannels <= AV_NUM_DATA_POINTERS,
+      "num_channels must be <= ",
+      AV_NUM_DATA_POINTERS,
+      ", got ",
+      numChannels);
+
+  audioStream_ = AudioStream{};
+  audioStream_->inSampleRate = sampleRate;
+  audioStream_->inNumChannels = numChannels;
+  audioStream_->options.bitRate = bitRate;
+  audioStream_->options.numChannels = desiredNumChannels;
+  audioStream_->options.sampleRate = desiredSampleRate;
+}
+
 void MultiStreamEncoder::initializeVideoStream() {
   // TODO MultiStreamEncoder: Iterate over all video streams
   auto& videoStream = *videoStream_;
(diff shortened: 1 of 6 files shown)


test/test_ops.py
```diff
@@ -1456,6 +1481,45 @@ def test_open_twice_errors(self, tmp_path, method):
...
+        encoder, _ = self._create_encoder(method, tmp_path, "mp4")
+        streaming_encoder_add_audio_stream(
+            encoder,
+            sample_rate=44100,
+            num_channels=2,
+            desired_num_channels=100,
+        )
+        with pytest.raises(
+            RuntimeError,
+            match="Desired number of channels.*is not supported|avcodec_open2 failed",
```
Dan-Flores: [validateNumChannels()](https://github.com/Dan-Flores/torchcodec/blob/3c865b251bc90678621d5249083ea651c477ee8d/src/torchcodec/_core/FFMPEGCommon.cpp#L103) cannot always validate the number of channels. In that case, an invalid `numChannels` arg will error during `avcodec_open2`.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1124,6 +1124,24 @@ void MultiStreamEncoder::addVideoStream(
   videoStream_->options.extraOptions = std::move(extraOptions);
 }
 
+void MultiStreamEncoder::addAudioStream(
+    int sampleRate,
```
mollyxu, line 1128: Would we want to have torch checks for `sampleRate` and `numChannels` being > 0?

test/test_ops.py
```diff
@@ -1261,6 +1262,20 @@ def test_add_video_stream_and_encode_frames(self, tmp_path, format, method, devi
...
+        streaming_encoder_add_audio_stream(
+            encoder,
+            sample_rate=44100,
+            num_channels=2,
+            desired_num_channels=1,
+            desired_sample_rate=16000,
+        )
+        streaming_encoder_open(encoder)
+        streaming_encoder_close(encoder)
+        streaming_encoder_close(encoder)  # double close is a no-op
```
mollyxu: Is there a reason we need a double close check? Would this be different from `test_double_close`

Dan-Flores: Its testing the same idea so its not necessary, I'll remove it.

src/torchcodec/_core/Encoder.cpp
Dan-Flores: Yes, added!

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1124,6 +1124,27 @@ void MultiStreamEncoder::addVideoStream(
...
+    int numChannels,
+    std::optional<int> bitRate,
+    std::optional<int> desiredNumChannels,
+    std::optional<int> desiredSampleRate) {
```
NicolasHug: I recommend you leave out the output num_channels and sample_rate functionality for now. They both add some complexity that we can avoid for the time being.

The output sample rate feature is particularly challenging, and it's responsible for all the FIFO logic documented in this long comment https://github.com/meta-pytorch/torchcodec/blob/9249d03adf69cf475783b9850bc5571d4f93763b/src/torchcodec/_core/Encoder.h#L70-L125

(it was introduced in [this commit](https://github.com/meta-pytorch/torchcodec/commit/eb51e0e97f456cf92c5c6987f6ebeae824ebab5a), which shows the before and after complexity)

NicolasHug (approved):

Dan-Flores:
The [failing lint signal](https://github.com/meta-pytorch/torchcodec/actions/runs/25471670297/job/74736640217?pr=1356) is unrelated, though I am not sure why it would surface now.

PR #1358, opened 2026-05-07
Encode audio in streaming encoder

Dan-Flores:
This PR enables basic audio encoding in the streaming encoder!

The encoding follows the existing flow of the `AudioEncoder`, see [this diffing](https://www.internalfb.com/intern/diffing/?paste_number=2311023992) to compare the main encoding functions.
* `MultiStreamEncoder::addSamples` is the top level encoding entry point that validates inputs and ensures correct API call order.
	* This can only be called once!! See inline TODOs and `addSamplesCalled`.
* `MultiStreamEncoder::encodeAudioSamples` processes the provided samples in portions, calls `maybeConvertAudioAVFrame` and `encodeAudioFrame`.
* `MultiStreamEncoder::maybeConvertAudioAVFrame` will potentially convert the sample format. 
	* Currently, no `output_num_channels` or `output_sample_rate` are supported.
* `MultiStreamEncoder::encodeAudioFrame` calls ffmpeg functions to encode the frame.
* `MultiStreamEncoder::maybeFlushAudioSwrBuffers` checks for remaining samples and flushes, similar to `AudioEncoder::maybeFlushSwrBuffers`.

### Testing
* `test_add_audio_and_video_streams_and_encode` tests multistream encoding with one audio and one video stream!
* `test_add_samples_twice_errors` documents that `addSamples()` can only be called once ATM.

### DIFF
--- a/src/torchcodec/_core/Encoder.cpp
+++ b/src/torchcodec/_core/Encoder.cpp
@@ -20,6 +20,10 @@ torch::stable::Tensor validateSamples(const torch::stable::Tensor& samples) {
       "samples must have float32 dtype, got ",
       (samples.scalar_type()));
   STD_TORCH_CHECK(
+      samples.device().type() == kStableCPU,
+      "samples must be on CPU, got ",
+      deviceTypeName(samples.device().type()));
+  STD_TORCH_CHECK(
       samples.dim() == 2,
       "samples must have 2 dimensions, got ",
       samples.dim());
@@ -1326,6 +1330,8 @@ void MultiStreamEncoder::initializeAudioStream() {
 
   validateSampleRate(*avCodec, audioStream.inSampleRate);
   audioStream.avCodecContext->sample_rate = audioStream.inSampleRate;
+  audioStream.avCodecContext->time_base =
+      AVRational{1, audioStream.inSampleRate};
 
   // Input samples are expected to be FLTP. Not all encoders support FLTP, so we
   // may need to convert the samples into a supported output sample format,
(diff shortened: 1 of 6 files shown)


src/torchcodec/_core/Encoder.cpp
```diff
@@ -19,6 +19,10 @@ torch::stable::Tensor validateSamples(const torch::stable::Tensor& samples) {
       samples.scalar_type() == kStableFloat32,
       "samples must have float32 dtype, got ",
       (samples.scalar_type()));
+  STD_TORCH_CHECK(
+      samples.device().type() == kStableCPU,
+      "samples must be on CPU, got ",
+      deviceTypeName(samples.device().type()));
```
Dan-Flores, line 25: Drive by fix: since `MultiStreamEncoder` accepts video frame tensors on GPU, it seems feasible that one could pass an audio samples tensor on GPU by accident (no audio codecs are hardware accelerated, that I am aware of). In this case, we should return an error early.

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1456,7 +1462,223 @@ void MultiStreamEncoder::encodeVideoFrame(
...
+    STD_TORCH_CHECK(
+        status >= 0,
+        "Error receiving packet: ",
+        getFFMPEGErrorStringFromErrorCode(status));
+
+    packet->stream_index = audioStream.avStream->index;
+    av_packet_rescale_ts(
+        packet.get(),
+        audioStream.avCodecContext->time_base,
+        audioStream.avStream->time_base);
```
Dan-Flores, line 1633: Without this `av_packet_rescale_ts` call, this test will fail in `test_add_audio_and_video_streams_and_encode` on `mkv`:

```
# One second of audio should contain exactly sample_rate samples.
assert samples_in_first_second.data.shape[1] == sample_rate
```

Now that we are encoding multiple streams to one container, we need to rescale to the stream's `time_base` to correctly interleave packets. The `AudioEncoder` did not need this logic as it is single stream.

test/test_ops.py
```diff
@@ -1490,6 +1510,103 @@ def test_add_audio_stream_invalid_bit_rate_errors(self, tmp_path, method):
...
+        streaming_encoder_add_samples(encoder, samples)
+        streaming_encoder_close(encoder)
+
+        source = self._get_decoder_source(encoder_output)
+        decoded = AudioDecoder(source).get_all_samples()
+        assert decoded.data.shape[0] == num_channels
+        assert decoded.sample_rate == sample_rate
+        assert_tensor_close_on_at_least(
+            decoded.data, samples, percentage=100, atol=1e-4
+        )
```
NicolasHug: when percentage is 100, that's equivalent to `torch.testing.assert_close(atol=1e-4)`

test/test_ops.py
```diff
@@ -1490,6 +1510,103 @@ def test_add_audio_stream_invalid_bit_rate_errors(self, tmp_path, method):
...
+            crf=0,
+        )
+        streaming_encoder_add_audio_stream(
+            encoder,
+            sample_rate=sample_rate,
+            num_channels=source_samples.shape[0],
+        )
+        streaming_encoder_open(encoder)
+        streaming_encoder_add_frames(encoder, source_frames)
+        streaming_encoder_add_samples(encoder, source_samples)
```
NicolasHug, line 1577: add more frames here, to test that we can interleave the calls?

src/torchcodec/_core/Encoder.cpp
```diff
@@ -1456,7 +1462,223 @@ void MultiStreamEncoder::encodeVideoFrame(
...
+          audioStream_->inNumChannels,
+      "Expected ",
+      audioStream_->inNumChannels,
+      " channels, got ",
+      validatedSamples.sizes()[0]);
+  // TODO MultiStreamEncoder: Support multiple addSamples() calls.
+  // Codecs have frame size requirements, only the final encoded sample batch
+  // can be smaller than the frame size, so to support multiple calls without
+  // requiring the user to know about frame size, we need to encode frames
+  // through a FIFO.
```
NicolasHug, line 1482: Are we able to relax this check by allowing multiple calls to `addSamples()` as long as the number of samples is constant? I am hoping that we would only need a fifo *iff* the number of samples varies across calls to `addSamples()`.

NicolasHug (approved):
Thanks @Dan-Flores , LGTM. Can you reshare the diffing link please, I think the current one is incorrect.
There are some test failures on mkv which, I suspect, should be easy to address (LMK how it goes please!)

src/torchcodec/_core/Encoder.cpp
Dan-Flores: We would need to require that the number of samples added is divisible by the codec's frame size (ex. AAC uses 1024). 
Without this, the final batch encoded in `MultiStreamEncoder::encodeAudioSamples` will be smaller than the frame size, which can only happen for the final encoded samples. 

We could enforce the frame size match, but I thought it would be simpler to only support one `addSamples()` call for now. This, along with resampling and remixing are reasons we need to implement the FIFO soon!

PR #1360, opened 2026-05-08
Encode audio through FIFO in streaming encoder

Dan-Flores:
This PR enables multiple `addSamples()` calls with varying frame sizes by implementing a FIFO.

* `MultiStreamEncoder::initializeAudioStream()` is updated to *always* create the FIFO ([diffing](https://www.internalfb.com/intern/diffing/?paste_number=2313544584))
	* `outputNumChannels` and `outputSampleRate` are still supported yet.
	* The FIFO is necessary to allow `addSamples()` to be called multiple times with an arbitrary number of samples (also called `frame_size`).
* `MultiStreamEncoder::encodeAudioFrameThroughFifo()` writes frames to the FIFO then encodes samples in batches of length `frame_size` ([diffing](https://www.internalfb.com/intern/diffing/?paste_number=2313569866))
* `MultiStreamEncoder::maybeFlushSwrAndFifo()`: flushes the swr if frames remain, and always flushes the FIFO ([diffing](https://www.internalfb.com/intern/diffing/?paste_number=2313575790)).
	* Currently, output number of samples always matches the input number of samples, so the swr buffer will always be empty. 
	
### Testing
* `test_add_audio_stream_and_encode_samples` is updated to encode varying frame lengths.
* `test_add_audio_and_video_streams_and_encode` is updated to interleave `addSamples()` and `addFrames()` calls.

### DIFF
--- a/test/test_ops.py
+++ b/test/test_ops.py
@@ -1601,6 +1601,8 @@ class TestMultiStreamEncoderOps:
         assert decoded_audio.data.shape[0] == source_samples.shape[0]
         # Codecs for lossy audio formats (not WAV or FLAC) can add padding which causes
         # sample count to differ, so we only compare the smaller sample count.
+        # TODO MultiStreamEncoder: The previous AudioEncoder didn't need
+        # padding after introducing a FIFO. Investigate why this is needed.
         num_samples_to_compare = min(
             decoded_audio.data.shape[1], source_samples.shape[1]
         )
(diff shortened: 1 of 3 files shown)


test/test_ops.py
NicolasHug, line 1608: I'm just seeing this now. I'm a bit surprised, the previous AudioEncoder didn't need padding. But I do remember having different number of samples before I started implemented the FIFO-based encoder. Can you add a TODO_MULTI_STREAM_ENCODER entry here to double check and investigate what's going on?

NicolasHug (approved):
Thanks @Dan-Flores !

PR #1361, opened 2026-05-08
Fix mypy CI job

NicolasHug:
Mypy linter job is [failing on main](https://github.com/meta-pytorch/torchcodec/actions/runs/25547537992/job/74987310418) with

```
"list" has incompatible type "list[int] | list[float]"; expected "Iterable[int]"
 [arg-type]

            all_clips_indices += frame_indices

                                 ^~~~~~~~~~~~~
```

This PR fixes it.

### DIFF
--- a/src/torchcodec/samplers/_common.py
+++ b/src/torchcodec/samplers/_common.py
@@ -1,30 +1,29 @@
 from collections.abc import Callable
+from typing import TypeVar
 
 from torchcodec import FrameBatch
 
-_LIST_OF_INT_OR_FLOAT = list[int] | list[float]
+_INT_OR_FLOAT = TypeVar("_INT_OR_FLOAT", int, float)
 
 
 def _repeat_last_policy(
-    values: _LIST_OF_INT_OR_FLOAT, desired_len: int
-) -> _LIST_OF_INT_OR_FLOAT:
+    values: list[_INT_OR_FLOAT], desired_len: int
+) -> list[_INT_OR_FLOAT]:
     # values = [1, 2, 3], desired_len = 5
     # output = [1, 2, 3, 3, 3]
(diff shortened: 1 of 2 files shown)


src/torchcodec/samplers/_index_based.py
```diff
@@ -112,7 +112,7 @@ def _build_all_clips_indices(
             range(start_index, frame_index_upper_bound, num_indices_between_frames)
         )
         if len(frame_indices) < num_frames_per_clip:
-            frame_indices = policy_fun(frame_indices, num_frames_per_clip)  # type: ignore[assignment]
+            frame_indices = policy_fun(frame_indices, num_frames_per_clip)
         all_clips_indices += frame_indices
```
NicolasHug, line 116: Drive-by: this `type: ignore` can now be removed.

Note that mypy was complaining on the line just below, i.e. this one:

```py
all_clips_indices += frame_indices
```

src/torchcodec/samplers/_common.py
```diff
@@ -1,38 +1,37 @@
 from collections.abc import Callable
+from typing import TypeVar
 
 from torchcodec import FrameBatch
 
-_LIST_OF_INT_OR_FLOAT = list[int] | list[float]
+_INT_OR_FLOAT = TypeVar("_INT_OR_FLOAT", int, float)
```
NicolasHug, line 6: this is the actual fix where, instead of declaring the return type of the policy function as 'list[int] | list[float]', they're now declared as 'list[SomeNewType]' where 'SomeNewType' *binds* the output type to the input type.

Previously, mypy wasn't able to understand that the type of the input list was the same as the type of the output list (despite this being absolutely clear from the code - but mypy can't read code).

scotts (approved):

PR #1364, opened 2026-05-11
Fix: Cuda backend now supports odd dimensions

NicolasHug:
This PR allows the CUDA backend (previously known as 'beta') to support videos with odd dimensions.

A frame with odd dimensions gets decoded on CPU, always, because our CUDA backend falls back to CPU as NVDEC doesn't support them. But on `main`, we were failing on such frames at the color-conversion step. The color-conversion step is mainly `transferCpuFrameToGpuNV12() -> convertNV12FrameToRGB()`

Both these functions required changes, the key ones being that:

- `transferCpuFrameToGpuNV12()` now always returns frame with *even* dimensions, even if the original dimensions are odd
- `convertNV12FrameToRGB()`, which takes as input this *always even* frame, knows to crop back the dimensions to the correct odd size via the addition of a new `outputDims` parameter.


src/torchcodec/_core/CudaDeviceInterface.cpp
NicolasHug, line 1: Note that this is the "ffmpeg" interface, which already supported odd sizes, because *everything* here happens on the CPU, including the color-conversion step. That's why we can validate below that the 'cuda' path of this interface actually won't ever receive odd sizes.

test/test_decoders.py
```diff
@@ -1535,6 +1538,30 @@ def test_bt601_colorspace(self, asset):
...
+    )
+    @pytest.mark.parametrize("device", ("cuda", "cuda:ffmpeg"))
+    def test_odd_sized_videos(self, asset, device):
+        decoder_gpu, _ = make_video_decoder(asset.path, device=device)
+        decoder_cpu = VideoDecoder(asset.path, device="cpu")
+
+        gpu_frame = decoder_gpu.get_frame_at(0).data.cpu()
+        cpu_frame = decoder_cpu.get_frame_at(0).data
+        assert gpu_frame.shape == cpu_frame.shape
+        assert_tensor_close_on_at_least(gpu_frame, cpu_frame, percentage=89, atol=3)
```
NicolasHug, line 1558: These are higher tolerances than what we usually have when we compare against the CPU ref. It's not because of some wrong cropping or bad color-conversion step. I think it's inherent to the input videos which are in 444 format. I validated that a video with even dimensions still required the same high tolerances.

test/utils.py
```diff
@@ -740,6 +740,33 @@ def get_empty_chw_tensor(self, *, stream_index: int) -> torch.Tensor:
     frames={},  # Automatically loaded from json file
 )
 
+TESTSRC2_ODD_WIDTH = TestVideo(
```
NicolasHug, line 745: TODO for myself before merging: add a comment showing how the vids were generated.

scotts:
I'm a little unsure which of these two possibilities this PR implements:

1. Fixes the existing color conversion path to handle odd dimensions and decoding still happens on the CPU. Or;
2. Fixes the existing color conversion path to promote odd dimensions to even dimensions so that decoding now happens on NVDEC.

src/torchcodec/_core/CUDACommon.cpp
```diff
@@ -402,8 +435,23 @@ torch::stable::Tensor convertNV12FrameToRGB(
...
+  // TODO we should log a warning if status > 0.
+  STD_TORCH_CHECK(status >= 0, "Failed to convert NV12 frame.");
 
+  if (needsCrop) {
+    if (outHeight != nv12Height) {
+      dst = torch::stable::narrow(dst, /*dim=*/0, /*start=*/0, outHeight);
+    }
+    if (outWidth != nv12Width) {
+      dst = torch::stable::narrow(dst, /*dim=*/1, /*start=*/0, outWidth);
+      dst = torch::stable::contiguous(dst);
```
scotts, line 447: So we're just chopping off a row or column of pixels to get the output frame back to what the user expects?

scotts (approved):
I've got some questions about how it all works, but looks good!

NicolasHug:
It's this one:

> 1. Fixes the existing color conversion path to handle odd dimensions and decoding still happens on the CPU.


The decoding happens before color-conversion, and always on the CPU. We have no control over that, NVDEC refuses to decode non-even frames so we automatically fallback to the CPU.

Then on `main`, after the frame was decoded on the CPU, the color-conversion would fail at one of these checks:

- [here](https://github.com/meta-pytorch/torchcodec/blob/b8cc33c45a96fe3e6605113dd862b746f1306362/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L789-L792) if both H and W are odd
- [here](https://github.com/meta-pytorch/torchcodec/blob/b8cc33c45a96fe3e6605113dd862b746f1306362/src/torchcodec/_core/BetaCudaDeviceInterface.cpp#L831-L833) if just H is odd
- [here](https://github.com/meta-pytorch/torchcodec/blob/b8cc33c45a96fe3e6605113dd862b746f1306362/src/torchcodec/_core/CUDACommon.cpp#L405) if W is odd

This PR removes those `STD_TORCH_CHECK` calls by enforcing that the (temporary) frames are always of even dimensions.

Issue #1323, opened 2026-04-07
TorchCodec 0.11.0 PYPI binary release broke external accelerators plugins support


dvrogozh:
TorchCodec plugins were supposed to work by adding support of non-CUDA hardware accelerators for the PYPI distributed pre-built TorchCodec binaries. This worked till release `0.10`. For example, for XPU:
```
$ pip install torch~=2.10.0 torchcodec~=0.10.0 --extra-index-url https://download.pytorch.org/whl/xpu
$ python3 -c "import torchcodec; print(torchcodec.__version__)"
0.10.0
$ python3 -c "import torchcodec_xpu; print(torchcodec_xpu.__version__)"
```
The last command was loading XPU plugin available at https://github.com/intel/torchlib-xpu (we do not yet release pre-build binaries). To make XPU loading work the `import torchcodec` must always work - that's a prerequisite.

However, starting from `0.11.0` TorchCodec libraries started to directly depend on CUDA and CUDA version of Pytorch. So, doing similar to above now fails:
```
$ pip install torch~=2.11.0 torchcodec~=0.11.0 --extra-index-url https://download.pytorch.org/whl/xpu
$ python3 -c "import torchcodec; print(torchcodec.__version__)"
...
OSError: libnvrtc.so.13: cannot open shared object file: No such file or directory
```

Below I added `ldd` outputs for comparison. I guess the changes happened on enabling Beta CUDA device interface in a built?

**What is the direction to enable non-CUDA plugins in the TorchCodec now?** I see few possible options:

1. Fix TorchCodec to isolate loading of CUDA specific dependency making "core" parts of TorchCodec to always load regardless of the installed version of Torch. If TorchCodec will be kept on PYPI, then this option makes sense as what is on PYPI should better be agnostic to the hardware in its core loading logic.
 
1. Build and distribute separate version of TorchCodec for each Torch HW backend. Similar to how we build and distribute torch now (https://download.pytorch.org/whl/xpu, https://download.pytorch.org/whl/cuda, etc.). Maybe we can reuse existing XPU Torch location for that. If we follow this path, we need further consider:
    * Should CUDA enabled version of TorchCodec be continued hosted on PYPI? I believe it should not and better if it will get moved under https://download.pytorch.org/whl/cuda

For TorchCodec `0.10`:
```
 $ ldd lib/python3.13/site-packages/torchcodec/libtorchcodec_core8.so
        linux-vdso.so.1 (0x00007138bc95d000)
[...]

PR #1371, opened 2026-05-12
Lazily load NPP libraries

NicolasHug:
### Context

Our current CUDA wheels directly depend on NPP, which can be seen with `ldd`:

```
~/dev/torchcodec-cuda (main*) » ldd src/torchcodec/libtorchcodec_core8.so | grep -i npp                                                 nicolashug@nicolashug-fedora-PW0H326Y
	libnppicc.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppicc.so.12 (0x00007f6131800000)
	libnppc.so.12 => /usr/local/cuda/targets/x86_64-linux/lib/libnppc.so.12 (0x00007f6130a00000)
```

This means that the NPP `.so` files must be found on the user's machine as soon as they do `import torchcodec`. If NPP isn't found, TorchCodec errors loudly. That's an issue (see the *why* section below).

### This PR

This PR moves this NPP library resolution from `import` time to decode time. That is, a user should now be able to `import torchcodec` and use the CPU-only decoding functionalities even if NPP isn't available on their system. They will see an error only if they try to run the CUDA decoding functionalities.

In terms of implementation, we're doing the exact same thing that we'd done for NVCUVID in https://github.com/meta-pytorch/torchcodec/pull/996.

### Why we're doing this

We want to be in a state where CUDA wheels are uploaded to PyPI, such that users running `pip install torch torchcodec`:

- have the CUDA-enabled torch and torchcodec versions
- can still use the CPU-only functionalities of torch and torchcodec even if they don't have CUDA libraries or even a GPU.

Our current CUDA wheels would fail if the user doesn't have the CUDA libraries on their machines, as explained above. This PR should solve that.

It's worth noting that with this PR, the TorchCodec CUDA wheel should work in CPU-only environments, but *only* alongside the `torch` CUDA wheel. It won't work with the torch CPU wheel: that's because there are still some dependencies on symbols from cudart and libvcrt that are resolved by `torch` itself.

Making the TorchCodec CUDA wheel work alongside the torch CPU wheel on CPU-only machines is a non-goal.


.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -132,6 +132,64 @@ jobs:
         run: |
           ${CONDA_RUN} time python benchmarks/decoders/gpu_benchmark.py --devices=cuda:0,cpu --resize_devices=none
 
+  install-and-test-on-cpu-only-machine:
+    # This job tests that CUDA wheels work fine on CPU-only machines.  Note that
+    # we still install a CUDA-enabled version of torch, and that's by design.
+    # Essentially, what we want to make sure is that
+    # `pip install torch torchcodec` works on CPU-only machines, and this
+    # command should install CUDA-enabled versions of both torch and torchcodec.
```
NicolasHug, line 140: https://github.com/meta-pytorch/torchcodec/pull/1324 shows that this new CI job is [failing on `main`](https://github.com/meta-pytorch/torchcodec/actions/runs/25735698197/job/75601957187?pr=1324), as expected, with:

```
E   OSError: libnppicc.so.12: cannot open shared object file: No such file or directory
```

So it is testing what it's supposed to test.

src/torchcodec/_core/NPPRuntimeLoader.cpp
```diff
@@ -0,0 +1,180 @@
+// Copyright (c) Meta Platforms, Inc. and affiliates.
+// All rights reserved.
+//
+// This source code is licensed under the BSD-style license found in the
+// LICENSE file in the root directory of this source tree.
```
NicolasHug, line 5: I'll consider merging this file with the NVCUVID loader https://github.com/meta-pytorch/torchcodec/blob/main/src/torchcodec/_core/NVCUVIDRuntimeLoader.cpp, as a follow-up PR.

dvrogozh (commented):
> It's worth noting that with this PR, the TorchCodec CUDA wheel should work in CPU-only environments, but only alongside the torch CUDA wheel.

If the intent is to upload the cuda wheels to PyPI after this PR landed, then we will have the #1323 issue again as to enable XPU plugin I need to both install non-CUDA torch and torchcodec. This however can be resolved if we will first address https://github.com/meta-pytorch/torchcodec/issues/1376, i.e. upload CPU-only torchcodec wheels to XPU index https://download.pytorch.org/whl/xpu.

For details see https://github.com/meta-pytorch/torchcodec/pull/1325#issuecomment-4434505012.

.github/workflows/linux_cuda_wheel.yaml
```diff
@@ -132,6 +132,64 @@ jobs:
...
+    runs-on: ubuntu-latest
```
scotts, line 142: Nit: maybe also comment that we need to ensure `runs-on` is a CPU-only machine? And maybe also put in a quick check that CUDA is not available?

scotts (approved):

atalman (approved):
lgtm

PR #1439, opened 2026-05-21
Add SPDX license expression to package metadata

maxi-seiringer:
This adds an explicit SPDX license expression to the package metadata.

The project already includes the LICENSE file via license-files, but without project.license some license-reporting tools may show less useful license metadata. BSD-3-Clause matches the repository license.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,6 +3,7 @@ name = "torchcodec"
 description = "A video decoder for PyTorch"
 readme = "README.md"
 requires-python = ">=3.10"
+license = "BSD-3-Clause"
 license-files = ["LICENSE"]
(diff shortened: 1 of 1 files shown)


NicolasHug:
Thanks for the PR @maxi-seiringer 

> without project.license some license-reporting tools may show less useful license metadata

Can you share which ones exactly?

maxi-seiringer:
> Thanks for the PR @maxi-seiringer
> 
> > without project.license some license-reporting tools may show less useful license metadata
> 
> Can you share which ones exactly?

Yes, I use pip-licenses. For your project the license name is the license text which messes up formatting when displaying all used licenses.

Issue #1438, opened 2026-05-20
Invalid delete in VideoDecoder backend


HowardLovekraft:
### 🐛 Describe the bug

Hello everyone. 
Recently my team has been trying to deploy a YOLO26 model on the virtual machine in `nvcr.io/nvidia/pytorch:26.04-py3`-based container, but stumbled upon a cryptic bug while using `torchcodec` for CUDA-accelerated video decoding. 
Upon further investigation, we managed to narrow this down to the situation when a `torchcodec.decoders.VideoDecoder` is instantiated inside a function which is released from the stack later and the breakage occurs exactly at that moment (presumably garbage collection).

Note that this issue pops up only when `device='cuda'`, i.e. everything works perfectly fine on CPU-based build.

Minimal reproducible environment for the bug:
```python
# test_dec.py
import torch
from torchcodec.decoders import VideoDecoder
from pathlib import Path

def test_lifecycle():
    video_path = Path("video.mkv")
    print("Creating Decoder...")
    decoder = VideoDecoder(str(video_path), device='cuda')
    print("Decoder Created.")

    print("Deleting Decoder...")
    del decoder
    print("Decoder Deleted.")

    torch.cuda.empty_cache()
    print("Done.")

if __name__ == "__main__":
    test_lifecycle()
```

Traceback: (that's all what i see in logs)
```text
Creating Decoder...
Decoder Created.
Deleting Decoder...
Decoder Deleted.
Done.
free(): invalid pointer
```

- - -

I tried to profile this program using `valgrind` by putting it inside our modified Docker Compose environment below:
```yaml
services:
  api:
    build: .
    restart: unless-stopped
    ipc: host
    deploy:
      resources:
        reservations:
          devices:
            - driver: nvidia
              count: all
              capabilities: [gpu, video]
```
```Dockerfile
FROM nvcr.io/nvidia/pytorch:26.04-py3

WORKDIR /app

RUN apt-get update && \
    apt-get install -y valgrind ffmpeg && \
    rm -rf /var/lib/apt/lists/*

RUN pip install torchcodec --index-url=https://download.pytorch.org/whl/cu132 --break-system-packages

COPY . .

ENV PYTHONMALLOC=malloc
CMD ["valgrind", "--tool=memcheck", "--track-origins=yes", "--trace-children=yes", "--show-mismatched-frees=yes", "python", "test_dec.py"]
```

[...]

PR #1441, opened 2026-05-21
Leak NVDECCache to avoid calling into CUDA during static destruction

NicolasHug:
Trying to address https://github.com/meta-pytorch/torchcodec/issues/1438

I can't reproduce https://github.com/meta-pytorch/torchcodec/issues/1438 directly, but hopefully this is a fix. `claude`'s analysis is that the `static` cache entries end up calling their `cuvidDestroyDecoder()` destructor at a point during process teardown where the  CUDA libraries have potentially already been torn down.

> The core issue is that shared library unload order at process exit is undefined. When the process is shutting down,
  the dynamic linker (ld.so) unloads .so files, and C++ runs static destructors within each library as it's unloaded.
  But there's no dependency graph — the linker doesn't know that our destructor needs to call into libnvcuvid.so, so it
   might unload libnvcuvid.so first, then unload our library and run our destructors, which now call into unmapped
  memory.

I don't understand process teardown enough to verify the quote - but wild stuff happen all the time, that I can believe.

The workaround is to heap-allocate the cache so its destructor never runs at exit, leaking all existing decoders in the cache, which the OS should reclaim anyway. `pytorch` itself has a habit of leaking lots of stuff, so this is probably OK in terms of practice.

### DIFF
--- a/src/torchcodec/_core/NVDECCache.cpp
+++ b/src/torchcodec/_core/NVDECCache.cpp
@@ -21,7 +21,9 @@ extern "C" {
 namespace facebook::torchcodec {
 
 NVDECCache* NVDECCache::getCacheInstances() {
-  static NVDECCache cacheInstances[MAX_CUDA_GPUS];
+  // Intentionally leaked to avoid calling into CUDA/NVCUVID during static
+  // destruction, when the CUDA runtime may already be torn down.
+  static NVDECCache* cacheInstances = new NVDECCache[MAX_CUDA_GPUS];
   return cacheInstances;
 }
 


scotts (approved):
Yeah, this looks legit. The profiler has to do the same thing in a few places:
https://github.com/pytorch/kineto/blob/5902263f641a8634af077031fb4befc7162cefbb/libkineto/src/Logger.h#L142-L145, https://github.com/pytorch/kineto/blob/5902263f641a8634af077031fb4befc7162cefbb/libkineto/src/CuptiActivityApi.cpp#L58-L61 and https://github.com/pytorch/kineto/blob/5902263f641a8634af077031fb4befc7162cefbb/libkineto/src/CuptiCallbackApi.cpp#L155-L158

Issue #1158, opened 2026-01-08
Loading Audio and Video


will-rice:
If I missed this functionality, please direct me to the correct documentation. I didn't see where you could load audio and video with the same decoder. If you cannot, then it's not really a replacement for the deprecated `torchvision.io.read_video` or `torchaudio.io.StreamReader`.

Dan-Flores:
Hi @will-rice,
Thanks for opening an issue. We have not yet implemented a decoder that handles audio and video simultaneously. 

Currently, a workaround would be to use `AudioDecoder` and `VideoDecoder` to get the audio samples and video frames separately: 
```py
video_decoder = VideoDecoder("video.mp4")
audio_decoder = AudioDecoder("video.mp4")

frames = video_decoder[:]
samples = audio_decoder.get_all_samples()
```

Would this work for your use case? If not, could you share a bit more about your use case so we can better understand what is needed?

will-rice:
@Dan-Flores Thanks for the reply! That's similar to what I ended up doing. It's already faster than my current pipeline so I think it is fine the way it is.

PR #1442, opened 2026-05-21
Add audio property to VideoDecoder

Samoed:
Previously, accessing audio from a video file required creating a separate `AudioDecoder` alongside `VideoDecoder` https://github.com/meta-pytorch/torchcodec/issues/1158. This split made it difficult to integrate with external libraries such as [huggingface datasets](https://github.com/huggingface/datasets/issues/8007) utilities that expect audio to be accessible directly from the video decoder - the absence of an audio attribute on VideoDecoder either required workarounds or left audio support missing entirely in those integrations.

This PR adds a lazy audio property to VideoDecoder that returns an AudioDecoder for audio stream in the same source, or None if the source contains no audio stream

### DIFF
--- a/src/torchcodec/decoders/_video_decoder.py
+++ b/src/torchcodec/decoders/_video_decoder.py
@@ -3,7 +3,7 @@
 #
 # This source code is licensed under the BSD-style license found in the
 # LICENSE file in the root directory of this source tree.
-
+from __future__ import annotations
 
 import io
 import json
@@ -11,7 +11,7 @@ import numbers
 from collections.abc import Sequence
 from dataclasses import dataclass, field
 from pathlib import Path
-from typing import Literal
+from typing import Literal, TYPE_CHECKING
 
 import torch
 from torch import device as torch_device, nn, Tensor
@@ -20,6 +20,9 @@ from torchcodec._core._decoder_utils import create_video_decoder
 from torchcodec._logging import _LG
 from torchcodec.decoders._decoder_utils import _get_cuda_backend
 from torchcodec.transforms import DecoderTransform
+
+if TYPE_CHECKING:
+    from torchcodec.decoders._audio_decoder import AudioDecoder
 
 
 @dataclass
@@ -146,6 +149,13 @@ class VideoDecoder:
             Alternative field names "pkt_pts" and "pkt_duration" are also supported.
             Read more about this parameter in:
             :ref:`sphx_glr_generated_examples_decoding_custom_frame_mappings.py`
+        audio_stream_index (int, optional): Stream index of the audio stream to
+            use when :attr:`audio` is accessed. If unspecified, the
+            :term:`best stream` is used. Ignored for ``torch.Tensor`` sources.
+        audio_sample_rate (int, optional): Desired output sample rate for the
+            :attr:`audio` decoder. Defaults to the stream's native sample rate.
+        audio_num_channels (int, optional): Desired number of channels for the
+            :attr:`audio` decoder. Defaults to the stream's native channel count.
 
     Attributes:
         metadata (VideoStreamMetadata): Metadata of the video stream.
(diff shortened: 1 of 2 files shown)


NicolasHug:
Hi @Samoed , can you share more about this:

> This split made it difficult to integrate with external libraries such as https://github.com/huggingface/datasets/issues/8007 utilities that expect audio to be accessible directly from the video decoder - the absence of an audio attribute on VideoDecoder either required workarounds or left audio support missing entirely in those integrations

Specifically I'd like to understand why having two seprate VideoDecoder and AudioDecoder objects lead to friction, and how does the proposed design solve it

Samoed:
The Datasets library outputs video inputs as `VideoDecoder`, and to be able to easily extract audio information it typically splits inputs into video and audio columns before uploading. Extracting audio input from such videos requires some hacks. Also, I think it would be good to be able to work with one object during data processing to handle both audio and video

Samoed:
Hi @NicolasHug! I thought about this feature more and found more use cases. Currently Transformers have a processor kwarg `load_audio_from_video` https://github.com/huggingface/transformers/blob/63d64d0be41940cb6383d109194a7dc323261252/src/transformers/processing_utils.py#L544 and `use_audio_in_video` https://github.com/huggingface/transformers/blob/63d64d0be41940cb6383d109194a7dc323261252/src/transformers/models/qwen2_5_omni/processing_qwen2_5_omni.py#L74 which extractiang audio only from video paths, but not `VideoDecoder` objects (overall, transformers don't support `VideoDecoder` inputs, but I think this could change in the future). If you add this feature it would help to standardize this and provide more opportunities to use `VideoDecoder`

NicolasHug:
Sorry I still don't get it - can you share a code example of what this would enable, and why it can't already be achieved by just keeping a ref to the source video (and creating a separate `AudioDecoder` from it)?

Samoed:
HuggingFace `datasets` decodes a `video` column directly into a `VideoDecoder`:

```python
from datasets import load_dataset

test_ds = load_dataset("Samoed/testds")
video = test_ds["train"][0]["video"]  # VideoDecoder without path

# Today: no clean way to get audio from `video`. Users either
#   - reach into `video._source` with Video(decode=False)
#   - split the dataset into separate `video` + `audio` columns upstream.

audio = video.audio
```

For huggingface transformers

```python
def __call__(self, videos: list[VideoDecoder | ...], ...):
    for v in videos:
        frames = v.get_frames_in_range(...)
        if self.use_audio_in_video and decoder.audio is not None:
            audio = v.audio.get_all_samples()
```

src/torchcodec/decoders/_video_decoder.py
```diff
@@ -171,8 +181,22 @@ def __init__(
...
+
+        # Normalize file-like objects to bytes so the source can be stored and
+        # later passed to AudioDecoder independently (file-likes have a single
+        # shared cursor that two decoders would stomp on each other).
+        if not isinstance(source, (str, Path, bytes, Tensor)) and hasattr(
+            source, "read"
+        ):
+            if hasattr(source, "seek"):
+                source.seek(0)
+            source = source.read()
```
NicolasHug, line 198: The above is one reason why we can't support the proposal as-is. We don't want to convert a file-like to bytes, because it forces materialization or full download of what is supposed to be a streaming object - it defeats the entire purpose of file-like support.

I thought of perhaps exposing the `.source` attribute publicly instead of exposing an `AudioDecoder`, but that would still be subject to the same issue with file-like objects. Let me ask the `datasets` maintainer to see if there's anything that can be done on that side

NicolasHug:
@lhoestq curious if you have any suggestion on `datasets` side to address @Samoed 's workflow from https://github.com/meta-pytorch/torchcodec/pull/1442#issuecomment-4734137992 ?

tomaarsen:
This also affects Sentence Transformers currently, e.g.:

```python
from datasets import load_dataset
from sentence_transformers import SentenceTransformer

test_ds = load_dataset("Samoed/testds")
video = test_ds["train"][0]["video"]

model = SentenceTransformer("...")  # a video-capable model
embedding = model.encode(video)
```

Internally, ST converts this `VideoDecoder` into 1) frame-data plus 2) video_metadata like fps, duration, etc., which is then passed to `transformers`. I can't easily also take the audio from this `video` variable, I believe. 
On the other hand, if I use `load_audio_from_video` (https://github.com/huggingface/transformers/blob/63d64d0be41940cb6383d109194a7dc323261252/src/transformers/processing_utils.py#L544), *and* I just pass a URL or path, then I can have `transformers` grab the audio as well (where `transformers` uses `AudioDecoder` for it, this could also perhaps be simplified if it could be grabbed directly from the `VideoDecoder`).

- Tom Aarsen

PR #1443, opened 2026-05-21
Activate strict mypy

Samoed:
In my project I use strict mypy and got an error `error: Module "torchcodec.decoders" does not explicitly export attribute "VideoDecoder"  [attr-defined]`. I activated strict mypy here to increase typing coverage


NicolasHug:
Hi @Samoed , can you share exactly what would address the "error: Module "torchcodec.decoders" does not explicitly export attribute "VideoDecoder"  [attr-defined]" error? This PR seems to do more than just that.

Generally though, I just want to set expectations, I don't think we'll want to go stricter on type annotations than what we're already doing. mypy is limited in what it can understand, and pleasing it often makes the code worse, not better, so we won't be following it blindly

Samoed:
> Can you share exactly what would address the "error: Module "torchcodec.decoders" does not explicitly export attribute "VideoDecoder" [attr-defined]" error? This PR seems to do more than just that.

Strict mypy requires explicitly listing imports in `__all__` or they should be exported inside the module's `__init__` as `from foo import bar as bar`.
> Generally though, I just want to set expectations, I don't think we'll want to go stricter on type annotations than what we're already doing. mypy is limited in what it can understand, and pleasing it often makes the code worse, not better, so we won't be following it blindly.

I agree, but overall this PR didn't add too much new code for handling strict mypy. Only some casting to tensors, but also adding typing in some untyped functions and exports in `__init__`, which helps different code analyzers. I can update the mypy config to disable strict mode, but keep some of the helpful [strict flags](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-strict), like [--disallow-untyped-defs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-untyped-defs), [--disallow-incomplete-defs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-disallow-incomplete-defs), and [--check-untyped-defs](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-check-untyped-defs), that would help keep the quality of typing.

NicolasHug:
> Strict mypy requires explicitly listing imports in __all__ or they should be exported inside the module's __init__ as from foo import bar as bar


All public APIs are already included in their corresponding module's `__init__.py`, what am I missing?

Samoed:
> what am I missing?

They should be in `__all__` or imported as `from foo import bar as bar` https://typing.python.org/en/latest/spec/distributing.html#library-interface-public-and-private-symbols

NicolasHug:
OK - then sorry, but no, we'll stick to the usual Python practice where `__all__` is meant for `import *` and `from X import foo` is a perfectly good enough import idiom.

Samoed:
I see, thanks for review

PR #1453, opened 2026-05-28
Move AVStream and AVFormatContext from initialize() to initializeVideo()

NicolasHug:
We used to pass 

```
    const AVStream* avStream,
    const UniqueDecodingAVFormatContext& avFormatCtx,
```

in the `DeviceInterface.initialize()` method, but those fields are mainly only needed in `initializeVideo()`. This helps resolving a TODO_HDR item where we had to store these as fields in the interface.initializeVideo()


test/third-party-interface/ThirdPartyInterfaceTest.cpp
```diff
@@ -16,10 +16,7 @@ class DummyDeviceInterface : public DeviceInterface {
 
   virtual ~DummyDeviceInterface() {}
 
-  void initialize(
-      const AVStream* avStream,
-      const UniqueDecodingAVFormatContext& avFormatCtx,
-      const SharedAVCodecContext& codecContext) override {
+  void initialize(const SharedAVCodecContext& codecContext) override {
```
NicolasHug, line 19: CC, @dvrogozh I suspect this might be BC-breaking for you?

dvrogozh: yes, it will be BC-breaking. As of today I am handling that by strictly tight XPU plugin version to specific torchcodec version. Unfortunately this requires manual work for each torchcodec release.

NicolasHug: Thanks for the flexibility. I don't think we're quite there yet since we're still figuring out the best interface, but at some point we might want to start thinking about providing BC guarantees for the device interface API, so that you get some degree of stability.

PR #1461, opened 2026-05-29
Add File-like and bytes support to WavDecoder

NicolasHug:
This PR adds file-like and [tensor of] bytes support to WavDecoder. To not reinvent the wheel I asked claude to re-use our AVIO logic which currently only exists for FFmpeg.

For that, we need to introduce normal read, seek and write methods on the `AvioContextHolder` objects, which previously only had those as static methods. 

The static `read` method is now `readCallback` and dispatches directly to the actual `read()` method, whose code is largely unchanged except for removal of ffmpeg-only parts. The ffmpeg-specific parts are now in `readCallback`. Previously, the `opaque` pointer of the static methods was the class-specific state (e.g. the underlying file-like object or the tensor). Now, it's always `this`, which has access to the file-like object and to the tensor as normal class members.

There will be some more  cleanups to the AVIO logic in follow-ups.


test/test_decoders.py
```diff
@@ -2983,10 +2965,12 @@ def test_non_wav_file_raises_error(self):
         "start_seconds,stop_seconds",
         [
             (0.0, 1.0),
+            (0.2, 0.6),
```
NicolasHug, line 2968: made some minor additions and refacs in the tests, now we have centralized most of the Wav testing into `test_against_audio_decoder`

src/torchcodec/_core/AVIOFileLikeContext.cpp
```diff
@@ -66,33 +92,27 @@ int AVIOFileLikeContext::read(void* opaque, uint8_t* buf, int buf_size) {
...
-  auto fileLike = static_cast<UniquePyObject*>(opaque);
-  py::gil_scoped_acquire gil;
-  py::bytes bytes_obj(reinterpret_cast<const char*>(buf), buf_size);
-
-  return py::cast<int>((*fileLike)->attr("write")(bytes_obj));
+int64_t AVIOFileLikeContext::getSize() {
+  int64_t currentPos = seek(0, SEEK_CUR);
+  int64_t size = seek(0, SEEK_END);
+  seek(currentPos, SEEK_SET);
+  return size;
```
NicolasHug: this defeats the streaming feature doesn't it? It will download the entire thing just to figure out the size.

PR #1467, opened 2026-06-01
Pass color-conversion matrix as kernel parameter

### LINKED ISSUE #1466: RGB→NV12 encode kernel and YUV→RGB decode kernel share __constant__ d_colorMatrix, breaking mixed decode+encode pipelines
### 🐛 Describe the bug

Follow-up to #1446 — the NPP-load-order issue I reported there.

Thanks @NicolasHug for going through with the full NPP removal in #1465. The compile-time fix in #1447 already solved the immediate problem on my side, so I didn't expect the bigger refactor to happen this quickly. That work does close the whole class of NPP DLL-resolution problems cleanly.

While testing 04f13005 on the same NVDEC + NVENC pipeline that originally surfaced #1446, I noticed what looks like a small leftover issue in the new RGB→YUV path. Filing as a separate issue since it seems unrelated to the NPP removal itself.

### Summary

In `src/torchcodec/_core/color_conversion.cu`, the new encode-side `rgbToNV12Kernel` and the existing decode-side `yuvToRgbKernel` / `writePairOfRGBPixels` share a single `__constant__ float d_colorMatrix[3][4]`. The two kernels need different matrices (RGB→YUV vs YUV→RGB), and the upload policies on the two paths are asymmetric, so a pipeline that runs both kernels in the same process appears to produce wrong decoded colors from the second frame onward.

### Reproducer

A pipeline that does NVDEC decode → GPU work → NVENC encode of the same stream, all in one process:

```python
import torchcodec
from torchcodec import _core

with torchcodec.decoders.set_cuda_backend("nvdec"):
    dec = torchcodec.decoders.VideoDecoder(input_path, device="cuda")

enc = _core.create_streaming_encoder_to_file(output_path)
_core.streaming_encoder_add_video_stream(
    enc, height=H, width=W, frame_rate=fps,
    device="cuda", codec="h264_nvenc", pixel_format=None,
    crf=23.0, preset="p4", extra_options=[])
_core.streaming_encoder_open_file(enc, output_path)

for frame in dec:                    # CHW uint8 on CUDA
    _core.streaming_encoder_add_frames(enc, frame.data.contiguous(), 0)

_core.streaming_encoder_close(enc)
```

Frame 1 of the output looks correct; every subsequent frame is heavily green/teal tinted. Switching to CPU decode (or any path that doesn't hit `yuvToRgbKernel`) makes the symptom disappear, which seems to point at the decode kernel — not at NVENC — as the source.

### What I think is going on

1. `launchRGBToNV12Kernel` unconditionally `cudaMemcpyToSymbol`s its RGB→YUV matrix into `d_colorMatrix` on every call.
2. `launchNV12ToRGBKernel` / `launchP016ToRGB16Kernel` go through `maybeUploadColorMatrix(matrix, colorMatrixChanged)`. `colorMatrixChanged` comes from `maybeUpdateColorMatrix`, which only tracks `(colorspace, color_range, bitDepth, outScale)` on the host, so it has no way to know the encode kernel just overwrote the GPU constant.

Per-frame sequence as I read it:

| Step | What happens | GPU `d_colorMatrix` | Decoded RGB |
|---|---|---|---|
| Frame 1, decode | `colorMatrixChanged=true` (cold cache) → upload YUV→RGB | YUV→RGB | correct |
| Frame 1, encode | upload RGB→YUV (overwrites) | RGB→YUV | — |
| Frame 2, decode | `colorMatrixChanged=false` → no upload | RGB→YUV (leftover) | wrong |
| Frame 2, encode | upload RGB→YUV again | RGB→YUV | — |
| … | … | … | … |

With BT.601 limited range, the wrong matrix's Y row carries no offset but its U/V rows carry the `+128` centering offset, so G picks up a constant +128 while R and B do not. That seems consistent with the heavy green/teal cast in the output.

### One possible fix

Giving the two kernels independent constant-memory symbols and routing each side's upload to its own symbol appears to work locally:

```cpp
__constant__ float d_colorMatrixDecode[3][4];
__constant__ float d_colorMatrixEncode[3][4];
```

- `writePairOfRGBPixels`, `yuvToRgbKernel`, and `maybeUploadColorMatrix` → `d_colorMatrixDecode`
- `rgbToNV12Kernel` and `launchRGBToNV12Kernel`'s `cudaMemcpyToSymbol` → `d_colorMatrixEncode`

This keeps the decode-side upload cache as-is and the encode-side upload is still cheap (48 bytes). The change is contained to `color_conversion.cu`. With this applied against 04f13005 I see:

- the green tint goes away in the mixed pipeline;
- CPU-decode + GPU-encode is unchanged;
- GPU-decode + CPU-encode is unchanged.

There may well be a better approach — sharing some bookkeeping between encode/decode caches, etc. Flagging the symptom and what I found in case it helps.

### Versions

Collecting environment information...
PyTorch version: 2.11.0+cu130
Is debug build: False
CUDA used to build PyTorch: 13.0
ROCM used to build PyTorch: N/A

OS: Microsoft Windows 11 Enterprise (10.0.26200 64 ビット)
GCC version: Could not collect
Clang version: Could not collect
CMake version: version 4.3.2
Libc version: N/A

Python version: 3.13.9 (tags/v3.13.9:8183fa5, Oct 14 2025, 14:09:13) [MSC v.1944 64 bit (AMD64)] (64-bit runtime)
Python platform: Windows-11-10.0.26200-SP0
Is CUDA available: True
CUDA runtime version: 13.2.78
CUDA_MODULE_LOADING set to:
GPU models and configuration: GPU 0: NVIDIA GeForce RTX 4060
Nvidia driver version: Could not collect (Actual 610.47)
cuDNN version: Could not collect
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True
Caching allocator config: N/A

CPU:
Name: AMD Ryzen 7 5700X 8-Core Processor
Manufacturer: AuthenticAMD
Family: 107
Architecture: 9
ProcessorType: 3
DeviceID: CPU0
CurrentClockSpeed: 3401
MaxClockSpeed: 3401
L2CacheSize: 4096
L2CacheSpeed: None
Revision: 8450

Versions of relevant libraries:
[pip3] Could not collect
[conda] Could not collect

----- relevant packages (from uv pip list) -----
nvidia-npp             13.1.0.48
nvidia-nvjpeg          13.1.0.48
nvidia-vfx             0.1.0.1
tensorrt               10.15.1.29
tensorrt-cu13          10.15.1.29
tensorrt-cu13-bindings 10.15.1.29
tensorrt-cu13-libs     10.15.1.29
torch                  2.11.0+cu130
torch-tensorrt         2.11.0+cu130
torchcodec             0.13.0a0     H:\test\torchcodec
torchvision            0.26.0+cu130

----- torchcodec version -----
.venv\Scripts\python.exe -c "import torchcodec; print(torchcodec.__version__)"
0.13.0a0+04f1300

----- FFmpeg version -----
& "$FFmpegRoot\bin\ffmpeg.exe" -version | Select-Object -First 1
ffmpeg version 8.1-full_build-www.gyan.dev Copyright (c) 2000-2026 the FFmpeg developers

NicolasHug:
We previously defined the color-conversion matrix for CUDA kernels as 

```
__constant__ float d_colorMatrix[3][4];
```

which came with all sorts of issues, mainly lack thread-safety and potential conflicts between decoder -> encoder pipelines. Thanks a lot @sh202603 for the report!

This PR passes the color-conversion matrix as kernel parameter, bypassing the issue entirely. `ncu` benchmarks report that the new kernel is actually even faster, maybe because we don't have to call `cudaMemcpyToSymbol()` anymore?

Fixes https://github.com/meta-pytorch/torchcodec/issues/1466

### DIFF
--- a/src/torchcodec/_core/color_conversion.cpp
+++ b/src/torchcodec/_core/color_conversion.cpp
@@ -118,7 +118,7 @@ void computeColorConversionMatrix(
   }
 }
 
-bool maybeUpdateColorMatrix(
+void maybeUpdateColorMatrix(
     CachedColorMatrix& cachedColorMatrix,
     AVColorSpace colorspace,
     AVColorRange colorRange,
@@ -128,7 +128,7 @@ bool maybeUpdateColorMatrix(
       cachedColorMatrix.colorRange == colorRange &&
       cachedColorMatrix.bitDepth == bitDepth &&
       cachedColorMatrix.outScale == outScale) {
-    return false;
+    return;
   }
 
   computeColorConversionMatrix(
@@ -138,7 +138,6 @@ bool maybeUpdateColorMatrix(
   cachedColorMatrix.bitDepth = bitDepth;
   cachedColorMatrix.outScale = outScale;
   cachedColorMatrix.valid = true;
-  return true;
 }
 
 void computeRGBToYUVMatrix(
@@ -221,7 +220,7 @@ torch::stable::Tensor convertYUVFrameToRGB(
   syncStreams(
       /*runningStream=*/nvdecStream, /*waitingStream=*/stream);
 
-  bool colorMatrixChanged = maybeUpdateColorMatrix(
+  maybeUpdateColorMatrix(
       cachedColorMatrix,
       avFrame->colorspace,
       avFrame->color_range,
@@ -240,7 +239,6 @@ torch::stable::Tensor convertYUVFrameToRGB(
         validateInt64ToInt(dst.stride(0) * 2, "dst.stride(0)*2"),
         bitDepth,
         cachedColorMatrix.matrix,
-        colorMatrixChanged,
         stream);
   } else {
     launchNV12ToRGBKernel(
@@ -253,7 +251,6 @@ torch::stable::Tensor convertYUVFrameToRGB(
         avFrame->linesize[1],
         validateInt64ToInt(dst.stride(0), "dst.stride(0)"),
         cachedColorMatrix.matrix,
-        colorMatrixChanged,
         stream);
   }
 
--- a/src/torchcodec/_core/color_conversion.cu
+++ b/src/torchcodec/_core/color_conversion.cu
@@ -17,9 +17,12 @@ namespace facebook::torchcodec {
 // each. P016 uses 16-bit samples (uint16_t) with actual data in the most
 // significant bits (right-shifted by 16 - bitDepth).
 
-// Color conversion matrix stored in constant memory for fast access.
-// It is available to every single thread.
-__constant__ float d_colorMatrix[3][4];
+// Wraps the 3x4 color-conversion matrix so it can be passed as a kernel param,
+// to ensure thread-safety (previous implem had it __global__, which wasn't
+// safe)
+struct ColorMatrix {
+  float m[3][4];
+};
 
 // Takes a pair of consecutive Y values, a pair of UV values, and writes the
 // corresponding two RGB values. Instead of writing each ra ga ba and rb gb bb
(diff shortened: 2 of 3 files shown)


sh202603:
Hi @NicolasHug ,

Verified `7f5c1951` in the environment I reported in #1466, against the original mixed-pipeline reproducer:

- **NVDEC decode + NVENC encode in the same process**: no more green/teal cast, output color matches CPU-decode reference.
- **CPU decode + NVENC encode**: unchanged (was already correct).
- **NVDEC decode + CPU encode**: unchanged (was already correct).

Thanks for the quick turnaround!

PR #1468, opened 2026-06-01
Fix Windows CUDA builds, maybe

NicolasHug:


sh202603:
Hi @NicolasHug,

Thanks for the quick follow-up here. 

Verified locally on the same Windows source-build setup that originally needed the `/Zc:preprocessor` env-var workaround (torch 2.11.0+cu130 / CUDA 13.2 / MSVC 14.44 / Win11).

Before this PR I had to set **both** `CXXFLAGS=/Zc:preprocessor` and `CUDAFLAGS=-Xcompiler=/Zc:preprocessor` before `uv pip install -e ../torchcodec --no-build-isolation`, otherwise the CCCL headers pulled in by `color_conversion.cu` failed with `error C1189`. With `54c4091e` I built cleanly with **both env vars unset** — the `COMPILE_LANGUAGE:CUDA`-gated option is sufficient on its own, and the `.cpp` side does not appear to need the flag at this revision either.

So the "maybe" can be promoted to "yes" at least for this configuration. For other Windows source-build users this is a strict simplification — no need to set the two `/Zc:preprocessor` env vars anymore!

PR #1470, opened 2026-06-01
Reduce size of built docs

NicolasHug:
We started getting a few warnings upon upload of our docs because they started getting really, really big - about 200Mb which is unreasonable for just a bunch of html files.

The issue was that when we use `play_audio` or `play_video`, the input is checked-in into the built docs artifacts. Meaning that we were storing large raw wav files, as well as large video files. This PR skips / reduces the amount of stuff we have to keep around. The audio example went from 60Mb to ~3, and the video example went from 45 to ~3.

This should account for 100Mb * 2 (since I think they were also duplicated). Let's see.

### DIFF
--- a/examples/decoding/audio_decoding.py
+++ b/examples/decoding/audio_decoding.py
@@ -22,8 +22,10 @@ import requests
 from IPython.display import Audio
 
 
-def play_audio(samples):
-    return Audio(samples.data, rate=samples.sample_rate)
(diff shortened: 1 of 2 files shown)


NicolasHug:
Built docs artifact is just ~20MB now, still not insignificant but mangeable, and 10X smaller than before

PR #1475, opened 2026-06-02
Minor updates to wavdecoder

NicolasHug:
Use `__restrict__` to tell compiler it can optimize further, and add another STD_TORCH_CHECK just to be sure


NicolasHug:
CBA making this work for windows, whatever

PR #1481, opened 2026-06-03
Remove old TODOs

NicolasHug:


src/torchcodec/_core/SingleStreamDecoder.cpp
```diff
@@ -1329,8 +1329,6 @@ bool SingleStreamDecoder::canWeAvoidSeeking() const {
   if (lastDecodedAvFramePts_ == cursor_) {
     // We are seeking to the exact same frame as we are currently at. Without
     // caching we have to rewind back and decode the frame again.
-    // TODO: https://github.com/pytorch/torchcodec/issues/84 we could
-    // implement caching.
```
NicolasHug, line 1333: This won't happen, corresponding issue was closed. Not the kind of optimization that is worth doing anymore with the existing seeking patterns.

test/test_decoders.py
```diff
@@ -1675,15 +1672,6 @@ def test_10bit_gpu_fallsback_to_cpu(self):
         frames_cpu = decoder_cpu.get_frames_at(frame_indices).data
         assert_frames_equal(frames_gpu.cpu(), frames_cpu)
 
-    @pytest.mark.parametrize("device", all_supported_devices())
-    @pytest.mark.parametrize("asset", (H264_10BITS, H265_10BITS))
-    def test_10bit_videos(self, device, asset):
```
NicolasHug, line 1680: The logic of this test is now entirely covered by the `*output_dtype*` tests

PR #1483, opened 2026-06-03
doc: add a note on XPU plugin availability

dvrogozh:
@NicolasHug , we finally started release XPU plugin for TorchCodec. With this PR I wanted to check with you if you would be ok to add a link to the XPU effort from the TorchCodec readme?


NicolasHug (approved):
Thanks @dvrogozh , LGTM. Just made some minor changes

PR #1487, opened 2026-06-16
Fix build failure on Python 3.14t without --enable-shared

leveretconey:
Free-threaded Python (3.13t+) uses mimalloc, whose initial-exec TLS is incompatible with shared libraries loaded via dlopen(). When Python is built without --enable-shared (the default), linking ${Python3_LIBRARIES} pulls in the static libpython3.14t.a and triggers a link error.

On Linux/macOS, extension modules don't need to link libpython — symbols resolve from the interpreter at runtime. Only Windows requires it (DLL import library).

This is consistent with how PyTorch handles it in torch/CMakeLists.txt (only links Python libraries on MSVC), and also consistent with pybind_ops in this repo which already avoids linking ${Python3_LIBRARIES}.

### DIFF
--- a/src/torchcodec/_core/CMakeLists.txt
+++ b/src/torchcodec/_core/CMakeLists.txt
@@ -185,8 +185,10 @@ function(make_torchcodec_libraries
     )
     set(custom_ops_dependencies
         ${core_library_name}
-        ${Python3_LIBRARIES}
     )
+    if(WIN32)
+        list(APPEND custom_ops_dependencies ${Python3_LIBRARIES})
+    endif()
     make_torchcodec_sublibrary(
         "${custom_ops_library_name}"
         SHARED


leveretconey:
rebased latest main because new commits fixed ci failures

NicolasHug:
Thanks for the PR @leveretconey , I pulled this in https://github.com/meta-pytorch/torchcodec/pull/1485 but there seem to still be build issues for the free-threaded builds on windows.

Did you try to confirm the build was working on your side? And can you share about your needs for free-threaded wheels on Windows specifically?

leveretconey:
Thanks @NicolasHug! Yes, I confirmed the build works on Linux with Python 3.14.5t (free-threaded, without --enable-shared). Our use case is Linux wheels only — no specific need for Windows.

This PR preserves ${Python3_LIBRARIES}` linking underif(WIN32), so it shouldn't change Windows behavior. I don't have a Windows environment to investigate further, but happy to adjust if needed.

NicolasHug:
I see, I got the context wrong, sorry. 

So in that case my question becomes: in what context do you need this? We *can* build linux 3.14t wheels on our CI, e.g. here: https://github.com/meta-pytorch/torchcodec/actions/runs/26814497648/job/79052728683?pr=1477

leveretconey:
@NicolasHug Your CI Python is likely built with --enable-shared. In our environment (and many others like pyenv, system Python, official python.org builds), --enable-shared is not enabled by default. Linking ${Python3_LIBRARIES}` then pulls in the static libpython3.14t.a and fails. Since extension modules don't need to link libpython on Linux anyway, this fix is correct regardless.

NicolasHug:
@leveretconey  I just landed https://github.com/meta-pytorch/torchcodec/pull/1516 which I hope should also address your use-case. Could you check if you still have the same issue or if this PR can be closed?

leveretconey:
Thanks @NicolasHug, I verified #1516 in our Linux Python 3.14t free-threaded environment without --enable-shared, and it addresses my original issue.

I'll close this PR now.

PR #1492, opened 2026-06-23
Use camel_case for cpp codebase

NicolasHug:
Something we've been wanting to do for a while now. With claude + clang tidy

### DIFF
--- a/test/third-party-interface/ThirdPartyInterfaceTest.cpp
+++ b/test/third-party-interface/ThirdPartyInterfaceTest.cpp
@@ -16,25 +16,25 @@ class DummyDeviceInterface : public DeviceInterface {
 
   virtual ~DummyDeviceInterface() {}
 
-  void initialize(const SharedAVCodecContext& codecContext) override {
+  void initialize(const SharedAVCodecContext& codec_context) override {
     // Access to CPU device interface is essential for CPU fallback
     // implementation
-    std::unique_ptr<DeviceInterface> cpuInterface =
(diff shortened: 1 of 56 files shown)


test/third-party-interface/ThirdPartyInterfaceTest.cpp
```diff
@@ -16,25 +16,25 @@ class DummyDeviceInterface : public DeviceInterface {
...
-        createDeviceInterface(kStableCPU);
+    std::unique_ptr<DeviceInterface> cpu_interface =
+        create_device_interface(kStableCPU);
   }
 
-  void convertAVFrameToFrameOutput(
-      UniqueAVFrame& avFrame,
-      FrameOutput& frameOutput,
-      std::optional<torch::stable::Tensor> preAllocatedOutputTensor =
+  void convert_av_frame_to_frame_output(
```
NicolasHug, line 26: @dvrogozh  will you hate me for this?  😬

dvrogozh: You doubt? Definitely! But wait... that's not you, that's Mr. Claude, is he? :)

Don't worry. I have immunity to this kind of stuff from the time of cherry-picking few hundreds Linux kernel patches some of which were adding whitespace changes here and there.

Thanks for heads up. That's a breaking change to handle on my side. This will go to 0.15, right?

NicolasHug: Yes this will be in 0.15 !

PR #1524, opened 2026-07-20
jpeg: fix setjmp/longjmp segfault

NicolasHug:
Fix a potential segfault in the jpeg decoder. The segfault manifests in TC which is compiled with O3, but not in TorchVision which is compiled with O3.

The main problem is that we had the `output` tensor declaration in the same function  where we would define the `setjmp()` point. This UB per the C standard.

The fix is to split `decode_jpeg()` into two helper functions: `read_header_and_start()` and `decode_rows()`. These two helpers define the `setjmp()` points while the `output` tensor is declared in `decode_jpeg()`: they're not in the same function anymore, so we're not in UB territory.

See the `Note [libjpeg error handling]` in the code for more details.

### DIFF
--- a/src/torchcodec/_core/DecodeJpeg.cpp
+++ b/src/torchcodec/_core/DecodeJpeg.cpp
@@ -37,7 +37,7 @@ torch::stable::Tensor decode_jpeg(
 #include <algorithm>
 #include <cstdint>
 #include <cstring>
-#include <optional>
+#include <tuple>
 
 #include "Exif.h"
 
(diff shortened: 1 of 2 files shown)


src/torchcodec/_core/DecodeJpeg.cpp
```diff
@@ -236,50 +246,11 @@ void set_source_ctx(
...
-  // won't run. We use std::optional to declare tensors before setjmp while
-  // deferring construction, and explicitly reset them on the error path.
-  error_ctx_t error_ctx;
-  jpeg_ctx.err = jpeg_std_error(&error_ctx.base);
-  error_ctx.base.error_exit = error_exit_cb;
-  // Establish the setjmp return context for error_exit_cb to use.
-  if (setjmp(error_ctx.setjmp_buffer)) {
-    // Release any tensors that may have been allocated after setjmp.
-    cmyk_line_tensor.reset();
-    output.reset();
```
NicolasHug, line 268: these two lines is what would segfault in `main`:  `cmyk_line_tensor` and `output` are indeterminate on this error path because they are:

- not volatile
- in the same function as the setjmp()
- modified within the setjmp / longjmp span.

PR #1535, opened 2026-07-23
Bundle license files of the image decoder libs (jpeg etc.) in our wheels

NicolasHug:
We now dynamically link against a bunch of libs like libjpeg[turbo], libavif, libpng etc. They're all permissive but we must ship their corresponding license and copyright files in our wheels.


NicolasHug:
Verified a few wheels on linux, macos and windows, this seems to work, merging

PR #1554, opened 2026-07-27
Add ROCJPEG decoder

NicolasHug:
Ported from TV.

Would need https://github.com/meta-pytorch/torchcodec/pull/1553 first but trying to make progress on this now


.github/workflows/linux_rocm.yaml
```diff
@@ -0,0 +1,140 @@
...
+        python-version: ['3.10']
+        rocm-version: ['7.1']
+        ffmpeg-version-for-tests: ['7']
+    uses: pytorch/test-infra/.github/workflows/linux_job_v2.yml@main
+    permissions:
+      id-token: write
+      contents: read
+    with:
+      repository: meta-pytorch/torchcodec
+      runner: "linux.rocm.gpu.gfx950.1"
```
jithunnair-amd:
```suggestion
      runner: "linux.rocm.gpu.meta-pytorch.mi350.1"
```

jithunnair-amd:
@NicolasHug For ROCm versions >= 7.14, the RPATH solution would work (for Linux at least). For ROCm 7.2 and earlier, would you want us to implement the same bundling logic here that we had in torchvision?
Cc @akashveramd

jithunnair-amd:
> @NicolasHug For ROCm versions >= 7.14, the RPATH solution would work (for Linux at least). For ROCm 7.2 and earlier, would you want us to implement the same bundling logic here that we had in torchvision? Cc @akashveramd

Pardon, I see that the intent in this PR is [*not* to bundle](https://github.com/meta-pytorch/torchcodec/pull/1554/changes#diff-db562eea79e950f9bba9ff52101d2b130aa2bda326a1436b07722d06403bf4e4R142), and instead install the rocjpeg dependency explicitly at build *and* test time: packaging/install_rocjpeg.sh

@akashveramd In that case, *assuming that this is a problem of locating librocjpeg*, it should be a matter of ensuring that the ROCm 7.2 and earlier builds can locate the librocjpeg.so being installed via the above script.

@NicolasHug This might come down to using LD_LIBRARY_PATH or RPATH (but the latter would need to assume an `/opt/rocm` installation, which is not always strictly true). The LD_LIBRARY_PATH should fix it for the CI env, but users of torchcodec wheels may still see the issue. The latter should fix it for most torchcodec users (except those who have ROCm installed in a non-`/opt/rocm` path; they would need to set `LD_LIBRARY_PATH` anyway).

Please let us know what you'd prefer.

NicolasHug:
Ideally the ROCm TorchCodec wheel would ship the relevant rocjpeg libs, just like the CUDA wheels are shipping libnvjpeg.

But claude is [saying this](https://github.com/meta-pytorch/torchcodec/pull/1554/changes#diff-aaa8b09b27cd12613b4ce57e17d1ed16ff5f2310e1048fbfec93742493204912R171-R177):

>  rocJPEG is not shipped by the torch-ROCm
        wheel, and bundling it would drag in torch's ROCm libs under mismatched
        (hashed) sonames. So we treat it as a runtime dependency provided by the
        ROCm install, like FFmpeg. decode_jpeg(device='cuda') therefore needs
        ROCm (with rocJPEG) present at runtime.


I haven't verified this claim - if you can find a way to properly bundle the rocjpeg libs, I'm happy to bundle them.

jithunnair-amd:
> Ideally the ROCm TorchCodec wheel would ship the relevant rocjpeg libs, just like the CUDA wheels are shipping libnvjpeg.
> 
> But claude is [saying this](https://github.com/meta-pytorch/torchcodec/pull/1554/changes#diff-aaa8b09b27cd12613b4ce57e17d1ed16ff5f2310e1048fbfec93742493204912R171-R177):
> 
> > rocJPEG is not shipped by the torch-ROCm
> > wheel, and bundling it would drag in torch's ROCm libs under mismatched
> > (hashed) sonames. So we treat it as a runtime dependency provided by the
> > ROCm install, like FFmpeg. decode_jpeg(device='cuda') therefore needs
> > ROCm (with rocJPEG) present at runtime.
> 
> I haven't verified this claim - if you can find a way to properly bundle the rocjpeg libs, I'm happy to bundle them.

@akashveramd please take note

akashveramd:
@NicolasHug: It seems the wheels is build and uploaded for ROCm 7.1 & 7.2. But the test runs only on 7.1. Do we want to run the test on both 7.1 & 7.2 or build only on one ROCm version?
FYI, I will be replacing ROCm 7.1/7.2 with 7.2/7.14 version.

NicolasHug:
I am happy to add more coverage if there is value there, but my priority is to make the 7.1 (or any other version) work correctly at the moment

PR #1585, opened 2026-07-30
Add `Packet.from_tensor` to build packets from raw data

oxkitsune:
### What

Packets could previously only be obtained from a `Demuxer`.

This adds `Packet.from_bytes` and `Packet.from_tensor` to construct a `Packet` directly from raw compressed data (e.g. from a network stream), with caller-provided `pts`, `duration`, and `is_key_frame`.

`pts` and `duration` are in the consuming decoder's stream time base. `dts` is set to `pts`, so this is only correct for streams without B-frame reordering.

### DIFF
--- a/src/torchcodec/_core/_ffmpeg_op_names.py
+++ b/src/torchcodec/_core/_ffmpeg_op_names.py
@@ -31,6 +31,7 @@ FFMPEG_OP_NAMES = frozenset(
         "get_json_metadata",
         "_blocks_create_demuxer",
         "_blocks_demuxer_next_packet",
+        "_blocks_packet_from_tensor",
         "_blocks_create_packet_decoder",
         "_blocks_packet_decoder_send_packet",
         "_blocks_packet_decoder_send_eof",
--- a/src/torchcodec/_core/_ffmpeg_ops.py
+++ b/src/torchcodec/_core/_ffmpeg_ops.py
@@ -96,6 +96,7 @@ _blocks_create_demuxer = torch.ops.torchcodec_ns._blocks_create_demuxer.default
 _blocks_demuxer_next_packet = (
     torch.ops.torchcodec_ns._blocks_demuxer_next_packet.default
 )
+_blocks_packet_from_tensor = torch.ops.torchcodec_ns._blocks_packet_from_tensor.default
 _blocks_create_packet_decoder = (
     torch.ops.torchcodec_ns._blocks_create_packet_decoder.default
 )
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -13,6 +13,8 @@
 #include <pybind11/pybind11.h>
 #pragma pop_macro("TORCH_TARGET_VERSION")
 #include <cstdint>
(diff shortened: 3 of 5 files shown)


NicolasHug:
Hi @oxkitsune , thanks for the PR and for your interest in the these new APIs! They are still very WIP and subject to change (and not public!) so I'd love to learn more about your use for it?

----

On the PR itself, I'm open to it and to support that feature in general, but I won't merge it now because I plan to continue working (and heavily changing) these APIs in the near future. I also suspect we'll be supporting a file-like input in the `Demuxer`, which should address the network stream use-case.

oxkitsune:
Hi @NicolasHug, these new APIs are awesome!
My usecase is to streaming in packets (typically one or two GOPs) in over the network and decoding them to tensors for training.

>  I also suspect we'll be supporting a file-like input in the Demuxer, which should address the network stream use-case.

This sounds great! Looking forward to it. If there's anything I can help out with, let me know!

NicolasHug:
Thanks! Can you share more on why the `VideoDecoder` (which supports file-like inputs already) isn't addressing your use-case? Is it because your streams don't have a duration metadata, which the `VideoDecoder` expects (it's not *truly* a stream decoder by that definition)?

oxkitsune:
You're right that missing duration metadata is one of the blockers (`exact` mode also fails on the rewind seek after its scan). We have PTS/duration, so we can work around it by wrapping each packet window (keyframe -> target) in an in-memory MP4 and using the file-like input.

> Can you share more on why the VideoDecoder (which supports file-like inputs already) isn't addressing your use-case?

Yea! Our use case is a bit different than the usual one I guess 😄

---

Some context on our setup: Rerun stores video as raw codec packets (Annex B / OBUs) in a database, one packet per row keyed by timestamp, with keyframe flags in a separate column. There is no file.
Our training dataloader samples random timestamps across many recordings, and for each sampled frame the server hands us exactly the packets from the preceding keyframe up to the target. So our ideal input would be a list of raw packets + the PTS/duration metadata which we can reconstruct.

With the current file-like input API we need to wrap the window of packets (keyframe to target frame) in an in-memory MP4 container in order to feed it to a `VideoDecoder` instance. This also means we cannot re-use a decoder across samples: consecutive targets often land in the same GOP, but since samples arrive one at a time we can't know the full window up front, each sample means muxing a fresh container, constructing a fresh `VideoDecoder`, and re-decoding from the keyframe.

---

So what I'd love from a packet-level API is the ability to feed raw packets (with keyframe/PTS info) into a decoder object, ask for decoded frames as tensors, and ideally be able to keep feeding packets to an existing instance.

NicolasHug:
Hey just giving you a quick update. I'll release the 'Blocks" APIs publicly soon. That first release won't have support for that just yet.

I think part of what you need is to be able to build a demuxer-free decoder. That is also what is needed for enabling multi-process pipelines. That's on my TODO list, and I'll keep both use-cases in mind when designing for that.

You also need a way to build packets from scratch as done here. I think this will tie the AVPacket data to one specific FFmpeg major version. Not sure there's a way around that, maybe that's fine.

PR #1592, opened 2026-07-31
Blocks API: CUDA support

NicolasHug:


benchmarks/bench_blocks.py
```diff
@@ -189,7 +194,7 @@ def main():
 
     results = {}
     for name, fn in methods.items():
-        times_ns, cpu = bench(fn, path)
+        times_ns, cpu = bench(fn, path, args.device)
```
NicolasHug, line 197: needs a cuda sync somewhere

src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -700,11 +789,60 @@ int BetaCudaDeviceInterface::receive_frame(UniqueAVFrame& av_frame) {
...
+  }
+
+  // NVDEC lays the UV plane right after the Y plane in a single allocation, so
+  // the whole frame is one contiguous copy. The Y plane always has an even
+  // number of rows, see convert_cuda_frame_to_av_frame().
+  int64_t pitch = av_frame->linesize[0];
+  int64_t even_height = round_up_to_even(av_frame->height);
+  int64_t num_bytes = pitch * even_height * 3 / 2;
+
+  cudaStream_t stream = get_current_cuda_stream(device_.index());
```
NicolasHug, line 823: shouldn't we wait on `info->producing_stream` (which i think is the nvdec_stream) before starting the copy?

PR #1595, opened 2026-08-03
Don't let convert_av_frame_to_frame_output() take ownership of the AVFrame

NicolasHug:
`convert_av_frame_to_frame_output()` took a non-const `UniqueAVFrame&`, which let implementations steal the caller's frame (by e.g. calling `std::move()` on it as done in the `BetaCudaDeviceInterface`)

This proved to cause subtle ownership issues as I started working on CUDA support for the "Blocks" API - the AVFrame that got `std::moved()` would be double-freed.

The parameter is now `const`, programatically enforcing that ownership remains to the caller.

---


NicolasHug:
Superseded by https://github.com/meta-pytorch/torchcodec/pull/1598

PR #1596, opened 2026-08-03
Store a UniqueAVFrame in the blocks frame handle

NicolasHug:
The frame handle held a raw AVFrame*, so _blocks_convert_frame() had to
build a UniqueAVFrame around a pointer the handle already owned and then
release() it again to undo the fabricated second owner. It only worked
because nothing downstream acted on that ownership.

Hold a UniqueAVFrame instead. It's an ordinary C++ object whose destructor
calls av_frame_free, so the generic wrap_pointer_to_tensor<T> handle used
by the other blocks types applies and the bespoke AVFrame wrapper goes
away. _blocks_convert_frame() then has a UniqueAVFrame to pass by
reference, with no temporary owner in sight.

Costs one small allocation per frame for the wrapper object.

---

### DIFF
--- a/src/torchcodec/_core/custom_ops.cpp
+++ b/src/torchcodec/_core/custom_ops.cpp
@@ -182,7 +182,7 @@ SingleStreamDecoder* unwrap_tensor_to_get_decoder(
 }
 
 // Generic pointer<->tensor laundering for the building-block handle types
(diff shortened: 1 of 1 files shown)


NicolasHug:
Superseded by https://github.com/meta-pytorch/torchcodec/pull/1598

PR #1602, opened 2026-08-05
Know which 'mode' the BetaCudaDeviceInterface is in

NicolasHug:
With recent changes to enable the Blocks APIs, the interfaces can now be in three modes:

- decoding-only
- color-conversion only
- both

Previously, with the SingleStreamDecoder, the only mode was always 'both'. This PR adds a mode() method on the BetaCudaDeviceInterface to so that we explicitly know how the current interface should behave - and so we can validate /enforce assumptions.

No behavior change, this is purely for robustness.

### DIFF
--- a/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
+++ b/src/torchcodec/_core/BetaCudaDeviceInterface.cpp
@@ -76,6 +76,14 @@ cudaVideoSurfaceFormat get_preferred_surface_format(OutputDtype output_dtype) {
                                               : cudaVideoSurfaceFormat_NV12;
 }
 
+// Whether a frame is a CPU-fallback frame rather than a GPU NVDEC surface,
+// inferred from its pixel format. This works today because our CPU fallback
+// never yields NV12/P016 frames, but it's only a proxy, and it's not super
+// robust to future changes.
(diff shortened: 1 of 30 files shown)


src/torchcodec/_core/BetaCudaDeviceInterface.cpp
```diff
@@ -867,56 +888,62 @@ void BetaCudaDeviceInterface::make_frame_standalone(UniqueAVFrame& av_frame) {
...
-
-  auto y_plane = static_cast<uint8_t*>(storage.mutable_data_ptr());
-  av_frame->data[0] = y_plane;
-  av_frame->data[1] = y_plane + (pitch * even_height);
 
   auto attached_data = new StandAloneFrameAttachedData();
   attached_data->producer_stream = current_stream;
-  // TODO_API_BREAKDOWN_CUDA P2: We don't *really* need to std::move it I guess?
-  attached_data->storage = std::move(storage);
+
```
NicolasHug, line 902: drive-by change where we also attach the consumer_stream on CPU frames for consistency. This may change later when we address the newly added TODO:

> TODO_API_BREAKDOWN P1: we should do this before the color-conversion..

PR #1612, opened 2026-08-07
Add TorchScript uniform video decode operator (#1612)

psaikaushik:
Summary:

Add a registered C++ operator that uniformly samples an encoded video from memory with TorchCodec. It returns decoded NCHW frames, sampling metadata, and an explicit validity status for serialized TorchScript callers.

Decode failures produce an invalid result so callers can handle corrupt inputs without propagating native exceptions.

Differential Revision: D115231531


NicolasHug (changes requested):
Discussion on the diff

PR #1671, opened 2026-08-24
fix: add missing num_threads param to decode_avif disabled stub

### LINKED ISSUE #1670: decode_avif stub has wrong signature when built with TORCHCODEC_BUILD_AVIF=0
## 🐛 Describe the bug

When building torchcodec with `TORCHCODEC_BUILD_AVIF=0`, importing torchcodec fails at runtime with:

```
OSError: .../torchcodec/libtorchcodec_image.so: undefined symbol: _ZN8facebook10torchcodec11decode_avifERKN5torch6stable6TensorElll
```

Demangled: `facebook::torchcodec::decode_avif(torch::stable::Tensor const&, long, long, long)`

The root cause is a signature mismatch between the header and the disabled stub:

- `DecodeAvif.h` declares `decode_avif` with 4 parameters: `(const Tensor&, int64_t mode, int64_t output_dtype, int64_t num_threads)`
- The `#if !TORCHCODEC_ENABLE_AVIF` stub in `DecodeAvif.cpp` only defines 3 parameters: `(const Tensor&, int64_t mode, int64_t output_dtype)` (missing `num_threads`)

Since `image_custom_ops.cpp` references `&decode_avif` (which resolves to the 4-parameter declaration from the header), the 4-parameter symbol is never defined when AVIF is disabled. The linker allows this for shared libraries, but `dlopen` fails at runtime.

**Reproducer:**

```bash
TORCHCODEC_BUILD_AVIF=0 pip install torchcodec --no-binary torchcodec
python -c "import torchcodec"
# OSError: undefined symbol
```

**Fix:** Add the missing `num_threads` parameter to the disabled stub. PR incoming.

## Versions

- torchcodec: 0.16.0 (also affects current `main`)
- torch: 2.11.0
- OS: RHEL 9 / Linux
- Python: 3.12

andre-motta:
Fixes #1670

## Summary

The `#if !TORCHCODEC_ENABLE_AVIF` stub for `decode_avif` in `DecodeAvif.cpp` is missing the `num_threads` parameter that the header (`DecodeAvif.h`) declares and the enabled implementation uses. This causes `libtorchcodec_image.so` to have an undefined symbol when built with `TORCHCODEC_BUILD_AVIF=0`, which makes `import torchcodec` fail at runtime.

The fix adds the missing `[[maybe_unused]] int64_t num_threads` parameter to the disabled stub so its signature matches the header declaration.

## Test plan

- Build with `TORCHCODEC_BUILD_AVIF=0` and verify `import torchcodec` succeeds
- Verify `decode_avif()` raises the expected "not compiled with libavif support" error instead of a symbol resolution crash

### DIFF
--- a/src/torchcodec/_core/DecodeAvif.cpp
+++ b/src/torchcodec/_core/DecodeAvif.cpp
@@ -19,7 +19,8 @@ namespace facebook::torchcodec {
 torch::stable::Tensor decode_avif(
     [[maybe_unused]] const torch::stable::Tensor& input,
     [[maybe_unused]] int64_t mode,
-    [[maybe_unused]] int64_t output_dtype) {
+    [[maybe_unused]] int64_t output_dtype,
+    [[maybe_unused]] int64_t num_threads) {
   STD_TORCH_CHECK(
       false,
       "decode_avif: torchcodec was not compiled with libavif support. Rebuild "


NicolasHug (approved):
Thanks @andre-motta

PR #1677, opened 2026-09-01
Let the C++ Demuxer follow several streams of one container

NicolasHug:
The demuxer selected exactly one stream at construction. It now opens and probes in the constructor, and follows streams through `add_stream()`, which can be called more than once; `av_read_frame`'s output is filtered against the set, and each packet is handed back with the stream index it came from.

`seek()` takes the stream to resolve against, since FFmpeg resolves a seek in one stream's time base and lands on that stream's keyframes: it's stream-dependent. By default, it picks the first followed stream as passed by the `streams=` parameter at Demuxer creation.

No Python API changes.
---

### DIFF
--- a/src/torchcodec/_core/Demuxer.cpp
+++ b/src/torchcodec/_core/Demuxer.cpp
@@ -32,18 +32,44 @@ std::string get_seek_error_message(
   return ss.str();
 }
 
-int read_next_packet(
+namespace {
+template <typename IsActive>
+int read_next_active_packet(
     AVFormatContext* format_context,
-    int active_stream_index,
-    ReferenceAVPacket& packet) {
+    ReferenceAVPacket& packet,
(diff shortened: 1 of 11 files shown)


src/torchcodec/_core/Demuxer.cpp
```diff
@@ -32,19 +32,45 @@ std::string get_seek_error_message(
...
+
+int read_next_packet(
+    AVFormatContext* format_context,
+    int active_stream_index,
+    ReferenceAVPacket& packet) {
+  return read_next_active_packet(
+      format_context, packet, [active_stream_index](int stream_index) {
+        return stream_index == active_stream_index;
+      });
+}
```
NicolasHug, line 60: I think this one can be removed when addressing the PR just above (scan)

PR #1678, opened 2026-09-01
Scan every followed video stream in one pass, and only before demuxing

NicolasHug:
A scan used to read the container once per stream and cache nothing, so indexing two video streams read the file twice and calling scan() twice read it twice again. It is now two phases: one pass over the container buckets the packets of every followed video stream, and the sort plus tensor build happens per stream, the first time that stream's scan is asked for. The I/O is what a scan costs, and it is now paid once. The sorting is paid per-stream (no way around that), but only if / when requested.

Scanning after the first packet has been demuxed raises.

Scanning a non-video stream now raises rather than quietly producing a frame index for something that has no frames or keyframes.

---

### DIFF
--- a/src/torchcodec/_core/Demuxer.cpp
+++ b/src/torchcodec/_core/Demuxer.cpp
@@ -249,29 +249,37 @@ void Demuxer::seek(double seconds, std::optional<int> stream_index) {
       get_seek_error_message(format_context_.get(), desired_pts, status));
 }
 
-namespace {
-struct ScannedPacket {
-  int64_t pts;
-  int64_t duration;
-  bool is_key_frame;
-};
-} // namespace
+void Demuxer::scan_all_video_streams() {
+  if (has_scanned_) {
+    return;
+  }
+  STD_TORCH_CHECK(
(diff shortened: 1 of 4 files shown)


src/torchcodec/_core/Demuxer.cpp
```diff
@@ -259,28 +259,37 @@ void Demuxer::seek(double seconds, std::optional<int> stream_index) {
...
-  seek(0, scanned_stream->index);
+  seek(0);
 
+  // One pass, however many video streams are being followed: the I/O is what a
+  // scan costs, and it is the same read for all of them.
   while (true) {
     ReferenceAVPacket packet(auto_packet_);
     int status =
-        read_next_packet(format_context_.get(), scanned_stream->index, packet);
+        read_next_packet(format_context_.get(), video_stream_indices, packet);
```
NicolasHug, line 282: Do we still need the `read_next_packet` that only takes a single stream index then??

NicolasHug: Yes, it's used in SingleStreamDecoder

PR #1719, opened 2026-09-14
[torchcodec] Add restricted in-memory media decoding

daniel-salib:
Summary:
This is the low-level TorchCodec foundation for safely restricting uploaded media. It does not add a server flag or change client requests; `D119264076` connects it to vLLM.

What changes:

- Adds optional `input_format` and `allowed_decoders` arguments for video bytes and tensors.
- When configured, tells FFmpeg exactly which container parser and decoders may process the input.
- Enforces the decoder policy during both stream probing and final decoder setup, and blocks nested file or URL opens.
- Allows another implementation of an approved codec (for example, `h264_cuvid` for `h264`) while still rejecting a different codec.
- Checks every decodable video, audio, and subtitle stream while allowing metadata/timecode tracks that have no decoder.
- Leaves existing decoding unchanged when the options are not configured.

The MOV/H264 policy used below is intentionally strict test configuration, not a production-wide recommendation. Deployments can allow the formats they support; this remains defense in depth rather than a sandbox, so every permitted parser and decoder still needs security review.

Differential Revision: D119264074


NicolasHug (changes requested):
Let's follow-up on phabricator D119264074

PR #1732, opened 2026-09-16
Fix alpha stripping for palette PNGs with tRNS

### LINKED ISSUE #1731: Palette PNG with tRNS decoded as RGB corrupts pixels and can overrun output buffer
### 🐛 Describe the bug

Decoding a palette PNG with a tRNS transparency chunk using `mode="RGB"` mixes alpha bytes into the color output and can corrupt heap memory. A 2×1 synthetic image is enough to show incorrect pixels; a larger image triggered SIGSEGV locally and `double free or corruption (out)` in another environment. No model or accelerator is involved.

Minimal reproducer (run in a disposable process because the affected native decoder can write past its output allocation):

```python
import io

import torch
from PIL import Image
from torchcodec.decoders import decode_png

image = Image.new("P", (2, 1), color=0)
image.putpalette([17, 34, 51, 68, 85, 102])
image.putpixel((1, 0), 1)
buffer = io.BytesIO()
image.save(buffer, format="PNG", transparency=bytes([0, 128]))

actual = decode_png(buffer.getvalue(), mode="RGB")
expected = torch.tensor([[[17, 68]], [[34, 85]], [[51, 102]]], dtype=torch.uint8)
print("Expected:", expected.tolist())
print("Actual:  ", actual.tolist())
torch.testing.assert_close(actual, expected)
```

Observed:

```text
Expected: [[[17, 68]], [[34, 85]], [[51, 102]]]
Actual:   [[[17, 0]], [[34, 68]], [[51, 85]]]
AssertionError: Tensor-likes are not equal!
Mismatched elements: 3 / 6 (50.0%)
```

`DecodePng.cpp` calls `png_set_palette_to_rgb()`, which also expands palette tRNS transparency into alpha. However, the RGB and GRAY branches only call `png_set_strip_alpha()` when the original PNG color type has `PNG_COLOR_MASK_ALPHA`. Palette PNGs do not have this flag. The decoder therefore allocates and advances rows for 3 channels (1 for GRAY), while libpng writes 4 (2 for GRAY). The last row can overrun the allocation. `png_read_update_info()` is called, but the manually tracked output channel count is not checked against the resulting libpng layout.

The same logic is present on main at `73cc2f309c2bee96857d7b24c46e34401e154a66`. I am preparing a focused fix and regression tests for palette tRNS decoded without alpha.

### Versions

- TorchCodec: 0.16.0+cpu
- PyTorch: 2.14.0+cpu
- Pillow: 11.3.0
- Python: 3.12.13
- OS: Ubuntu 26.04 LTS, Linux x86_64
- glibc: 2.43

The original larger-image crash was also observed with torch 2.14.0 and torchcodec 0.16.0 on Linux x86_64 / Python 3.12.

Yanko-7:
Fixes #1731.

When a palette PNG has tRNS transparency, libpng expands the palette to RGBA. The RGB and GRAY paths only stripped alpha when it was present in the original color type, so the decoder could write four channels into a three-channel tensor (or two into one), corrupting pixels and potentially writing past the allocation.

Strip alpha for tRNS inputs in these conversion paths as well. Add a synthetic 2×1 PNG regression test covering RGB/GRAY output, palette bit depths 1/2/4/8, and uint8/uint16/auto output dtypes.

Tests:

- Built from main at `73cc2f309c2bee96857d7b24c46e34401e154a66`, with PNG enabled and torch 2.14.0+cpu.
- The regression test fails on unmodified main with 3/6 RGB elements incorrect; all 24 parameter combinations pass after the fix.
- `python -m pytest test/test_decoders.py test/test_encoders.py -k png -q`: 166 passed, 12 skipped, 2751 deselected.
- `pre-commit run --files src/torchcodec/_core/DecodePng.cpp test/test_decoders.py`: passed.
- The original 497×315 palette+tRNS PNG decodes successfully for 12 iterations after the fix.


NicolasHug (approved):
Thanks for the report and for the fix @Yanko-7 . I made some minor changes to also fix the same issue with 16-bit pngs.

